PR #11, opened 2014-05-28
Fixed race condition discovered by Intel Inspector. 

thender:
This change also matches the thread-safe version of this file in the WRF repository.


mgduda:
The changes proposed in this PR were merged to the private repository that was used for primary MPAS development up to the v6.0 release. The relevant commit is a8cd3e6f . We can safely close this PR now.

PR #29, opened 2018-04-26
Commented out two flags and updated bgc parameter defaults.

njeffery:
BFB. Changes two "stops" to warnings in the bgc code.

This PR concerns bgc in the seaice core and is needed for the CBGC v1 simulations.


njeffery:
@akturner , @maltrud , @jonbob : seaice in this PR should be identical to the version @maltrud is running in the E3SM bgc simulation.  Also, Registry has been updated with the necessary bgc namelist defaults.  Please let me know if this is not the case.

PR #35, opened 2018-05-02
Added BGC testing and corrected 0 dimension

njeffery:
BFB in the physics only.
1) Corrects regression, parallelism and restartability tests for v1 BGC.
2) Changes nDIC value in Registry


testing_and_setup/seaice/testing/tests/restartability_bgc.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
```
akturner: There is no need to create a whole new test script when it is almost identical to the original restartability script, simply add an option for bgc in the test suite, e.g.

			<test name="restartability_bgc">
			  <option name="bgc" value="True"/>
			</test>	

Then copy the functionality from the parallelism scrip for the multipleBlocks option e.g.

    bgcRestart = False
    if ("bgc" in options.keys() and options["bgc"] == "True"):
        bgcRestart = True

if (not bgcRestart):
    nmlChanges = {"seaice_model": {"config_start_time":"file"},
                  "restart":    {"config_do_restart":True}}
else:
    nmlChanges = {"seaice_model": {"config_start_time":"file"},
                  "restart":    {"config_do_restart":True,
                                 "config_do_restart_bgc":True, 
                                 "config_do_restart_hbrine":True}}

PR #38, opened 2018-05-03
Update Makefile to enable building with recent XL compilers under Linux

mgduda:
This merge adds several changes that have been found necessary for building
MPAS with recent (V15.1.5, but not the newest, V16) versions of the XL compilers
under Linux. In summary, the names of the compilers and MPI compiler wrappers
have been updated, pre-processing directives have been cleaned up, and a flag
has been added to set the unformatted record endianness to big-endian.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,11 @@ xlf:
 	"CC_SERIAL = xlc" \
 	"CXX_SERIAL = xlcxx" \
 	"FFLAGS_PROMOTION = -qrealsize=8" \
-	"FFLAGS_OPT = -O3" \
+	"FFLAGS_OPT = -O3 -qufmt=be" \
 	"CFLAGS_OPT = -O3" \
 	"CXXFLAGS_OPT = -O3" \
 	"LDFLAGS_OPT = -O3" \
-	"FFLAGS_DEBUG = -O0 -g -C" \
+	"FFLAGS_DEBUG = -O0 -g -C -qufmt=be" \
 	"CFLAGS_DEBUG = -O0 -g" \
 	"CXXFLAGS_DEBUG = -O0 -g" \
 	"LDFLAGS_DEBUG = -O0 -g" \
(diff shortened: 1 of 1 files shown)


trudeaun:
@mgduda Hi Michael. I think you might be using a version of the IBM XL Fortran for Linux compiler that is 15.1.5 or older. In those older releases, the `-D` option does not define macros, please find more details [here](https://www.ibm.com/support/knowledgecenter/en/SSAT4T_15.1.5/com.ibm.xlf1515.lelinux.doc/compiler_ref/bigd.html). From [your comment](https://github.com/MPAS-Dev/MPAS-Model/pull/38/commits/afae72d1ceb17f54cc85f0802e44c05f45dd450d), `-DUSE_PIO2` is actually interpreted as:
`-D` generate preprocessed files
`-U` make Fortran case-sensitive
**`-S` generate assembly files**
... and so on. The S generates assembly files (*.s), so that's why you were noticing this behaviour. In V15.1.6 and later releases, `-D` defines macros just like in C/C++, and `-DUSE_PIO2` and other `-D` options will act the way you expect. The latest release (V16.1) was made available recently, on Apr 27.

Alternatively, if you would like to continue using the older release, you may prefix it with `-W`.  e.g. `-W,-DUSE_PIO2`.

mgduda:
@trudeaun Thanks very much for your comments. It looks as though we're using the XL 15.1.5 release, and we'll likely need to stick with that for the very near future.

The MPAS build system doesn't have a clean way at present to add a prefix to macro definitions. However, we can build with an option `GEN_F90=true` that causes all source files to be explicitly pre-processed with cpp, and that's what we've been using on POWER8/Linux with the XL compilers. So, I think the changes in this PR are sufficient, and should we have access to newer versions of the XL compilers in future, we should be able to simply omit the `GEN_F90=true` option and compile directly from `*.F` files as we do with other compilers.

PR #40, opened 2018-05-10
Add AIS meshes to COMPASS

matthewhoffman:
This merge adds two variable resolution Antarctica meshes made using JIGSAW to COMPASS.  One is high res 20 to 2 km including high-res across all marine basins.  The other is ultra low-res 200-40 km for running on not-so-super computers.


matthewhoffman:
PR opened to incorrect branch.  Replaced by #68

PR #41, opened 2018-05-10
add halo reuse calls to ocean btr subcycle

Larofeticus:
This is PR 1521 which got left behind in the old repo. It's been tested (mostly documented in old PR 1458, there was a memory leak which was found and corrected at that time)

### DIFF
--- a/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
@@ -372,6 +372,11 @@ module ocn_forward_mode
 
       endif
 
+      call mpas_dmpar_exch_group_create(domain, 'subcycleFields')
+      call mpas_dmpar_exch_group_add_field(domain, 'subcycleFields', 'sshSubcycle')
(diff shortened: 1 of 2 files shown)


mark-petersen:
@Larofeticus , thanks for updating.  I had been meaning to move the relevant PRs and issues over.  I changed the merge branch into ocean/develop here.  

The previous memory leak was found and fixed, thanks for that.  This PR needs to be fully tested in E3SM v1.0 with a full-scale simulation for performance gain and any other potential problems.  I am out the next two weeks, not sure if others have time.

philipwjones:
@mark-petersen I've tested this in short runs and found it to be b4b and documented performance gains similar to what @Larofeticus had reported (~13% overall at upper end of scaling curve). In the previous tests, how long did it take before problems (i.e. memory leak) showed up? I can try to set something up in E3SM.

jonbob:
@philipwjones - we could see the memory impact fairly quickly, like a matter of days using the oEC60to30 configuration. If it's still happening, it will show up in the memory footprint printed out in the cpl log. Or I can always run it under the totalview memory debugger (if I can find my notes)....

mark-petersen:
OK, spoke with @jonbob yesterday, and it is time to merge.

PR #42, opened 2018-05-14
Fix GL flux & SMB stats

matthewhoffman:
This merge updates the global and regional analysis members to calculate correct values for the grounding line flux and surface mass balance.  

For both variables, new 2d model variables are added that save their values at the time those things are calculated.  Reconstructing them at the end of the time step after masks have changed is problematic.

Testing also revealed that the mask calculation for the location of the grounding line on edges was inconsistent with a recent update to how it is calculated for cells (namely, including grounded ice that discharges into open ocean).  That discrepancy had to be rectified to get the GL flux stat to work correctly.

I confirmed that totalGroundedSfcMassBal is the same as groundingLineFlux
for a simple test case run out to steady state.  Prior to these changes
that was NOT the case - the difference was about 15%.  The test
configuration led to a noisy steady state groundingLineFlux, but
averaging over about 250 time steps yielded agreement between the two
variables to the 4th decimal place.

Additionally a third statistic is added quantifying the mass of ice that went from grounded to floating or floating to grounded during a time step.  This term is required to close the grounded ice mass balance during transient conditions.  (It is zero at steady state.)  I used the same simple test case to show I could close the transient mass balance after the addition of this term.  It indeed fixed errors that occurred when the grounding line position moved.

With these changes, it is possible to close the mass balance for a grounded ice sheet for both steady state and transient situations.  This merge does not assess the ability to close the mass balance of the floating ice sheet (or the ice sheet as a whole).

This merge also sets the default regionMask value to be 1 everywhere, so without a regionMasks input file it will default to the same behavior as global stats.  Additional minor cleanup and minor bug fixes are also included.

Application to more complicated configurations has shown that the grounded ice budget still does not entirely close with these changes, but that will be addressed in a future branch.

### DIFF
--- a/src/core_landice/analysis_members/mpas_li_global_stats.F
+++ b/src/core_landice/analysis_members/mpas_li_global_stats.F
@@ -448,8 +448,16 @@ contains
 
          iceThicknessMean = totalIceVolume / totalIceArea
          avgNetAccumulation = totalSfcMassBal / totalIceArea / rhoi
-         avgGroundedBasalMelt = -1.0_RKIND * totalGroundedBasalMassBal / groundedIceArea / rhoi
-         avgSubshelfMelt = -1.0_RKIND * totalFloatingBasalMassBal / floatingIceArea / rhoi
+         if (groundedIceArea > 0.0_RKIND) then
+            avgGroundedBasalMelt = -1.0_RKIND * totalGroundedBasalMassBal / groundedIceArea / rhoi
+         else
+            avgGroundedBasalMelt = 0.0_RKIND
+         endif
+         if (floatingIceArea > 0.0_RKIND) then
+            avgSubshelfMelt = -1.0_RKIND * totalFloatingBasalMassBal / floatingIceArea / rhoi
+         else
+            avgSubshelfMelt = 0.0_RKIND
+         endif
 
          block => block % next
       end do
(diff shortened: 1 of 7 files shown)


src/core_landice/analysis_members/mpas_li_global_stats.F
```diff
@@ -355,25 +351,9 @@ subroutine li_compute_global_stats(domain, memberName, timeLevel, err)!{{{
...
-               do k = 1, nVertLevels
-                  ! Flux across GL, units = kg/yr
-                  blockGLflux = blockGLflux + fluxSign * layerNormalVelocity(k, iEdge) * dvEdge(iEdge) * &
-                      layerThicknessEdge(k, iEdge) * scyr * config_ice_density  ! convert from m^3/s to kg/yr
-               end do ! end loop over levels
-            end if ! if GL
-
+            ! Flux across GL, units = kg/yr
+            blockGLflux = blockGLflux + fluxAcrossGroundingLine(iEdge) * dvEdge(iEdge) &
+                      * scyr * rhoi ! convert from m^3/s to kg/yr
```
tongzhangice: convert from m^2/s to kg/yr?

matthewhoffman: Yes, good catch - I'll fix this.

matthewhoffman: Fixed in 0866a435

matthewhoffman:
@tongzhangice , @stephenprice , I updated this PR with the additional calculations to determine the grounded mass change that occurs when the grounding line migrates.  See time series figure below.  The black line is the `groundedIceVolume` variable in Gt.  The blue line is the cumulative sum of SMB+BMB+GLflux after the changes in this PR.  The orange also adds in the new variable for grounding line migration.  You can see how important it can be for transient simulations, and how it allows the mass balance to be completely closed.

I also confirmed that the regional stat run for a region covering the whole domain matches the global stat. 

screen shot 2018-05-17 at 2 58 10 pm

Issue #50, opened 2018-05-30
In Compass, path to model invalid with MPICH 3.3a2


xylar:
When I run a Compass test case with MPICH 3.3a2 installed on my laptop (running Ubuntu 18.04), I get the error (the first line repeated once for each MPI task):
```
[proxy:0:0@eleven] HYDU_create_process (utils/launch/launch.c:74): execvp error on file ocean_model (No such file or directory)
Traceback (most recent call last):
  File "./run.py", line 37, in <module>
    'namelist.ocean', '-s', 'streams.ocean'])
  File "/home/xylar/miniconda3/envs/mpas/lib/python2.7/subprocess.py", line 190, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['mpirun', '-n', '4', 'ocean_model', '-n', 'namelist.ocean', '-s', 'streams.ocean']' returned non-zero exit status 255
```

The solution seems to be changing from `ocean_model` to `./ocean_model`.

I was able to get this result by changing this line:
https://github.com/MPAS-Dev/MPAS-Model/blob/master/testing_and_setup/compass/setup_testcase.py#L1086
as follows
```
-                        grandchild.text = executable_link
+                        grandchild.text = './{}'.format(executable_link)
```

Before I make a pull request, I thought I'd get feedback on whether this is an issue others have run into and also whether this fix seems appropriate.

xylar:
@mark-petersen, @matthewhoffman, @vanroekel and @pwolfram, you seem the most likely to have run into this problem.  Does it look familiar or am I deviating from known recipes by using MPICH?

pwolfram:
My understanding is that MPICH is to be supported.  I vaguely recall running into this issue a long time ago in grad school and I believe I did what you did here.  

If you check for the executable "x" bit on ocean_model is it there?  I think this is a pretty strange issue but don't believe the `./ocean_model` syntax will cause a problem in general.

xylar:
> If you check for the executable "x" bit on ocean_model is it there?
```
-rwxr-xr-x  1 xylar xylar  11M May 30 19:45 ocean_model
```
Yep, that's not it.

vanroekel:
First, I have used mpich2 previously to build/run MPAS without issue.  I don't recall seeing this issue, but I think I may have always invoked with ./ocean_model

mgduda:
[...]

PR #60, opened 2018-08-03
Fixes mpich execution in compass

pwolfram:
Ensures that mpich can find executable for `ocean_model`.

PR related to issue #50.


xylar:
@pwolfram, this looks like exactly what worked for me. I didn't, however, try it out again since I don't have time right now.  if you can verify that the change actually works for you, I think you can merge.

xylar:
Thanks for taking care of this!

xylar:
@pwolfram, I didn't realize this until just now, but this change needs to be made off of `develop`, not `ocean/develop`.  Please rebase and change the destination repo.  all `compass` changes above the `ocean` subfolder should not go into `ocean/develop`!

pwolfram:
@xylar, thanks for catching that this should be on develop.  I've migrated the commit to develop as requested.  Thanks!

PR #65, opened 2018-08-14
Added MPAS-Seaice license

akturner:
Added the new MPAS-Seaice license to core_seaice and removed old copyright statements that came into code through thoughless copying


matthewhoffman:
@akturner , do you actually want review from us on this?  It's entirely in the sea ice core, so that's not necessary.

akturner:
@matthewhoffman Just making sure all cores are cool with this LANL only license happening in the sea ice core

PR #69, opened 2018-08-23
Refactoring and GPU optimization of the ocean equation of state

philipwjones:
Large refactoring and GPU optimization of the ocean equation of state routines. Minor performance improvements on KNL, traditional clusters, but enables significant GPU (>2x) improvement on Summit. Roundoff level differences with QU240 standalone tests on Cori and Summit. Did not check Jackett and McDougall reference values.

Changes include:
   - moving calculation of reference pressure to initialization
   - separating EOS calls to routines with and without expansion coefficients
   - merged some loops to reduce arrays to scalars
   - moved tracer modification to separate loop for optimization
   - changed loop limits to fixed nVertLevels rather than maxLevelCell
   - pushed extraction of pool variables up one call level
   - added OpenACC directives to enable computations on GPU
   - convert JM allowable T,S range to parameters
   - introduce enums for EOS types to reduce string manipulation
   - changed scalar pointers to scalar variables
   - corrections and updates to a number of comments
   - general code cleanup

### DIFF
--- a/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
+++ b/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
@@ -102,7 +102,7 @@ module ocn_analysis_mode
       call mpas_timer_stop('io_reset_alarms')
 
       ! Initialize submodules before initializing blocks.
-      call ocn_equation_of_state_init(err_tmp)
+      call ocn_equation_of_state_init(domain, err_tmp)
       ierr = ior(ierr, err_tmp)
 
       call ocn_analysis_init(domain, err_tmp)
--- a/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
@@ -242,7 +242,7 @@ module ocn_forward_mode
       call ocn_vmix_init(domain, err_tmp)
       ierr = ior(ierr, err_tmp)
 
-      call ocn_equation_of_state_init(err_tmp)
+      call ocn_equation_of_state_init(domain, err_tmp)
       ierr = ior(ierr, err_tmp)
 
       call ocn_tendency_init(err_tmp)
--- a/src/core_ocean/mode_init/mpas_ocn_init_mode.F
+++ b/src/core_ocean/mode_init/mpas_ocn_init_mode.F
@@ -137,7 +137,7 @@ module ocn_init_mode
       call mpas_timer_stop('reset_io_alarms')
 
       ! Initialize submodules before initializing blocks.
-      call ocn_equation_of_state_init(err_tmp)
+      call ocn_equation_of_state_init(domain, err_tmp)
       ierr = ior(ierr, err_tmp)
       if(ierr.eq.1) then
           call mpas_log_write('An error was encountered while initializing the MPAS-Ocean init mode', MPAS_LOG_CRIT)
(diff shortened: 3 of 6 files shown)


philipwjones:
Will get those numbers today - most changes were actually b4b, but one of the loop changes caused a small difference from sqrt, so must have triggered a different version of the library call.

Regarding GPUs, we have a couple of different versions of the transport that are GPU-enabled, but we are still evaluating which version to use/commit. I've gotten pretty far along on vmix, but that's not quite ready yet either. We have other bits here and there. We added an OPENACC build flag to turn the GPU portions on, but those Makefile changes were not part of this commit since no one else has access to Summit and these changes will not be effective on Titan due to the higher transfer overhead on that machine.

philipwjones:
On Cori after one timestep:
Differences in field: density
    max  rel diff:   2.219540115956395E-015
    mean rel diff:   5.773551112341405E-017
Where relative difference is abs((density1 - density2)/density1)
Summit's been down most of the day, but can do something similar with GPU vs CPU.

mark-petersen:
@philipwjones, could you add a blank line for line 2 of your commit message?

If you use
```
git log --oneline
```
you will see that a lack of a blank line 2 makes the whole message appear.  You can use a `git commit --amend` and then `git push --force`.

mark-petersen:
A few notes for new code additions: Doug wrote a nice tool to help us write uniform code.  It is here:
https://github.com/MPAS-Dev/MPAS-Tools/tree/master/source_code_processing/mpas_source_linter
and can be run with:
```
linter -f file.F
```
it produces a new file that notifies you of 'delint' clean-up items, that sometimes cause a problem on odd compilers.  These can also be helpful for brand new code:
```
sed -i 's/    /\t/g' $file    # change space to tab, for *.xml files
sed -i 's/[ \t]*$//g' $file   # remove trailing white spaces
```
though it is best to clean up just the new code you are submitting, and not include white space alterations to lots of other code in this PR.

mark-petersen:
For something non-bfb like this, we need to make an E3SM PR with this updated MPAS submodule and test it.  I would say an EC60to30 G case for two years, compared to the current repo, would be fine.  @jonbob, this would be a nice task for Kristin to work on.   (As an aside, we also have new values for EC60to30 with dt=30 minutes and revised del2 & del4 from Mat that could be tested with that bunch).

@philipwjones I am merging a bunch of bfb compass updates to ocean/develop this week.  I'd like to merge this one after it is tested in E3SM.

philipwjones:
@mark-petersen - fixed the commit message. Also fixed an untyped constant that lint found and some trailing whitespace that lint also complained about.

mark-petersen:
@philipwjones I rebased and tested. Passes nightly regression suite on grizzly with gnu and intel, optimized and debug.  Tests with nonlinear EOS do not match bit-for-bit with ocean/develop.  Differ by max of 1e-9 after 6 time steps, which is expected.

Can I merge and make PR into E3SM?

philipwjones:
@mark-petersen - fine by me

mark-petersen:
@philipwjones, this did not pass the tests on E3SM next branch, so we had to revert this merge in order to proceed with other pull requests.  See comments here:
https://acmeclimate.slack.com/archives/C04B3QGEQ/p1547562515073700
https://acmeclimate.slack.com/archives/C04B3QGEQ/p1547570643077900

It passed my G-case tests on edison, but it's failing these: 

- PET_Ln9_PS.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel.allactive-mach-pet
- SMS_P12x2.ne4_oQU240.A_WCYCL1850.sandiatoss3_intel.allactive-mach_mods

PR #76, opened 2018-09-04
Support data iceberg fluxes from ICE in E3SM

matthewhoffman:
This merge introduces changes to the ocean model to support coupling of data iceberg fluxes from the sea ice component in E3SM.  First, new fields are added (icebergFreshWaterFlux and icebergHeatFlux) to use the new fluxes within MPAS-O.  Second, an option is added that will disable solid and liquid runoff from Antarctica coming from E3SM's runoff model.  (This option does nothing in standalone.)

The existing MPAS-O support for iceRunoffFlux (solid flux from the runoff model) is retained as we anticipate using a mixture of the two approaches for some time (specifically, solid discharge from the runoff model for GIS and iceberg melt flux from the sea ice model for AIS).

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -1676,10 +1676,12 @@
 			<var name="longWaveHeatFluxUp"/>
 			<var name="longWaveHeatFluxDown"/>
 			<var name="seaIceHeatFlux"/>
+			<var name="icebergHeatFlux"/>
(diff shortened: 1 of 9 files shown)


mark-petersen:
I removed the red labels.  If you all are happy with the behavior in E3SM, go ahead and merge this into ocean/develop, then merge ocean/develop into e3sm/develop, and point to that merge commit in E3SM mpas-source submodule.  Thanks!

PR #77, opened 2018-09-13
Use default initialization in framework DDTs to simplify registry-generated code

mgduda:
This merge adds default initialization for components of several framework DDTs 
to avoid the need to initialize these components explicitly in registry-generated
code (found in the `inc/structs_and_variables.inc` file). This results in
a significant reduction in the size of the registry-generated routines, which in 
turn translates into faster build times for MPAS core interface modules.

The use of default initialization is also arguably better practice, since it
eliminates the possibility of accidentally forgetting to initialize DDT
components when instantiating one of these types.

### DIFF
--- a/src/framework/mpas_attlist_types.inc
+++ b/src/framework/mpas_attlist_types.inc
@@ -10,8 +10,8 @@
 
    ! Derived type for holding field attributes
    type att_list_type
-      character (len=StrKIND) :: attName
-      integer :: attType
+      character (len=StrKIND) :: attName = ''
+      integer :: attType = -1
       integer :: attValueInt
       integer, dimension(:), pointer :: attValueIntA => null()
       real (kind=RKIND) :: attValueReal
--- a/src/framework/mpas_field_types.inc
+++ b/src/framework/mpas_field_types.inc
@@ -10,10 +10,10 @@
    type field5DReal
   
       ! Back-pointer to the containing block
-      type (block_type), pointer :: block
+      type (block_type), pointer :: block => null()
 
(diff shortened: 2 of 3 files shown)


mgduda:
To put some specific numbers to the reduction in generated code size and compile time, in the atmosphere core, we previously had 21732 lines of code in `inc/structs_and_variables.inc`, and the `mpas_atm_core_interface.F` file took about 1m42s to compile with the ifort compiler. After the changes in this PR, the `inc/structs_and_variables.inc` file is only 15422 lines and takes about 1m12s to compile.

For parallel builds (e.g., with `-j 4`), the `mpas_atm_core_interface.F` file still takes almost 20% of the total build time (1m12s out of a total build time of 5m31s).

mgduda:
@mark-petersen Pointers in Fortran can have three statuses: associated, unassociated, and undefined. In some of the framework code, we have lines like:
```
if (associated(cursor % next)) then
    ! some code here...
end if
```
If the `next` component of, e.g., a linked-list node type is not explicitly nullified (which sets its status to unassociated), or if the node type doesn't provide default initialization (what we're switching to in this PR), code like the above will have undefined behavior becuase the status of the `next` pointer is neither associated nor unassociated, but undefined (unless, of course, we can guarantee that some earlier code would have allocated the `next` pointer, giving it associated status).

We previous used the registry to generate explicit code to nullify many pointer components, which led to very long `inc/structs_and_variables.inc` files. Using default initialization for pointer components in our derived type definitions obviates the need for this code, and is probably safer practice anyway.

PR #82, opened 2018-09-14
MYNN Surface Layer and PBL Schemes Unified with WRFV4.0

kkeene44:
Toward the effort to unify physics schemes between the MPAS and WRF models, the MYNN surface and PBL schemes were modified to include changes necessary to allow them to run in both models. In areas where the models differ (e.g., in their handling of 'dx'), there are if-defs put around the code to ensure that it's handled correctly for the model in which it is running. Additionally, it was necessary to make some mods to the MPAS PBL driver (mpas_atmphys_driver_pbl.F) to incorporate some of the mods. Results before and after are NOT bit-for-bit, however, as this update includes some updates to the actual physics - in order to bring MPAS physics up-to-date with recently modified WRF physics.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
@@ -646,9 +646,9 @@
                  th       = th_p        , dz       = dz_p        , u        = u_p        , &
                  v        = v_p         , qv       = qv_p        , qc       = qc_p       , &
                  qi       = qi_p        , qni      = ni_p        , rho      = rho_p      , &
-                 du       = rublten_p   , dv       = rvblten_p   , dth      = rthblten_p , &
-                 dqv      = rqvblten_p  , dqc      = rqcblten_p  , dqi      = rqiblten_p , &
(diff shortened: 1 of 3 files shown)


ldfowler58:
Since these schemes are not part of the mesoscale reference suite, I suggest that we put these two modules on the side and delete the PRs. This will help keep a clean list of commits to the atmosphere/unified_physics branch.

mgduda:
@ldfowler58 Not merging these changes to `atmosphere/unified_physics` is fine with me. But, if we ultimately want these changes in the next release of MPAS, should we instead change the PR to target the `develop` branch?

ldfowler58:
I cannot think so far ahead and am currently focusing the atmosphere/unified_physics first. With the current timeline, it looks like the develop branch may be updated using WRF-4.1 which will certainly contains some changes in those 2 modules. The changes are not lost since they are on Kelly's local repository.

mgduda:
After some discussion, we'll close this PR for now. At such time as we update the full MPAS-A `convection_permitting` suite to a newer and consistent version of the WRF schemes, we may revisit the MYNN schemes.

PR #83, opened 2018-09-14
MM5/Monin-Obukhov Surface Layer Scheme Unified with WRFV4.0

kkeene44:
Toward the effort to unify physics schemes between the MPAS and WRF models, the MM5 (or Monin-Obukhov) surface layer scheme was modified to include changes necessary to allow it to run in both models. In areas where the models differ (e.g., in their handling of 'dx'), there are if-defs put around the code to ensure that it's handled correctly for the model in which it is running. Results before and after are bit-for-bit.


kkeene44:
@mgduda @ldfowler58 
I've made all the modifications that were suggested. All mods build/run okay in both MPAS and WRF.

ldfowler58:
After comparing the pull request again the original MPAS sourcecode, we are now missing the variables isftcflx to cda(ims,j) when using MPAS since those variables are bracketed inside #if (EM_CORE == 1).

I suggest that we use #if ( EM_CORE == 1 ) || defined(mpas) to include both WRF and MPAS.

kkeene44:
@ldfowler58 
The suggested modification has been made and pushed to the PR. I tested both WRF/MPAS to verify that they build/run correctly.

PR #85, opened 2018-09-17
YSU PBL physics unified with WRFV4.0

kkeene44:
Toward the effort to unify physics schemes between the MPAS and WRF models, the YSU PBL
scheme was modified to include changes necessary to allow it to run in both models. In areas where the models differ (e.g., in their handling of 'dx'), there are if-defs
put around the code to ensure that it's handled correctly for the model in which it is
running. Results before and after are NOT bit-for-bit, however, as this update includes
some updates to the actual physics - in order to bring MPAS physics up-to-date with recently
modified WRF physics.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
@@ -66,6 +66,9 @@
 ! * since we removed the local variable pbl_scheme from mpas_atmphys_vars.F, now defines pbl_scheme as a pointer
 !   to config_pbl_scheme.
 !   Laura D. Fowler (laura@ucar.edu) / 2917-02-16.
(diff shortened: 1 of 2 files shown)


weiwangncar:
I have two comments:
1. All of the code changes made in WRF relative to MPAS version are related to topo_wind option which MPAS doesn't support. Therefore the MPAS code should produce identical results, or at least close to identical results.
2. If MPAS is willing to remove the extra output arrays for the K coefficients, it will make the shared code a lot cleaner. Else is WRF willing to add those output arrays? Note also that one of them, exch_h and kzhout are duplicate.

weiwangncar:
I chatted with Jimy, and he agreed that we could output the K coefficients from WRF's YSU. Since WRF's variable names are exch_h and exch_m, can MPAS also adopt these names? The K for q is the same as for h, hence it seems it doesn't need the extra output for q.

kkeene44:
@ldfowler58 
I included the updated comment in the mpas_atmphys_driver_pbl.F file, per your request.

PR #86, opened 2018-09-21
Noah LSM physics unified with WRFV4.0

kkeene44:
Toward the effort to unify physics schemes between the MPAS and WRF
models, the Noah LSM scheme was modified and updated to include
changes necessary to allow it to run in both models. It was necessary
to include updated options (e.g., seaice, glacial) and to modify
several of the mpas_atmphys* files to allow it to build and run without
errors. The new code is comparable to the 2017 HWT branch, which was
proven to provide reasonable results for the HWT project. Results before
and after are not bit-fot-bit, which is to be expected, as the actual
physics were updated.

* A modified version of the VEGPARM.TBL file is necessary to run
with these updates.


src/core_atmosphere/Registry.xml
```diff
@@ -336,6 +336,18 @@
...
+                <nml_option name="config_max_median_time" type="real" default_value="1.0"
+                     units="s"
+                     description="Maximum allowable median time for a timestep, beyond which the model will terminate"
+                     possible_values="Positive real values"/>
+
+                <nml_option name="config_median_sample_size" type="integer" default_value="100"
+                     units="-"
+                     description="Number of timesteps that are sampled before computing median timestep time"
+                     possible_values="Positive integer values"/>
+        </nml_record>
```
mgduda: These `self_monitoring` namelist options were included to catch real-time HWT simulations that were progressing too slowly, and are not part of the changes to the Noah scheme.

src/core_atmosphere/Registry.xml
```diff
@@ -1348,6 +1379,10 @@
                              description="Graupel mixing ratio"
                              packages="mp_thompson_in;mp_wsm6_in"/>
 
+                        <var name="nc" array_group="number" units="nb kg^{-1}"
+                             description="Cloud water number concentration"
+                             packages="bl_mynn_in"/>
+
```
mgduda: This looks like something that is part of a different parameterization and not Noah. Can you verify this?

src/core_atmosphere/Registry.xml
```diff
@@ -1847,6 +1882,36 @@
...
+                <nml_option name="config_gfconv_closure_midlc" type="integer" default_value="0" in_defaults="false"
+                     units="-"
+                     description="closure option for mid-level convection in Grell-Freitas convection scheme"
+                     possible_values="0"/>
+
+                <nml_option name="config_gfconv_closure_shalc" type="integer" default_value="0" in_defaults="false"
+                     units="-"
+                     description="closure option for shallow convection in Grell-Freitas convection scheme"
+                     possible_values="0"/>
+
```
mgduda: The names of all of these suggest they are part of the GF scheme, and not Noah.

src/core_atmosphere/Registry.xml
```diff
@@ -1857,6 +1922,16 @@
...
+                <nml_option name="config_mynn_cloudpdf" type="integer" default_value="-1" in_defaults="false"
+                     units="-"
+                     description="switch to different cloud PDFs to represent subgrid clouds"
+                     possible_values="0,1,2"/>
+
+                <nml_option name="config_mynn_edmf" type="integer" default_value="0" in_defaults="false"
+                     units="-"
+                     description="option to activate the PBL eddy diffusivity mass flux scheme"
+                     possible_values="0,1"/>
+
```
mgduda: These look like they belong to the set of changes to the MYNN scheme, and not Noah.

src/core_atmosphere/Registry.xml
```diff
@@ -1891,6 +1966,11 @@
                      units="s"
                      description="Relaxation time to initial values in 1-d OML"
                      possible_values="Non-negative real values"/>
+
+                <nml_option name="config_lightning_flash" type="logical" default_value="false" in_defaults="false"
+                     units="-"
+                     description="Turns on/off the calculation of storm parameters to compute lightning flash rates"
+                     possible_values=".true. or. false"/>
```
mgduda: I think the lightning flash rate diagnostics were something we added to the HWT runs, but these should not be included in the changes for Noah.

src/core_atmosphere/Registry.xml
```diff
@@ -1971,6 +2051,22 @@
...
+                     packages="mp_thompson_in"/>
+
+                <var name="vqs" type="real" dimensions="nVertLevels nCells Time" units="m s^{-1}"
+                     description="mass-weighted mean velocity for snow"
+                     packages="mp_thompson_in"/>
+
+                <var name="vqg" type="real" dimensions="nVertLevels nCells Time" units="m s^{-1}"
+                     description="mass-weighted mean velocity for graupel"
+                     packages="mp_thompson_in"/>
+
```
mgduda: The package attached to these suggests they are needed by the Thompson scheme, and not Noah.

src/core_atmosphere/Registry.xml
```diff
@@ -2113,6 +2209,14 @@
...
 
                 <!-- MYNN PBL SCHEME: -->
+                <var name="cldfrac_bl" type="real" dimensions="nVertLevels nCells Time" units="unitless"
+                     description="Cloud fraction in the Planetary Boundary Layer"
+                     packages="bl_mynn_in"/>
+
+                <var name="qc_bl" type="real" dimensions="nVertLevels nCells Time" units="kg kg^{-1}"
+                     descrition="Cloud liquid water mixing ratio in the Planetary Boundary Layer"
+                     packages="bl_mynn_in"/>
+
```
mgduda: Are these needed by MYNN PBL scheme instead of Noah?

src/core_atmosphere/Registry.xml
```diff
@@ -2161,6 +2265,30 @@
...
+                <var name="edmf_thl" type="real" dimensions="nVertLevels nCells Time" units="K"
+                     description="EDMF liquid water potential temperature - moist updrafts"
+                     packages="bl_mynn_in"/>
+
+                <var name="edmf_ent" type="real" dimensions="nVertLevels nCells Time" units="m^{-1}"
+                     description="EDMF entrainment - moist updrafts"
+                     packages="bl_mynn_in"/>
+
```
mgduda: These variables, too, look like they belong with the MYNN scheme.

src/core_atmosphere/physics/mpas_atmphys_vars.F
```diff
@@ -250,7 +263,11 @@ module mpas_atmphys_vars
     recloud_p,        &!
     reice_p,          &!
     resnow_p,         &!
-    refl10cm_p         !
+    refl10cm_p,       &!
+    vqr_p,            &!mass-weighted mean velocity for rain                                                 [m/s]
+    vqi_p,            &!mass-weighted mean velocity for cloud ice                                            [m/s]
+    vqs_p,            &!mass-weighted mean velocity for snow                                                 [m/s]
+    vqg_p              !mass-weighted mean velocity for graupel                                              [m/s]
```
mgduda: Are these needed by Noah?

src/core_atmosphere/physics/mpas_atmphys_vars.F
```diff
@@ -361,14 +374,48 @@ module mpas_atmphys_vars
...
+!                                               != 2: for TEMF.
+ integer,parameter:: bl_mynn_edmf_mom  = 0      !option to activate momentum transport (bl_mynn_edmf >0).
+                                                != 0: option is turned off.
+                                                != 1: option is turned on.
+ integer,parameter:: bl_mynn_edmf_tke  = 0      !option to activate TKE transport (bl_mynn_edmf >0).
+                                                != 0: option is turned off.
+                                                != 1: option is turned on.
+ integer,parameter:: bl_mynn_edmf_part = 0      !option activate areal partitioning of ed and mf.
+                                                != 0: option is turned off.
+                                                != 1: option is turned on.
```
mgduda: This all looks like code for a PBL scheme.

src/core_atmosphere/Registry.xml
```diff
@@ -442,6 +442,8 @@
 			<var name="isltyp"/>
 			<var name="ivgtyp"/>
 			<var name="mminlu"/>
+			<var name="isice_lu"/>
+			<var name="iswater_lu"/>
```
mgduda: Are there additional changes needed in the `init_atmosphere` core to produce the `isice_lu` and `iswater_lu` fields?

kkeene44: @mgduda 
I'm not sure about this. As I've modified it, everything compiles/runs okay, but maybe @ldfowler58 would have more insight?

kkeene44:
@mgduda 
Per the offline conversation @ldfowler58 and I had, I've modified the files we agreed on. The code changes should now reflect those updates.

kkeene44:
@mgduda @ldfowler58 
I verified again that (even when built with full optimization), we get bit-for-bit results prior to the mods and after the mods in the WRF model. This means that changing the variables from OUT to INOUT does not affect WRF negatively and we can proceed on the MPAS side.

kkeene44:
@mgduda @ldfowler58 
I committed the agreed-upon changes to modify some of the MPAS intent(inout) variables to intent(out). I verified the code still builds okay.

mgduda:
The PR description mentions that an updated `VEGPARM.TBL` file is needed. From which version of WRF should we get this `VEGPARM.TBL` file?

kkeene44:
@

kkeene44:
> The PR description mentions that an updated `VEGPARM.TBL` file is needed. From which version of WRF should we get this `VEGPARM.TBL` file?

@mgduda 
The version I used for testing MPAS is identical to the file in the tagged V4.0.3 of WRF.

mgduda:
Without additional changes, I think we'll run into problems with the changes in this PR when testing with MODIS land-cover data. Since the `isice_lu` and `iswater_lu` variables will not be present in any initial conditions files, their default values, which are only appropriate for USGS land cover, will be used.

mgduda:
@ldfowler58 @kkeene44 Would it be feasible to just revert the changes to `landuse_init_forMPAS` where the logic to set `iswater_lu` and `isice_lu` based on `mminlu` was removed? Then, we wouldn't need to add `iswater_lu` and `isice_lu` to the `Registry.xml` file or to initial conditions files.

ldfowler58:
There are differences in the files LANDUSE.TBL, SOILPARM.TBL, and VEGPARM.TBL currently used in MPAS and those from WRF-4.0.3. Since we are updating the NOAH LSM scheme to that in WRF-4.0.3, we should also update LANDUSE.TBL, SOILPARM.TBL, and VEGPARM.TBL in the tagged V4.0.3 of WRF.

kkeene44:
@mgduda
I have implemented and committed the changes that @ldfowler58 made to the necessary files, regarding iswater_lu and isice_lu.

mgduda:
@ldfowler58 @kkeene44 Thinking a little more about this, I'm wondering whether we should not include the changes to the `init_atmosphere` core that introduce new fields `isice_lu` and `iswater_lu` into the model initial conditions, and instead, maybe we should just continue to set `isice` and `iswater` inside `landuse_init_forMPAS`? I don't think I completely understand why the setting of `isice_lu` and `iswater_lu` are being moved to the `init_atmosphere` core; and, why were we not doing the same for `isurban`?

ldfowler58:
In my view, now that we have two options for config_landuse_data, my main motivation is to avoid having to define isice_lu and iswater_lu in core_init_atmosphere and core_atmosphere, separately.

1. In the MPAS release (MPAS v6.2), isice_lu and iswater_lu are defined twice, once in mpas_init_atm_static.F and once in mpas_atmphys_initialize_real.F. This is cumbersome.
When building the initial conditions, we always call the subroutine physics_initialize_real after we call the subroutine init_atm_static, so having both variables in the static file makes them readily available.

2. In the MPAS release (MPAS v6.2) isice_lu and iswater_lu are also defined in the subroutine landuse_init_forMPAS and used in the subroutine physics_update_sst. Again, having the two variables in the initial conditions makes them readily available. It also may lead to less errors since they are only defined in one spot for the whole physics.

3. To date, we do not have plans to run the urban lsm physics. That is the only reason why I did not add it to the initial conditions. I think that this is consistent with WRF is doing where isice_lu and iswater_lu are only defined in one place.

ldfowler58:
Note that in this case, we also need to update mpas_atmphys_update_surface.F in this branch.

mgduda:
@ldfowler58 Thanks for the explanation. I agree that setting `isice` and `iswater` in one place only is much cleaner.

kkeene44:
@mgduda @ldfowler58 
I've now added in the module_sf_bem file, and made the changes to the noahdrv file to update to V4.0.3 of WRF. Testing can now begin! ❇️

mgduda:
@kkeene44 I think we're missing changes to add `isice_lu` and `iswater_lu` to the `src/core_init_atmosphere/Registry.xml` file. Without these additions to the init_atmosphere core's `Registry.xml` file, the code will segfault.

kkeene44:
@mgduda 
I committed the modified src/core_init_atmosphere/Registry.xml file, but there are conflicts. I'm afraid to mess with those because I  don't think they are completely specific to this PR.

mgduda:
@kkeene44 The conflict will be easy to resolve when merging the branch. @MiCurry added `units` and `description` attributes to all fields in the init_atmosphere `Registry.xml` file in PR #119 , so we just want to add `isice_lu` and `iswater_lu`; when merging, I'll add `units` and `description` attributes for these as well.

mgduda:
@ldfowler58 @kkeene44 When I run with different MPI task counts, the `smstav` field is different between the two runs. This is the only field that differs, and it does appear to be a diagnostic.

mgduda:
The differences appear to only occur over snow/ice land use. I'll keep searching to see what I can uncover.

ldfowler58:
Do you want me to look more specifically as why smstav is diferent with different MPI task counts. Although it is diagnostic only, it should not be? Let me know.

mgduda:
@ldfowler58 I'll spend a little more time seeing if I can find the source of the differences over snow/ice. If I'm not able to turn up anything today, any help would be much appreciated!

mgduda:
I think the issue may be that the initialization of `SOILW` at [L.673](https://github.com/MPAS-Dev/MPAS-Model/blob/bfab9d1073bd2aa681cb99fcdbdde071c17302d0/src/core_atmosphere/physics/physics_wrf/module_sf_noahdrv.F#L673) of `module_sf_noahdrv.F` needs to be moved inside the I-loop at [L.739](https://github.com/MPAS-Dev/MPAS-Model/blob/bfab9d1073bd2aa681cb99fcdbdde071c17302d0/src/core_atmosphere/physics/physics_wrf/module_sf_noahdrv.F#L739). It's possible that the other initializations just above that of `SOILW` should be moved as well. I'll do some more testing, and then we can check with the Noah developers to see whether they agree.

ldfowler58:
You are probably correct since the variables FLX4, FVB, FBUR, FGSN, and SOILW are only needed in the subroutine sflx for land points. With a differ. All 4 variables may need to be initialized inside the ILOOP.

mgduda:
Just an update for posterity: the current plan is to first merge PR #193 , then to rebase this PR and merge it after.

PR #92, opened 2018-10-09
Bug fix to the seaice warning system.

njeffery:
Corrects the passing of warnings from the column package.
In stand alone runs on IC machines (grizzly), column warnings failed to write to the sea ice log.

### DIFF
--- a/src/core_seaice/column/ice_warnings.F90
+++ b/src/core_seaice/column/ice_warnings.F90
@@ -28,7 +28,7 @@ contains
 
   subroutine add_warning(warning)
 
-    character(len=*), intent(in) :: &
+    character(len=char_len_long), intent(in) :: &
          warning ! warning to add to array of warnings
 
     ! number of array elements to increase size of warnings array if that array has run out of space


akturner:
Why does the previous implementation not work??

njeffery:
I'm not sure.  I compiled on grizzly with DEBUG=true.  The warnings failed to write when I removed the specific conditional.  Even the abort flag failed to signal an abort.  But after this change, warnings and aborts started working.


n



akturner:
What do you mean by specific conditional?

njeffery:
Warnings are in a conditional loop.  So I commented out the if end if to trigger a warning.

n

akturner:
Do you get the same problem if you set config_full_abort_write=.false.?


Cheers

Adrian

njeffery:
With the flag false and the original code base, the abort works and there's a short error message with just the first flux error warning. I assume that's what the flag is supposed to do.


n

akturner:
So, is the code working as you would hope it would with this flag false?

njeffery:
If a false flag is intended to only write one warning message, then yes, it appears to work properly.

akturner:
No, that's not what that does in any way.

akturner:
So, this cant be merged without a proper understanding of what the root issue is and why this fixes it.

njeffery:
That's fine.  PR #93 appears to fix the problem in a more complete way.

akturner:
@njeffery: Can this PR be closed now?

njeffery:
yes

PR #98, opened 2018-10-17
bgc fesedflux bug

mark-petersen:
Remove factor of 5, mistakenly left in code.


mark-petersen:
Passed compile and nightly regression suite.  @maltrud please view and hit 'approve'

maltrud:
thanks for handling this, @mark-petersen

jonbob:
@mark-petersen - is there any reason you couldn't have made a clean update to the mpas source with just this ocean bug fix? All tests in e3sm that include MAIL fail with this submodule.

PR #99, opened 2018-10-18
Add ice shelf melt param. based on ocean temp. profile

matthewhoffman:
This merge adds a new ice shelf basal melt parameterization based on a prescribed ocean temperature profile, developed by Xylar Asay-Davis.  The parameterization takes a ocean temperature profile and sill depth as input. The temp. profile is adjusted for cavity temperature behind the sill, assuming constant temp. below the sill depth. Then a local temp profile in the boundary layer is estimated assuming exponential decay from the cavity profile to a plume profile at infinite distance downstream from the grounding line, accounting for the pressure-melt temperature. The melt rate is calculated as proportional to the  product of the boundary layer temperature profile and the mean cavity temperature.  

This merge includes some reorganization of the ice shelf melt set of code to streamline the addition of the new parameterization.  It also includes updates to the Thwaites test case that complete the simulations and analysis for the paper "Effect of subshelf melt variability on sea level rise contribution from Thwaites Glacier, Antarctica".


xylar:
I guess an "ice shelt" is half way between an "ice sheet" and an "ice shelf"?

darincomeau:
> I guess an "ice shelt" is half way between an "ice sheet" and an "ice shelf"?

or a contraction of "ice shelf" and "melt"?

xylar:
If so, I don't think it's going to catch on.

matthewhoffman:
@xylar , as of 63e80d3 the MALI param. (left) matches your python script (right) .
screen shot 2018-10-18 at 8 05 43 pm
I still need to calculate a global average of the GL depth (right now it's hardcoded) and fix a few other cleanup items, as well as introduce the ability to add variability to the temperature profile depth.  I'll ask for a code review when all the pieces are together, but wanted to let you know that it is largely functional now.  Follow-up commit  26d95a2 makes a few modifications, including changing the depth range over which TFmean is calculated.  This slight adjustment makes more sense to me, but we can discuss further.
(cc: @stephenprice )

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -0,0 +1,690 @@
...
+         TSill = (TThermoclineTop - TThermoclineBot) * &
+                 (zSill - zThermoclineBot) / (zThermoclineTop - zThermoclineBot) + zThermoclineBot
+      endif
+      if (print_debug) then
+         call mpas_log_write("TSill=$r at zSill=$r", realArgs=(/TSill, zSill/))
+      endif
+
+      ! calculate temperature at GL (allowing for no sill - sill deeper than GL depth)
+      if (zGL < zSill) then
+         TGL = TSill
```
xylar: In theory, if the grounding line is above the sill, I would think you would want to keep TGL as it is, wouldn't you?  Deeper, warmer water isn't going to flow "uphill" to the grounding line.  I believe, for example, this is the case for Abbot (at least in some places).  It's a surprisingly cold-cavity ice shelf for its region because the GL is so shallow.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -0,0 +1,690 @@
...
+      zThermoclineBot = -600.0_RKIND
+      TThermoclineTop = -1.0_RKIND
+      TThermoclineBot = 1.2_RKIND
+
+      ! Calculate variables independent of depth
+      kappa = 6.0_RKIND ! likely to become nl option
+      zSill = -700.0_RKIND ! likely to become nl option
+      zGL = -1000.0_RKIND ! to come from global reduction of GL depth
+      zeta = slopeDraft * plumeThickness / (E0 * slopeDraft + Stanton)
+      zCavityDepth = -200.0_RKIND ! depth below which mean thermal forcing is averaged
```
xylar: I would call this something like `zMaxCavity` to make clear that it is a limit.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
xylar: Sorry, I mixed up the sign of z, I think.  This now makes sense to me.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -0,0 +1,690 @@
...
+      real (kind=RKIND), dimension(:), intent(out) :: floatingBasalMassBal !< BMB for ice shelves
+      integer, intent(out) :: err
+
+
+      !-----------------------------------------------------------------
+      ! local variables
+      !-----------------------------------------------------------------
+      ! Some constants - could be made into namelist parameters as needed
+      real(kind=RKIND), parameter :: slopeDraft = 1.0e-2_RKIND  !< characteristic slope of ice shelf draft
+      real(kind=RKIND), parameter :: plumeThickness = 30.0_RKIND  !< characteristic plume thickness (m)
```
xylar: I would think `slopeDraft` and `plumeThickness` should eventually become namelist options if `kappa` is a namelist parameter.  `plumeThickness` in particular might be something to tune to observations.  I guess the point is that I wouldn't make these two constants in the same sense as `E0` and `Stanton`, which really are intended to be universal constants.

xylar:
Looks very good so far!  I'll keep an eye out for your modifications to the mean thermal forcing but everything looks as expected so far.  My comments above (where they're not completely wrong) are very minor.

PR #102, opened 2018-10-22
Add variable-resolution gm bolus parameter

mark-petersen:
Currently the GM bolus parameter is a constant.  This adds a ramp function that depends on dcEdge.  Constant remains the default, so this is a bit-for-bit change.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -2583,6 +2583,10 @@
 			 description="GM stream function"
 			 packages="forwardMode;analysisMode"
 		/>
+		<var name="gmBolusKappa" type="real" dimensions="nEdges" units="m^2 s^{-1}"
+			 description="GM Bolus Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
+			 packages="gm"
+		/>
 		<var name="surfaceFluxAttenuationCoefficient" type="real" dimensions="nCells Time" units="m"
 			description="The spatially-dependent length scale of exponential decay of surface fluxes. Fluxes are multiplied by $e^{z/\gamma}$, where this coefficient is $\gamma$."
 		/>
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -395,22 +395,22 @@
 					description="Control what type of function is used for GM Bolus kappa"
 					possible_values="'constant' and 'ramp'"
 		/>
-		<nml_option name="config_standardGM_tracer_kappa" type="real" default_value="0.0" units="m^2 s^{-1}"
+		<nml_option name="config_standardGM_tracer_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$. Used when config_GM_Bolus_kappa_function is set to constant."
 		/>
 		<nml_option name="config_GM_Bolus_kappa_min" type="real" default_value="0.0" units="m^2 s^{-1}"
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$, min value in ramp as a function of dcEdge. Used when config_GM_Bolus_kappa_function is set to ramp."
 					possible_values="Any positive real value."
 		/>
-		<nml_option name="config_GM_Bolus_kappa_max" type="real" default_value="0.0" units="m^2 s^{-1}"
+		<nml_option name="config_GM_Bolus_kappa_max" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$, min value in ramp as a function of dcEdge. Used when config_GM_Bolus_kappa_function is set to ramp."
 					possible_values="Any positive real value."
 		/>
-		<nml_option name="config_GM_Bolus_cell_size_min" type="real" default_value="0.0" units="m"
+		<nml_option name="config_GM_Bolus_cell_size_min" type="real" default_value="20000.0" units="m"
 					description="Minimum value in grid cell size for GM $kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_GM_Bolus_kappa_function is set to ramp."
 					possible_values="Any positive real value."
 		/>
-		<nml_option name="config_GM_Bolus_cell_size_max" type="real" default_value="0.0" units="m"
+		<nml_option name="config_GM_Bolus_cell_size_max" type="real" default_value="30000.0" units="m"
 					description="Maximum value in grid cell size for GM $kappa$ ramp function.  Here cell size refers to dcEdge"
 					possible_values="Any positive real value."
 		/>
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -1223,6 +1223,7 @@
 		<package name="initMode" description="This package controls variables that are intended to be used within the init run mode of the ocean model."/>
 		<package name="cullCells" description="This package controls variables that provide information on what cells should be culled."/>
 		<package name="tracerBudget" description="This package controls variables that record the tracer budget."/>
+		<package name="gm" description="This package is for GM variables, which are only needed at lower resolutions."/>
 	</packages>
 
 	<streams>
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -406,11 +406,11 @@
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$, min value in ramp as a function of dcEdge. Used when config_GM_Bolus_kappa_function is set to ramp."
 					possible_values="Any positive real value."
 		/>
-		<nml_option name="config_GM_Bolus_cell_size_min" type="real" default_value="20000.0" units="m"
+		<nml_option name="config_GM_Bolus_cell_size_min" type="real" default_value="20e3" units="m"
 					description="Minimum value in grid cell size for GM $kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_GM_Bolus_kappa_function is set to ramp."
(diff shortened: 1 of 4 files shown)


xylar:
@mark-petersen, I think this is great!  But I want to caution that `dcEdge` varies in space quite a lot, even on a uniform mesh (in this example, on the QU120 mesh, between 0.0152367 and 0.0199597, or about 25%).  This is especially problematic around pentagons but it's really a pretty ubiquitous feature even elsewhere.  So your bolus coefficient might have a strong spatial pattern like dcEdge has, see below:

dcedge_uq120

pwolfram:
Could an effective edge scale be build from averaging the cell areas on each edge?  I think this would produce a smoother field.

mark-petersen:
Good point.  I could also compute a length scale from the average areaCell on each side of the edge.  That would help.  Yes, I see @pwolfram just said the same thing.  I am running a test right now, will post an image in a bit.

xylar:
A smoother solution would be something like
```
edgeSize = sqrt(0.5*(areaCell(cellOnEdge1) + areaCell(cellOnEdge2)))
```
sqrt_areacell_qu120

xylar:
So we all came up with the same solution, and I posted it without bothering to read the previous comments...  But, yeah, now you have a picture...

pwolfram:
Glad to see we are on the same page-- thanks for sharing the visual @xylar!

mark-petersen:
**Left:** before, using dcEdge
**Right:** after, using areaCell
using SO60to15, with
```
    config_GM_Bolus_kappa_function = 'ramp'
    config_standardGM_tracer_kappa = 600.0
    config_GM_Bolus_kappa_min = 0.0
    config_GM_Bolus_kappa_max = 600.0
    config_GM_Bolus_cell_size_min = 35e3
    config_GM_Bolus_cell_size_max = 55e3
```

image

Obviously much better.  I changed the code to produce the one on the right.  Thanks for your input. 

I would say this serves to show that the variable gmBolusKappa is initialized correctly.

Here is a revised image:
variablebolusgmcoeff_areacell
[variableBolusGMCoeff_areaCell.pdf](https://github.com/MPAS-Dev/MPAS-Model/files/2502446/variableBolusGMCoeff_areaCell.pdf)

mark-petersen:
This passes nightly regression suite with gnu and intel, dubug and optimized, on wolf, with bfb match to previous.

@xylar or @pwolfram if either of you could do a five minute visual inspection, that would be great.  I am confident code is sufficiently tested for functionality.  

@milenaveneziani will probably help with test the operation in E3SM.

src/core_ocean/Registry.xml
```diff
@@ -391,8 +391,27 @@
 					description="Vertical component of Redi mixing limited in bottom config_Redi_surface_layer_tapering_depth"
 					possible_values="Any positive real value, typically between 100 and 500 m."
 		/>
-		<nml_option name="config_standardGM_tracer_kappa" type="real" default_value="0.0" units="m^2 s^{-1}"
-					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$"
+		<nml_option name="config_GM_Bolus_kappa_function" type="character" default_value="constant" units="unitless"
+					description="Control what type of function is used for GM Bolus kappa"
+					possible_values="'constant' and 'ramp'"
+		/>
+		<nml_option name="config_standardGM_tracer_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
```
xylar: Having non-zero GM kappa doesn't mean GM is necessarily on, right?  I just want to make sure we're not making GM on by default with this change.

src/core_ocean/Registry.xml
```diff
@@ -391,8 +391,27 @@
...
+		<nml_option name="config_GM_Bolus_kappa_max" type="real" default_value="600.0" units="m^2 s^{-1}"
+					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$, min value in ramp as a function of dcEdge. Used when config_GM_Bolus_kappa_function is set to ramp."
+					possible_values="Any positive real value."
+		/>
+		<nml_option name="config_GM_Bolus_cell_size_min" type="real" default_value="20e3" units="m"
+					description="Minimum value in grid cell size for GM $kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_GM_Bolus_kappa_function is set to ramp."
+					possible_values="Any positive real value."
+		/>
+		<nml_option name="config_GM_Bolus_cell_size_max" type="real" default_value="30e3" units="m"
+					description="Maximum value in grid cell size for GM $kappa$ ramp function.  Here cell size refers to dcEdge"
```
xylar: Do we think that 20 to 30 km is the range over which we want GM to turn on in practice?  That seems surprisingly narrow to me but I could be wrong.  The deformation radius around Antarctica is tiny -- order 5 to 9 km -- so maybe I have a skewed perspective.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -730,6 +753,48 @@ subroutine ocn_gm_init(err)!{{{
...
+            !$omp do schedule(runtime)
+            do iEdge = 1, nEdges
+               gmBolusKappa(iEdge) = config_standardGM_tracer_kappa
+            end do
+            !$omp end do
+         else if (config_GM_Bolus_kappa_function == 'ramp') then
+            sqrtPiInv = 1.0_RKIND / sqrt(pii)
+            !$omp do schedule(runtime)
+            do iEdge = 1, nEdges
+               avgCellDiameter = sqrtPiInv*(sqrt(areaCell(cellsOnEdge(1,iEdge))) + sqrt(areaCell(cellsOnEdge(2,iEdge))))
```
xylar: I guess this seems slightly weird to me, since our cells aren't circular.  But it's probably a fractional difference.

milenaveneziani:
> are we planning to make the GM kappa a function of cell size and local deformation radius (either a simple function of latitude or a function of stratification and Coriolis parameter) down the road?

@xylar: we talked about this with Mark and one reason why Mark thought of keeping it simple was to reduce mesh-dependent computations.

src/core_ocean/shared/mpas_ocn_gm.F
mark-petersen: That is a common way to approximate a cell diameter.  The circle would be between the inscribed and circumscribed circles.
image

src/core_ocean/Registry.xml
mark-petersen: Good point. I didn't choose it too carefully.  The span of the ramp is something we need to test.

I thought that, if 1/3 degree needs GM on, but 1/10 has GM off, then switch should be somewhere between 10 and 30 km. See more discussion below.

milenaveneziani: Short of doing a local Rossby Radius dependency, I think that this is the best compromise (20-30 km range). For example, the metoffice is not using GM at all for their latest 0.3deg simulations, as far as I understand. I agree that this is not ideal for the high-latitudes, but seems good for these simplified settings.
@maltrud: what do you think about this?

milenaveneziani: This is definitely something that we'll need to test. The idea is to help testing along using the E3SM-ARRM (Arctic refined) configuration, where the lower resolution grid is EC60to30 in the Southern hemisphere and 30to10 in the Northern hemisphere. This grid has already been run for ~10 years or so without GM.

mark-petersen:
In this PR I was thinking of how to control GM-Bolus for regional variable-resolution meshes, so you would have GM on for a broad swath of the low-resolution global ocean.  @xylar you are right that we could make it also depend on the local Rossby Radius, like this map:
image
from Chelton, D. B., R. A. deSzoeke, M. G. Schlax, K. El Naggar and N. Siwertz, 1998: Geographical variability of the first-baroclinic Rossby radius of deformation. J. Phys. Oceanogr., 28, 433-460.
[ftp://numbat.coas.oregonstate.edu/pub/chelton/papers/chelton_etal_1998_jpo.pdf](url)

or maybe just a simpler function of latitude.  I think that is more complexity than we want to deal with at this point.  Since the current implementation has options
```
config_GM_Bolus_kappa_function = 'constant'
config_GM_Bolus_kappa_function = 'ramp'
```
it would be straightforward to add something like
```
config_GM_Bolus_kappa_function = 'RossbyRadiusDependant'
```
and have GM turn on with a ramp above c*RR, where c might be 2.

xylar:
> one reason why Mark thought of keeping it simple was to reduce mesh-dependent computations, at least for the time being.

Perhaps I misunderstood.  I would think that having Rossby-radius dependence would allow you to have a set of coefficients that are appropriate for almost any mesh.  GM is off when the RR is kMin or more times the local resolution but gradually turns on as resolution gets coarser until it's fully on when RR is only kMax times the local resolution, where kMin and kMax would need to be tuned once but then would hopefully be constant across meshes and latitudes.  As @mark-petersen says, there are a lot of options about how you treat the RR -- a fn. of lat. only; a constant-in-time map; computed from the simulation -- but my intuition would be that any of these would lead to better GM transition than a RR-unaware approach.

mark-petersen:
@jonbob, I wanted to make sure there was a bit-for-bit match before/after this change, with the default flags in E3SM.  I compared ocean/develop 95042e2 against this branch 60e0d01 with these on all:
```
    config_use_standardGM = .true.
    config_standardGM_tracer_kappa = 600.0
```
and on this branch, the default for the new flag:
```
config_GM_Bolus_kappa_function = 'constant'
```

I get bfb match using both gnu and intel on grizzly.  I am confident this additional will be bfb using the standard flags in E3SM.

mark-petersen:
@milenaveneziani, it would be good to have a five-minute review by you. If you could just read the flags listed at the top of this PR, please approve if you agree with the design.  Thanks!

mark-petersen:
Rebased on head of ocean/develop.  Tested one more time, bfb match with ocean/develop on grizzly intel using debug flag.

PR #103, opened 2018-10-31
Increase AM dimension from 32 to 36

mark-petersen:
This is a bug fix for running with additional flux fields in two analysis members.


mark-petersen:
This bug made a CMIP-6 test on next fail.  The correction here passed the test.  Thanks @jonbob.  Merging in.

PR #104, opened 2018-11-05
Add NAQU60to30cr8 mesh

sbrus89:
This PR adds the NAEC60to30cr8 test case to COMPASS, which is the US North American refined coastlines plus Labrador Sea, where the background mesh is EC60to30 and the refined part is 8 km.  This includes two COMPASS cases, init and spin-up. Init uses jigsaw and adds initial conditions.  Spin up has six steps with a sequence of increasing dt and decreasing Rayleigh drag values.


pwolfram:
Produces mesh with 8km (yellow) refined resolution for NA
image

pwolfram:
Discussion about mesh at https://acme-climate.atlassian.net/wiki/spaces/ATM/pages/831620987/Regionally+Refined+North+American+Grids

pwolfram:
@sbrus89, based on discussions at the E3SM meeting today (https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/879656961/2018-11-09+Meeting+notes+-+Water+Cycle), I think we are ok for now (cc @maltrud).  We can follow up next week more to discuss any minor tweaks that may be needed.

sbrus89:
Updated to include West Central America coast...

resolution

mark-petersen:
@sbrus89 I added this last commit to fix those specs on the date line on the mesh density variable.

mark-petersen:
@sbrus89 I rebased onto the head of ocean/develop to make sure all my testing is up to date.  When you start again, you should fetch and then rebase your local branch on this remote one for those recent changes you added.

mark-petersen:
I squashed some of mine and Steve's recent commits.  Merged into ocean/develop locally and retested.

pwolfram:
@mark-petersen, does this mean we are now ready to merge and spin-up the E3SM case?

maltrud:
@sbrus89 , what is the bathymetry dataset that is being used?  i'm not familiar with it.

xylar:
> so we only import those libraries if they are needed (i.e. depending on an option flag)? If they are always needed, that's OK.

@mark-petersen, best practices in python are to import libraries at the top of python files without extra logic to the extent possible.  My understanding is that this actually leads to more efficient execution in general.  Part of what the PEP8 standard checks for is that imports happen before anything else.  I couldn't find these particular imports but I suspect the same should apply here.

The exception would be if the logic is checking whether the package even exists, in which case that's reasonable.

sbrus89:
@maltrud, The dataset I'm using to find the coastline contours is called SRTM 15 plus. It's a 1/2 degree global resolution and is availiable here: https://topex.ucsd.edu/WWW_html/srtm30_plus.html (first link at the top)

At this point, it's not being used in init mode to set the bottom depth variable because the file is too large to be read in and interpolated without running out of memory. We have a script called inject_bathymetry.py that does the interpolation in chunks to get around this.

sbrus89:
@mark-petersen, Somehow the file lost the climate group permissions. Sorry about that; I went ahead and set it back. I tried to copy the `/users/sbrus/climate/bathy_data/SRTM15_plus/earth_relief_15s.nc` file to `/usr/projects/regionalclimate/COMMON_MPAS/ocean/topo` and got a permissions error. I think because I don't have write permissions to the `topo` directory. For now I copied it to `/usr/projects/regionalclimate/COMMON_MPAS/ocean`. Would you be able to move it into `topo` for me? Thanks!

sbrus89:
@mark-petersen, Unfortunately, those two python libraries are always needed. I believe they were added to e3sm-unified via this issue: https://github.com/E3SM-Project/e3sm-unified/issues/25

mark-petersen:
@sbrus89 and @pwolfram thanks for your work on this.  A standardized and tested CONUS mesh is a big help to E3SM.

PR #106, opened 2018-11-06
Add support for MPAS as a lib used in an external program

jjguerrette:
This is the first of several PR's for interfacing MPAS with the JEDI data assimilation framework. 
 There are multiple modifications to enable an external program to call top-level MPAS subroutines as part of a shared library, including  (1) the MPI communicator can be initialized and destroyed outside of MPAS, (2) multiple instances of the domain and corelist can be used, and (3) the mpas_log_info pointer is associated with the appropriate domain % logInfo before it is used in lower level subroutines.

The new SHARELIB option in the top-level Makefile enables users to build a shared library with Position Independent Code.

This code has been tested within JEDI with code that will be submitted in several other pull requests (#107 and #108).  There should be no impact to non-JEDI applications, but I have no method to test that fact at this time.

### DIFF
--- a/src/driver/mpas.F
+++ b/src/driver/mpas.F
@@ -14,11 +14,11 @@ program mpas
    type (core_type), pointer :: corelist
    type (domain_type), pointer :: domain_ptr
 
-   call mpas_init(corelist,domain_ptr)
+   call mpas_init(corelist, domain_ptr)
 
    call mpas_run(domain_ptr) 
 
-   call mpas_finalize(corelist,domain_ptr)
+   call mpas_finalize(corelist, domain_ptr)
 
    stop
 
--- a/src/driver/mpas.F
+++ b/src/driver/mpas.F
@@ -11,11 +11,14 @@ program mpas
 
    implicit none
 
-   call mpas_init()
+   type (core_type), pointer :: corelist
+   type (domain_type), pointer :: domain_ptr
 
-   call mpas_run() 
+   call mpas_init(corelist,domain_ptr)
 
-   call mpas_finalize()
+   call mpas_run(domain_ptr) 
+
+   call mpas_finalize(corelist,domain_ptr)
 
    stop
 
--- a/src/driver/mpas_subdriver.F
+++ b/src/driver/mpas_subdriver.F
@@ -154,8 +154,8 @@ module mpas_subdriver
 
       call mpas_allocate_domain(domain_ptr)
 
-      if (present(mpi_comm)) domainID = domainID + 1
-      domain_ptr % domainID = domainID + 1
+      domainID = domainID + 1
+      domain_ptr % domainID = domainID
 
       !
       ! Initialize infrastructure
@@ -362,7 +362,7 @@ module mpas_subdriver
    end subroutine mpas_run
 
 
-   subroutine mpas_finalize(corelist,domain_ptr)
+   subroutine mpas_finalize(corelist, domain_ptr)
 
       use mpas_stream_manager, only : MPAS_stream_mgr_finalize
       use mpas_log, only : mpas_log_finalize, mpas_log_info
--- a/src/driver/mpas_subdriver.F
+++ b/src/driver/mpas_subdriver.F
@@ -35,14 +35,12 @@ module mpas_subdriver
    use test_core_interface
 #endif
 
-   type (core_type), pointer :: corelist => null()
    type (dm_info), pointer :: dminfo
-   type (domain_type), pointer :: domain_ptr
 
    contains
 
 
-   subroutine mpas_init()
+   subroutine mpas_init(corelist, domain_ptr, mpi_comm)
 
       use mpas_stream_manager, only : MPAS_stream_mgr_init, MPAS_build_stream_filename, MPAS_stream_mgr_validate_streams
       use iso_c_binding, only : c_char, c_loc, c_ptr, c_int
@@ -52,6 +50,10 @@ module mpas_subdriver
       use mpas_log
  
       implicit none
+
+      type (core_type), intent(out), pointer :: corelist
+      type (domain_type), intent(out), pointer :: domain_ptr
+      integer, intent(in), optional :: mpi_comm
 
       integer :: iArg, nArgs
       logical :: readNamelistArg, readStreamsArg
@@ -81,6 +83,7 @@ module mpas_subdriver
       character(len=StrKIND) :: iotype
       logical :: streamsExists
       integer :: mesh_iotype
+      integer, save :: domainID = -1
 
       interface
          subroutine xml_stream_parser(xmlname, mgr_p, comm, ierr) bind(c)
@@ -151,10 +154,13 @@ module mpas_subdriver
 
       call mpas_allocate_domain(domain_ptr)
 
+      if (present(mpi_comm)) domainID = domainID + 1
+      domain_ptr % domainID = domainID + 1
+
       !
       ! Initialize infrastructure
       !
-      call mpas_framework_init_phase1(domain_ptr % dminfo)
+      call mpas_framework_init_phase1(domain_ptr % dminfo, mpi_comm)
 
 
 #ifdef CORE_ATMOSPHERE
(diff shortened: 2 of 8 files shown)


src/driver/mpas_subdriver.F
```diff
@@ -151,10 +154,13 @@ end subroutine xml_stream_get_attributes
 
       call mpas_allocate_domain(domain_ptr)
 
+      if (present(mpi_comm)) domainID = domainID + 1
```
mgduda: Why do we only increment `dominID` if `mpi_comm` is present? Wouldn't it be possible in principle to run multiple instances of MPAS sequentially using `MPI_COMM_WORLD` as the communicator?

src/driver/mpas.F
```diff
@@ -11,11 +11,14 @@ program mpas
 
    implicit none
 
-   call mpas_init()
+   type (core_type), pointer :: corelist
+   type (domain_type), pointer :: domain_ptr
 
-   call mpas_run() 
+   call mpas_init(corelist,domain_ptr)
```
mgduda: This is nitpicking to be sure, but almost everywhere we use a single space after each comma in an argument list. For consistency, it might be nice to do the same in all new code.

src/framework/mpas_log.F
```diff
@@ -823,13 +839,15 @@ subroutine log_abort()
...
+      call mpas_release_unit(mpas_log_info % errorLog % unitNum)
 
       deallocate(mpas_log_info % errorLog)
       deallocate(mpas_log_info % outputLog)
       deallocate(mpas_log_info)
 
 #ifdef _MPI
-      call MPI_Abort(MPI_COMM_WORLD, mpi_errcode, mpi_ierr)
+      if ( mpas_log_info % domainID == 0 ) &
+         call MPI_Abort(MPI_COMM_WORLD, mpi_errcode, mpi_ierr)
```
mgduda: Does this have any implications for situations in which domainID 0 does not encounter any fatal runtime issues, while some other domainID does?

pwolfram:
FYI @jd-moulton and @#dasvyat

pwolfram:
FYI @jd-moulton and @dasvyat

src/driver/mpas_subdriver.F
jjguerrette: That is a very good point.  My assumption was that the external program would initialize MPI, which is true in the case of JEDI, but that is not strictly necessary.  I did not think of the use-case you suggest here.

The two requirements that I am trying to meet are that (1) MPI is not aborted when it was initialized outside of MPAS (i.e., when mpi_comm is passed to mpas_init) and (2) that each instance of domain has its own log file.   Would you prefer that we add your use-case as a requirement?

src/framework/mpas_log.F
jjguerrette: As this PR is currently written, domainID 0 is only used when MPI is initialized internally within MPAS, which does not allow more than one domain to be instantiated.  Thus domainID != 0 only occurs when mpi_comm is initialized externally, in which case we assume the external program will handle shutdown of MPI.  If we are to add the use-case of allowing multiple instances of domain to use an internally initialized MPI communicator, then this would need to change.  Maybe I could add a logical to domain and mpas_log_info such as "internal_mpi" that knows how MPI was initialized.  What do you think?

src/driver/mpas_subdriver.F
```diff
@@ -151,10 +154,13 @@ end subroutine xml_stream_get_attributes
 
       call mpas_allocate_domain(domain_ptr)
 
+      domainID = domainID + 1
+      domain_ptr % domainID = domainID
```
mgduda: As written, the logic for setting `domain_ptr % domainID` seems odd to me. At least in my opinion, it might make more sense to increment `domainID` after assigning it to `domain_ptr % domainID`. That way, the initial value of `domainID` could be 0, which is the first domainID to be assigned.

jjguerrette: Yes, that makes way more sense.  I missed that when updating for multiple internally initialized MPAS domains.

src/framework/mpas_io_units.F
```diff
@@ -38,14 +46,29 @@ module mpas_io_units
...
+        if ( present(unformatted) ) then
+           if ( unformatted ) limit_type = 2
+        end if
+        if ( limit_type == 2 ) then
+           minsearch = unformatted_min
+           maxsearch = unformatted_max
+        else
+           minsearch = 1
+           maxsearch = maxUnits
+        end if
```
mgduda: The logic here involving `limit_type` seems unnecessarily complicated to me (but perhaps the idea was that this would be more easily generalized in future?). Instead, it seems simpler in my opinion to use something like the following:
```fortran
        minsearch = 1
        maxsearch = maxUnits
        if ( present(unformatted) ) then
           if ( unformatted ) then
              minsearch = unformatted_min
              maxsearch = unformatted_max
           end if
        end if
```

mgduda:
@jjguerrette I think we had discussed this via e-mail, but I'd be glad to help out with the generalization of the `-fPIC` changes to work (to the best of our ability to test) with other compilers. One possibility is that I could simply push changes related to this directly to this branch. Would that be fine for you?

mgduda:
Since we've got a chance to define the terminology from the outset, are there any opinions on whether we call different instances of MPAS "domains", "instances", or something different? I'm wondering whether the "domains" terminology that's used in, e.g., the `mpas_log` module may have certain misleading connotations for some users? But, I'm not familiar enough with exactly what different domains/instances mean in the context of OOPS or JEDI to offer informed alternatives.

jjguerrette:
> @jjguerrette I think we had discussed this via e-mail, but I'd be glad to help out with the generalization of the `-fPIC` changes to work (to the best of our ability to test) with other compilers. One possibility is that I could simply push changes related to this directly to this branch. Would that be fine for you?

@mgduda that would be very helpful.  Please push changes as they seem appropriate.

mgduda:
@jjguerrette I'm finally getting back to this. My thinking is that we should split the changes here into several PRs, and this PR could focus on adding a `SHAREDLIB` build option. I can try to refactor the `framework/external_lib_support` branch and add some generalization for other compilers. In order not to lose the other changes, I could push the `framework/external_lib_support` branch as it currently exists to a branch with another name on your fork or on my fork. Then, subsequent PRs could pick commits from this saved branch. Does that sound reasonable?

mgduda:
@jjguerrette I think we could create four separate PRs to handle:
1. Shared-library flags
2. `domain_ptr` and `corelist` changes
3. Changes for I/O units
4. Changes to add a domain ID

This PR could be refactored to include just the shared-library changes. I should be able to handle the refactoring this week, assuming you don't object to my modifying the `framework/external_lib_support` branch on your fork.

jjguerrette:
@mgduda, would you mind doing the split under new branch names?  Whether under my fork or yours is not important to me.  Currently we have a develop-for-jedi branch that depends on all of the other branches with modifications (including framework/external_lib_support).  I kept it that way so that it is easy to track our differences with develop and individual merges into develop.

If needed, I can help more after next week.  I will be out all of next week, but feel free to consult this week or when I return.

mgduda:
@jjguerrette No problem. I may need to close this PR and to create all new PRs, since (as far as I'm aware), GitHub doesn't allow changing the source branch for a PR.

jjguerrette:
That is OK with me as long as the code changes are not lost (which they should not be).

mgduda:
@jjguerrette No problem -- I'll create the PRs from new branches on my fork.

mgduda:
PR #307 should take care of the `-fPIC` changes.

mgduda:
PR #309 should take care of the unformatted I/O units 95 - 99.

mgduda:
PR #308 should handle the remaining changes, but I'd like to think a little more on the consequences of not calling `MPI_Abort`. I think we may need to carefully go through our code to identify places where we expect that execution will never get past, e.g., `call mpas_log_write(..., messageType=MPAS_LOG_CRIT)`.

mgduda:
@jjguerrette With PR #307 and PR #309 merged, and with the remaining changes captured by PR #308 (which will require more work), I think we can safely close this PR. If you notice anything from this PR that isn't handled elsewhere, though, please feel free to re-open.

PR #107, opened 2018-11-06
Update clock and mpas_log_info in core_atmosphere when there are multiple instances of domain

jjguerrette:
This PR supports interfacing MPAS with the JEDI data assimilation framework. 

When there are multiple instances of the MPAS domain, the clock and mpas_log_info
pointers need to be updated for each call to dependent high-level subroutines in
mpas_atm_core.F.

This code has been tested within JEDI with code that will be submitted in several other pull requests.  The effect of this PR is apparent when considered with #106 .  There should be no impact to non-JEDI applications, but I have no method to test that fact at this time.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core.F
+++ b/src/core_atmosphere/mpas_atm_core.F
@@ -10,7 +10,7 @@ module atm_core
    use mpas_derived_types
    use mpas_pool_routines
    use mpas_dmpar
-   use mpas_log, only : mpas_log_write
+   use mpas_log, only : mpas_log_write, mpas_log_info
 
    type (MPAS_Clock_type), pointer :: clock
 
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/mpas_atm_core.F
```diff
@@ -851,6 +855,9 @@ subroutine atm_do_timestep(domain, dt, itimestep)
       real (kind=RKIND) :: xtime_s
       integer :: ierr
 
+      clock => domain % clock
+      mpas_log_info => domain % logInfo
+
```
mgduda: There's no problem to have these assignments here, but just out of curiosity, does JEDI call `atm_do_timestep` outside of a call to `atm_core_run`? In stand-alone MPAS-A, the only call to `atm_do_timestep` is inside of `atm_core_run`, where the `clock` and `mpas_log_info` pointers are also being set.

jjguerrette: Yes that is the case.  JEDI calls observation operators between time steps and so needs the capability to integrate single time steps without performing a full run.

mgduda: Got it! I'll go ahead and merge this PR now.

PR #109, opened 2018-11-07
Ocean/fix mld for land ice

vanroekel:
This fixes MLD for land ice cavities. IN the original code pressure is used to find a reference depth above 10 deci-bars, which does not work under land ice cavities and results in large MLD.  This commit subtracts land ice pressure at the start of the analysis and adds it back at the end.

### DIFF
--- a/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
+++ b/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
@@ -78,7 +78,7 @@
 		/>
 	</var_struct>
 <var_struct name="mixedLayerDepthAMScratch" time_levs="1">
-▸\▸\<var name="workArrayPressure"
+▸\▸\<var name="pressureAdjustedForLandIceScratch"
 ▸\▸\▸\ persistence="scratch"
 ▸\▸\▸\ type="real"
 ▸\▸\▸\ dimensions="nVertLevels nCells Time"
--- a/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
+++ b/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
@@ -78,7 +78,7 @@
 		/>
 	</var_struct>
 <var_struct name="mixedLayerDepthAMScratch" time_levs="1">
-▸\▸\<var name="pressureAdjustedForLandIceScratch"
+  <var name="pressureAdjustedForLandIceScratch"
 ▸\▸\▸\ persistence="scratch"
 ▸\▸\▸\ type="real"
 ▸\▸\▸\ dimensions="nVertLevels nCells Time"
--- a/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
+++ b/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
@@ -77,7 +77,16 @@
 			 description="mixed layer depth based on gradient of density"
 		/>
 	</var_struct>
-	<streams>
+<var_struct name="mixedLayerDepthAMScratch" time_levs="1">
+▸\▸\<var name="workArrayPressure"
+▸\▸\▸\ persistence="scratch"
+▸\▸\▸\ type="real"
+▸\▸\▸\ dimensions="nVertLevels nCells Time"
+▸\▸\▸\ units="various"
+▸\▸\▸\ description="temporary array to hold pressure " 
+▸\▸\/>
+</var_struct>
+  <streams>
 		<stream name="mixedLayerDepthsOutput" type="output"
 				mode="forward;analysis"
 				filename_template="analysis_members/mixedLayerDepths.$Y-$M-$D.nc"
--- a/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
+++ b/src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
@@ -77,16 +77,16 @@
 			 description="mixed layer depth based on gradient of density"
 		/>
 	</var_struct>
-<var_struct name="mixedLayerDepthAMScratch" time_levs="1">
-  <var name="pressureAdjustedForLandIceScratch"
-▸\▸\▸\ persistence="scratch"
-▸\▸\▸\ type="real"
-▸\▸\▸\ dimensions="nVertLevels nCells Time"
-▸\▸\▸\ units="various"
-▸\▸\▸\ description="temporary array to hold pressure " 
-▸\▸\/>
-</var_struct>
-  <streams>
+	<var_struct name="mixedLayerDepthAMScratch" time_levs="1">
+		<var name="pressureAdjustedForLandIceScratch"
+			persistence="scratch"
+			type="real"
+			dimensions="nVertLevels nCells Time"
+			units="various"
+			description="temporary array to hold pressure " 
+		/>
+	</var_struct>
+  	<streams>
 		<stream name="mixedLayerDepthsOutput" type="output"
 				mode="forward;analysis"
 				filename_template="analysis_members/mixedLayerDepths.$Y-$M-$D.nc"
--- a/src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
+++ b/src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
@@ -177,8 +177,8 @@ contains
       type(field2DReal), pointer :: pressureAdjustedForLandIceField
       real (kind=RKIND), dimension(:,:), pointer :: pressureAdjustedForLandIce 
 
-      real (kind=RKIND), dimension(:), pointer :: tThresholdMLD, tGradientMLD, landIceDraft
-      real (kind=RKIND), dimension(:), pointer :: dThresholdMLD, dGradientMLD, landIcePressure
+      real (kind=RKIND), dimension(:), pointer :: tThreshMLD, tGradientMLD, landIceDraft
+      real (kind=RKIND), dimension(:), pointer :: dThreshMLD, dGradientMLD, landIcePressure
       integer, dimension(:), pointer :: landIceMask 
       real (kind=RKIND), dimension(:,:,:), pointer :: tracers
       real (kind=RKIND), dimension(:,:), pointer :: zTop, zMid, pressure
@@ -263,7 +263,7 @@ contains
          endif
 
          if(tThresholdFlag) then
-            call mpas_pool_get_array(mixedLayerDepthsAMPool, 'tThreshMLD',tThresholdMLD)
+            call mpas_pool_get_array(mixedLayerDepthsAMPool, 'tThreshMLD',tThreshMLD)
             do iCell = 1,nCellsSolve
 
                !Initialize RefIndex for cases of very shallow columns
@@ -294,26 +294,26 @@ contains
                         call interp_bw_levels(localVals(2),localVals(3), dV, dVp1, tempThresh,     &
                                    interp_local, mldTemp)!,dVm1, localVals(1))
                         mldTemp=max(mldTemp,zMid(k+1,iCell)) !make sure MLD isn't deeper than zMid(k+1)
-                        tThresholdMLD(iCell)=abs(min(mldTemp,zMid(k,iCell))) !MLD should be deeper than zMid(k)
+                        tThreshMLD(iCell)=abs(min(mldTemp,zMid(k,iCell))) !MLD should be deeper than zMid(k)
                         found_temp_mld = .true.
                         exit
                      endif
                 enddo
 
 ! if the mixed layer depth is not found, it is set to the depth of the bottom most level
-                   if(.not. found_temp_mld) tThresholdMLD(iCell) = abs(zMid(maxLevelCell(iCell),iCell))
+                   if(.not. found_temp_mld) tThreshMLD(iCell) = abs(zMid(maxLevelCell(iCell),iCell))
                enddo !iCell
 
                if (associated(landIceMask) ) THEN
                   do iCell = 1, nCellsSolve
-                     tThresholdMLD(iCell) = tThresholdMLD(iCell) - abs(landIceDraft(iCell))*landIceMask(iCell)
+                     tThreshMLD(iCell) = tThreshMLD(iCell) - abs(landIceDraft(iCell))*landIceMask(iCell)
                   enddo
                endif
 
           endif !end tThresh MLD search
 
          if(dThresholdFlag) then
-            call mpas_pool_get_array(mixedLayerDepthsAMPool, 'dThreshMLD',dThresholdMLD)
+            call mpas_pool_get_array(mixedLayerDepthsAMPool, 'dThreshMLD',dThreshMLD)
 
             do iCell = 1,nCellsSolve
 
@@ -343,7 +343,7 @@ contains
                         call interp_bw_levels(localVals(2),localVals(3), dV, dVp1, denThresh,     &
                                    interp_local, mldTemp)!, dVm1, localVals(1))
                         mldTemp=max(mldTemp,zMid(k+1,iCell)) !make sure MLD isn't deeper than zMid(k+1)
-                        dThresholdMLD(iCell)=abs(min(mldTemp,zMid(k,iCell))) !MLD should be deeper than zMid(k)
+                        dThreshMLD(iCell)=abs(min(mldTemp,zMid(k,iCell))) !MLD should be deeper than zMid(k)
                         found_den_mld = .true.
                         exit
                      endif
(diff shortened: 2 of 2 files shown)


vanroekel:
NOTE: not yet tested.  DO NOT MERGE.

@xylar here is the MLD branch that should work under land ice cavities.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
```diff
@@ -230,11 +232,18 @@ subroutine ocn_compute_mixed_layer_depths(domain, timeLevel, err)!{{{
...
+         call mpas_pool_get_array(forcingPool, 'landIcePressure', landIcePressure)
          call mpas_pool_get_array(diagnosticsPool, 'zMid', zMid)
          call mpas_pool_get_array(diagnosticsPool, 'zTop', zTop)
          call mpas_pool_get_array(meshPool, 'latCell', latCell)
          call mpas_pool_get_array(meshPool, 'lonCell', lonCell)
 
+         !If landice cavity remove land ice pressure to search for ML depth
+         if ( associated(landIcePressure) ) THEN
+           do iCell = 1,nCellsSolve
+             pressure(1,iCell) = pressure(1,iCell) - landIcePressure(iCell)
```
xylar: @vanroekel, a couple of things.  First, it strikes me as maybe a bad idea to modify pressure and then change it back but maybe this is a common practice in MPAS-Ocean that I hadn't noticed.  It would seem better to me (though obviously more work) to use a scratch field. @mark-petersen, what do you think?

Second, I think `landIcePressure` needs to be subtracted from pressure at all levels, not just at `k = 1`.

vanroekel: @xylar 
1) Using a separate field does make a lot of sense and removes the latter loop over pressure
2) Yep, you're right.  I was not fully understanding the code in the pressure computation.

I'll address both of these comments

vanroekel:
@xylar I've updated to using a scratch variable for pressure and corrected the subtraction of landIcePressure.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
```diff
@@ -216,9 +221,16 @@ subroutine ocn_compute_mixed_layer_depths(domain, timeLevel, err)!{{{
...
          call mpas_pool_get_subpool(block % structs, 'diagnostics', diagnosticsPool)
+         call mpas_pool_get_subpool(block % structs, 'forcing', forcingPool)
          call mpas_pool_get_subpool(block % structs, 'mixedLayerDepthsAM', mixedLayerDepthsAMPool)
          call mpas_pool_get_subpool(statePool, 'tracers', tracersPool)
 
+         !Use work array for storing of pressure, potentially modified by landIce
+         call mpas_pool_get_subpool(block % structs, 'mixedLayerDepthsAMScratch', scratchPool)
+         call mpas_pool_get_field(scratchPool, 'workArrayPressure', workArrayField)
+         call mpas_allocate_scratch_field(workArrayField, .true.)
+         workArray => workArrayField % array
```
xylar: I know this is picky, but could you rename `workArray*` to `pressureAdjustedLandIce` or something else that's more readable?

vanroekel: not a problem.  Sounds like a good change for clarity.

xylar:
Great, thanks, @vanroekel.  We should set up a simple QU240 test with ice shelves to test it.  This is a question of running one of the COMPASS test cases but setting up the monthly output as expected by MPAS-Analysis.  I'll try to get to this later today or tomorrow.

xylar:
@vanroekel, I'm seeing these errors:

```
mpas_ocn_mixed_layer_depths.F(230): error #6404: This name does not have a type, and must have an explicit type.   [PRESSUREADJUSTEDFORLANDICEFIELD]
         call mpas_pool_get_field(scratchPool, 'pressureAdjustedForLandIceScratch', pressureAdjustedForLandIceField)
------------------------------------------------------------------------------------^
mpas_ocn_mixed_layer_depths.F(230): error #6285: There is no matching specific subroutine for this generic subroutine call.   [MPAS_POOL_GET_FIELD]
         call mpas_pool_get_field(scratchPool, 'pressureAdjustedForLandIceScratch', pressureAdjustedForLandIceField)
--------------^
mpas_ocn_mixed_layer_depths.F(231): error #6285: There is no matching specific subroutine for this generic subroutine call.   [MPAS_ALLOCATE_SCRATCH_FIELD]
         call mpas_allocate_scratch_field(pressureAdjustedForLandIceField, .true.)
--------------^
mpas_ocn_mixed_layer_depths.F(232): error #6460: This is not a field name that is defined in the encompassing structure.   [ARRAY]
         pressureAdjustedForLandIce => pressureAdjustedForLandIceField % array
-------------------------------------------------------------------------^
mpas_ocn_mixed_layer_depths.F(232): error #6796: The variable must have the TARGET attribute or be a subobject of an object with the TARGET attribute, or it must have the POINTER attribute.   [PRESSUREADJUSTEDFORLANDICEFIELD]
         pressureAdjustedForLandIce => pressureAdjustedForLandIceField % array
---------------------------------------^
mpas_ocn_mixed_layer_depths.F(451): error #6285: There is no matching specific subroutine for this generic subroutine call.   [MPAS_DEALLOCATE_SCRATCH_FIELD]
         call mpas_deallocate_scratch_field(pressureAdjustedForLandIceField, .true.)
--------------^
```

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
```diff
@@ -172,8 +174,11 @@ subroutine ocn_compute_mixed_layer_depths(domain, timeLevel, err)!{{{
       logical,pointer :: tThresholdFlag, dThresholdFlag
       logical,pointer :: tGradientFlag, dGradientFlag
 !      real (kind=RKIND), dimension(:), pointer ::  areaCell
+      type(field2DReal), pointer :: pressureAdjustedForLandIceScratchField
```
xylar: This should be:
```suggestion
      type(field2DReal), pointer :: pressureAdjustedForLandIceField
```

xylar:
The change above fixes the problem.

xylar:
Update: I rebased onto `e3sm/develop` and the MPAS-Seaice compilation error is now gone.

xylar:
I've got a job in the queue to test this.  1.5 days to go.  Hopefully I didn't accidentally tell it to restart like last time my job sat in the queue for 3 days....

src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
```diff
@@ -77,7 +77,16 @@
 			 description="mixed layer depth based on gradient of density"
 		/>
 	</var_struct>
-	<streams>
+<var_struct name="mixedLayerDepthAMScratch" time_levs="1">
+▸\▸\<var name="pressureAdjustedForLandIceScratch"
```
xylar: something weird happened here with the "tabs".  I think this is causing errors in the run, but I can't be sure.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
```diff
@@ -216,9 +221,16 @@ subroutine ocn_compute_mixed_layer_depths(domain, timeLevel, err)!{{{
          call mpas_pool_get_subpool(block % structs, 'state', statePool)
          call mpas_pool_get_subpool(block % structs, 'mesh', meshPool)
          call mpas_pool_get_subpool(block % structs, 'diagnostics', diagnosticsPool)
+         call mpas_pool_get_subpool(block % structs, 'forcing', forcingPool)
          call mpas_pool_get_subpool(block % structs, 'mixedLayerDepthsAM', mixedLayerDepthsAMPool)
          call mpas_pool_get_subpool(statePool, 'tracers', tracersPool)
 
+         !Use work array for storing of pressure, potentially modified by landIce
+         call mpas_pool_get_subpool(block % structs, 'mixedLayerDepthsAMScratch', scratchPool)
+         call mpas_pool_get_field(scratchPool, 'pressureAdjustedForLandIceScratch', pressureAdjustedForLandIceField)
```
xylar: @vanroekel, I'm getting an error: "forrtl: severe (71): integer divide by zero" from this line.

We should double check based on a different AM that uses a scratch variable that we did this correctly here.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
```diff
@@ -216,9 +221,16 @@ subroutine ocn_compute_mixed_layer_depths(domain, timeLevel, err)!{{{
          call mpas_pool_get_subpool(block % structs, 'state', statePool)
          call mpas_pool_get_subpool(block % structs, 'mesh', meshPool)
          call mpas_pool_get_subpool(block % structs, 'diagnostics', diagnosticsPool)
+         call mpas_pool_get_subpool(block % structs, 'forcing', forcingPool)
          call mpas_pool_get_subpool(block % structs, 'mixedLayerDepthsAM', mixedLayerDepthsAMPool)
          call mpas_pool_get_subpool(statePool, 'tracers', tracersPool)
 
+         !Use work array for storing of pressure, potentially modified by landIce
+         call mpas_pool_get_subpool(block % structs, 'mixedLayerDepthsAMScratch', scratchPool)
```
xylar:
```suggestion
         call mpas_pool_get_subpool(block % structs, 'mixedLayerDepthAMScratch', scratchPool)
```

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
xylar: This was caused by the fact that the Registry has `Depth` but the subpool call above is with `Depths` (plural).  The pool didn't exist and therefor had zero size and gave a divide by zero error.

Re-testing with this fix.

vanroekel:
@xylar I'm testing this now with your suggested fixes on anvil.  It's running.  I'll post an antarctic MLD plot once it completes.

vanroekel:
@xylar here is a very rough figure from my test.  Does this look reasonable?  I'm not sure what the analysis was giving previously.  I'm not convinced I setup the with Land Ice test correctly (I used the EC6030 with land ice test from compass)
mldantarctica

xylar:
@vanroekel, that certainly looks better than what we had before.  I'm curious if the very deep MLD under Flichner-Ronne is real or if that's still and artifact.  But maybe we don't have an obvious, independent way of confirming that.  Any suggestions for how we might know?

xylar:
In any case, I'm glad to see that your test ran successfully.

vanroekel:
The only possible verification I could think of would be to plot profiles of T&S&rho in the Filchner Ronne and see if the diagnosed MLD is consistent visually.

xylar:
Sounds good.  If you point me to your output, I'll run MPAS-Analysis with climatologyMapMLD and geojsonTransects to see what we find.  But I'm heading out soon so that'll probably happen over the weekend or on Monday.

vanroekel:
@xylar the test is here on anvil
```
/lcrc/group/acme/lvanroe/MPAS-Model/testing_and_setup/compass/ocean/global_ocean_V1/EC60to30/with_land_ice/forward
```
I've only run 9 months, so I'm not sure mpas-analysis will work.  I'm running the last 3 months now.

xylar:
@vanroekel, I'm working on testing this now.  I need a new python environment (with `shapely`) for the latest MPAS-Analysis so that's taking a little time to set up.

xylar:
@vanroekel, I'm all set to test but don't have read permission on the files in that directory. Could you fix those?

vanroekel:
@xylar permissions should be set now.

xylar:
@vanroekel, I ran analysis and it's here:
http://portal.nersc.gov/project/acme/xylar/test_mld_fix/ocean/index.html#mld

MLD is quite deep under ice shelves but density transects suggest that this *might* be correct, especially this early in a run when the ice pump hasn't really got going yet:
http://portal.nersc.gov/project/acme/xylar/test_mld_fix/ocean/index.html#geojson

I'll come by your office quickly and then approve the PR unless you're concerned about what you see.

vanroekel:
@xylar quick question -- is zTop(1,iCell) the depth of the landIce - Ocean interface?  I think I found why the MLs are so deep.  I should have a fix running soon.

xylar:
>  is zTop(1,iCell) the depth of the landIce - Ocean interface?

@vanroekel, I think it should be but I'm not 100% certain.  I'll dig into the code a bit to be sure.

xylar:
Yes, looks like it should be:
https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/core_ocean/shared/mpas_ocn_diagnostics.F#L671-L682

You could also use `ssh(iCell)` for the same thing, at least it looks that way in the diagnostics routines.

xylar:
Is that part of the issue?  You explicitly *don't* want the ssh included?  If so, I think you'd need some reference state ssh from an initial condition or something like that.  Not really a well defined concept once ice shelves are allowed to depress the ice-ocean interface.

xylar:
@vanroekel, could you use `modifySSHMask` instead of `landIceMask` to determine where to reference to `ssh`?  The former tells you where I allowed the iterative algorithm to change the SSH so there's a smooth transition at ice-shelf calving fronts.  Obviously, this shouldn't affect runs without ice-shelf cavities.

xylar:
@vanroekel, ah ha! The variable we want is `landIceDraft`.  It should be masked with `modifySSHMask` rather than `landIceMask` if at all possible.

vanroekel:
@xylar I've made the changes to modifySSHMask and landIceDraft.  Test is underway now.

xylar:
Excellent!  Fingers crossed.

vanroekel:
@xylar here is the result of the latest test.  MLD under the ice shelves are quite changed (much shallower).    Note that ```modifySSHMask``` is not saved to the restart file, so i had to use ```landIceMask```.  The data is on IC at 
```
/lustre/scratch4/turquoise/lvanroekel/MPAS-Model/testing_and_setup/compass/ocean/global_ocean_V1/EC60to30/with_land_ice/forward
```
if you want to look at it.  It's run 1 year.
mldantarctica

xylar:
That looks great!  Much more what I was expecting and potentially quite useful.  I don't see any major artifacts of using `landIceMask` (and obviously that's the only feasible option) so that's fine.  Have you tested with a run without ice-shelf cavities to make sure this doesn't break the original behavior?

src/core_ocean/analysis_members/Registry_mixed_layer_depths.xml
```diff
@@ -77,7 +77,16 @@
 			 description="mixed layer depth based on gradient of density"
 		/>
 	</var_struct>
-	<streams>
+<var_struct name="mixedLayerDepthAMScratch" time_levs="1">
+  <var name="pressureAdjustedForLandIceScratch"
+▸\▸\▸\ persistence="scratch"
```
xylar: I'm still seeing weird non-tab characters here.  Could you see if you can fix those?

vanroekel: @xylar I think I got the white space figured out and fixed.

vanroekel:
I've just tested this in a 30to10 no landIce case and the MLD output looks fine.

mark-petersen:
Sorry, I wasn't following the conversation.  Looking now, will also test.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_depths.F
```diff
@@ -230,11 +243,24 @@ subroutine ocn_compute_mixed_layer_depths(domain, timeLevel, err)!{{{
...
+           do iCell = 1,nCellsSolve
+             do k = 1,maxLevelCell(iCell)
+                pressureAdjustedForLandIce(k,iCell) = pressureAdjustedForLandIce(k,iCell)   &
+                                                        - landIcePressure(iCell)
+             enddo
+           enddo
+         endif
 
          if(tThresholdFlag) then
             call mpas_pool_get_array(mixedLayerDepthsAMPool, 'tThreshMLD',tThresholdMLD)
```
mark-petersen: Strangely, we used tThreshMLD in `Registry_mixed_layer_depths.xml` and tThresh**old**MLD in the code. Same with dThreshMLD.  I think that is unnecessarily confusing, and easy to change here as *ThresholdMLD only appears in 8 lines, all in this module.

vanroekel:
@mark-petersen and @xylar Thanks for the reviews, I've updated the variable names and merged.

PR #110, opened 2018-11-07
New Tiedtke Cumulus Scheme Unified with WRFV4.0

kkeene44:
Toward the effort to unify physics schemes between the MPAS and WRF models, the New Tiedtke routine was modified to include changes necessary to allow it to run in both models. In areas where the models differ (e.g., in their handling of 'dx'), there are if-defs put around the code to ensure that it's handled correctly for the model in which it is running. Results before and after are NOT bit-for-bit, however, as this update includes some updates to the actual physics. The original MPAS version of New Tiedtke was updated to V3.8.1 of WRF, and then was modified by @weiwangncar to update to WRFV3.9. The mods I made stemmed from those.

### DIFF
--- a/src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
+++ b/src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
@@ -1,15 +1,3 @@
-!=================================================================================================================
-! copied for implementation in MPAS from WRF version 3.8.1:
-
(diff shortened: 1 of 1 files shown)


kkeene44:
@ldfowler58 
I verified that your mods incorporating the dx2d variable and reducing the "if defined(mpas)" statements works for WRF, as well. I pushed the changes to this PR and opened a new one on the WRF side.

PR #117, opened 2018-11-30
Add 'llvm' make target to support compilation with clang+flang

mgduda:
This merge adds a new target, `llvm`, to the top-level Makefile to support the compilation
of MPAS with clang, clang++, and flang.


mgduda:
@matthewhoffman This is a small enough change that I didn't think it necessary to add @mark-petersen and @akturner as reviewers.

PR #124, opened 2018-12-20
Data iceberg forcing restart bug fix.

darincomeau:
The current implementation of data icebergs passes testing in MPAS-Seaice standalone, including restartability, however in E3SM the model dies on a restart initialization due to data icebergs. Compsets without data icebergs active do not seem to be affected.

This PR is a work-in-progress, opened for visibility. Currently this branch passes testing in MPAS-Seaice standalone, runs in E3SM for an initial run, and fails on a restart run.

### DIFF
--- a/src/core_seaice/shared/mpas_seaice_forcing.F
+++ b/src/core_seaice/shared/mpas_seaice_forcing.F
@@ -345,9 +345,7 @@ contains
 
        call data_iceberg_forcing(&
             streamManager, &
-            domain, &
-            simulationClock, &
-            firstTimeStep)
+            domain)
 
     endif
 
@@ -1700,21 +1698,16 @@ contains
 
   subroutine data_iceberg_forcing(&
        streamManager, &
-       domain, &
-       simulationClock, &
-       firstTimeStep)
+       domain)
 
     type (MPAS_streamManager_type), intent(inout) :: streamManager
 
     type (domain_type) :: domain
 
-    type (MPAS_clock_type) :: simulationClock
-
-    logical, intent(in) :: firstTimeStep
-
     type (MPAS_time_type) :: currentForcingTime
 
-    character(len=strKIND) :: currentForcingTimeStr
+    ! for debugging
+    ! character(len=strKIND) :: currentForcingTimeStr
 
     logical, pointer :: &
          config_do_restart
@@ -1723,7 +1716,6 @@ contains
          config_dt
 
     call mpas_pool_get_config(domain % configs, 'config_dt', config_dt)
-    call MPAS_pool_get_config(domain % configs, "config_do_restart", config_do_restart)
 
     ! get the current time
     call MPAS_forcing_get_forcing_time(&
@@ -1731,8 +1723,9 @@ contains
        "seaice_data_iceberg_forcing_monthly", &
        currentForcingTime)
 
-    call MPAS_get_time(currentForcingTime, dateTimeString=currentForcingTimeStr)
-    call MPAS_log_write('Get Data icebergs at: '//trim(currentForcingTimeStr))
+    ! for debugging
+    ! call MPAS_get_time(currentForcingTime, dateTimeString=currentForcingTimeStr)
+    ! call MPAS_log_write('Get Data icebergs at: '//trim(currentForcingTimeStr))
 
     ! use the forcing layer to get
     call MPAS_forcing_get_forcing(&
--- a/src/core_seaice/shared/mpas_seaice_forcing.F
+++ b/src/core_seaice/shared/mpas_seaice_forcing.F
@@ -1704,9 +1704,8 @@ contains
 
     type (domain_type) :: domain
 
-    type (MPAS_time_type) :: currentForcingTime
-
     ! for debugging
+    ! type (MPAS_time_type) :: currentForcingTime
     ! character(len=strKIND) :: currentForcingTimeStr
 
     logical, pointer :: &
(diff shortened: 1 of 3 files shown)


src/core_seaice/shared/mpas_seaice_forcing.F
```diff
@@ -1622,29 +1626,41 @@ end subroutine post_oceanic_coupling
...
+    call mpas_get_time(curr_time=currTime, dateTimeString=timeStamp, ierr=ierr)
+    timeStamp = '0000'//trim(timeStamp(5:))
+
     ! create own data iceberg forcing group
     call MPAS_forcing_init_group(&
          seaiceForcingGroups, &
          "seaice_data_iceberg_forcing_monthly", &
          domain, &
-         '0000-01-01_00:00:00', &
+         timeStamp, &
```
akturner: This only effects an initial run not a restart run. I guess the code assumes the initial simulation (but not restarts) start on the year start. Is this the issue?

akturner:
MPAS_forcing_init_group is called but that argument is only relevant when config_do_restart is false (although you might want to check this)

src/core_seaice/shared/mpas_seaice_forcing.F
akturner: @darincomeau: Can you email me the preprocessed files mpas_forcing.f90 and mpas_seaice_forcing.f90 so I can see what line numbers the traceback corresponds to?

akturner:
@darincomeau: @jonbob and myself are investigating if the below is the problem: https://github.com/MPAS-Dev/MPAS-Model/pull/126

jonbob:
@akturner - the test is submitted to anvil but wallowing in the queue....

jonbob:
My test overnight failed writing the forcing restart time -- I'm debugging right now

jonbob:
I'm going to run an exact restart test over months instead of days and see if that works

akturner:
No, that should be fine. So, I assume this works in the ocean model with forcing and restarts? @darincomeau: would it make sense to entrain an ocean person into this, such as Mat who has dealt with the forcing system in the coupled system?

jonbob:
Even with the test using months instead of days, the comparison part of the test fails. I'll check the logs for clues, but otherwise @akturner's suggestion is probably best

jonbob:
I'm running now outputting the forcing time as the data icebergs are read -- hopefully there will be some clues in there

jonbob:
The output showed no obvious issues and now it's passing the 11-day ERS test. I may clean up the build and test one more time -- otherwise it's good to go, @darincomeau? Are you going to push on this over the holidays?

darincomeau:
@akturner this is now ready for review.

I tested this on my laptop with the QU120km domain, and this passes the MPAS-Seaice testing suite with both config_use_data_icebergs = true and false.
I also ran two tests with E3SM that use data icebergs with the EC60to30v3wLI grid  on Edison (A_WCYCL1850-DIB-ISMF_CMIP6 and GMPAS-DIB-IAF-ISMF), to confirm that restarts run (though I haven't run any tests from the E3SM testing suite).

src/core_seaice/shared/mpas_seaice_forcing.F
```diff
@@ -1685,21 +1698,35 @@ end subroutine init_data_iceberg_forcing
...
+    ! for debugging
+    ! ! get the current time
+    ! call MPAS_forcing_get_forcing_time(&
+    !    seaiceForcingGroups, &
+    !    "seaice_data_iceberg_forcing_monthly", &
+    !    currentForcingTime)
+
+    ! call MPAS_get_time(currentForcingTime, dateTimeString=currentForcingTimeStr)
+    ! call MPAS_log_write('Get Data icebergs at: '//trim(currentForcingTimeStr))
+
```
darincomeau: @akturner - These commented out lines were added by @jonbob during debugging to make sure the forcing and simulation times were aligned correctly. While they're extra commented out lines in the code, they were useful to have, and could be useful in other situations. Another option would be to have such debugging messages implemented more robustly using the framework (for another day).

akturner: @darincomeau: Thats fine, but maybe a more detailed comment about their purpose?

src/core_seaice/shared/mpas_seaice_forcing.F
```diff
@@ -1685,21 +1698,35 @@ end subroutine init_data_iceberg_forcing
...
 
     type (domain_type) :: domain
 
-    type (MPAS_clock_type) :: simulationClock ! dc is this being used?
+    ! for debugging
+    ! type (MPAS_time_type) :: currentForcingTime
+    ! character(len=strKIND) :: currentForcingTimeStr
+
+    logical, pointer :: &
+         config_do_restart
```
darincomeau: @akturner This argument is unused in this subroutine - I'll remove it with the next commit with better comments on the debugging lines.

darincomeau:
Just to wrap this up, I've also tested this branch in E3SM (along with an E3SM change in components/mpas-seaice/driver/ice_comp_mct.F), and the following exact restart tests passed:

ERS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF.edison_intel
ERS.ne30_oECv3wLI.A_WCYCL1850-DIB_CMIP6.edison_intel

jonbob:
@darincomeau - did the G-case pass with salinity-restoring on or off?

darincomeau:
@jonbob good point - salinity-restoring was off:
config_use_surface_salinity_monthly_restoring = .false.
config_use_activetracers_surface_restoring = .false.

PR #126, opened 2018-12-20
Fixes bug in iceberg forcing write restart times

akturner:
Fixes bug in iceberg forcing write restart times


darincomeau:
I tested this on edison but the job also failed on restart.

akturner:
Yeah @jonbob also needs to add something to ice_comp_mct. It seems this routine isn't currently called there.

PR #127, opened 2018-12-21
Add finalize subroutine for physics alarms

jjguerrette:
This subroutine can be used to reset alarms between
multiple simulations that use the same domain.

This subroutine is used in the MPAS interface to the JEDI data assimilation.
tool.  The alarms need to be destroyed between multiple calls to 
atm_mpas_init_block, which can be used to re-initialize the atmospheric 
core of a single domain instance using multiple initial condition files.  This 
avoids the need to call mpas_init more than once for two identical domains.

This is a follow-on to MPAS-Dev PR #106

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_manager.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_manager.F
@@ -20,7 +20,7 @@
 
  implicit none
  private
- public:: physics_timetracker,physics_run_init
+ public:: physics_timetracker,physics_run_init,physics_run_finalize
 
  integer, public:: year                 !Current year.
  integer, public:: julday               !Initial Julian day.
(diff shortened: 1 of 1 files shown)


mgduda:
Is there any need or plan to add a call to `physics_run_finalize` in the stand-alone MPAS-Atmosphere code, e.g., in the `atm_core_finalize` routine somewhere near the call to `atmphys_finalize`? Or does this routine only need to be called from OOPS code?

Also, could you add a mention of the new `physics_run_finalize` routine around line 78 to match what was done for other public routines in the `mpas_atmphys_manager` module?

Otherwise, the addition of this routine seems fine to me. I've also requested a review from @ldfowler58 .

jjguerrette:
This `physics_run_finalize` subroutine is used in OOPS in order to account for the way that the preparation for model integration is handled when multiple instances of MPAS are run in serial.  The alarms have to be removed and then created at the beginning of each integration.

For `atm_core_finalize`, the alarms are destroyed in `mpas_destroy_clock`.

ldfowler58:
Not being familiar with development of MPAS within the JEDI framework, it is difficult for me to completely understand the issue. I assume that calling physics_run_finalize in the stand-alone MPAS-Atmosphere does not affect these alarms? Has this been tested in the stand-alone version?

jjguerrette:
@ldfowler58 as the code stands, there is not a call to `physics_run_finalize` in the stand-alone MPAS-Atmosphere.  Normally MPAS-Atmosphere only needs to remove the physics alarms one time at the end of the simulation, which is accomplished with `mpas_destroy_clock`.  This new subroutine would have no impact in the run-time behavior of MPAS-Atmosphere, but it does add extra code that may be unnecessary.

We can probably just deallocate the alarms in the alarm linked list directly in JEDI instead of creating this extra subroutine to manage.

I'm going to close this PR for now.

PR #129, opened 2018-12-28
PIO version compatibility check

MiCurry:
This commit adds a conditional check in the Makefile to check to see if the
user has the correct version of PIO insatalled against the one which was
specified in the make command and gives an error message and suggestions for
fixes if a mismatch is found.

First, it checks to see if either PIO1 or PIO2 are installed and reports an
error if neither of them are installed. It then checks PIO1 or PIO2 depending
on whether or not USE_PIO2 is set to true.

--

@mgduda, let me know if you have any suggestions, additions/edits to the help messages if you have any! I included a few, but I'm sure I missed one that we talked about earlier.

As well, if anyone has any suggestions to the shell commands please let me know, I'm not too fluent in it still!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -424,32 +424,51 @@ llvm:
 CPPINCLUDES = 
 FCINCLUDES = 
 LIBS = 
-ifneq ($(wildcard $(PIO)/lib), ) # Check for newer PIO version
+
+#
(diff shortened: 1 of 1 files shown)


mark-petersen:
@mgduda I'm out early this week, will look at this when I get back in.

PR #130, opened 2019-01-08
SST Interval Update Check

MiCurry:
This commit adds a pre-run check that determines if the user has set the sea
surface update interval if `config_sst_update == true`.

If `config_sst_update == true` and the update interval is `none` the model will
abort its run. Conversely, if `config_sst_update == false` and the update
interval is set, a warning will be produced to inform the user that they can
turn on updates by setting `config_sst_update` to true in the namelist file.


src/core_atmosphere/physics/mpas_atmphys_control.F
```diff
@@ -493,6 +495,79 @@ subroutine physics_compatibility_check(dminfo, blockList, streamManager, ierr)
...
+ logical,pointer:: config_sst_update
+ character(len=StrKIND) :: stream_interval
+
+
+ call mpas_pool_get_config(blocklist % configs, 'config_sst_update', config_sst_update)
+ call MPAS_stream_mgr_get_property(streamManager, 'surface', MPAS_STREAM_PROPERTY_RECORD_INTV, stream_interval, &
+                                   direction=MPAS_STREAM_INPUT, ierr=ierr)
+
+ write(0,*) "stream_interval is: ", stream_interval
+ write(0,*) "config_sst_update is: ", config_sst_update
```
mgduda: Let's remove these `write` statements.

src/core_atmosphere/physics/mpas_atmphys_control.F
```diff
@@ -493,6 +495,79 @@ subroutine physics_compatibility_check(dminfo, blockList, streamManager, ierr)
...
+!> \details
+!>  This routine checks that the user has correctly set the SST update options,
+!>  by setting the `config_sst_update` namelist option equal to `true` .AND.
+!>  by setting the `input_interal` surface stream variable in streams.atmosphere
+!>  to a valid HH:MM:SS update time.
+!>
+ subroutine sst_update_check(blockList, streamManager, ierr)
+!=================================================================================================================
+
+ use mpas_timekeeping, only : mpas_split_string
```
mgduda: It looks like the use of `mpas_split_string` is unnecessary. Perhaps it was used during development for debugging or something else? In any case, it seems safe to remove this `use` statement.

src/core_atmosphere/physics/mpas_atmphys_control.F
```diff
@@ -493,6 +495,79 @@ subroutine physics_compatibility_check(dminfo, blockList, streamManager, ierr)
...
+                    messageType=MPAS_LOG_ERR)
+    call mpas_log_write("Please set the update interval to a valid DD_HH:MM:SS interval or turn 'config_sst_update'", &
+                    messageType=MPAS_LOG_ERR)
+    call mpas_log_write("to 'false'.", &
+                    messageType=MPAS_LOG_ERR)
+    call mpas_log_write('*******************************************************************************', &
+                        messageType=MPAS_LOG_ERR)
+    call mpas_log_write("FATAL ERROR: SST Interval Update Error - see above", &
+                        messageType=MPAS_LOG_CRIT)
+    ierr = ierr + 1
```
mgduda: Because `ierr` is declared as `intent(out)`, I think the compiler is allowed to let the variable be set to undefined values upon entry to the subroutine (with the expectation that it will always be assigned a value. As such, this statement may give `ierr` an undefined value. It appears that the `physics_compatibility_check` routine may have the same problem (but that is likely my fault!). It should be safe here to simply set `ierr` to 1.

mgduda: Update: I see that `ierr` is initialized to zero in the `physics_compatibility_check`, so there's no problem there.

src/core_atmosphere/mpas_atm_core.F
```diff
@@ -1245,6 +1245,9 @@ subroutine mpas_atm_run_compatibility(dminfo, blockList, streamManager, ierr)
       call physics_compatibility_check(dminfo, blockList, streamManager, local_ierr)
       ierr = ierr + local_ierr
 
+      ! SST Update Interval Check - /physics/mpas_atmphys_control.F
+      call sst_update_check(blockList, streamManager, ierr)
```
mgduda: I think `ierr` should be `local_ierr` in the argument list here.

src/core_atmosphere/physics/mpas_atmphys_control.F
```diff
@@ -493,6 +495,79 @@ subroutine physics_compatibility_check(dminfo, blockList, streamManager, ierr)
...
+    call mpas_log_write("file has not been set.", &
+                    messageType=MPAS_LOG_ERR)
+    call mpas_log_write("Please set the update interval to a valid DD_HH:MM:SS interval or turn 'config_sst_update'", &
+                    messageType=MPAS_LOG_ERR)
+    call mpas_log_write("to 'false'.", &
+                    messageType=MPAS_LOG_ERR)
+    call mpas_log_write('*******************************************************************************', &
+                        messageType=MPAS_LOG_ERR)
+    call mpas_log_write("FATAL ERROR: SST Interval Update Error - see above", &
+                        messageType=MPAS_LOG_CRIT)
```
mgduda: I think we may actually want to omit the critical error message here so that the error can bubble back up the call stack, possibly letting other errors be reported before the model stops. A non-zero return code from this routine will cause a non-zero return code in `mpas_atm_run_compatibility`, which will cause a non-zero return code in `atm_core_init`, which will finally stop the model in `mpas_init` (in `mpas_subdriver.F`).

MiCurry:
@mgduda I pushed those review changes as an amendment to my first commit and then pushed a new commit for the `gwdo_static_check`. Let me know if there is anything else you see wrong with it.

mgduda:
@ldfowler58 This PR is intended to stop users when they set `sst_update = true` but don't set an input interval for the `surface` stream in the `streams.atmosphere` file. Is there any particular issue that we should warn users about when they set up the `surface` stream, so that SST and seaice fields are read, but `sst_update = false`?

mgduda:
@ldfowler58 Just checking in to see whether you had any thoughts on the check in this PR to verify that the `surface` stream receives a valid interval whenever `update_sst = true` in the `&physics` namelist.

ldfowler58:
If "sst_update = true", then I agree that the model should abort if the update interval is none or the surface update file is not found.

If "sst_update = false" and the update interval is different than none and the surface update file exists, then a minimal warning can be added, but not really necessary. At some level, it is up to the user to be responsible for checking his own namelist.atmosphere.

I agree with this PR.

PR #133, opened 2019-01-16
Add Langmuir mixing parameterization via CVMix

qingli411:
This merge adds Langmuir mixing parameterization via CVMix
   
1. Change the CVMix repository to use the 'theory-wave' approximation of Li et al., 2017 to estimate the Langmuir enhancement factor from 10-meter wind, surface friction velocity and boundary layer depth
2. New entries in mpaso namelist allowing different options of Langmuir mixing parameterization


vanroekel:
Hi @qingli411, could you close this PR and make it into ocean/develop instead?  We need to merge there first.  If you need help with this, come by and I can show you.

PR #134, opened 2019-01-16
Add Langmuir mixing parameterization via CVMix

qingli411:
This merge adds Langmuir mixing parameterization via CVMix

1. Change the CVMix repository to use the 'theory-wave' approximation of Li et al., 2017 to estimate the Langmuir enhancement factor from 10-meter wind, surface friction velocity and boundary layer depth
2. New entries in mpaso namelist allowing different options of Langmuir mixing parameterization

MPAS-Analysis:

[G-case without Langmuir turbulence parameterization](http://portal.nersc.gov/project/acme/qingli/GMPAS-IAF_T62_oEC60to30v3_CTRL/)
```
config_cvmix_kpp_langmuir_mixing_opt = 'NONE'
config_cvmix_kpp_langmuir_entrainment_opt = 'NONE'
```

[G-case with Langmuir (LWF16)](http://portal.nersc.gov/project/acme/qingli/GMPAS-IAF_T62_oEC60to30v3_TWAV/)
```
config_cvmix_kpp_langmuir_mixing_opt = 'LWF16'
config_cvmix_kpp_langmuir_entrainment_opt = 'LWF16'
```

[G-case with Langmuir (LF17)](http://portal.nersc.gov/project/acme/qingli/GMPAS-IAF_T62_oEC60to30v3_LF17/)
```
config_cvmix_kpp_langmuir_mixing_opt = 'LWF16'
config_cvmix_kpp_langmuir_entrainment_opt = 'LF17'
```

[B-case with Langmuir (LF17, 40 years)](http://portal.nersc.gov/project/acme/qingli/20181129.Langmuir-LF17.A_WCYCL1850S.ne30_oECv3_ICG.anvil_years1-40/)
```
config_cvmix_kpp_langmuir_mixing_opt = 'LWF16'
config_cvmix_kpp_langmuir_entrainment_opt = 'LF17'
```

### DIFF
--- a/src/core_ocean/get_cvmix.sh
+++ b/src/core_ocean/get_cvmix.sh
@@ -1,15 +1,15 @@
 #!/bin/bash
 
 ## CVMix Tag for build
-CVMIX_TAG=lt_v1.3
+CVMIX_TAG=v0.94b-beta
 ## Subdirectory in CVMix repo to use
 CVMIX_SUBDIR=src/shared
 
 ## Available protocols for acquiring CVMix source code
-CVMIX_GIT_HTTP_ADDRESS=https://github.com/qingli411/CVMix-src.git
-CVMIX_GIT_SSH_ADDRESS=git@github.com:qingli411/CVMix-src.git
-CVMIX_SVN_ADDRESS=https://github.com/qingli411/CVMix-src/tags
-CVMIX_WEB_ADDRESS=https://github.com/qingli411/CVMix-src/archive
+CVMIX_GIT_HTTP_ADDRESS=https://github.com/CVMix/CVMix-src.git
+CVMIX_GIT_SSH_ADDRESS=git@github.com:CVMix/CVMix-src.git
+CVMIX_SVN_ADDRESS=https://github.com/CVMix/CVMix-src/tags
+CVMIX_WEB_ADDRESS=https://github.com/CVMix/CVMix-src/archive
 
 GIT=`which git`
 SVN=`which svn`
--- a/src/core_ocean/get_cvmix.sh
+++ b/src/core_ocean/get_cvmix.sh
@@ -1,15 +1,15 @@
 #!/bin/bash
 
 ## CVMix Tag for build
-CVMIX_TAG=v0.84-beta
+CVMIX_TAG=lt_v1.3
 ## Subdirectory in CVMix repo to use
 CVMIX_SUBDIR=src/shared
 
 ## Available protocols for acquiring CVMix source code
-CVMIX_GIT_HTTP_ADDRESS=https://github.com/CVMix/CVMix-src.git
-CVMIX_GIT_SSH_ADDRESS=git@github.com:CVMix/CVMix-src.git
-CVMIX_SVN_ADDRESS=https://github.com/CVMix/CVMix-src/tags
-CVMIX_WEB_ADDRESS=https://github.com/CVMix/CVMix-src/archive
+CVMIX_GIT_HTTP_ADDRESS=https://github.com/qingli411/CVMix-src.git
+CVMIX_GIT_SSH_ADDRESS=git@github.com:qingli411/CVMix-src.git
+CVMIX_SVN_ADDRESS=https://github.com/qingli411/CVMix-src/tags
+CVMIX_WEB_ADDRESS=https://github.com/qingli411/CVMix-src/archive
 
 GIT=`which git`
 SVN=`which svn`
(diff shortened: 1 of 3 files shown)


src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -872,6 +917,9 @@ subroutine ocn_vmix_cvmix_init(domain,err)!{{{
       call mpas_pool_get_config(ocnConfigs, 'config_cvmix_convective_diffusion', config_cvmix_convective_diffusion)
       call mpas_pool_get_config(ocnConfigs, 'config_cvmix_convective_viscosity', config_cvmix_convective_viscosity)
       call mpas_pool_get_config(ocnConfigs, 'config_cvmix_kpp_use_enhanced_diff', config_cvmix_kpp_use_enhanced_diff)
+      call mpas_pool_get_config(ocnConfigs, 'config_cvmix_kpp_langmuir_mixing_opt', config_cvmix_kpp_langmuir_mixing_opt)
+      call mpas_pool_get_config(ocnConfigs, 'config_cvmix_kpp_langmuir_entrainment_opt', &
+                                config_cvmix_kpp_langmuir_entrainment_opt)
```
vanroekel: please add a check below that user input values are acceptable values of your two character strings.  You can follow L1022 and following for an example check.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -596,6 +623,18 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, diagnost
...
+            if (config_cvmix_kpp_use_theory_wave .and. iceFraction(iCell) .lt. 0.05_RKIND) then
+               langmuirEnhancementFactor =  &
+                       cvmix_kpp_EFactor_model(windSpeed10m(iCell), &
+                                               surfaceFrictionVelocity(iCell), &
+                                               boundaryLayerDepth(iCell), &
+                                               cvmix_global_params)
+            else
+               langmuirEnhancementFactor = 1.0_RKIND
+            end if
+
```
vanroekel: please move this down after L651, you may miss some boundary layer changes in your computation

vanroekel:
@mark-petersen two questions on this PR
1) this should be BFB with ocean/develop as this is off by default and will be climate changing if on.  How do you want to test this for merging to MPAS-Model?  @qingli411 has run a core cycle with these changes and 40 years of a B-case.  Results are as expected.

2) is it okay to merge this even though we have to point to @qingli411 personal cvmix repo?  He has issued a PR into cvmix for the changes. Also asking @jonbob

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
qingli411: @vanroekel Added.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
qingli411: @vanroekel Done.

mark-petersen:
Also good to run a performance test to catch anything major.  Run your standard test with standard cvmix and your new option.  Compare the vertical mixing timers at the end of the ocean log file.  We should know if time is much larger.  This is the time to catch major performance changes, like loops nested in the wrong order.

src/core_ocean/get_cvmix.sh
```diff
@@ -1,15 +1,15 @@
...
 
 ## Available protocols for acquiring CVMix source code
-CVMIX_GIT_HTTP_ADDRESS=https://github.com/CVMix/CVMix-src.git
-CVMIX_GIT_SSH_ADDRESS=git@github.com:CVMix/CVMix-src.git
-CVMIX_SVN_ADDRESS=https://github.com/CVMix/CVMix-src/tags
-CVMIX_WEB_ADDRESS=https://github.com/CVMix/CVMix-src/archive
+CVMIX_GIT_HTTP_ADDRESS=https://github.com/qingli411/CVMix-src.git
+CVMIX_GIT_SSH_ADDRESS=git@github.com:qingli411/CVMix-src.git
+CVMIX_SVN_ADDRESS=https://github.com/qingli411/CVMix-src/tags
+CVMIX_WEB_ADDRESS=https://github.com/qingli411/CVMix-src/archive
```
mark-petersen: We can't use @qingli's repo here, as then all mainline E3SM simulations would depend on a nonstandard CVMix tag.  Worse, we need to be able to go back to any previous E3SM/master commit and run it in future years.  That means all submodules also need to be permanent.

The solution is to simply not make the above changes.  We can still merge this PR.  In order to run with @qingli411's non-default option, one would need to make the above change with your local repo copy.  

After Qing's changes are merged into CVMIX, we make an MPAS PR to update the above `CVMIX_TAG` line.

qingli411:
@vanroekel and @mark-petersen, here are some updates on this PR:

1. My changes to CVMix have been merged into CVMix with tag v0.94b-beta. I have updated the git repo of CVMix in this PR. 

2. I've conducted two tests (six hour tests at QU240 in COMPASS) on Grizzly with target 'ifort-gcc': 1) exact restart and 2) comparison across different partition numbers, which are all passed.

3. I've also compared the vertical mixing timers in these two tests and the timers are not significantly different from each other (within 3%). But I haven't tried comparing the timers in a longer run.

qingli411:
I just checked the vertical mixing timers (vmix imp) in my G test cases listed above. It takes 6.9% and 13.8% longer than the standard CVMix for vertical mixing with the first (LWF16) and the second (LF17) options, respectively, in a 5-year run. This is closer to what I was expecting, as there are additional computations in the second option and it should take longer time than the first option.

vanroekel:
@qingli411 could you also include the percentage increase in the overall ocean timestep timer?  I can't remember how much of the timestep is CVMix.

qingli411:
@vanroekel For ocean timestep timer ('o:se timestep') it's 0.16% and 0.64% increase respectively for the two options.

PR #138, opened 2019-01-22
Update to fix salinity restoring time interval

jonbob:
This PR updates the salinity restoring code to use the config_surface_salinity_monthly_restoring_compute_interval for the forcing time interval instead of config_dt. It does some minor clean-up and removes some unused code.

Tested with:
* ERS.T62_oEC60to30v3.GMPAS-IAF.anvil_intel

To work inside E3SM, it will require a corresponding modification to the ocn_comp_mct driver as well.

@maltrud , @vanroekel - I have a version that prints the time of the forcing update to the ocn log, if that would be an improvement. I should point out that not much goes to the log by default, so I'm hesitant to start filling it up...

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_surface_restoring.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_surface_restoring.F
@@ -211,7 +211,7 @@ contains
         character(len=strKind) :: timeStamp
 
         logical, intent(in) :: firstTimeStep
-        character(len=strKind), pointer :: config_dt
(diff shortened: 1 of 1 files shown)


maltrud:
@jonbob it looks good to me.  i could go either way on printing out the forcing time to the log file, leaning towards doing it.

mark-petersen:
I merged locally, this passes nightly regression suite on ocean stand-alone.  Of course that doesn't test time-varying salinity restoring, it's just a basic test.

@vanroekel when you are satisfied, please approve and I will merge into ocean/develop.

PR #141, opened 2019-01-28
Add Schoof friction law from Albany

matthewhoffman:
This merge introduces the Schoof friction law added to Albany.  It makes the corresponding interface changes and updates tests cases accordingly.


mperego:
I also noticed small changes relative to the baseline (1e-12 or smaller). I think it makes sense because we have changed the order of some operations.
I also agree it's better to merge this first and then try to fix the restart tests.

PR #143, opened 2019-02-01
Well-defined "extra" neighbors for cellsOnCell etc.

MiCurry:
This commit creates a well-defined "extra" neighbors for the fields:
`cellsOnCell`, `edgesOnCell`, `edgesOnEdge`, and `verticiesOnCell`.

In this commit all of the well-defined extra neighbors are set to 0. A future
commit could alter the value of any of these extra neighbors by altering the
parameters `UNUSED_CELL`, `UNUSED_EDGE` or `UNUSED_VERTEX` accordingly.

### DIFF
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -4787,6 +4787,10 @@ module mpas_stream_manager
 
         implicit none
 
+        integer, parameter :: UNUSED_CELL = 0
+        integer, parameter :: UNUSED_EDGE = 0
+        integer, parameter :: UNUSED_VERTEX = 0
+
         type (mpas_pool_type), pointer :: allFields
         type (mpas_pool_type), pointer :: streamFields
 
@@ -4909,7 +4913,7 @@ module mpas_stream_manager
                            cellsOnCell % array(j,i) = indexToCellID % array(cellsOnCell_ptr % array(j,i))
                        end do
 
-                       cellsOnCell % array(nEdgesOnCell%array(i)+1:maxEdges,i) = nCells+1
+                       cellsOnCell % array(nEdgesOnCell%array(i)+1:maxEdges,i) = UNUSED_CELL
                    end do
 
                    cellsOnCell => cellsOnCell % next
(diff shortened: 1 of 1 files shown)


xylar:
@MiCurry, I think `nCells+1`, etc. were deliberately used as "invalid" values.  @mark-petersen or @mgduda, could you comment on why these values were chosen instead of zero?

mgduda:
I think the original intent was to use `nCells+1` so that any references to unused elements of, e.g., `cellsOnCell` would be directed to the "garbage cell".

However, I don't think the changes in this PR will have any impact on code *inside of MPAS*: the changes simply set the values for unused indexing elements when these indexing fields are written to output files. After writing, e.g., the `cellsOnCell` field that has been reindexed to global index space by the `prewrite_reindex` routine, we reset the field to the local index space with the `postwrite_reindex` routine, thereby preserving whatever value was placed in unused elements by the `postread_reindex` routine.

With this PR, the intent is that the indexing fields (`cellsOnCell`, `verticesOnCell`, `edgesOnCell`, and `edgesOnEdge`) should be identical in output files produced by model runs using different MPI task counts. With the current code, we use the task-local value of `nCells+1` to fill in unused indexing elements, leading to differences in these fields when they are written to output files. For example, if a task owns a pentagon and has `nCells` = 200 locally, we get values of 201 for the unused, sixth element of the `cellsOnCell` field for that pentagon; running again with a different decomposition that gives the task 198 cells, we then get values of 199 in the output field.

We could have chosen the global `nCells+1` value to fill unused elements of the indexing arrays, but we don't already have that available (as far as I'm aware), and rather than performing a global sum just to fill in output arrays, a value of zero seemed the next best choice.

Because we're changing just the `prewrite_reindex` routine, but not the `postwrite_reindex` or `postread_reindex` routine, I don't think there should be any impact on code within MPAS. But, post-processing tools working with model output file could be impacted -- though that seems unlikely, since the unused element value was a local `nCells+1` value that depended on which task owned the cell with unused elements.

mgduda:
@mark-petersen Did you want a chance to look over the changes in this PR?

PR #144, opened 2019-02-05
Update ISOMIP+ test case and add new parameter studies

xylar:
Variables are now averaged with `timeSeriesStatsMonthlyOutput` to ISOMIP+ test cases.  Evaporation rates are computed based on monthly mean melt rates.  Visualization and output for MISOMIP are computed from this output as well.

New default parameter values are used in ISOMIP+ based on redoing the Ocean0 parameter study (varying the heat and salt transfer coefficients and finding the value that gives 30 m/yr mean melt rate for `landIceDraft < -300`).

Scripts have been added for generating 3 different parameter studies that will be used in Gwyther et al., "The importance of vertical processes and resolution on simulated basal melt of an ice shelf: a multi-model study within the ISOMIP+ framework", in prep.

PR #146, opened 2019-02-05
Fix core count in compass regression suites

### LINKED ISSUE #145: Compass test suite reports max. MPI tasks for all configurations, not just those used in driver
When setting up a Compass test suite, the maximum number of MPI tasks is reported as part of the final output after the suite has been set up.  However, the driver script (`config_driver.xml` that produces `run_test.py`) may only include a subset of the configurations available in a given test case (e.g. `forwrad_short` but not `forward`).  In many cases, it might be desirable to have a larger number of MPI tasks be used for longer runs but only a few for regression testing.

`manage_regression_suite.py` should, therefore, report the maximum MPI tasks for only those configurations that are part of `config_driver.xml`.

xylar:
The core count now only involves those configurations that are part of the driver script of each test case, not all configurations.

Without this fix, core counts are sometimes far too high, making it hard to decide how many nodes are needed to run the regression suite.

closes #145

### DIFF
--- a/testing_and_setup/compass/manage_regression_suite.py
+++ b/testing_and_setup/compass/manage_regression_suite.py
@@ -354,6 +354,19 @@ def summarize_suite(suite_tag):  # {{{
 
             test_path = '{}/{}/{}/{}'.format(test_core, test_configuration,
                                              test_resolution, test_test)
+            driver_path = '{}/config_driver.xml'.format(test_path)
+            config_tree = ET.parse(driver_path)
+            config_root = config_tree.getroot()
+
+            cases = []
+            assert(config_root.tag == 'driver_script')
+            for case in config_root.iter('case'):
+                name = case.attrib['name']
+                cases.append(name)
+
+            del config_root
+            del config_tree
+
             # Loop over all files in test_path that have the .xml extension.
             for file in os.listdir('{}'.format(test_path)):
                 if fnmatch.fnmatch(file, '*.xml'):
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

In combination with #144, I tested that the regression test suite  `ocean/regression_suites/land_ice_fluxes.xml` reports needing 16 cores rather than 136 cores (only required for the `forward` configuration, which is not part of the driver script).

xylar:
@mark-petersen and @matthewhoffman, I would appreciate any feedback you might have and if you could verify that this doesn't break your standard test suites.

matthewhoffman:
@xylar , I'm not familiar with this part of the COMPASS code, but I'll make sure our test suite still works with it sometime in the next few days.

xylar:
@matthewhoffman, once you've had a chance to make sure this doesn't break your code, I'd like to merge.  Any idea when that might be?

xylar:
@mark-petersen, could you remind me what we consider best practices in terms of merging to `develop`?  I was about to hit the "Merge pull request" button but I think that's not allowed.

mark-petersen:
Correct.  The button does not use --no-ff, so can end up with odd things.  Do this: check out the develop branch locally, and:
```
git merge --no-ff my_branch
```
Then use the first comment on this PR for the commit message.  I like to put the PR number on the merge subject, so it is easy to find from the git log.

xylar:
@mark-petersen, Interesting!  I think the merge button normally *does* use `--no-ff`: https://help.github.com/en/articles/merging-a-pull-request.  But it looks like there may be circumstances where it does a fast-forward so I guess the command line is safest.  I've never seen a fast-forward commit in all my 300+ merges into MPAS-Analysis with the button...

PR #147, opened 2019-02-05
Fixes for bfb restart and partitions tests in E3SM

mark-petersen:
There are three changes.  
1. Widen halo computation ring for kinetic energy.
2. Widen halo computation ring for melt fluxes.
3. Add halo exchange for surfaceFrictionVelocity

When running E3SM with melt fluxes on, the first is needed for bfb restarts (https://github.com/E3SM-Project/E3SM/issues/2702) and all three for bfb partition tests (https://github.com/E3SM-Project/E3SM/issues/2739).

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -313,8 +313,8 @@ contains
       end do
       !$omp end do
 
-      ! Need 0 and 1 halo cells
-      nCells = nCellsArray( 2 )
+      ! Need 0,1,2 halo cells
+      nCells = nCellsArray( 3 )
       !
       ! Compute divergence, kinetic energy, and vertical velocity
       !
@@ -379,8 +379,8 @@ contains
       end do
       !$omp end do
 
-      nCells = nCellsArray( size(nCellsArray) )
-
+      ! Need 0,1,2 halo cells
+      nCells = nCellsArray( 3 )
       !
       ! Compute kinetic energy
       !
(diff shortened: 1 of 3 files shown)


src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -406,6 +407,7 @@ subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnostic
            end do
          end do
          !$omp end do
+         call mpas_threading_barrier()
```
amametjanov: This barrier is not necessary, because there is an implied synchronization among threads at the end of the loop. When needed, such barriers can be removed with
```
!$omp end do nowait
```
and threads do not wait after exiting the loop.

mark-petersen: Thanks for watching this repo, @amametjanov! That is really helpful to know.  I added that thread barrier because I thought it was needed, not because I saw a difference in a test.

mark-petersen:
@jonbob These are the fixes for https://github.com/E3SM-Project/E3SM/issues/2702.  Please approve when you are ready, and I will merge in.

mark-petersen:
Passes bfb match with ocean/develop using gnu with nightly regression suite.  That includes restart and partition test in MPAS-O standalone.

src/core_ocean/shared/mpas_ocn_diagnostics.F
amametjanov: Update: [this PR](https://github.com/amametjanov/MPAS-Model/pull/3) (a follow-up to PR #152) should reduce the number thread barriers in the remaining ocean sources and improve threading throughput. Thanks.

PR #151, opened 2019-02-06
Tidy up Makefiles for framework and atmosphere core

mgduda:
This PR is a work-in-progress to clean up Makefiles by fixing automatic variables in implicit
rules and correcting dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -401,41 +401,37 @@ FCINCLUDES =
 LIBS = 
 ifneq ($(wildcard $(PIO)/lib), ) # Check for newer PIO version
 ifeq "$(USE_PIO2)" "true"
-	CPPINCLUDES = -DUSE_PIO2 -I$(PIO)/include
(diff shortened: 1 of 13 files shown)


mark-petersen:
This passes MPAS-O nightly regression suite on grizzly with gnu/optimized, is bfb with previous.

I am not able to compile with `ifort` on grizzly. It is not able to find the pio include path. I messed around with the include lines that were changed, but was not able to figure it out.  

@mgduda it looks like this PR is not on your schedule for the 6.2 release, so I can ignore this problem for right now?

mgduda:
@mark-petersen Thanks so much for taking a look! This PR is not slated for inclusion in the v6.2 release, and at the moment only serves as a collection place for some Makefile changes that had come up when working on out-of-source builds of MPAS-A for CESM. It would be no problem to ignore this PR for now, and once I've worked on it a bit more, I'll check-in with you, @matthewhoffman, and others.

PR #152, opened 2019-02-07
Update ocean BGC threading

amametjanov:
Update ocean BGC threading to make multi-threaded runs BFB with single-threaded runs in coupled E3SM cases.

Checked with E3SM test `SMS.ne30_oECv3.BGCEXP_CNTL_CNPRDCTC_1850.anvil_intel.clm-bgcexp` on latest `maint-1.1` branch:
* pure-MPI run on 20 nodes with 18 tasks/node, all components stacked
* multi-threaded run on 20 nodes with 18x2 tasks x threads

Coupler history files `cpl.hi.0001-01-02-00000.nc` at the end of 1-day runs are BFB.

[BFB]

Fixes E3SM-Project/E3SM#1931

### DIFF
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -439,7 +439,6 @@ module ocn_time_integration_rk4
                     * highFreqThicknessTend(:, iCell)
               end do
               !$omp end do
-              call mpas_threading_barrier()
               block => block % next
            end do
 
@@ -642,7 +641,6 @@ module ocn_time_integration_rk4
             normalTransportVelocity(:, iEdge) = normalVelocityNew(:, iEdge)
          end do
          !$omp end do
-         call mpas_threading_barrier()
 
          ! Compute normalGMBolusVelocity and the tracer transport velocity
          if (config_use_standardGM) then
@@ -1161,7 +1159,6 @@ module ocn_time_integration_rk4
          normalTransportVelocity(:, iEdge) = normalVelocityProvis(:, iEdge)
       end do
       !$omp end do
-      call mpas_threading_barrier()
 
       ! Compute normalGMBolusVelocity, relativeSlope and RediDiffVertCoef if respective flags are turned on
       if (config_use_standardGM) then
@@ -1176,7 +1173,6 @@ module ocn_time_integration_rk4
          end do
          !$omp end do
       end if
-      call mpas_threading_barrier()
       ! ------------------------------------------------------------------
       ! End: Accumulating various parametrizations of the transport velocity
       ! ------------------------------------------------------------------
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
@@ -560,7 +560,7 @@ module ocn_time_integration_split
 
                allocate(uTemp(nVertLevels))
 
-               !$omp do schedule(runtime) private(cell1, cell2, k, normalThicknessFluxSum, thicknessSum)
+               !$omp do schedule(runtime)
                do iEdge = 1, nEdges
                   cell1 = cellsOnEdge(1,iEdge)
                   cell2 = cellsOnEdge(2,iEdge)
@@ -813,7 +813,7 @@ module ocn_time_integration_split
                      nEdges = nEdgesPtr
                      nEdges = nEdgesArray( edgeHaloComputeCounter )
 
-                     !$omp do schedule(runtime) private(cell1, cell2, CoriolisTerm, i, eoe)
+                     !$omp do schedule(runtime)
                      do iEdge = 1, nEdges
 
                         temp_mask = edgeMask(1, iEdge)
@@ -1025,7 +1025,7 @@ module ocn_time_integration_split
 
                    nEdges = nEdgesArray( edgeHaloComputeCounter )
 
-                   !$omp do schedule(runtime) private(cell1, cell2, eoe, CoriolisTerm, i, sshCell1, sshCell2)
+                   !$omp do schedule(runtime)
                    do iEdge = 1, nEdges
 
                      ! asarje: added to avoid redundant computations based on mask
@@ -1338,7 +1338,7 @@ module ocn_time_integration_split
                   useVelocityCorrection = 0
                endif
 
-               !$omp do schedule(runtime) private(k, normalThicknessFluxSum, thicknessSum, normalVelocityCorrection)
+               !$omp do schedule(runtime)
                do iEdge = 1, nEdges
 
                   ! velocity for normalVelocityCorrectionection is normalBarotropicVelocity + normalBaroclinicVelocity + uBolus
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -285,12 +285,13 @@ contains
       !    used -1e34 so error clearly occurs if these values are used.
       !
 
-      !$omp single
+      !$omp master
       normalVelocity(:,nEdges+1) = -1e34
       layerThickness(:,nCells+1) = -1e34
       activeTracers(indexTemperature,:,nCells+1) = -1e34
       activeTracers(indexSalinity,:,nCells+1) = -1e34
-      !$omp end single
+      !$omp end master
+      call mpas_threading_barrier()
 
       call ocn_relativeVorticity_circulation(relativeVorticity, circulation, meshPool, normalVelocity, err)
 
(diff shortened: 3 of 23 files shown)


src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
```diff
@@ -439,7 +439,6 @@ subroutine ocn_time_integrator_rk4(domain, dt)!{{{
                     * highFreqThicknessTend(:, iCell)
               end do
               !$omp end do
-              call mpas_threading_barrier()
```
amametjanov: Unnecessary/duplicate.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -559,7 +559,7 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
 
                allocate(uTemp(nVertLevels))
 
-               !$omp do schedule(runtime) private(cell1, cell2, k, normalThicknessFluxSum, thicknessSum)
+               !$omp do schedule(runtime)
```
amametjanov: `private` can be omitted, `firstprivate` is default.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -285,12 +285,13 @@ subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnostic
...
 
-      !$omp single
+      !$omp master
       normalVelocity(:,nEdges+1) = -1e34
       layerThickness(:,nCells+1) = -1e34
       activeTracers(indexTemperature,:,nCells+1) = -1e34
       activeTracers(indexSalinity,:,nCells+1) = -1e34
-      !$omp end single
+      !$omp end master
+      call mpas_threading_barrier()
```
amametjanov: Bug: `omp master` is needed here, because of the `firstprivate` default.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1450,8 +1449,7 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
                                          densitySurfaceDisplaced, err, thermalExpansionCoeff, salineContractionCoeff, &
                                          timeLevel)
 
-      !$omp do schedule(runtime) private(invAreaCell, deltaVelocitySquared, i, iEdge, factor, delU2, fracAbsorbed, &
-      !$omp                              fracAbsorbedRunoff)
+      !$omp do schedule(runtime)
```
amametjanov: Bug: continuing `&` is missing on the second line:
```
!$omp do ... &
!$omp& ...
```

src/core_ocean/shared/mpas_ocn_frazil_forcing.F
```diff
@@ -451,9 +451,7 @@ subroutine ocn_frazil_forcing_build_arrays(domain, meshPool, forcingPool, diagno
       nCells = nCellsArray( 2 )
 
       ! loop over all columns
-      !$omp do schedule(runtime) private(kBottomFrazil, k, columnTemperatureMin, sumNewFrazilIceThickness, &
-      !$omp                              oceanFreezingTemperature, potential, freezingEnergy, meltingEnergy, &
-      !$omp                              newFrazilIceThickness, meltedFrazilIceThickness)
+      !$omp do schedule(runtime)
```
amametjanov: Bug

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -418,11 +418,10 @@ subroutine ocn_gm_compute_Bolus_velocity(diagnosticsPool, meshPool, scratchPool)
       ! use relativeSlopeTaperingCell as a temporary space for smoothing of relativeSlopeTopOfCell
       relativeSlopeTaperingCell = relativeSlopeTopOfCell
       do iter = 1, 5
-         !$omp barrier
```
amametjanov: Implied barriers make this unnecessary.

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -673,6 +676,9 @@ subroutine ocn_tend_tracer(tendPool, statePool, forcingPool, diagnosticsPool, me
                      nTracerGroup, nCellsSolve, zMid, indexTemperature, indexSalinity, tracerGroupSurfaceFlux, err)
                endif
 
+               !$omp end master
+               call mpas_threading_barrier()
+
```
amametjanov: Bug. This will run `ocn_tracer_*_compute` single-threaded.

rljacob:
What is the status of this PR?   An E3SM PR is waiting on it.

jonbob:
It's next on my list, but complicated by having to come in through mpas

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
mark-petersen: @amametjanov could you explain all the changes that remove private(var1, var2,...)?  

I thought we always need to declare indices within an openmp loop to be private, as in the removed line.  It looks like my understand of this is incorrect.  Doesn't each thread need it's own copy of index variables, so the private clause is required?

amametjanov: Because of `!$omp parallel default(firstprivate)`, all variables inside `!$omp do` regions are `firstprivate` by default: meaning each thread's copy/storage of a private variable inherits master thread's value. There is no need to make these variables here `private`, because their values are defined/initialized prior to the first use.

Overall, two possible scenarios in an omp-do region: def-before-use, use-before-def. If `firstprivate` variable `var1` is defined before use:
* `private(var1)` is redundant; `var1` is overwritten anyway (this is the reason for removal above)

If `firstprivate` `var1` is used before def:
* `private(var1)` is a bug, intent was to inherit master thread's value

Instead of enumerating all locally defined variables in private clauses, I'd like to simplify and omit them when possible. Otherwise, when a new variable is added to an omp-do region, it also needs to be added to the private clause; when a variable is removed from a region, it also needs to be removed from the private clause. This is hard to do/enforce.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -812,7 +812,7 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
                      nEdges = nEdgesPtr
                      nEdges = nEdgesArray( edgeHaloComputeCounter )
 
-                     !$omp do schedule(runtime) private(cell1, cell2, CoriolisTerm, i, eoe)
+                     !$omp do schedule(runtime)
                      do iEdge = 1, nEdges
 
                         temp_mask = edgeMask(1, iEdge)
```
amametjanov: Here is an example: `temp_mask` was added to the region, but not declared in the private clause. So `private(cell1, cell2, CoriolisTerm, i, eoe)` is now outdated/incomplete.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
mark-petersen: Thanks for the explanation.  Let's make sure I understand - is that another way to say it: 
- In MPAS, threads are spawned upon initialization, and run throughout.
- Upon entry to this subroutine, each thread declares its own set of local variables such as `cell1, cell2, k, normalThicknessFluxSum, thicknessSum`, so the memory space is unique to each thread.
- At each loop, `private` is not needed because each thread already has its own copy.

But why do you say "each thread's copy of a private variable inherits master thread's value"? When does that occur - only on initialization?

amametjanov: Yes.

The copying occurs just before entering an omp-do loop.

src/core_ocean/shared/mpas_ocn_equation_of_state_jm.F
```diff
@@ -309,7 +309,7 @@ subroutine ocn_equation_of_state_jm_density(meshPool, scratchPool, nCells, k_dis
          enddo
       endif
 
-      !$omp do schedule(runtime) private(k, DRDT0, DKDT, DRHODT, DRDS0, DKDS, DRHODS)
+      !$omp do schedule(runtime)
```
mark-petersen: FYI, I merged this locally and got conflicts with the previous local merge #170. I resolved the conflicts by using the version from #170 but removing all omp private statements from @philipwjones's EOS update.

mark-petersen:
@amametjanov could you rebase on the head of ocean/develop?

After that, I will make a temporary MPAS branch with #152, #170, #194 and make an E3SM branch for people to test with all these changes. That way we have the opportunity to edit these PRs while testing in E3SM. Any other threading or performance changes that are ready could also be tested.

amametjanov:
Rebased

rljacob:
This fix needs to go to the maint-1.1 branch of E3SM which is not using the head of MPAS-Dev.  Is there a branch here tracking the version of MPAS used in E3SM maint-1.1?

jonbob:
@rljacob - yes, we do have branches for each E3SM maint-1.X

PR #156, opened 2019-02-12
Eliminate need for -DUNDERSCORE by using F2003's ISO_C_BINDING

MiCurry:
This pull requests creates `ISO_C_BINDING` interfaces for c functions in the atmosphere core as to eliminate the need for `-DUNDERSCORE` within the makefile (Though `-DUNDERSCORE` is still present within the makefile).

It also updates each C file to be more standards complaint. For instance by passing in arrays using the `c_loc` `ISO_C_BINDING` function (see`read_geogrid.c`). As well, where applicable it uses the `mpas_f_to_c_string` from `mpas_c_interfacing` to create null terminated strings for C, which eliminates the need to pass in the length of a string in some circumstances (`read_geogrid.c` and `pool_hash.c`). 

It also passes in most elements by value (where possible), to avoid the need to deference them in C.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_gwd.F
+++ b/src/core_init_atmosphere/mpas_init_atm_gwd.F
@@ -7,14 +7,34 @@
 !
 module mpas_init_atm_gwd
 
+   use iso_c_binding, only : c_char, c_int, c_float, c_ptr, c_loc
+
    use mpas_derived_types, only : MPAS_LOG_ERR
    use mpas_framework
    use mpas_timekeeping
    use mpas_log, only : mpas_log_write
+   use mpas_c_interfacing, only : mpas_f_to_c_string
 
    public :: compute_gwd_fields
 
    private
+
+   interface
+      subroutine read_geogrid(fname, rarray, nx, ny, nz, isigned, endian, &
+                              scalefactor, wordsize, status) bind(C)
+         use iso_c_binding, only : c_char, c_int, c_float, c_ptr
+         character (c_char), dimension(*), intent(in) :: fname
+         type (c_ptr), value :: rarray
+         integer (c_int), intent(in), value :: nx
+         integer (c_int), intent(in), value :: ny
+         integer (c_int), intent(in), value :: nz
+         integer (c_int), intent(in), value :: isigned
+         integer (c_int), intent(in), value :: endian
+         real (c_float), intent(in), value :: scalefactor
+         integer (c_int), intent(in), value :: wordsize
+         integer (c_int), intent(inout) :: status
+      end subroutine read_geogrid
+   end interface
 
    integer, parameter :: I1KIND = selected_int_kind(2)
 
@@ -314,14 +334,17 @@ module mpas_init_atm_gwd
       integer, parameter :: tile_y = 1200       ! y-dimension of each tile of global 30-arc-second topography
       integer, parameter :: tile_bdr = 3        ! number of layers of border/halo points surrounding each tile
 
-      integer :: istatus
+      integer (c_int) :: istatus
       integer :: ix, iy, ishift, ix_shift
-      integer :: isigned, endian, wordsize, nx, ny, nz
-      real (kind=R4KIND) :: scalefactor
-      real (kind=R4KIND), dimension(:,:,:), allocatable :: tile
+      integer (c_int) :: isigned, endian, wordsize, nx, ny, nz
+      real (c_float) :: scalefactor
+      real (c_float), dimension(:,:,:), pointer, contiguous :: tile
+      type (c_ptr) :: tile_ptr
       character(len=StrKIND) :: filename
+      character(kind=c_char), dimension(StrKIND+1) :: c_filename
 
       allocate(tile(tile_x+2*tile_bdr,tile_y+2*tile_bdr,1))
+      tile_ptr = c_loc(tile)
 
       isigned  = 1
       endian   = 0
@@ -345,7 +368,8 @@ module mpas_init_atm_gwd
       do ix=1,topo_x,tile_x
          write(filename,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(path)//trim(sub_path), ix, '-', (ix+tile_x-1), '.', &
                                                                                       iy, '-', (iy+tile_y-1)
-         call read_geogrid(filename, len_trim(filename), tile, nx, ny, nz, isigned, endian, &
+         call mpas_f_to_c_string(filename, c_filename)
+         call read_geogrid(c_filename, tile_ptr, nx, ny, nz, isigned, endian, &
                            scalefactor, wordsize, istatus)
          if (istatus /= 0) then
             call mpas_log_write('Error reading topography tile '//trim(filename), messageType=MPAS_LOG_ERR)
@@ -389,14 +413,17 @@ module mpas_init_atm_gwd
       integer, parameter :: tile_x = 1200       ! x-dimension of each tile of global 30-arc-second landuse
       integer, parameter :: tile_y = 1200       ! y-dimension of each tile of global 30-arc-second landuse
 
-      integer :: istatus
+      integer (c_int) :: istatus
       integer :: ix, iy
-      integer :: isigned, endian, wordsize, nx, ny, nz
-      real (kind=R4KIND) :: scalefactor
-      real (kind=R4KIND), dimension(:,:,:), allocatable :: tile
+      integer (c_int) :: isigned, endian, wordsize, nx, ny, nz
+      real (c_float) :: scalefactor
+      real (c_float), dimension(:,:,:), pointer, contiguous :: tile
+      type (c_ptr) :: tile_ptr
       character(len=StrKIND) :: filename
+      character(kind=c_char), dimension(StrKIND+1) :: c_filename
 
       allocate(tile(tile_x,tile_y,1))
+      tile_ptr = c_loc(tile)
 
       isigned  = 1
       endian   = 0
@@ -410,7 +437,8 @@ module mpas_init_atm_gwd
       do ix=1,topo_x,tile_x
          write(filename,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(path)//'/landuse_30s/', ix, '-', (ix+tile_x-1), '.', &
                                                                                       iy, '-', (iy+tile_y-1)
-         call read_geogrid(filename, len_trim(filename), tile, nx, ny, nz, isigned, endian, &
+         call mpas_f_to_c_string(filename, c_filename)
+         call read_geogrid(c_filename, tile_ptr, nx, ny, nz, isigned, endian, &
                            scalefactor, wordsize, istatus)
          if (istatus /= 0) then
             call mpas_log_write('Error reading landuse tile '//trim(filename))
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -15,8 +15,11 @@
  use mpas_log, only : mpas_log_write
  use init_atm_hinterp
  use init_atm_llxy
+ use mpas_c_interfacing, only : mpas_f_to_c_string
 
  use mpas_atmphys_utilities
+
+ use iso_c_binding, only : c_char, c_int, c_float, c_loc, c_ptr
 
  implicit none
  private
@@ -25,6 +28,23 @@
           init_atm_check_read_error, &
           nearest_cell,              &
           sphere_distance
+
+ interface
+    subroutine read_geogrid(fname, rarray, nx, ny, nz, isigned, endian, &
+                            scalefactor, wordsize, status) bind(C)
+       use iso_c_binding, only : c_char, c_int, c_float, c_ptr
+       character (c_char), dimension(*), intent(in) :: fname
+       type (c_ptr), value :: rarray
+       integer (c_int), intent(in), value :: nx
+       integer (c_int), intent(in), value :: ny
+       integer (c_int), intent(in), value :: nz
+       integer (c_int), intent(in), value :: isigned
+       integer (c_int), intent(in), value :: endian
+       real (c_float), intent(in), value :: scalefactor
+       integer (c_int), intent(in), value :: wordsize
+       integer (c_int), intent(inout) :: status
+    end subroutine read_geogrid
+ end interface
 
  contains
 
@@ -41,6 +61,7 @@
  type(proj_info):: proj
 
  character(len=StrKIND) :: fname
+ character(kind=c_char), dimension(StrKIND+1) :: c_fname
  character(len=StrKIND), pointer :: config_geog_data_path
  character(len=StrKIND), pointer :: config_landuse_data
  character(len=StrKIND), pointer :: config_topo_data
@@ -49,16 +70,17 @@
 
  integer:: isice_lu,iswater_lu,ismax_lu
 
- integer:: nx,ny,nz
- integer:: endian,isigned,istatus,wordsize
+ integer(c_int):: nx,ny,nz
+ integer(c_int):: endian,isigned,istatus,wordsize
  integer:: i,j,k
  integer:: iCell,iEdge,iVtx,iPoint,iTileStart,iTileEnd,jTileStart,jTileEnd
  integer,dimension(5) :: interp_list
  integer,dimension(:),allocatable  :: nhs
  integer,dimension(:,:),allocatable:: ncat
       
- real(kind=4):: scalefactor
- real(kind=4),dimension(:,:,:),allocatable:: rarray
+ real(kind=c_float):: scalefactor
+ real(kind=c_float),dimension(:,:,:),pointer,contiguous :: rarray
+ type(c_ptr) :: rarray_ptr
 
  real(kind=RKIND):: start_lat
  real(kind=RKIND):: start_lon
@@ -239,6 +261,8 @@
  nhs(:) = 0
  ter(:) = 0.0
 
+ rarray_ptr = c_loc(rarray)
+
  start_lat = -89.99583
  select case(trim(config_topo_data))
     case('GTOPO30')
@@ -266,8 +290,9 @@
        write(fname,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(geog_data_path)//trim(geog_sub_path), &
                     iTileStart,'-',iTileEnd,'.',jTileStart,'-',jTileEnd
        call mpas_log_write(trim(fname))
+       call mpas_f_to_c_string(fname, c_fname)
 
-       call read_geogrid(fname,len_trim(fname),rarray,nx,ny,nz,isigned,endian, &
+       call read_geogrid(c_fname,rarray_ptr,nx,ny,nz,isigned,endian, &
                          scalefactor,wordsize,istatus)
        call init_atm_check_read_error(istatus, fname)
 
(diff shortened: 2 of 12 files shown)


src/core_init_atmosphere/mpas_init_atm_gwd.F
```diff
@@ -314,14 +334,17 @@ function read_global_30s_topo(path, sub_path) result(iErr)
...
       integer :: ix, iy, ishift, ix_shift
-      integer :: isigned, endian, wordsize, nx, ny, nz
-      real (kind=R4KIND) :: scalefactor
-      real (kind=R4KIND), dimension(:,:,:), allocatable :: tile
+      integer (c_int) :: isigned, endian, wordsize, nx, ny, nz
+      real (c_float) :: scalefactor
+      real (c_float), dimension(:,:,:), pointer, contiguous :: tile
+      type (c_ptr) :: tile_ptr
       character(len=StrKIND) :: filename
+      character(len=StrKIND, kind=c_char) :: c_filename
```
mgduda: I think `c_filename` should be have size `StrKIND+1` to allow for the addition of a C null character in the `mpas_f_to_c_string` routine. Also, I think `c_filename` should be declared as an array, i.e.,
```
character(kind=c_char), dimension(StrKIND+1) :: c_filename
```
I'll check on this last point (declaration as an array), though, as my memory of interoperability of C character arrays may be fuzzy.

src/core_init_atmosphere/mpas_init_atm_gwd.F
```diff
@@ -7,15 +7,35 @@
...
 
    public :: compute_gwd_fields
 
    private
 
+   interface
+      subroutine read_geogrid(fname, rarray, nx, ny, nz, &
+                             & isigned, endian, scalefactor, wordsize, status) bind(C)
+         use iso_c_binding, only : c_char, c_int, c_float, c_ptr
+         character (c_char), intent(in) :: fname
```
mgduda: Depending on how we ultimately need to declare `c_filename`, we may need to adjust the declaration of `fname` here.

src/framework/mpas_pool_routines.F
```diff
@@ -5992,5 +5992,30 @@ integer function pool_get_member_decomp_type(dimName) result(decompType)!{{{
...
+           character (c_char), intent(in) :: key
+        end subroutine c_pool_hash
+      end interface
+
+      integer (c_int), intent(inout) :: hash
+      character(len=*), intent(in) :: key
+
+      character(len=StrKIND, kind=c_char) :: c_key
+
+      call mpas_f_to_c_string(key, c_key)
```
mgduda: Since we're now copying the Fortran string to a C character array, it might be interesting to check the timing to see whether this impacts the performance of the pool hashing in any significant way. It could that, if Fortran's `len_trim` is an O(n) operation, there's essentially no impact in the O(n) operation of copying to a null-terminated C array.

src/framework/pool_hash.c
```diff
@@ -1,21 +1,25 @@
...
+
+void c_pool_hash(int* hash, char* key)
 {
 	int i;
 	unsigned int whash;
 
 	whash = 0;
 
-	for (i=0; i<(*len); i++) {
+	for (i=0; i < strlen(key); i++) {
```
mgduda: If we are guaranteed that the `key` that is passed in is a null-terminated string, then we should be able to avoid an extra O(n) operation here by looping until we hit a `\0` character.

src/core_init_atmosphere/mpas_init_atm_gwd.F
```diff
@@ -7,15 +7,35 @@
...
    use mpas_log, only : mpas_log_write
+   use mpas_c_interfacing, only : mpas_f_to_c_string
 
    public :: compute_gwd_fields
 
    private
 
+   interface
+      subroutine read_geogrid(fname, rarray, nx, ny, nz, &
+                             & isigned, endian, scalefactor, wordsize, status) bind(C)
```
mgduda: Stylistically, in MPAS we tend to only put the `&` continuation character at the end of the line to be continued. For consistency, it may be nice to remove the beginning `&` character here and anywhere else this occurs (e.g., in the comments at the top of `read_geogrid.c`).

src/core_init_atmosphere/read_geogrid.c
```diff
@@ -20,55 +20,65 @@
...
+       integer (c_int), intent(in), value :: nz
+       integer (c_int), intent(in), value :: isigned
+       integer (c_int), intent(in), value :: endian
+       real (c_float), intent(in), value :: scalefactor
+       integer (c_int), intent(in), value :: wordsize
+       integer (c_int), intent(inout) :: status
+    end subroutine read_geogrid
+ end interface
+
+ use iso_c_binding, only : c_char, c_int, c_float, c_ptr, c_loc
```
mgduda: Since `use` statements must come before `interface` statements in Fortran, I'd propose we either remove this line or move it above the example interface block.

src/framework/mpas_stream_manager.F
```diff
@@ -4148,8 +4153,18 @@ end subroutine mpas_build_stream_filename !}}}
...
 
+        use iso_c_binding, only : c_int, c_char
+
         implicit none
 
+        interface
+          subroutine gen_random(len, id) bind(c)
+             use iso_c_binding, only : c_int, c_char
+             integer (c_int), intent(in), value :: len
+             character (c_char), intent(inout) :: id
```
mgduda: I think `id` should be a character array, as we did for `fname` in the `read_geogrid` interface.

src/framework/pool_hash.c
```diff
@@ -1,21 +1,25 @@
...
-#endif
+/*
+ use iso_c_binding, only : c_int, c_char
 
-void pool_hash(int* hash, char* key, int* len)
+ interface
+   subroutine pool_hash(hash, key, len) bind(c)
+      use iso_c_binding, only : c_int, c_char
+      integer (c_int), intent(inout) :: hash
+      character (c_char), intent(in) :: key
```
mgduda: I think `key` should also be declared as a character array.

src/core_init_atmosphere/mpas_init_atm_gwd.F
```diff
@@ -314,14 +334,17 @@ function read_global_30s_topo(path, sub_path) result(iErr)
...
       integer :: ix, iy, ishift, ix_shift
-      integer :: isigned, endian, wordsize, nx, ny, nz
-      real (kind=R4KIND) :: scalefactor
-      real (kind=R4KIND), dimension(:,:,:), allocatable :: tile
+      integer (c_int) :: isigned, endian, wordsize, nx, ny, nz
+      real (c_float) :: scalefactor
+      real (c_float), dimension(:,:,:), pointer, contiguous :: tile
+      type (c_ptr) :: tile_ptr
       character(len=StrKIND) :: filename
+      character(len=StrKIND+1, kind=c_char) :: c_filename
```
mgduda: A `character(len=StrKIND+1, kind=c_char)` may be interoperable with a `character(kind=c_char), dimension(StrKIND+1)`, but since the latter is closer to what is used in the interface for `read_geogrid`, I'd propose we change the declarations of `c_filename` to be character arrays.

mgduda:
@MiCurry Sorry about the minor, nit-picky nature of the comments -- I think this is a sign that we're essentially there, though!

src/framework/mpas_pool_routines.F
MiCurry: As we said early, and for the PR. The original `pool_hash` function ran slightly faster then `mpas_f_to_c_string`, but `mpas_f_to_c_string` was faster then the intrinsics `len` and `len(adjustl)`.

src/framework/mpas_stream_manager.F
MiCurry: Sounds good to me. Updated the deceleration here as updated the Fortran deceleration interface within `random_id.c`.

src/framework/pool_hash.c
MiCurry: Also sounds good. I updated this reference to the interface here, and updated the interface within `mpas_pool_routines.F` (lines ~6000).

src/core_init_atmosphere/mpas_init_atm_gwd.F
MiCurry: Good catch. I updated this deceleration and the deceleration to `c_fname` within `mpas_init_atm_static.F`.

src/core_init_atmosphere/mpas_init_atm_gwd.F
```diff
@@ -7,15 +7,35 @@
...
+       integer (c_int), intent(in), value :: nx
+       integer (c_int), intent(in), value :: ny
+       integer (c_int), intent(in), value :: nz
+       integer (c_int), intent(in), value :: isigned
+       integer (c_int), intent(in), value :: endian
+       real (c_float), intent(in), value :: scalefactor
+       integer (c_int), intent(in), value :: wordsize
+       integer (c_int), intent(inout) :: status
+    end subroutine read_geogrid
+ end interface
```
mgduda: Let's add two columns of indentation to this block here so that it matches the indentation of the surrounding code.

src/core_init_atmosphere/mpas_init_atm_gwd.F
```diff
@@ -389,14 +413,17 @@ function read_global_30s_landuse(path) result(iErr)
...
       integer :: ix, iy
-      integer :: isigned, endian, wordsize, nx, ny, nz
-      real (kind=R4KIND) :: scalefactor
-      real (kind=R4KIND), dimension(:,:,:), allocatable :: tile
+      integer (c_int) :: isigned, endian, wordsize, nx, ny, nz
+      real (c_float) :: scalefactor
+      real (c_float), dimension(:,:,:), pointer, contiguous :: tile
+      type (c_ptr) :: tile_ptr
       character(len=StrKIND) :: filename
+      character(len=StrKIND+1, kind=c_char) :: c_filename
```
mgduda: Let's change the definition of `c_filename` here to match what was done in the `read_global_30s_topo` routine.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -26,12 +29,30 @@ module mpas_init_atm_static
...
+    end subroutine read_geogrid
+ end interface
+
  contains
 
 !==================================================================================================
  subroutine init_atm_static(mesh, dims, configs)
 !==================================================================================================
 
+
```
mgduda: Let's remove this extra newline.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -49,16 +71,17 @@ subroutine init_atm_static(mesh, dims, configs)
...
  integer:: i,j,k
  integer:: iCell,iEdge,iVtx,iPoint,iTileStart,iTileEnd,jTileStart,jTileEnd
  integer,dimension(5) :: interp_list
  integer,dimension(:),allocatable  :: nhs
  integer,dimension(:,:),allocatable:: ncat
       
- real(kind=4):: scalefactor
- real(kind=4),dimension(:,:,:),allocatable:: rarray
+ real(kind=c_float):: scalefactor
+ real(c_float),dimension(:,:,:),pointer,contiguous :: rarray
```
mgduda: I'm not sure if the Fortran standard expresses a preference for either style (e.g., `real(kind=c_float)` versus `real(c_float)`), but we should declare `scalefactor` and `rarray` consistently.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -769,6 +817,7 @@ subroutine init_atm_static_orogwd(mesh, dims, configs)
 
  character(len=StrKIND) :: mess
  character(len=StrKIND) :: fname
+ character(len=StrKIND+1, kind=c_char) :: c_fname
```
mgduda: Let's change the declaration here to match what was done in `init_atm_static`.

src/framework/pool_hash.c
```diff
@@ -1,21 +1,25 @@
...
-#ifdef DOUBLEUNDERSCORE
-#define pool_hash pool_hash__
-#endif
-#endif
+/*
+ use iso_c_binding, only : c_int, c_char
 
-void pool_hash(int* hash, char* key, int* len)
+ interface
+   subroutine c_pool_hash(hash, key, len) bind(c)
```
mgduda: Should `len` be removed from the argument list here?

src/framework/mpas_stream_manager.F
```diff
@@ -4184,8 +4199,8 @@ subroutine build_stream(stream, direction, allFields, allPackages, timeLevelIn,
 
         integer :: local_ierr
 
-        integer, parameter :: idLength = 10
-        character (len=idLength) :: file_id
+        integer (c_int), parameter :: idLength = 10
+        character (len=idLength, kind=c_char) :: file_id
```
mgduda: For consistency with the declarations of C character strings elsewhere, I'd propose we change the declaration to:
```
character(kind=c_char), dimension(idLength) :: file_id
```

src/framework/mpas_pool_routines.F
```diff
@@ -5992,5 +5992,30 @@ integer function pool_get_member_decomp_type(dimName) result(decompType)!{{{
...
+           use iso_c_binding, only : c_int, c_char
+           integer (c_int), intent(inout) :: hash
+           character (c_char), dimension(*), intent(in) :: key
+        end subroutine c_pool_hash
+      end interface
+
+      integer (c_int), intent(inout) :: hash
+      character(len=*), intent(in) :: key
+
+      character(len=StrKIND, kind=c_char) :: c_key
```
mgduda: For consistency, I'd propose we declare this as:
```
character(kind=c_char), dimension(StrKIND+1) :: c_key
```
Note that the `+1` is probably important in the length to account for a C null character that will be added by `mpas_f_to_c_string`.

MiCurry:
@mgduda Thanks for looking things over and all the fixes. I just updated and pushed a new version. Hopefully it will be the last!

MiCurry:
Erp, this doesn't compile. There appears to be a problem in `mpas_stream_manager.F`, specifically the most recent change. Compiling with `file_id` defined as: `character(kind=c_char), dimension(idLength) :: file_id` will error out (for some reason) with Intel, but compiling with the original deceleration  ```character (len=idLength, kind=c_char) :: file_id``` compiles without issues. 🤔

mgduda:
@MiCurry I see the error you're talking about in `mpas_stream_manager.F`. I think the compiler has a legitimate complaint here, and I think the solution may be to just call `mpas_c_to_f_string` to convert the `file_id` variable that is set by C code to a Fortran string that can be used as an argument to `MPAS_writeStreamAtt`.

MiCurry:
@mgduda Right. Intel gave a mess of errors, but gfortran was able to drill it down a little bit closer. Thanks for picking it up.

I fixed it and gave a compile with GNU and Intel and ran a small test of the init core and things look good to go!

mgduda:
@matthewhoffman and @mark-petersen For a little more historical background, before the `ISO_C_BINDING` module was widely available, we had used cpp macros `UNDERSCORE` and `DOUBLEUNDERSCORE` to try to manage the name mangling for C routines to match the calls generated by the Fortran compiler. Using `ISO_C_BINDING`, we can now simply declare interfaces to C routines, adding `bind(c)`. We can further guarantee that the Fortran types are interoperable with C types using, e.g., `real (kind=c_float)`.

There are quite a few changes in this PR, but the majority of them are in the `init_atmosphere` core, which you can feel free to ignore. If you browse through the commit history for this PR, it might be easier to get a handle on which changes affect the shared framework.

mgduda:
@mark-petersen There are just a few routines in the ocean core that still rely on the `UNDERSCORE` macro. If you'd be interested, I could help you all to use `ISO_C_BINDING` with the code in `src/core_ocean/analysis_members/mpas_ocn_okubo_weiss_eigenvalues.c` so that we could completely eliminate the need to define `UNDERSCORE` or `DOUBLEUNDERSCORE` anywhere in MPAS. If now is not a good time to take on these changes in MPAS-Ocean, though, that's no problem.

mark-petersen:
@mgduda thanks for your offer.  I just looked, and agree that we should eliminate `UNDERSCORE` to be more standards compliant.  I also only see it in the ocean in 
`src/core_ocean/analysis_members/mpas_ocn_okubo_weiss_eigenvalues.c`
here:
```
 24 #ifdef UNDERSCORE
 25 #define compute_ev_2 compute_ev_2_
 26 #define compute_ev_3 compute_ev_3_
 27 #else
 28 #ifdef DOUBLEUNDERSCORE
 29 #define compute_ev_2 compute_ev_2__
 30 #define compute_ev_3 compute_ev_3__
 31 #endif
 32 #endif
```
Since you guys want to help me by writing the update for those two, that would be great.  You've had practice and would be faster.  If you just get MPAS to compile with `CORE=ocean`, I can run a little before/after test on that routine.

MiCurry:
@mark-petersen, I have pushed a new commit that compiles with CORE=ocean with the gnu compiler. We had some troubles compiling the Ocean module on Cheyenne due to the MPT installation on Cheyenne, so we weren't able to compile at all with Intel. 

Please let me know if you would like anything changed, and if it or runs correctly or not.

mgduda:
@mark-petersen Just a couple of details regarding the compilation issues: these appear to be due to a bug in SGI's (now HPE's) MPI module from MPT, where the `mpi.mod` file doesn't provide a definition of `MPI_STATUSES_IGNORE`. In principle we can substitute `include 'mpif.h'` for `use mpi` in the lagrangian particle tracking analysis member and the code will compile; but, given that the changes to MPAS-Ocean are so small, and that the code does compile on a different (non-SGI) system with the GNU compilers, we thought it was in good enough shape for your testing.

mgduda:
@MiCurry With commit e04f23f now in this PR, should we go ahead and remove the addition of `-DUNDERSCORE` from the top-level `Makefile` and any other parts of the code where it appears? If so, it might be better to do this before @mark-petersen does any testing on his end, so that he'll see the final contents of the PR.

MiCurry:
@mgduda Right. I forgot we mentioned to do that. I'll go ahead and remove `-DUNDERSCORE` from the `Makefile`.

MiCurry:
Okay, I just removed `-DUNDERSCORE` from the top level `Makefile` as well as `Makefile.in.ACME` and `Makefile.in.CESM` and pushed another commit to this PR.

@mark-petersen Thanks for running the tests. I was able again to compile this last commit with `CORE=ocean`. I'm sure your tests will still pass, but it might be worth you compile and run them again. I apologize about forgetting about this additional commit.

mark-petersen:
I compiled and ran again.  All looks good.

mark-petersen:
@matthewhoffman I compiled this branch with CORE=landice.  No problems, as expected, because your routines don't use this underscore interface for c.  I'm taking you off the reviewers list.

@MiCurry and @mgduda this is approved to merge from the LANL side.

mgduda:
Thanks, @mark-petersen ! We'll merge this later this afternoon.

mgduda:
@MiCurry I just noticed this now when grepping for 'underscore' throughout the code, but there is one mention of needing to adjust the name of the `read_geogrid` routine [here](https://github.com/MPAS-Dev/MPAS-Model/blob/develop/src/core_init_atmosphere/read_geogrid.c#L12-L14). Since this comment will be obviated by the changes in this PR, and to avoid making a separate PR to remove this comment, would you mind just pushing one more commit to delete these comment lines?

NB: the changes mentioned above are to comments in the `init_atmosphere` core code only. No additional review should be needed.

MiCurry:
@mgduda It has been removed and the changes are now in this PR.

mgduda:
@MiCurry Thanks!

PR #157, opened 2019-02-19
Point global_ocean_V1 test cases to correct templates

xylar:
Many test cases are trying to use templates from `global_ocean`, which works for critical passages (but is not necessary, since `global_ocean_V1` has a copy of the same template, and could lead to issues in the future) but not for `init2`, since `global_ocean` does not use this template.

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean_V1/EC120to60/default/config_init2.xml
+++ b/testing_and_setup/compass/ocean/global_ocean_V1/EC120to60/default/config_init2.xml
@@ -33,13 +33,13 @@
 	<add_link source_path="initial_condition_database" source="chlorophyllA_monthly_averages_1deg.151201.nc" dest="swData.nc"/>
 
 	<namelist name="namelist.ocean" mode="init">
-		<template file="template_init2.xml" path_base="script_core_dir" path="global_ocean"/>
-		<template file="template_critical_passages.xml" path_base="script_core_dir" path="global_ocean"/>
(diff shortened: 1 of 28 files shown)


xylar:
@mark-petersen, I was trying to set up `global_ocean_V1/EC60to30/default` this morning and found that it's broken.  I decided to fix it and other test cases with the same issue.  I'm guessing many of these test cases have not been run since the name change to `global_ocean_V1`, so it might be good to make sure `./setup_testcase.py ...` at least works on all of them even if there's not time to test that each one runs successfully.

xylar:
I will make sure I can set up all test cases in `global_ocean_V1` on Cori.  For now, I'm only going to test whether I can run `EC60to30/default` because that's what I need for other purposes.

xylar:
The mesh `mesh.EC.120-60km.160115.nc` appears not to have been uploaded to `oceans11`.  @mark-petersen, can you upload it?  `EC120to60` fails on Cori without it.

xylar:
## Testing

With these changes, I am able to set up all test cases except EC120to60 (see above) on Cori.

sbrus89:
@xylar, I had noticed this as well. Thanks for catching and fixing it!

mark-petersen:
Thanks. I uploaded that file to oceans11:
```
-rw-rw-rw- 1 mpetersen users 93M Feb 21 12:07 /raid/mpas_data/mesh_database/mesh.EC.120-60km.160115.nc
```

xylar:
Thanks again!

PR #158, opened 2019-02-19
Remove C null characters from strings read by PIO

mgduda:
This merge adds a new routine, `MPAS_sanitize_string`, to convert C null characters to spaces
in Fortran strings, and it applies this routine to all strings read via PIO. The PIO2 library converts
unused characters of strings to C null characters when writing, and converts C null characters
back to spaces when reading. This can create problems if a file written with PIO2 is read by PIO1,
in which case the C null characters will remain in the string, potentially creating problems for
MPAS infrastructure.

### DIFF
--- a/src/framework/mpas_c_interfacing.F
+++ b/src/framework/mpas_c_interfacing.F
@@ -5,6 +5,37 @@ module mpas_c_interfacing
 
 
     !-----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mgduda:
Actually, it may be better to target a v6.2 release with this PR. Let's close this for now until a hotfix-v6.2 branch has been created.

PR #160, opened 2019-02-21
Change length of xtime variables to ShortStrKIND to match netCDF files

mgduda:
This merge changes the declared length of xtime and similar timestamp
variables in the mpas_io_streams module from StrKIND to ShortStrKIND to
match the length of xtime in netCDF files.

Prior to this change, because the in-memory xtime variable was larger
than the in-file version, reading xtime could result in a string with
garbage characters at the end (beginning at character ShortStrKIND+1).
These gargbage characters, in turn, could lead to errors in the timekeeping
module when trying to parse the string as a timestamp. By matching
the in-memory size of the xtime variable with its size in files,
we avoid this problem.

### DIFF
--- a/src/framework/mpas_io_streams.F
+++ b/src/framework/mpas_io_streams.F
@@ -148,12 +148,12 @@ module mpas_io_streams
       integer :: io_err
       integer :: i
       integer :: timeDim
-      character (len=StrKIND), dimension(:), pointer :: xtimes
-      character (len=StrKIND) :: strTemp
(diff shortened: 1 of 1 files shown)


mgduda:
This PR, along with PR #161 will be included in the MPAS v6.2 release.

mark-petersen:
@mgduda sorry for the delay.  I will test all your proposed framework changes together in MPAS-Ocean later today, so you can proceed with the release.

mgduda:
@akturner Did you want to do any final testing of this branch, as well as PR #160?

PR #161, opened 2019-02-21
Remove C null characters from strings read by PIO

mgduda:
This merge adds code to remove C null characters from strings read by PIO.

The PIO2 library fills in unused characters of a string variable with
C null characters, and replaces C null characters with spaces when reading
strings. However, the PIO1 library does not perform these conversions.
Consequently, if a file written by the PIO2 library is read by the PIO1 library,
C null characters may be present in the resulting Fortran string.

Under the assumption that C null characters will generally not be expected
or handled in MPAS (e.g., in timestamp strings), this commit makes calls to
MPAS_sanitize_string to convert any C null characters in strings read by PIO
into spaces.

### DIFF
--- a/src/framework/mpas_c_interfacing.F
+++ b/src/framework/mpas_c_interfacing.F
@@ -5,6 +5,37 @@ module mpas_c_interfacing
 
 
     !-----------------------------------------------------------------------
+    !  routine mpas_sanitize_string
+    !
+    !> \brief Converts C null characters in a Fortran string to spaces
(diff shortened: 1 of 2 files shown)


mgduda:
This PR, along with PR #160, will be included in the MPAS v6.2 release

mgduda:
@mark-petersen and @matthewhoffman Thanks very much for the reviews.

@akturner Did you want to do any testing of these changes with MPAS-Sea Ice?

PR #108, opened 2018-11-06
Use appropriate units for core_atmosphere file I/O

jjguerrette:
This PR supports interfacing MPAS with the JEDI data assimilation framework.

The atmospheric core is updated to use the MPAS framework subroutines mpas_new_unit and mpas_release_unit for defining appropriate file units for I/O that do not conflict with other units already used throughout MPAS or in a surrounding program when MPAS is used as a library.

This code has been tested within JEDI with code that will be submitted in several other pull requests. This PR depends on the code changes in #106 , and should be considered after that merge is completed. There should be no impact to non-JEDI applications, but I have no method to test that fact at this time.

PR #162, opened 2019-02-26
Add temperature and spechum fields in JEDI DA package

jjguerrette:
These fields are used to produce analysis fields in the JEDI
data assimilation framework.  They are only used in that
external framework and thus need to be in a package.

All related DA tests pass.

This PR is related to https://github.com/MPAS-Dev/MPAS-Model/pull/106, https://github.com/MPAS-Dev/MPAS-Model/pull/107, and https://github.com/MPAS-Dev/MPAS-Model/pull/108.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -336,7 +336,7 @@
                      possible_values="Non-negative real values"/>
 	</nml_record>
 
-        <nml_record name="assimilation" in_defaults="true">
+        <nml_record name="assimilation" in_defaults="false">
                 <nml_option name="config_jedi_da" type="logical" default_value="false"
                      units="-"
                      description="Whether this run is within the JEDI data assimilation framework; used to add temperature and specific humidity as diagnostics"
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -336,6 +336,12 @@
                      possible_values="Non-negative real values"/>
 	</nml_record>
 
+        <nml_record name="assimilation" in_defaults="true">
+                <nml_option name="config_jedi_da" type="logical" default_value="false"
+                     units="-"
+                     description="Whether this run is within the JEDI data assimilation framework; used to add temperature and specific humidity as diagnostics"
+                     possible_values=".true. or .false."/>
+        </nml_record>
 
 <!-- **************************************************************************************** -->
 <!-- **************************************  Packages  ************************************** -->
@@ -352,6 +358,7 @@
                 <package name="bl_mynn_in" description="parameterization of MYNN Planetary Boundary Layer."/>
 
                 <package name="iau" description="Incremental Analysis Update"/>
+                <package name="jedi_da" description="Data Assimilation in JEDI framework"/>
         </packages>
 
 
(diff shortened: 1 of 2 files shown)


jjguerrette:
@mgduda , this is the next PR for JEDI.  Can you check that the packaging looks correct?  Also I am not sure if the namelist variable "config_jedi_da" follows MPAS conventions or if it warrants a new namelist record called "assimilation" that is added.

src/core_atmosphere/Registry.xml
```diff
@@ -336,6 +336,12 @@
                      possible_values="Non-negative real values"/>
 	</nml_record>
 
+        <nml_record name="assimilation" in_defaults="true">
```
mgduda: If MPAS-A is supported to the general community in JEDI, then I think it makes sense to keep the `assimilation` namelist record in the default namelist; otherwise, I'd propose to set `in_defaults="false"` so that the `&assimilation` namelist record will still be valid, but it just won't appear by default in the `namelist.atmosphere` file that is auto-generated at build time.

jjguerrette: That makes sense.  I admit that I did not know what that entry meant.  Fixing it now.

src/core_atmosphere/mpas_atm_core_interface.F
```diff
@@ -134,6 +135,14 @@ function atm_setup_packages(configs, packages, iocontext) result(ierr)
...
       end if
 
+      nullify(config_jedi_da)
+      call mpas_pool_get_config(configs, 'config_jedi_da', config_jedi_da)
+
+      nullify(jedi_daActive)
+      call mpas_pool_get_package(packages, 'jedi_daActive', jedi_daActive)
+
+      jedi_daActive = config_jedi_da
+
```
mgduda: It's not at all mandatory, but in case we'd like to be really thorough here (I know we didn't do this for IAU, which is my fault), we could set `jedi_daActive` only if both `jedi_daActive` and `config_jedi_da` are associated. With pools, we pass in a pointer, requesting that it be set to point to the named entity; but, if that entity doesn't exist for some reason (e.g., because someone accidentally removed it from the `Registry.xml` file), the pointer is left unchanged, which in this case would leave the pointer as unassociated. So, the code could look like this:
```
      if (associated(config_jedi_da) .and. associated(jedi_daActive)) then
         jedi_daActive = config_jedi_da
      else
         ierr = ierr + 1
         call mpas_log_write('Package setup failed for ''jedi_da''. '// &
              'Either ''jedi_da'' is not a package, or ''config_jedi_da'' is not a namelist option.', &
              messageType=MPAS_LOG_ERR)
      end if
```

src/core_atmosphere/Registry.xml
mgduda: No worries -- our documentation is non-existent in many cases!

src/core_atmosphere/mpas_atm_core_interface.F
mgduda: Actually, even with this extra logic, there's still the issue that the physics package setup further down will simply overwrite the value of `ierr` before the routine returns. If you're willing to add some extra logic for the `jedi_da` package, I'll push an additional commit to this branch before merging to address the physics package setup problem.

jjguerrette: If by "add some extra logic for the `jedi_da` package", you mean insert the code above, then I am very willing.  Is there anything else you need me to change?

mgduda: Sorry, I was a little unclear. If you wouldn't mind committing the code section from my comment to check that both `jedi_daActive` and `config_jedi_da` are both associated, and to print an error if not, that would be excellent. I'll then push a commit to address the potential issue of the call to `atmphys_setup_packages` overwriting any non-zero value of `ierr`. Then I think we're all set for this PR.

jjguerrette: Ok, all set.

PR #164, opened 2019-02-26
Update sub-ice-shelf melt to remove coupled flag

### LINKED ISSUE #163: Bug in melt rate computation in "coupled" mode
There is a bug in "coupled" mode here:
https://github.com/MPAS-Dev/MPAS-Model/blob/e3sm/develop/src/core_ocean/shared/mpas_ocn_surface_land_ice_fluxes.F#L810-L813
where T0 is accessed before it is computed.

xylar:
The melt computation is never called with the fields from the ice sheet (this will be handled in the coupler instead) so there is no value in keeping this code.  We have the past history if it is ever needed in aiding coupler development.

closes #163

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_surface_land_ice_fluxes.F
+++ b/src/core_ocean/shared/mpas_ocn_surface_land_ice_fluxes.F
@@ -731,10 +731,7 @@ contains
     outOceanHeatFlux, &
     outIceHeatFlux, &
     nCells, &
-    err, &
-    iceTemperature, &
-    iceTemperatureDistance, &
-    kappa_land_ice) !{{{
+    err) !{{{
 
     !-----------------------------------------------------------------
     !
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran all but the QU120wLI test case from the land-ice-fluxes test suite.  (The QU120 uses more cores than my laptop has.)  All tests passed and are bit-for-bit identical to `ocean/develop`, as expected.

@mark-petersen, could you run both nightly and land-ice-fluxes regression tests (compared with `ocean/develop`) to make sure you also see bit-for-bit identical results?

xylar:
BTW, I will hesitate to do ocean clean-up pull requests in the future because of the hideous green/blue color combination of the tags :-p

xylar:
@mark-petersen, this needs attention only so it doesn't languish too long.  It's simply to remove some buggy and unused code.

xylar:
A good test would be the land-ice fluxes regression suite against current ocean/develop to make sure everything is bit-for-bit the same.

PR #165, opened 2019-02-26
Change start index from 1 to 2 on activeTracerVerticalAdvectionTopFlux

mark-petersen:
Vertical index k=1 on activeTracerVerticalAdvectionTopFlux is the ocean surface, so is always zero.  It was using some uninitialized values when using k=1, so start loop at k=2.


mark-petersen:
E3SM with this change passes
```
```
It was failing that before, in the monthly time average calculation.

mark-petersen:
@vanroekel could you give this a quick visual sanity check?

PR #167, opened 2019-02-27
removed need for -DROTATED_GRID in the build for the mtn_wave test case

skamaroc:
This merge removed the need for -DROTATED_GRID in the build for the mtn_wave test case.
This was accomplished by adding a new 1D array (v_init) that compliments u_init, and adopting the standard vector definition V_init = u_init * cos(angleEdge) + v_init * sin(angleEdge) in the code.  Registry additions add the v_init(nVertLevels) array and other changes introduce the vector definition.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -1834,7 +1834,7 @@ module init_atm_cases
       real (kind=RKIND), dimension(nVertLevels ) :: zu, dzw, rdzwp, rdzwm
 
       real (kind=RKIND) :: d1, d2, d3, cof1, cof2
-      real (kind=RKIND) :: um, vm, us, vs, rcp, rcv
+      real (kind=RKIND) :: um, vm,rcp, rcv
       real (kind=RKIND) :: xmid, temp, pres, a_scale
 
       real (kind=RKIND) :: xi, xa, xc, xla, zinv, xn2, xn2m, xn2l, sm, dzh, dzht, dzmin, z_edge, z_edge3 
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -75,29 +75,6 @@ module init_atm_cases
 
       call mpas_pool_get_config(domain % blocklist % configs, 'config_init_case', config_init_case)
 
-      !
-      ! Do some quick checks to make sure compile options are compatible with the chosen test case
-      !
-      if (config_init_case == 6) then
-#ifndef ROTATED_GRID
-         call mpas_log_write('*****************************************************************',  messageType=MPAS_LOG_ERR)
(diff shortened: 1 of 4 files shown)


src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -1857,7 +1834,7 @@ subroutine init_atm_case_mtn_wave(dminfo, mesh, nCells, nVertLevels, state, diag
       real (kind=RKIND), dimension(nVertLevels ) :: zu, dzw, rdzwp, rdzwm
 
       real (kind=RKIND) :: d1, d2, d3, cof1, cof2
-      real (kind=RKIND) :: um, us,  rcp, rcv
+      real (kind=RKIND) :: um, vm, us, vs, rcp, rcv
```
mgduda: It looks like we can safely remove the `us` and `vs` variables.

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -2185,13 +2163,15 @@ subroutine init_atm_case_mtn_wave(dminfo, mesh, nCells, nVertLevels, state, diag
...
                ztemp = .25*( zgrid(k,cell1 )+zgrid(k+1,cell1 )  &
                             +zgrid(k,cell2)+zgrid(k+1,cell2))
-               u(k,i) = um
-               if(i == 1 ) u_init(k) = u(k,i) - us
-#ifdef ROTATED_GRID
-               u(k,i) = sin(angleEdge(i)) * (u(k,i) - us)
-#else
-               u(k,i) = cos(angleEdge(i)) * (u(k,i) - us)
-#endif
+               u(k,i) = vm
```
mgduda: This line of code has no effect, since `u(k,i)` is set just below here and independent of any existing value for `u(k,i)`.

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -2185,13 +2163,15 @@ subroutine init_atm_case_mtn_wave(dminfo, mesh, nCells, nVertLevels, state, diag
...
+
+               if(i == 1 ) then 
+                 v_init(k) = vm
+                 u_init(k) = 0.
+               end if
```
mgduda: It might be cleaner to move this block of code out of the loop over `nEdges`, since `u_init` and `v_init` are independent of horizontal location. We could also set `u_init(:)` to `um` rather than `0.`.

src/core_init_atmosphere/mpas_init_atm_cases.F
skamaroc: us and vs removed.

src/core_init_atmosphere/mpas_init_atm_cases.F
skamaroc: This code has been moved outside the nEdges loop, as suggested.

PR #168, opened 2019-02-27
Add two new upper absorbing layer options

skamaroc:
2 additional upper absorbing layer options are added:

(1) The absorbing layer often used in the MPAS/CAM configuration.  This option
activates a fixed second-order horizontal diffusion in the topmost
three layers on the dynamics variables (u, w and theta).
(2) Rayleigh damping on the horizontal mumentum (u) with a user-specified
damping timescale and a linear ramp over a number of levels from the top
(also user specified).

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -256,22 +256,17 @@
                      description="Maximum w-damping coefficient at model top"
                      possible_values="0 $\leq$ config_xnutr $\leq$ 1"/>
 
-                <nml_option name="config_horiz_absorbing_layer" type="character" default_value="none" in_defaults="false"
-                     units="-"
-                     description="h_diffusion configuration for upper absorbing layer"
-                     possible_values="`none' or `mpas_cam'"/>
-
-                <nml_option name="config_mpas_cam_coef" type="real" default_value="0.2" in_defaults="false"
+                <nml_option name="config_mpas_cam_coef" type="real" default_value="0.0" in_defaults="false"
                      units="-"
                      description="coefficient for scaling the 2nd-order h-mixing in mpas_cam absorbing layer"
-                     possible_values="0 $\leq$ config_mpas_cam_coef $\leq$ 1"/>
+                     possible_values="0 $\leq$ config_mpas_cam_coef $\leq$ 1, standard value is 0.2"/>
 
                 <nml_option name="config_rayleigh_damp_u" type="logical" default_value="false" in_defaults="false"
                      units="-"
                      description="use Rayleigh damping on horizontal velocity topmost model levels"
                      possible_values=".true. or .false."/>
 
-                <nml_option name="config_rayleigh_damp_u_timescale" type="real" default_value="5." in_defaults="false"
+                <nml_option name="config_rayleigh_damp_u_timescale_days" type="real" default_value="5." in_defaults="false"
                      units="days"
                      description="Rayleigh damping timescale in days"
                      possible_values="config_rayleigh_damp_u_timescale must be greater than 0"/>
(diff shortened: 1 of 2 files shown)


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -4360,6 +4382,24 @@ subroutine atm_compute_dyn_tend_work(nCells, nEdges, nVertices, nVertLevels_dumm
...
+
+            do iCell = cellStart,cellEnd
+
+            !
+            ! 2nd-order filter for top absorbing layer as in CAM-SE :  WCS 10 May 2017
+            ! From MPAS-CAM V4.0 code, with addition to config-specified coefficient (V4.0_coef = 0.2; SE_coef = 1.0)
+            !
+            kdiff(nVertLevels-2,iCell) = max(kdiff(nVertLevels-2,iCell),   2.0833*config_len_disp*config_mpas_cam_coef)
+            kdiff(nVertLevels-1,iCell) = max(kdiff(nVertLevels-1,iCell),2.*2.0833*config_len_disp*config_mpas_cam_coef)
+            kdiff(nVertLevels  ,iCell) = max(kdiff(nVertLevels  ,iCell),4.*2.0833*config_len_disp*config_mpas_cam_coef)
```
mgduda: Where do the `2.0833` values come from?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -4661,6 +4701,24 @@ subroutine atm_compute_dyn_tend_work(nCells, nEdges, nVertices, nVertLevels_dumm
          end do
       end do
 
+!  Rayleigh damping on u 
+
+      if(config_rayleigh_damp_u) then
+         do k=nVertlevels-config_number_rayleigh_damp_u_levels+1,nVertLevels
+            rayleigh_damp_coef(k) = (real(k - (nVertLevels-config_number_rayleigh_damp_u_levels))/ &
+                                     real(config_number_rayleigh_damp_u_levels))/                  &
+                                        (config_rayleigh_damp_u_timescale*86400.)
```
mgduda: Would it be better to specify the damping timescale in seconds in the namelist so we don't need to multiply by 86400 here? That would also have the benefit of keeping our namelist options in mks units as much as possible. We could also consider pre-computing the inverse timescale to avoid a division here.

src/core_atmosphere/Registry.xml
```diff
@@ -255,6 +255,31 @@
...
+                     units="-"
+                     description="h_diffusion configuration for upper absorbing layer"
+                     possible_values="`none' or `mpas_cam'"/>
+
+                <nml_option name="config_mpas_cam_coef" type="real" default_value="0.2" in_defaults="false"
+                     units="-"
+                     description="coefficient for scaling the 2nd-order h-mixing in mpas_cam absorbing layer"
+                     possible_values="0 $\leq$ config_mpas_cam_coef $\leq$ 1"/>
+
+                <nml_option name="config_rayleigh_damp_u" type="logical" default_value="false" in_defaults="false"
```
mgduda: Rather than having a separate logical to activate the Rayleigh damping on u, we could just apply the damping whenever the specified timescale is greater than zero.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
skamaroc: This coefficient comes from CAM - will check and include in documentation.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
skamaroc: appended _days to the config timescale so the units are clear.  The loop setting the coefficient has been moved outside the edge loop (it didn't depend on the edge), and pre-computed the inverse.

src/core_atmosphere/Registry.xml
skamaroc: As per our discussion, a smaller coefficient increases damping, so we'll leave the logical for now.

PR #170, opened 2019-03-01
Update threading in ocean EOS

amametjanov:
Update threading in ocean EOS to fix issues in multi-threaded runs. Tested with `SMS_P12x2.ne4_oQU240.A_WCYCL1850.anvil_intel.allactive-mach_mods`.

[BFB]

### DIFF
--- a/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
+++ b/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
@@ -102,7 +102,7 @@ module ocn_analysis_mode
       call mpas_timer_stop('io_reset_alarms')
 
       ! Initialize submodules before initializing blocks.
-      call ocn_equation_of_state_init(err_tmp)
+      call ocn_equation_of_state_init(domain, err_tmp)
       ierr = ior(ierr, err_tmp)
 
       call ocn_analysis_init(domain, err_tmp)
--- a/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
@@ -243,7 +243,7 @@ module ocn_forward_mode
       call ocn_vmix_init(domain, err_tmp)
       ierr = ior(ierr, err_tmp)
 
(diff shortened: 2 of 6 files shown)


mark-petersen:
@amametjanov Thanks for adding code. Before this PR, does the code potentially cause an incorrect answer, or just slower performance with threading?

@philipwjones how does this overlay with your performance work on EOS - any preferred order for merging?

philipwjones:
@mark-petersen This is fixing a bug in my new EOS version. I had placed allocates of some temps within an omp single construct when shared allocated arrays must be allocated by the master thread. For some reason, a lot of machines were fine with the version I had committed, but it correctly failed on the sandiatoss machine during testing.  This should fix that issue while still retaining the performance improvement. @amametjanov can correct any misstatement or add any additional detail.  I reviewed the mods - they simply replace the single directive with a master directive and add the required barriers that were implicit in single - and this should be merged after my earlier (reverted) PR for the EOS change.

amametjanov:
Yes, it should go after Phil's PR.

mark-petersen:
These are next in line:
Remerge https://github.com/MPAS-Dev/MPAS-Model/pull/69
Merge this PR.
Test in MPAS-O, then E3SM. Is there anything I should know, or is it all ready to go?

philipwjones:
Should all be good to go.

mark-petersen:
Sorry for the delay. https://github.com/E3SM-Project/E3SM/pull/2843 got higher priority. I will do all the threading-related PRs after that, so if you have other changes or potential PRs to add for threading, now is a good time to make them.

mark-petersen:
@amametjanov this PR is on top of https://github.com/MPAS-Dev/MPAS-Model/pull/69 which was reverted. I'd like to make this PR the re-merge of the #69 commits and your commit on top of it, since they are all related.  May I do that, and then push to this branch?
You need to change the target branch of this PR to `ocean/develop`.

amametjanov:
No problem. Done.

mark-petersen:
OK, I rebased, added #69, and put the 'fix' commit after that. @philipwjones and @amametjanov please browse the code and see if it looks right.

PR #174, opened 2019-03-06
Spelling corrections

mark-petersen:
This is my collection bin of spelling corrections.  They are in comments and RK4 timers.  Also, remove legacy CESM reference from comments!

PR #175, opened 2019-03-06
Fix snowFlow misspelling

mark-petersen:
The variable snowFlux is misspelled as snowFlow.  I think snowFlow causes a warning when the code tries to access that field, could cause an error on some machines.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -2167,8 +2167,8 @@ contains
                call ocn_write_field_statistics(debugUnit, fieldName, minValue, maxValue)
             end if
 
-            ! Test snowFlow
-            fieldName = 'snowFlow'
+            ! Test snowFlux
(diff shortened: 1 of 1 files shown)


mark-petersen:
Originally in #174.  Extracted bug for clarity.

mark-petersen:
Retested local merge with MPAS nightly regression suite, and on E3SM:
```
PASS PEM_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_gnu RUN time=955
PASS PEM_Ln9.ne30_oECv3wLI.A_WCYCL1850-DIB-ISMF_CMIP6.cori-knl_gnu RUN time=665
PASS SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-knl_gnu RUN time=1114
PASS SMS_D_Ln9.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-knl_intel RUN time=720
```

PR #179, opened 2019-03-07
Create new script to set cullCell

### LINKED ISSUE #176: Bathymetry from JIGSAW being used in base_mesh.nc
In COMPASS global_ocean test cases there is a step in build_mesh.py:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py#L60-L64
where we "inject" the bathymetry from a topography file into the mesh.  This topography appears to come from JIGSAW:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/ocean/global_ocean/config_files/config_base_mesh.xml#L15
so it's not clear to me why this is a topography we would trust or want to be using.

The critical bug is that not only do we use it to set a variable called `bathymetry` (never used again as far as I can tell), but we also set `cullCell` based on where this is above 20.  (How was this threshold chosen?!)

This `cullCell` field propagates to the `cull_mesh` stage and is treated as the starting land mask onto which other land masks are added ("maxed" as it were).  I had to dig into the cell-culler code to reming myself that this was another possible source of land masking.

Anyway, long story short, the solution is to skip the `inject_bathymetry.py` stage of `build_mesh.py`.  I don't see any reason to do this stage and am curious how this crept into the workflow.

sbrus89:
Currently, the cullCell variable is being set in the base_mesh step inside inject_bathymetry.py (#176). This PR puts this functionality in a new script, so it doesn't affect the global_ocean mesh creation.

closes #176

### DIFF
--- a/testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_bathymetry.py
+++ b/testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_bathymetry.py
@@ -113,10 +113,7 @@ if __name__ == "__main__":
     nc_vars = nc_mesh.variables.keys()
     if 'bathymetry' not in nc_vars:
         nc_mesh.createVariable('bathymetry', 'f8', ('nCells'))
-    if 'cullCell' not in nc_vars:
-        nc_mesh.createVariable('cullCell', 'i', ('nCells'))
 
     # Write to mesh file
     nc_mesh.variables['bathymetry'][:] = bathymetry
-    nc_mesh.variables['cullCell'][:] = nc_mesh.variables['bathymetry'][:] > 20.0
     nc_mesh.close()
--- a/testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_cullCell.py
+++ b/testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_cullCell.py
@@ -1,0 +1,12 @@
+import netCDF4 as nc4
+
+if __name__ == "__main__":
+
+    mesh_file = sys.argv[1]
+    nc_mesh = nc4.Dataset(mesh_file, 'r+')
+
+    if 'cullCell' not in nc_vars:
+        nc_mesh.createVariable('cullCell', 'i', ('nCells'))
+    
+    mesh.variables['cullCell'][:] = nc_mesh.variables['bathymetry'][:] > 20.0
+    nc_mesh.close()


xylar:
@sbrus89, I think the new script didn't get `git add`ed.  Could you do that?

sbrus89:
@xylar, I just added it.

xylar:
@mark-petersen and @pwolfram, I know your availability is limited right now.  Please make it a priority to review this (maybe just by inspection).  I will make sure @sbrus89 and I test this carefully on the major test cases that use this script before it gets merged to `ocean/develop`.

sbrus89:
This should remove this from the global_ocean workflow. From here, we can just add it back to the cases (i.e. Delaware Bay) where it is needed later on.

xylar:
@mark-petersen,
> I see that you are not adding a call to inject_cullCell anywhere, which is fine. You are making the user add it by hand.

We believe that the Delaware Bay test case may need to call this new script but no current test cases need that capability.

xylar:
Thanks for the reviews while you're on travel!  Much appreciated!

xylar:
@pwolfram, I'm going to merge this PR and #180 into `ocean/develop` tomorrow morning.  If you would like to give feedback, please do so before then.

mark-petersen:
Yes, please merge those two into ocean/develop today when you are ready.  These are a small number of COMPASS changes, and we can do an E3SM PR for these and the added atmospheric forcing, which also has associated COMPASS changes.  We are trying to keep E3SM PRs in small chunks of MPAS changes, but I think this group is reasonable.

@xylar, @matthewhoffman and @jonbob, if you guys have time today today, please proceed.  I'm busy during the day, but could otherwise do it soon.
1. merge #179 and #180 into ocean/develop
2. single merge of ocean/develop into e3sm/develop (no PR required)
3. E3SM PR with updated mpas-source to head of e3sm/develop. This is only adds #123, #179, #180.  

You can use this text for the PR description:
This pull request adds capability for stand-alone MPAS-Ocean to include time varying wind and pressure forcing.  The added code is not called during E3SM simulations. Also adds associated COMPASS files for testing.
bfb

and first comment, mention testing here:
https://github.com/MPAS-Dev/MPAS-Model/pull/123#issuecomment-469720815

xylar:
##Testing

I tested the nightly regression suite on badger with GNU and this branch is bit-for-bit identical to the current `ocean/develop`

PR #180, opened 2019-03-07
Make coastal refined meshes produced with JGSAW BFB between runs

### LINKED ISSUE #178: JIGSAW-produced meshes not BFB reproducable
Meshes produced with the JIGSAW workflow vary between runs. This is most likely due to using an approximate nearest neighbor search to find the distances to the coastline which are then used to specify resolution input to JIGSAW.

sbrus89:
This PR fixes the consistency issues observed when producing coastal refined meshes (issue #178). The cause was not setting the random seed used in the approximate nearest neighbor library (pyflann). This led to varying resolution specification inputs going into JIGSAW each time the COMPASS workflow was run. Setting the random seed to the same specified value resolves the issue.

closes #178

### DIFF
--- a/testing_and_setup/compass/ocean/jigsaw_to_MPAS/coastal_tools.py
+++ b/testing_and_setup/compass/ocean/jigsaw_to_MPAS/coastal_tools.py
@@ -453,7 +453,8 @@ def distance_to_coast(coastlines, lon_grd, lat_grd, origin, nn_search, smooth_wi
         flann.build_index(
             coast_pts_xyz,
             algorithm='kdtree',
-            target_precision=1.0)
+            target_precision=1.0,
+            random_seed=0)
 
     # Convert  backgound grid coordinates to x,y,z and put in a nx_grd x 3 array for kd-tree query
     Lon_grd, Lat_grd = np.meshgrid(lon_grd, lat_grd)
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen and @pwolfram, again, a quick review by inspection is what is needed here.  Once we get this onto `ocean/develop`,  we should be able to produce new EC60to30 and CONUS meshes that are reproducible.

sbrus89:
@xylar, If there were reproducibility issues with generating EC60to30, we may need to look into that further. That mesh does not use my coastal_tools.py to produce the resolution distribution input for JIGSAW, it's just specified by a latitude-dependent function.

https://github.com/MPAS-Dev/MPAS-Model/blob/0040864d4a2c16d7268af8eb644c6bd1ed24f313/testing_and_setup/compass/ocean/global_ocean/EC60to30/init/define_base_mesh.py#L13-L20

cc @mark-petersen

sbrus89:
I'll test it right now.

sbrus89:
I found that the JIGSAW EC60to30 case is BFB between runs.

xylar:
@sbrus89, as we discussed in person, we suspect that the reproducibility issues with the EC60to30 mesh are likely to do with @mark-petersen running on two different machines (grizzly and badger) with two different compilations of JIGSAW (perhaps different versions).

xylar:
The reproducibility issues are likely going to be diminished by using a common version of JIGSAW from conda in the future but for now we probably need to document details like which machine was used, which modules (and versions) were loaded and which JIGSAW hashtag was used.

xylar:
##Testing

I tested the nightly regression suite on badger with GNU and this branch is bit-for-bit identical to the current `ocean/develop`

PR #181, opened 2019-03-07
Add missing 'recursive' attribute to MPAS_log_write(...) routine

mgduda:
This merge adds the 'recursive' attribute to the MPAS_log_write routine.

The MPAS_log_write routine can generate indirectly recursive calls to itself,
for example, when MPAS_log_write calls log_abort, which calls
MPAS_log_finalize, which then calls MPAS_log_write. Without the recursive
attribute, several compilers will halt execution when MPAS is compiled with
DEBUG=true if a recursive call is detected in a routine not marked as recursive.

### DIFF
--- a/src/framework/mpas_log.F
+++ b/src/framework/mpas_log.F
@@ -461,7 +461,7 @@ module mpas_log
 !
 !-----------------------------------------------------------------------
 
-   subroutine mpas_log_write(message, messageType, masterOnly, flushNow, &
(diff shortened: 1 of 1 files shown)


mgduda:
@matthewhoffman @mark-petersen I thought it would be worth including this small fix in the MPAS v6.2 release as well.

mgduda:
Thanks, @matthewhoffman and @mark-petersen , for the quick review!

PR #142, opened 2019-01-31
Memory leak cleanup

mark-petersen:
- Add needed deallocation lines.  
- Remove unused variable declarations and allocations.

PR #185, opened 2019-03-14
Clone pool memory leak

### LINKED ISSUE #137: Memory leak in RK4
We don't run RK4 very often. Running EC60to30 on 576 cores on grizzly with gnu optimized, it runs for 7 days and then dies with memory errors.

mark-petersen:
Currently, the subroutine `mpas_pool_clone_pool` double-allocates arrays due to an index error, so that `mpas_duplicate_field` is called twice (by mistake) but deallocate is only called once. This PR fixes that index. This error previously caused ocean simulations using RK4 time stepping to die with out-of-memory errors.

fixes #137

### DIFF
--- a/src/framework/mpas_pool_routines.F
+++ b/src/framework/mpas_pool_routines.F
@@ -710,7 +710,7 @@ module mpas_pool_routines
                         deallocate(newmem % data % r0)
                      end do
 
-                     do j = minTimeLevels, newmem % data % contentsTimeLevs
+                     do j = minTimeLevels+1, newmem % data % contentsTimeLevs
                         call mpas_duplicate_field(dptr % r0a(dptr % contentsTimeLevs), newmem % data % r0)
                         newmem % data % r0a(j) = newmem % data % r0
                         deallocate(newmem % data % r0)
@@ -728,7 +728,7 @@ module mpas_pool_routines
                         deallocate(newmem % data % r1)
                      end do
 
-                     do j = minTimeLevels, newmem % data % contentsTimeLevs
+                     do j = minTimeLevels+1, newmem % data % contentsTimeLevs
                         call mpas_duplicate_field(dptr % r1a(dptr % contentsTimeLevs), newmem % data % r1)
                         newmem % data % r1a(j) = newmem % data % r1
                         deallocate(newmem % data % r1)
@@ -746,7 +746,7 @@ module mpas_pool_routines
                         deallocate(newmem % data % r2)
                      end do
 
-                     do j = minTimeLevels, newmem % data % contentsTimeLevs
+                     do j = minTimeLevels+1, newmem % data % contentsTimeLevs
                         call mpas_duplicate_field(dptr % r2a(dptr % contentsTimeLevs), newmem % data % r2)
                         newmem % data % r2a(j) = newmem % data % r2
                         deallocate(newmem % data % r2)
@@ -764,7 +764,7 @@ module mpas_pool_routines
                         deallocate(newmem % data % r3)
                      end do
 
-                     do j = minTimeLevels, newmem % data % contentsTimeLevs
+                     do j = minTimeLevels+1, newmem % data % contentsTimeLevs
                         call mpas_duplicate_field(dptr % r3a(dptr % contentsTimeLevs), newmem % data % r3)
                         newmem % data % r3a(j) = newmem % data % r3
                         deallocate(newmem % data % r3)
@@ -782,7 +782,7 @@ module mpas_pool_routines
                         deallocate(newmem % data % r4)
                      end do
 
-                     do j = minTimeLevels, newmem % data % contentsTimeLevs
+                     do j = minTimeLevels+1, newmem % data % contentsTimeLevs
                         call mpas_duplicate_field(dptr % r4a(dptr % contentsTimeLevs), newmem % data % r4)
                         newmem % data % r4a(j) = newmem % data % r4
                         deallocate(newmem % data % r4)
@@ -800,7 +800,7 @@ module mpas_pool_routines
                         deallocate(newmem % data % r5)
                      end do
 
-                     do j = minTimeLevels, newmem % data % contentsTimeLevs
+                     do j = minTimeLevels+1, newmem % data % contentsTimeLevs
                         call mpas_duplicate_field(dptr % r5a(dptr % contentsTimeLevs), newmem % data % r5)
                         newmem % data % r5a(j) = newmem % data % r5
                         deallocate(newmem % data % r5)
@@ -818,7 +818,7 @@ module mpas_pool_routines
                         deallocate(newmem % data % i0)
                      end do
 
-                     do j = minTimeLevels, newmem % data % contentsTimeLevs
+                     do j = minTimeLevels+1, newmem % data % contentsTimeLevs
                         call mpas_duplicate_field(dptr % i0a(dptr % contentsTimeLevs), newmem % data % i0)
                         newmem % data % i0a(j) = newmem % data % i0
                         deallocate(newmem % data % i0)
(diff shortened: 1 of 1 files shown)


mark-petersen:
I got an email confirmation from Doug, who wrote this module:
> Yes, that would be a memory leak, and your fix looks correct.

mark-petersen:
@xylar and @sbrus89 could you cherry-pick this commit onto ocean/develop and see if it fixes your problem? 

@matthewhoffman and @mgduda your cores don't use `mpas_pool_clone_pool`, but please review as a framework change.

mark-petersen:
@hrmoncada, please redo the memory leak report shown here: https://github.com/MPAS-Dev/MPAS-Model/issues/137#issuecomment-472394442 with the identical test case, but cherry-pick this commit d53484f from my fork. Previous memory leak with RK4 was 4.4MB for your test case. Please comment here what that number is with this change. Then add #142 and see if that reduces it a little more.

mark-petersen:
Starting with `ocean/develop` and merging in this PR, passes ocean nightly regression suite and is bfb with previous. Note that includes all recent additions to framework as well.

sbrus89:
@mark-petersen, absolutely. I'm re-running the case I had memory issues with now. I will let you know what happens. Thanks for doing such a thorough of looking into this!

xylar:
I probably won't get to this until after the E3SM meeting next week.  Feel free to go with @sbrus89's testing if need be.

sbrus89:
@mark-petersen, I was able to confirm that with the cherry-picked commit, the model runs well past the point where it had previously failed with a memory error (prior to your fix). To me, this would indicate that you've fixed the issue. However, this is on a Delaware Bay hurricane case that is very much a work in progress. Do you think this is enough to show the problem is fixed? I can do additional testing if needed.

cc @pwolfram

mark-petersen:
@sbrus89 that is great news. Since I am fixing what is (now) an obvious bug, and it fixes your specific problem, that's all I need from you. I'm almost sure we have fixed the right thing here. Thanks for the quick testing.

sbrus89:
@mark-petersen, perfect. I'll go ahead and approve.

mgduda:
I've tried cloning a pool in MPAS-Atmosphere, verifying the existence of a memory leak, applying the changes in this PR, and noting that the memory leak largely disappears. There does still appear to be a small memory leak. Considering the updated code below for cloning a 3-d real field in a pool:
```
L.759                allocate(newmem % data % r3a(newmem % data % contentsTimeLevs))
L.760                minTimeLevels = min(dptr % contentsTimeLevs, newmem % data % contentsTimeLevs)
L.761                do j = 1, minTimeLevels
L.762                   call mpas_duplicate_field(dptr % r3a(j), newmem % data % r3)
L.763                   newmem % data % r3a(j) = newmem % data % r3
L.764                   deallocate(newmem % data % r3)
L.765                end do
L.766
L.767                do j = minTimeLevels+1, newmem % data % contentsTimeLevs
L.768                   call mpas_duplicate_field(dptr % r3a(dptr % contentsTimeLevs), newmem % data % r3)
L.769                   newmem % data % r3a(j) = newmem % data % r3
L.770                   deallocate(newmem % data % r3)
L.771                end do
```
When we deallocate the temporary field (`r3`) at lines 764 and 770, any memory allocated for the `constitutentNames` and `attLists` members of the field type will probably be leaked. However, I think the reason we didn't just use `mpas_deallocate_field` is because we don't want to deallocate the `array` member of the field type, since that is what is pointed to by the cloned field. If we really wanted to eliminate every last memory leak in this code, I think the `deallocate(newmem % data % r3)` lines could be replaced with something like:
```
nullify(newmem % data % r3 % array)  ! prevent mpas_deallocate_field from seeing this memory
call mpas_deallocate_field(newmem % data % r3)  ! deallocate all parts of the field except the array storage
```
Complicating matters further, I don't think the `mpas_deallocate_field` routine properly deallocates `constituentNames`.

@mark-petersen I'd propose we add a little more detail to the commit message. If you're able to summarize the bug and its fix based on your discussion with @douglasjacobsen , that would be great. But, if needed I can contribute some text based on my understanding, too. Also, if you would really like to address every last memory leak in the `mpas_pool_clone_pool` routine, just let me know and I can help with that, too.

mark-petersen:
@mgduda thanks for the detailed message.  I'll go ahead and change all the deallocate lines as you say above.  Are you suggesting I make that change for all
```
deallocate(newmem % data % *)
```
where * is r0,r1,r2,r3,r4,r5,i0,i1,i2,i3,c0,c1,l0 ?

pwolfram:
Excellent work tracking this down everyone!

mgduda:
@mark-petersen I think that's correct: everywhere we were freeing a field type with a simple Fortran `deallocate` rather than a call to `mpas_deallocate_field` represents a small memory leak, since the allocated members of the field type are not deallocated.

matthewhoffman:
I removed myself as a reviewer, as this appears to be adequately vetted from multiple directions at this point.

xylar:
@mark-petersen, I'm going to try to test this again in the next couple of days.  If I don't get to it now, it'll be mid-April at the earliest.

mark-petersen:
I updated all lines as follows:
`deallocate(newmem % data % r3)`
to 
```
nullify(newmem % data % r3 % array)
call mpas_deallocate_field(newmem % data % r3)
```
etc, based on @mgduda's advice above (thanks). One exception is that r0,i0,c0,l0 are scalars, so `nullify` command is not applicable, since there is no array.

I merged this branch into ocean/develop locally and tested.  Passes nightly regression suite with gnu, bfb with ocean/develop, including the "RK4 Blocks Test".  For reference, the temporary merge branch is on my fork at `merge_clone_pool_memory_leak`

@mgduda, please take a look.

mark-petersen:
Successfully passed gnu-debug and intel-debug on ocean nightly test suite, with this PR merged into ocean/develop. I'm trying to be extra thorough here, since we are messing with low-level framework routines.

mark-petersen:
Thanks @hrmoncada for help testing. Running the identical test with RK4 through Totalview, the results are:
 **Before**
image

**After**
image

The total leaks are reduced from 4MB to 7KB!

The later was tested with my branch `merge_clone_pool_memory_leak`, which includes this PR, all recent framework changes, and small ocean memory clean-ups in #142.

mark-petersen:
@mgduda any comments, or is it OK to merge?

mgduda:
@mark-petersen I'll give this one last test this afternoon. Do you think we should address the issue that `mpas_deallocate_field` doesn't deallocate the `constituentNames` member of fields? I'd be willing to add a commit to your branch for this PR if so.

mark-petersen:
@mgduda, I would just like to merge this in by Monday. If you can add the extra deallocate commit by then, that would be great to button everything up.

mgduda:
@mark-petersen No problem. I think I can have this added by tomorrow mid-day. I'd also propose to add the comments that I had in my code comment to explain to future developers why we're nullifying the array components before deallocating the field that contains them, as this might not be at all obvious.

mgduda:
@mark-petersen I'm really sorry for the confusion -- I was looking through the code more carefully, and I was mistaken in my suggestion to nullify the array components before calling `mpas_deallocate_field`. The memory leak that I was seeing was actually due to a bug in `mpas_pool_destroy_pool`, which I can fix with a separate PR. Since we do a shallow copy of the source fields to the destination fields in a pool clone, if we ever access the `attLists` or `constituentNames` of members of the clone, we'd run into problems by calling `mpas_deallocate_field` in the `mpas_pool_clone_pool` routine.

TLDR; I think we just need to reset the HEAD of this branch to d7029e70 and we're good to go
(though I'll still try to fix the bug in `mpas_pool_destroy_pool` with a follow-on PR).

mark-petersen:
@mgduda Done. Thanks for checking again. It's better than merging in unnecessary changes.

mark-petersen:
Removing @xylar as a review, as this has been sufficiently tested by others.

PR #186, opened 2019-03-14
WSM6 correction for accretion for freezing rain

kkeene44:
Prior to these modifications, this scheme would allow rain to freeze too easily,
with even very small amounts of ice. The changes reduce the efficiency
proportional to the sqaure of the ratio of ice to water during the accretion
process. This maintains liquid precipitation longer, which is in better
agreement with observations.


mgduda:
@kkeene44 I don't quite remember what we had discussed during last week's physics unification meeting, but did we want to target the `atmosphere/physics_unification` branch with this PR, or perhaps the `develop` branch?

@ldfowler58 Do you recall where we intended these changes to go?

kkeene44:
@mgduda 
I believe this should at least go in the develop branch, as this is an improvement to the code, unrelated to unification. This change went into the develop branch for WRF. I'm not certain the group decided that it would NOT be going into the physics_unification branches, but that is likely the case. I think @weiwangncar and @ldfowler58 can answer definitively.

PR #188, opened 2019-03-26
Change default topo and landuse to GMTED2010 and MODIFIED_IGBP_MODIS_NOAH

mgduda:
This merge changes the default values for config_topo_data
and config_landuse_data to GMTED2010 and MODIFIED_IGBP_MODIS_NOAH in
the Registry.xml file for the init_atmosphere core. These default should
give closer static fields to those used in the ARW model.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -101,7 +101,7 @@
 
                 <nml_option name="config_dynamics_split_steps" type="integer" default_value="3"
                      units="-"
-                     description="When config\_split\_dynamics\_transport = T, the number of RK steps per transport step"
+                     description="When config_split_dynamics_transport = T, the number of RK steps per transport step"
                      possible_values="Positive integer values"/>
 
                 <nml_option name="config_h_mom_eddy_visc2" type="real" default_value="0.0"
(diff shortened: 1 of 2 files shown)


src/core_atmosphere/Registry.xml
```diff
@@ -101,7 +101,7 @@
 
                 <nml_option name="config_dynamics_split_steps" type="integer" default_value="3"
                      units="-"
-                     description="When config\_split\_dynamics\_transport = T, the number of RK steps per transport step"
+                     description="When config_split_dynamics_transport = T, the number of RK steps per transport step"
                      possible_values="Positive integer values"/>
```
weiwangncar: @mgduda I assume this is a no-impact change...

mgduda: Correct. There are actually two commits in this PR. The second commit simply fixes some documentation strings to match what we need when creating LaTeX source for the users' guide. This is mentioned in the commit message for the second commit.

ldfowler58:
I do not have objections making GMTED2010 and MODIFIED_IGBP_MODIS_NOAH the default options.

ldfowler58:
No issues about config_split_dynamics_transport either.

PR #189, opened 2019-03-26
Use mminlu rather than config_landuse_data when initializing sea ice

mgduda:
This merge makes changes to use mminlu rather than config_landuse_data when
initializing sea ice in the init_atmosphere core.

In the physics_init_seaice routine, we were previously using config_landuse_data
to determine the category to be used for ice. However, in cases where
the static file is produced with config_landuse_data = MODIFIED_IGPB_MODIS_NOAH,
but the model initial conditions file is produced with
config_landuse_data = USGS, the resulting sea ice points will have the wrong
category.

The solution to this issue implemented in this commit is to use mminlu rather
than config_landuse_data to determine the ice category. The mminlu value comes
from the input static file (or, in the case where static fields and initial
conditions are being processed all at once, from the config_landuse_data option),
and will therefore be consistent with the land use classification scheme used in
the static file.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
@@ -599,7 +599,7 @@
  real(kind=RKIND),dimension(:,:),pointer:: tslb,smois,sh2o,smcrel
 
  logical, pointer :: config_frac_seaice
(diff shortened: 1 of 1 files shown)


mgduda:
PR #236 (and, more specifically, c4724ad and e9dab3b) eliminates the need for the changes in this PR.

PR #191, opened 2019-04-01
E3SM and CESM are now using TRUE|FALSE for compile_threaded

jgfouca:
This change will need to be reflected in the MPAS submodule of E3SM as soon as possible.

### DIFF
--- a/src/Makefile.in.ACME
+++ b/src/Makefile.in.ACME
@@ -1,11 +1,11 @@
 # Duplicate logic from Tools/Makefile to set compile_threaded
-compile_threaded = false
+compile_threaded = FALSE
 ifeq ($(strip $(SMP)),TRUE)
-   compile_threaded = true
+   compile_threaded = TRUE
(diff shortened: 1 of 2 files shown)


mgduda:
At some point, would it also be worth renaming `Makefile.in.ACME` to `Makefile.in.E3SM`? And, if CESM is no longer being actively tested with MPAS components, would it be worth removing `Makfile.in.CESM`?

jgfouca:
@mgduda , yes and yes. You may want to check with the CESM folks first.

jgfouca:
@jonbob

jonbob:
Thanks @jgfouca

mark-petersen:
@jgfouca please change the target branch to MPAS-Dev:develop

@mgduda and @matthewhoffman this is a framework PR, so I asked you to review. A quick glance is enough.

jgfouca:
@mark-petersen , done.

PR #192, opened 2019-04-02
Update checkout_data_files.sh to obtain new physics tables for WRF 4.0.3 physics

mgduda:
This merge updates the checkout_data_files.sh script to obtain new physics tables
compatible with the WRF 4.0.3 physics.

The new checkout_data_files.sh looks for physics lookup tables that are
compatible with MPAS-Atmosphere version 7.0. In the new checkout script, we look
for a tag in the MPAS-Data repository that matches the version of the physics
files matching the MPAS-Atmosphere version.

The plan is that, moving forward, new editions of physics lookup tables will
always be found in a tag named according to the version of the tables, and
the 'master' branch in the MPAS-Data repository will essentially remain
unchanged since the version 3.0 physics tables to maintain compatibility with
older versions of MPAS-Atmosphere, whose checkout script only looks in
the master branch of the repository.

### DIFF
--- a/src/core_atmosphere/physics/checkout_data_files.sh
+++ b/src/core_atmosphere/physics/checkout_data_files.sh
@@ -5,64 +5,94 @@
 #
 # The purpose of this script is to obtain lookup tables used by the WRF physics
 #   packages.  At present, the only method for acquiring these tables is through
-#   the MPAS-Dev github repository using either git, svn, or curl.
+#   the MPAS-Dev GitHub repository using either git, svn, or curl.
 #
 # If none of the methods used in this script are successful in acquiring the 
 #   tables, please attempt to manually download the files from the MPAS-Data 
 #   repository at https://github.com/MPAS-Dev/MPAS-Data/.  All *.TBL and *DATA* 
 #   files, as well as the COMPATIBILITY file, should be copied into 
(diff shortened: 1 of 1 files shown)


mgduda:
@ldfowler58 For you review, could you verify that the lookup tables in the `v7.0` tag of the MPAS-Data repository are those that we want to use with the unified physics? To get the `v7.0` tag, you can:
```
git clone https://github.com/MPAS-Dev/MPAS-Data
cd MPAS-Data
git checkout v7.0
cd atmosphere/physics_wrf/files
```

ldfowler58:
I verified that all the files in the v7.0 tag of the MPAS-Data repository are the ones we need to run the unified physics. All the files are good to go.

ldfowler58:
It is actually difficult for me to check that the script works, but I agree that we need to keep older versions of the same files to run the older physics schemes for now. Let's try it out. Does this mean that the unified_physics branch will be a version 7.0 as well?

mgduda:
@ldfowler58 My thinking was that the `atmosphere/unified_physics` branch would be effectively v7.0 from the perspective of physics. I've done quite a bit of testing of the script itself, and I'll go over all options for downloading the tables again before merging.

ldfowler58:
Sounds great. Thanks.

PR #193, opened 2019-04-02
Update use of isice_lu and iswater_lu in core_atmosphere

ldfowler58:
isice_lu and iswater_lu are now defined in the initial conditions file instead of being defined in core_atmosphere. Changes to the different files in this PR reflect that the 2 variables are now in the init file.


mgduda:
@ldfowler58 I think these changes (with the exception of some comments at the top of files) have already been included in PR #86 from @kkeene44 . In this case, should we just close this PR?

mgduda:
Changes related to `isice_lu` and `iswater_lu` will be removed from PR #86 , which we will merge after this PR.

PR #194, opened 2019-04-03
Reduce threading barriers in ocean

amametjanov:
Reduce the number of threading barriers in ocean

[BFB]


mark-petersen:
@amametjanov could you rebase on the head of ocean/develop? If you prefer to add these commits to #152 and close this PR, that is fine. Whatever is easiest for you. The two PRs look similar, but they each edit a large number of files.

amametjanov:
Rebased

PR #195, opened 2019-04-03
E3SM Makefile name update

### LINKED ISSUE #66: Rename Makefile.in.ACME to Makefile.in.E3SM
Cleanup of naming to match E3SM project name

Pat Worley reported this in E3SM [Issue #2467](https://github.com/E3SM-Project/E3SM/issues/2467). Though it is not critical, it would be good to do at some point.

mark-petersen:
Change Makefile name from ACME to E3SM.
Remove Makefile.in.CESM

fixes #66


mark-petersen:
This is for a few changes to the makefiles that will end up in the next E3SM PR.

Is anyone using Makefile.in.CESM? Any votes on whether to keep it or delete it?

If it is unused, my vote is to delete it.

matthewhoffman:
@mgduda , any objections at NCAR to removing Makefile.in.CESM?

jonbob:
My vote is to delete it as well

mark-petersen:
Thanks! It is nearly identical to Makefile.in.E3SM, so it would be easy recreate, and we will keep that one up to date.

PR #197, opened 2019-04-04
Fix implicit declaration of generate_immutable_streams in gen_inc code

mgduda:
This merge fixes an implicit declaration of `generate_immutable_streams` in registry code.

The generate_immutable_streams function had no declaration, leading to compiler
warnings about an implicitly defined function. This commit adds a declaration
of generate_immutable_streams to gen_inc.h.


mgduda:
I think it should be reasonable to test PRs #196 #197 #198 and #199 together.

PR #198, opened 2019-04-04
Fix warnings for implicitly declared functions in xml_stream_parser.c

mgduda:
This merge fixes warnings for three implicitly declared functions in xml_stream_parser.c.

There were three functions defined in mpas_stream_manager.F whose declarations
in xml_stream_parser.c inadvertently contained the prefix "mpas_", effectively
declaring non-existent functions and implicitly declaring the actual functions
being called.

This commit simply corrects the names of these three functions in their
declarations:
 * stream_mgr_add_field_c
 * stream_mgr_add_pool_c
 * stream_mgr_add_immutable_stream_fields_c


mgduda:
I think it should be reasonable to test PRs #196 #197 #198 and #199 together. There's also no particular rush with these four PRs -- I mostly just wanted to get them off of my to-do list by submitting them.

PR #199, opened 2019-04-04
Don't process trigraphs with the XL compilers

mgduda:
This merge disables processing of trigraphs with the XL compilers.

The XL Fortran compiler invokes the XL C preprocessor, which by default processes
trigraphs (https://en.wikipedia.org/wiki/Digraphs_and_trigraphs#C). This can
cause problems in MPAS code, e.g., if a string '??' appears in a non-comment
context in the source, where the ??' characters are turned into a ^ character,
leading to an un-terminated character string.

To avoid processing of trigraphs, the `-qnotrigraph` option is passed to the C
preprocessor via `-WF` in the FFLAGS for the XL compiler only. There is apparently
no need to add `-qnotrigraph` to the CPPFLAGS, since the 'cpp' that is invoked
when explicitly preprocessing files with `GEN_F90=true` is the system's
C preprocessor, which doesn't process trigraphs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,11 @@ xlf:
 	"CC_SERIAL = xlc_r" \
 	"CXX_SERIAL = xlc++_r" \
 	"FFLAGS_PROMOTION = -qrealsize=8" \
-	"FFLAGS_OPT = -O3 -qufmt=be" \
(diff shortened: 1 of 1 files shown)


mgduda:
@mark-petersen and @matthewhoffman Sorry in advance for the testing load; but, I think it should be reasonable to test PRs #196 #197 #198 and #199 together.

PR #200, opened 2019-04-08
Ignore the word vector during preprocessing for E3SM Makefile

mark-petersen:
Gnu CPP will treat "vector" as a macro on the PowerPC64 architecture. Add -Uvector to ignore.

This was already done in the standard MPAS Makefile, but is now being added to the E3SM Makefile.

fixes https://github.com/E3SM-Project/E3SM/issues/2725

### DIFF
--- a/src/Makefile.in.E3SM
+++ b/src/Makefile.in.E3SM
@@ -67,6 +67,10 @@ ifeq ($(compile_threaded), TRUE)
     override CPPFLAGS += -DMPAS_OPENMP
 endif
 
+ifeq ($(GEN_F90), TRUE)
+    override CPPFLAGS += -Uvector
+endif
+
 all:
 	@echo $(CPPINCLUDES)
 	@echo $(FCINCLUDES)


mark-petersen:
Reviewed by minxu, who notified us of this problem. He confirmed that this works:
https://github.com/E3SM-Project/E3SM/issues/2725#issuecomment-480474056

mark-petersen:
@jonbob could you do a visual inspection, and approve based on the resolution of https://github.com/E3SM-Project/E3SM/issues/2725#issuecomment-480474056. Since this is the E3SM makefile, it is more relevant to us on the E3SM side.

@matthewhoffman - any comment on this one?

After merging this PR and https://github.com/MPAS-Dev/MPAS-Model/pull/185 (memory leak index), we will merge the develop branch into all the cores, and update e3sm/develop. The E3SM PR will have all these makefile updates, but also all MPAS framework updates from the develop branch.

mark-petersen:
Changed to caps TRUE format like other Makefile flags.

mgduda:
Just for future reference, we also added similar changes to the top-level Makefile in 7a66ba3 .

PR #202, opened 2019-04-09
Fix formatting on E3SM Makefile

mark-petersen:
I used the wrong case. GEN_F90 uses lowercase throughout.

### DIFF
--- a/src/Makefile.in.E3SM
+++ b/src/Makefile.in.E3SM
@@ -67,7 +67,7 @@ ifeq ($(compile_threaded), TRUE)
     override CPPFLAGS += -DMPAS_OPENMP
 endif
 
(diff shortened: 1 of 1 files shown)


mark-petersen:
@jonbob this is the same as yesterday's https://github.com/MPAS-Dev/MPAS-Model/pull/200. I was thinking we were changing all Makefile logicals to caps, but this one stays lower case, as it is lower case in all Makefiles in lower directories.

mark-petersen:
This alters E3SM Makefile only, and format was originally approved in #200. I'm merging these 8 characters (my mistake) so we can proceed with https://github.com/E3SM-Project/E3SM/pull/2843.

PR #203, opened 2019-04-11
Add options for using MODIS-based vegetation fraction, albedo, and max snow albedo

mgduda:
This merge introduces the capability to use MODIS-based datasets for monthly vegetation fraction,
monthly albedo, and maximum snow albedo. The use of these new datasets is controlled via three
new namelist options: `config_vegfrac_data`, `config_albedo_data`, and `config_maxsnowalbedo_data`. Each of these options may be set to `MODIS` (the default) to use
the new MODIS-based datasets, or `NCEP` to use the old default datasets.

Also included in this merge is a new option to supersample the MODIS-based monthly albedo
and maximum snow albedo fields. Supersampling is enabled by setting the hidden namelist
option `config_supersample_factor` to an integer larger than 1. This option is useful for MPAS
meshes with a smaller grid distance than about 0.05-degrees (i.e., the resolution of the MODIS
albedo datasets).

### DIFF
--- a/src/core_init_atmosphere/Registry.xml
+++ b/src/core_init_atmosphere/Registry.xml
@@ -153,7 +153,7 @@
                      description="The maximum snow albedo dataset to use (case 7 only)"
                      possible_values="`MODIS' or `NCEP'"/>
 
-                <nml_option name="config_supersample_factor"    type="integer"       default_value="1"    in_defaults="false"
+                <nml_option name="config_supersample_factor"    type="integer"       default_value="3"
                      units="-"
                      description="The supersampling factor to be used for MODIS maximum snow albedo and monthly albedo datasets (case 7 only)"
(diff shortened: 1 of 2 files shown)


weiwangncar:
@mgduda Perhaps a suggested value for config_supersample_factor can be added. My test shows 3 appears to work.

mgduda:
@weiwangncar The MODIS datasets that are supersampled are at 0.05-degree resolution. A supersampling factor of 3 should give us 0.01667-degree pixels, which works out to about 1.85 km pixels along the equator. This should be sufficient for a 3-km MPAS mesh. Anyone running on a finer mesh will have quite a few other options to change, anyway. So, I think a default supersampling factor of 3 seems reasonable. I can change this default, and also include the option in the default `namelist.init_atmosphere` file so it's visible to users.

PR #204, opened 2019-04-11
Enable regional static field and vertical grid processing in the init_atmosphere core

mgduda:
This merge enables the production of static fields and vertical grids for limited-area meshes
in the init_atmosphere core.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -3191,6 +3191,15 @@ module init_atm_cases
       do iEdge = 1,nEdges
          cell1 = cellsOnEdge(1,iEdge)
          cell2 = cellsOnEdge(2,iEdge)
+
+         ! Avoid referencing the garbage cell for exterior edges
+         if (cell1 == nCells+1) then
+            cell1 = cell2
+         end if
+         if (cell2 == nCells+1) then
+            cell2 = cell1
+         end if
+
          if (cell1 <= nCellsSolve .or. cell2 <= nCellsSolve ) then
 
             do k = 1, nVertLevels
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -2884,6 +2884,14 @@ module init_atm_cases
             hs(iCell) = 0.
             if(ter(iCell) .ne. 0.) then
                do j = 1,nEdgesOnCell(iCell)
+
+                  ! For smoothing at cells along the boundary of the mesh, set the terrain value
+                  ! for non-existent neighbors, which map to the "garbage cell", to the same as
+                  ! the terrain in the cell being smoothed
+                  if (cellsOnCell(j,iCell) == nCells+1) then
+                     ter(nCells+1) = ter(iCell)
+                  end if
+
                   hs(iCell) = hs(iCell) + dvEdge(edgesOnCell(j,iCell))    &
                                         / dcEdge(edgesOnCell(j,iCell))    &
                                         * (ter(cellsOnCell(j,iCell))-ter(iCell))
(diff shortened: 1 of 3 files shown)


mgduda:
This PR should probably be merged after #203 , and after adding in checks like [those introduced here](https://github.com/mgduda/MPAS-Model/blob/eb416b71bc61a2d6c8d97a5c719a7dfdd05498f1/src/core_init_atmosphere/mpas_init_atm_static.F#L291-L306) for the MODIS monthly albedo, vegetation fraction, and maximum snow albedo fields added by #203.

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -2907,6 +2907,12 @@ subroutine init_atm_case_gfs(block, mesh, nCells, nEdges, nVertLevels, fg, state
             hs(iCell) = 0.
             if(ter(iCell) .ne. 0.) then
                do j = 1,nEdgesOnCell(iCell)
+
+                  ! For smoothing at the edge of the mesh, effectively ignore non-existent neighbors
```
MiCurry: I know its getting a bit into the 'what' side of comments, but it might be worth adding the method for ignoring non-existent neighbors i.e. setting them equal to the edge cell.
```
! For smoothing at the edge of the mesh, effectively ignore non-existent neighbors by setting them equal
! to the edge cell
```

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -281,8 +292,22 @@ subroutine init_atm_static(mesh, dims, configs)
...
-          ter(iPoint) = ter(iPoint) + rarray(i,j,1)
-          nhs(iPoint) = nhs(iPoint) + 1
+          if (bdyMaskCell(iPoint) < nBdyLayers) then
+             ter(iPoint) = ter(iPoint) + rarray(i,j,1)
+             nhs(iPoint) = nhs(iPoint) + 1
+          else
+             zPixel = sphere_radius * sin(lat_pt)
+             xPixel = sphere_radius * cos(lon_pt) * cos(lat_pt)
+             yPixel = sphere_radius * sin(lon_pt) * cos(lat_pt)
+
```
MiCurry: I think it might be worth having a short comment here (and at further calls to `in_cell`) to explain this section. For instance:

```
! Avoid falsely inflating (or deflating) a regional mesh's border cells by determine if a pixel should be interpolated or ! not
```

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -827,6 +879,147 @@ real (kind=RKIND) function sphere_distance(lat1, lon1, lat2, lon2, radius)
...
+!  routine rotate_about_vector
+!
+!> \brief Rotates a point about a vector in R3
+!> \author Michael Duda
+!> \date   7 March 2019
+!> \details
+!>  Rotates the point (x,y,z) through an angle theta about the vector
+!>  originating at (a, b, c) and having direction (u, v, w).
+!
+!>  Reference: http://inside.mines.edu/~gmurray/ArbitraryAxisRotation/ArbitraryAxisRotation.html
```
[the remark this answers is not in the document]
mgduda: I've found what I think is the equivalent page, and I've updated the link.

src/core_init_atmosphere/mpas_init_atm_cases.F
mgduda: New comments have been added.

src/core_init_atmosphere/mpas_init_atm_static.F
mgduda: New comments have been added.

mgduda:
@MiCurry I've rebased the branch onto the latest `develop` so that I can add similar changes to the processing of monthly albedo, monthly vegetation fraction, and maximum snow albedo. I think I've addressed all of your comments with amendments during the rebasing.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -69,6 +74,7 @@ subroutine init_atm_static(mesh, dims, configs)
  integer, pointer :: supersample_fac
 
  real(kind=RKIND):: lat,lon,x,y
+ real(kind=RKIND):: lat,lon,x,y,z
```
mgduda: This line should replace the line above. As is, the code fails to compile with ifort. I'm going to correct this with a rebased branch before merging.

PR #205, opened 2019-04-12
Import netcdf constants for netcdf-only builds

amametjanov:
When PIO is built without pnetcdf, fill value constants NF_FILL_INT, NF_FILL_CHAR etc. are not included. This update fixes build errors in such cases.

[BFB]

### DIFF
--- a/src/framework/mpas_io.F
+++ b/src/framework/mpas_io.F
@@ -17,6 +17,10 @@ module mpas_io
    use pionfatt_mod
    use pio_types
 
+#ifndef USE_PIO2
+include 'netcdf.inc'
+#endif
+
 #ifdef SINGLE_PRECISION
    integer, parameter :: PIO_REALKIND = PIO_REAL
 #else


mgduda:
We've found it necessary to make similar changes when using the MPAS-A dycore in CAM/CESM: https://github.com/mgduda/MPAS-Model/commit/291176d . I'd like to see whether we can reconcile the changes in this PR with those made in commit 291176d.

amametjanov:
That also eliminated build errors. Can you push that to `develop`?

mgduda:
@amametjanov Thanks very much for checking! I'll make a PR to `develop` with a more descriptive commit message.

mgduda:
@amametjanov The problem with the change in 291176d is that it apparently doesn't work when the PIO1 library was compiled with support for only the parallel-netcdf library:
```
mpif90 -D_MPI -DUNDERSCORE -DCORE_INIT_ATMOSPHERE -DMPAS_NAMELIST_SUFFIX=init_atmosphere -DMPAS_EXE_NAME=init_atmosphere_model -DMPAS_NATIVE_TIMERS -DMPAS_GIT_VERSION=v6.2-122-ge8b87ad-dirty -O3 -m64 -ffree-line-length-none -fconvert=big-endian -ffree-form -fdefault-real-8 -fdefault-double-8 -c mpas_io.F -I/dev/shm/pio-gnu/include -I/dev/shm/pnetcdf-gnu/include -I/dev/shm/pio-gnu/include -I/dev/shm/pnetcdf-gnu/include -I../external/esmf_time_f90
mpas_io.F:30:43:

    integer, parameter :: MPAS_INT_FILLVAL = NF90_FILL_INT
                                           1
Error: Parameter ‘nf90_fill_int’ at (1) has not been declared or is a variable, which does not reduce to a constant expression
mpas_io.F:31:46:

    character, parameter :: MPAS_CHAR_FILLVAL = NF90_FILL_CHAR
                                              1
Error: Parameter ‘nf90_fill_char’ at (1) has not been declared or is a variable, which does not reduce to a constant expression
mpas_io.F:44:54:

    real (kind=RKIND), parameter :: MPAS_REAL_FILLVAL = NF90_FILL_DOUBLE
                                                      1
Error: Parameter ‘nf90_fill_double’ at (1) has not been declared or is a variable, which does not reduce to a constant expression
```
So, it seems that something a little more sophisticated may be needed to handle the various ways that the PIO1 library might be built.

amametjanov:
In E3SM, pio1+parallel_netcdf (no serial netcdf) option is not supported: either pio1+netcdf, or pio1+netcdf+pnetcdf only.

mgduda:
We do support PIO1 + pnetcdf for stand-alone MPAS-A, so I think we'll need to come up with a solution that supports all three variants of PIO1 builds (netcdf, netcdf+pnetcdf, pnetcdf). My first thought is that we could set additional cpp macros (to be added to `CPPFLAGS`) based on information gathered in the top-level `Makefile`. The difficulty is that, although a netCDF library may be present in the build environment, PIO might not necessarily have been compiled with netCDF support. I'll give this more thought.

jonbob:
@mgduda - have you had any luck with this? We're running into a situation in E3SM in which we need this to function on some specific platforms.

mgduda:
@jonbob @amametjanov Sorry for losing track of this issue. Does the E3SM build environment already define macros that would allow us to determine whether we're building within E3SM? If so, would it work for you all to wrap the `include 'netcdf.inc'` statement inside an E3SM-specific macro?

jonbob:
@mgduda - thanks for jumping back in. Right now we don't define any E3SM macros for the MPAS build, though we do for CIME and some other places -- so it is possible. If you feel strongly, I can pursue adding the flags to the MPAS build as well.

PR #206, opened 2019-04-12
Adds boundary tidal forcing capability

pwolfram:
This allows restoring of the layerThickness to correspond to relaxation toward a tidally-enforced free-surface boundary condition.  Note, this is tested for RK4.  Split explicit was not as stable.

This solves for a layer thickness tendency of the form

M*(tidal_boundary - ssh) / tidal_tau * layerThickness/sum_k(layerThickness)

where

* M is a boundary condition mask for application of the tidal boundary
* tidal_boundary is the value of the tide to be applied on the boundary
* ssh is the sea surface height
* tidal_tau is the time scale corresponding to application of the tendency
* layerThickness is the iCell,k varying layerThickness variable
* sum_k(layerThickness) is the depth-averaged layerThickness (total water thickness)

The tendency is used to apply tidal forcing to the prognostic equations corresponding to a mask designating the location of the applied tidal boundary.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -771,6 +771,32 @@
 					possible_values="Real Values greater than zero less than bottomDepth"
 		/>
 	</nml_record>
+	<nml_record name="tidal_forcing" mode="init;forward">
+		<nml_option name="config_use_tidal_forcing" type="logical" default_value=".false." units="unitless"
+					description="Controls if tidal forcing is used."
+					possible_values=".true. or .false."
+		/>
+		<nml_option name="config_use_tidal_forcing_tau" type="real" default_value="10000" units="s"
+					description="Controls time scale for relaxation of tidal forcing."
+					possible_values="Real non-zero value."
+		/>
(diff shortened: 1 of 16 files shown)


pwolfram:
# Testing
A 25km long test case is included that makes a comparison in an analysis step against the forced tidal signal vs the actual MPAS-O free surface:

Screen Shot 2019-04-12 at 4 18 29 PM

pwolfram:
Tested locally and on IC at `/lustre/scratch4/turquoise/pwolfram/PR_tests/MPAS-Model_tidal_forcing`.

pwolfram:
@caozd999, here is the test case with tidal forcing you need for your sediment transport work as we discussed.

mark-petersen:
Passes nightly regression, bfb with ocean/develop. Of course, that does not exercise the new code. @pwolfram or @caozd999, please compile with DEBUG=true and run with your standard test case to create the new initial condition for tidal forcing and run the forward model with 
`config_use_tidal_forcing=.true.`
That should catch any typos or simple bugs.

pwolfram:
@mark-petersen, thank you.  I've compiled and run the case with DEBUG=true and it produces the output as expected.

PR #209, opened 2019-04-15
Add ability to output max and min quantities in timeSeriesMonthly

vanroekel:
This adds the ability to output maximum and minimum observed quantities for each month.  It utilizes the current timeseries AM functionality.

For now, default output for these streams is set to 2-D surface fields.


vanroekel:
@jonbob graciously tested this in E3SM.  A few results below

January Mixed Layer Depth mean
janmeanmld

January MLD max
julmeanmax

Jan MLD min
janmeanmin

July Mean MLD
julmeanmean

July Max MLD
julmeanmax

July Min MLD
julmeanmin

xylar:
This seems like a really useful capability.  Let me know when you have runs with it and I can add it to MPAS-Analysis.  I presume it'll be easy to detect through the namelist options whether this is available/on?  

It looks like this changes things so mean, min and max go to separate streams, and that the old monthly stream get renamed.  I want to make sure we're all on board with this change.  At the moment, this will break MPAS-Analysis so we'll need a strategy for updating the analysis that's still backwards compatible with the "old" way of doing things.

vanroekel:
great point.  I had renamed the streams for clarity, but it is not necessary.  I can revert those changes.

xylar:
> I can revert those changes.

Don't do that just yet.  I'd like to make sure we discuss that.  @milenaveneziani, what do you think?

If the min, max and mean could go to the same stream, that would mean no changes needed in MPAS-Analysis but is less clear and more unwieldy.

vanroekel:
sounds good on naming

Regarding streams, when I looked through the timeseries code it does not appear possible to have a unified stream.  For each stream, the operation is passed in the init phase, so I needed to initialize three times.

mark-petersen:
@vanroekel Thanks for this addition.

I currently have two E3SM PRs in line: 
1. https://github.com/E3SM-Project/E3SM/pull/2843
2. https://github.com/E3SM-Project/E3SM/pull/2852

After that we could make a PR for BFB MPAS changes.  

Do you want to turn the new monthly min and monthly max on by default for some E3SM Comp sets "out of the box?"  If so, for all or just some? 

Regardless, it looks like the E3SM PR will need corresponding changes to the default namelist and streams file.

xylar:
> Regarding streams, when I looked through the timeseries code it does not appear possible to have a unified stream. For each stream, the operation is passed in the init phase, so I needed to initialize three times.

Hmm, that seems counter to the flexibility of streams elsewhere in MPAS.  But time series stat has always had its own set of ... idiosyncrasies.

xylar:
> Do you want to turn the new monthly min and monthly max on by default for some E3SM Comp sets "out of the box?" If so, for all or just some?

I think this should be on by default for all.  The extra data will be small (because this is only for 2D fields).

mark-petersen:
I see. On the topic of streams flexibility, we could make a stream that has no i/o, like "min_max_surface_variables", and then include that stream in both streams:
- timeSeriesStatsMonthlyMinOutput
- timeSeriesStatsMonthlyMaxOutput

That would be a way to avoid repeating the variable list. We do that with the mesh stream. But including a list of variables twice is honestly OK too.

vanroekel:
I agree about on by default.  Also, @jonbob has all the E3SM script changes done and ready to go.

@xylar and @mark-petersen  -- thinking about this a bit more, I think it might be possible to unify the streams.  I think if I point all the output streams in the namelist to the same place it might work. to have everything in one place.  The issue would be that we would then get min max of ALL variables in the monthly mean, which I don't think we want.  What do you think?

vanroekel:
@mark-petersen I didn't see your message before posting.  I could make those changes too.  It gets around my issue.

mark-petersen:
Adding min/max on ALL monthly mean variables is just too much i/o.  Making it default on but for 2D variables is perfect. Then if someone wants to add 3D variables by hand, that is fine.  @vanroekel I think that was your intention anyway.

xylar:
If there's a solution where all these variables end up in the existing `timeSeriesStatsMonthlyOutput` stream, that would by far be my preference.  A lot less work to handle this in MPAS-Analysis and presumably a lot of other E3SM-related code.

xylar:
@mark-petersen wrote:
> I see. On the topic of streams flexibility, we could make a stream that has no i/o, like  "min_max_surface_variables", and then include that stream in both streams:
> timeSeriesStatsMonthlyMinOutput
> timeSeriesStatsMonthlyMaxOutput
>That would be a way to avoid repeating the variable list. We do that with the mesh stream. But including a list of variables twice is honestly OK too.

This suggestion was seemingly aimed at combining just the Min and Max but could we also combine the existing (Mean) and just keep the existing name?

vanroekel:
This suggestion was seemingly aimed at combining just the Min and Max but could we also combine the existing (Mean) and just keep the existing name?

-- I'm testing this on a smaller (16 cell) test case to see if I can keep the base name and get all variables in one file.

milenaveneziani:
Frankly I don't think we need these for ALL 2-d variables. Just the mixed/boundary layer depth variables would be enough.

If we make this change, we will have to change all variable names in MPAS-Analysis from `timeSeriesStatsMonthly` to `timeSeriesStatsMonthlyMean`.

xylar:
> Frankly I don't think we need these for ALL 2-d variables. Just the mixed/boundary layer depth variables would be enough.

I'm not sure I agree.  I think this is useful for getting a sense of variability within a month of various fields, which could be potentially rather important.

> If we make this change, we will have to change all variable names in MPAS-Analysis from timeSeriesStatsMonthly to timeSeriesStatsMonthlyMean.

The fields are named things like `timeMonthly_avg_layerThickness`.  If I understand right, the new ones are `timeMonthly_min_layerThickness` and `timeMonthly_max_layerThickness`, so there should be no conflict.

milenaveneziani:
> I think this is useful for getting a sense of variability within a month of various fields, which could be potentially rather important.

oh yes, I am sure. I guess I wouldn't know which additional fields I would look at if asked point-blank :)
Anyway, not a big deal, since these are 2d fields.

About the variable names: I think you are right, my bad. Would the output filenames change though?

vanroekel:
@xylar and @mark-petersen and @milenaveneziani I have tried a few different ways to try and get all variables in one file.  There are some idiosyncrasies of the time series stats AM I don't understand.  I tried to look through the time series stats code to see if there is another way, but I can't follow that code.  So for this PR I would propose

1. keeping all monthly average variables and files named as they are now so as not to break mpas-analysis
2. add timeSeriesStatsMonthlyMin/Max files with all 2-d fields for now.
3. Variable naming for new min/max would be timeMonthly_min_XXXX or timeMonthly_max_XXXX

Does that sound reasonable?  If pulling all variables into one file is desired, I will need some assistance.

milenaveneziani:
This is fine by me. It wouldn't affect the current version of MPAS-Analysis; we'd just have to keep it in mind when adding a new capability to it.

xylar:
@vanroekel, thanks for trying alternatives.  Your plan sounds fine to me.  If we can combine the streams in the long run, that would be great.  Unfortunately, I don't know if anyone on the team understands the code very well...

vanroekel:
okay.  I have it such that the monthly means are identical to what was there before, so maps-analysis should be fine.  In the new max/min files, variables will have a take a slightly different convention for naming.  in the max file it will be timeSeriesMonthlyMax_max_XXXX and in min it is timeSeriesMonthlyMin_min_XXXX. this seems required based on the current construction of the time series AM.

xylar:
@vanroekel, is there some example output somewhere (preferably 1 year +) that I could use to test out MPAS-Analysis and add these new min/max plots?  Then, if MPAS-Analysis works as expected, I could approve the PR.

vanroekel:
@xylar yes, I have a 5-year G-case with output from this PR on edison here
```
/scratch2/scratchdirs/lvroekel/acme_scratch/edison/20190420.GMPAS-IAF.T62_oEC60to30v3_control.edison/run
```
I think permissions should be open, but let me know if there are problems.

mark-petersen:
I'd like to merge this PR today, and put the corresponding PR into E3SM for tonight if we can. @vanroekel Is it ready? @xylar are you waiting on something for approval?

xylar:
Yes, please wait. I'll test analysis on the 5 years of data above.

xylar:
I'm running the analysis now.  I don't foresee any problems and I'll approve the PR as soon as it completes successfully.

xylar:
I'm going to approve the PR.  The analysis went fine.  However I notice the following enormous file that looks troubling:
```
-rw-rwx--- 1 lvroekel acme  99G Apr 23 04:46 mpaso.rst.am.timeSeriesStatsMonthly.0001-01-01_00000.nc
```
I assume that was part of debugging the streams and not something that will get written out in future runs?

vanroekel:
@xylar yes that was a mistake in how I setup the streams.  That file (time series restart) should not be written normally.  But we will be sure this is not part of the final configuration for the e3sm PR.

mark-petersen:
Merged locally, did one last compile and test.

PR #210, opened 2019-04-16
Remove WRF if-defs around 2D dx declarations for New Tiedtke and MM5 sfclay

kkeene44:
For the purpose of WRF/MPAS physics unification,
'if defined(mpas/wrfmodel)' statements were originally added to
module_cu_ntiedtke and module_sf_sfclay because MPAS used a
2-dimensional dx, while WRF used a 1D dx. WRF code has now been modified
to use a 2d grid distance, and therefore these
'if defined(mpas/wrfmodel)' statements are no longer needed.

Confirmed that MPAS compiles with these modifications.


mgduda:
@kkeene44 It might be good to break this PR into two commits: one to modify the NTiedtke scheme and another to modify the MM5 surface-layer scheme.

Issue #207, opened 2019-04-13
PIO version check fails with PGI 18.10 compiler


mgduda:
The check in the top-level `Makefile` for a working PIO library fails with the PGI 18.10 compiler, apparently because the source files need to be specified at the beginning of the compiler's command-line arguments.

As-is, we get the following error (for what should be a valid PIO2 library):
```
bash-4.2$ mpif90  -I/glade/work/duda/pio-2.3.1-pgi18.10/include -I/glade/work/duda/libs-pgi18.10-skylake/include -I/glade/work/duda/libs-pgi18.10-skylake/include -I/glade/work/duda/libs-pgi18.10-skylake/include -O3 -byteswapio -Mfree -O3 -L/glade/work/duda/pio-2.3.1-pgi18.10/lib -lpiof -lpioc -L/glade/work/duda/libs-pgi18.10-skylake/lib -lnetcdff -lnetcdf -L/glade/work/duda/libs-pgi18.10-skylake/lib -lpnetcdf -L/glade/work/duda/libs-pgi18.10-skylake/lib -lhdf5_hl -lhdf5 -ldl -lz -o pio2.out pio2.f90
/glade/scratch/duda/pgf90vUZc-XbvgynC.o: In function `.C1_283':
pio2.f90:(.data+0x4): undefined reference to `pio_'
pio2.f90:(.data+0x1c): undefined reference to `pio_support_'
pio2.f90:(.data+0x2c): undefined reference to `pio_kinds_'
pio2.f90:(.data+0x34): undefined reference to `pio_types_'
pio2.f90:(.data+0x3c): undefined reference to `pionfget_mod_'
pio2.f90:(.data+0x44): undefined reference to `pionfput_mod_'
pio2.f90:(.data+0x4c): undefined reference to `pionfatt_mod_'
pio2.f90:(.data+0x54): undefined reference to `pio_nf_'
pio2.f90:(.data+0x5c): undefined reference to `piodarray_'
pio2.f90:(.data+0x64): undefined reference to `piolib_mod_'
```

Moving the `pio2.f90` argument to the beginning of the compiler arguments seems to resolve the issue:
```
[...]

PR #212, opened 2019-04-17
PIO Version Compatibility Check - PGI Compiler Bug Fix

MiCurry:
This commit fixes the bug addressed in issue #207, that caused the PGI compiler 18.10 to fail upon reaching the PIO version compatibility check. This single commit moves the target file to the front of the compiler arguments, which fixes the problems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -744,8 +744,8 @@ pio_test:
 	@# See whether either of the test programs can be compiled
 	@#
 	@echo "Checking for a usable PIO library..."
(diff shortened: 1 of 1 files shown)


mgduda:
@matthewhoffman @mark-petersen The changes in this PR only update the small test programs that are generated by the top-level Makefile, compiled, then deleted. Accordingly, I think it may not be necessary to run actual model simulations to test these changes. Verifying that the build process still works should be sufficient.

mgduda:
@matthewhoffman Was there any testing of this PR that you'd like to do with the land-ice core specifically?

PR #214, opened 2019-04-19
Multiple output intervals

t-brown:
This adds the ability to have multiple different output frequencies/intervals per stream.

The key points are

* Streams can have multiple alarms.
* The alarms now has an end timer.
* Parsing of the XML stream attributes `input_interval` and `output_interval`.

The XML stream attributes `input_interval` and `output_interval` are backwards compatible. However the new parser looks for `interval,start,end` 3-tuples. 
For example: 

```
output_interval="1:00:00,start,+1_11:00:00;6:00:00,+1_12:00:00,end"
```

Means there are two output intervals.

1. Hourly, starting at the model start time, and going for 35 hours.
2. Every six hours, starting at 36 hours, and going to the model end time.

Note, that the time interval argument is still a time interval (`MPAS_TimeInterval_type`), while the start and end arguments can be times (`MPAS_Time_type`) if specified as `YYYY-MM-DD_hh:mm:ss` or time intervals (`MPAS_TimeInterval_type`) if prefixed with either `+` or `-`.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_manager.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_manager.F
@@ -470,7 +470,7 @@
  if(trim(config_radtlw_interval) /= "none") then
     call mpas_set_timeInterval(alarmTimeStep,timeString=config_radtlw_interval,ierr=ierr)
     alarmStartTime = startTime
-    call mpas_add_clock_alarm(clock,radtlwAlarmID,alarmStartTime,alarmTimeStep,ierr=ierr)
+    call mpas_add_clock_alarm(clock,radtlwAlarmID,alarmStartTime,alarmTimeInterval=alarmTimeStep,ierr=ierr)
     if(ierr /= 0) &
        call physics_error_fatal('subroutine physics_run_init: error creating radtlwAlarmID')
 
@@ -489,7 +489,7 @@
  if(trim(config_radtsw_interval) /= "none") then
     call mpas_set_timeInterval(alarmTimeStep,timeString=config_radtsw_interval,ierr=ierr)
     alarmStartTime = startTime
-    call mpas_add_clock_alarm(clock,radtswAlarmID,alarmStartTime,alarmTimeStep,ierr=ierr)
+    call mpas_add_clock_alarm(clock,radtswAlarmID,alarmStartTime,alarmTimeInterval=alarmTimeStep,ierr=ierr)
     if(ierr /= 0) &
        call physics_error_fatal('subroutine physics_run_init: error creating alarm radtsw')
 
@@ -508,7 +508,7 @@
  if(trim(config_conv_interval) /= "none") then
     call mpas_set_timeInterval(alarmTimeStep,timeString=config_conv_interval,ierr=ierr)
     alarmStartTime = startTime
-    call mpas_add_clock_alarm(clock,convAlarmID,alarmStartTime,alarmTimeStep,ierr=ierr)
+    call mpas_add_clock_alarm(clock,convAlarmID,alarmStartTime,alarmTimeInterval=alarmTimeStep,ierr=ierr)
     if(ierr /= 0) &
        call physics_error_fatal('subroutine physics_init: error creating convAlarmID')
 
@@ -527,7 +527,7 @@
  if(trim(config_pbl_interval) /= "none") then
     call mpas_set_timeInterval(alarmTimeStep,timeString=config_pbl_interval,ierr=ierr)
     alarmStartTime = startTime
-    call mpas_add_clock_alarm(clock,pblAlarmID,alarmStartTime,alarmTimeStep,ierr=ierr)
+    call mpas_add_clock_alarm(clock,pblAlarmID,alarmStartTime,alarmTimeInterval=alarmTimeStep,ierr=ierr)
     if(ierr /= 0) &
        call physics_error_fatal('subroutine physics_init: error creating pblAlarmID')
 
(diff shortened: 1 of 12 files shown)


MiCurry:
@t-brown you had mostly the same gist here as you did in your email. I just enjoyed seeing the additional example.

> interval1,start1,end1;interval2,start2,end2
> 
> A real example:
> 1:00:00,start,+1_11:00:00;6:00:00,+1_12:00:00,end
> Which means hourly output from the model run start to the 35th hour, then 6 hourly output to the model run end. I'm using "±" to indicate a time interval from model start and the two key words "start" and "end".

MiCurry:
Thanks again for the PR @t-brown. I think interval streams will be much appreciated by users. @mgduda and I have some thoughts on the syntax you proposed and some of the changes around them.

We really think having syntax similar to how streams are defined and adding additional `start_time` and `end_time` attributes  to specify an interval aligns most with our goals for MPAS. Mainly that being user convince. Here’s an example of that syntax we think would work:

```
<stream name="output-hourly"          
        type="output"          
        filename_template="history.$Y-$M-$D_$h.$m.$s.nc"
         start_time_on="2010-10-23_00:00:00"
         end_time="2010-10-24_12:00:00"
         output_interval="1:00:00" >      
         <file name="stream_list.atmosphere.output"/>
</stream>
 
<stream name="output-6-hourly"          
         type="output"          
         filename_template="history.$Y-$M-$D_$h.$m.$s.nc"
         start_time_after="2010-10-24_12:00:00"
         end_time='model-end'
         output_interval="6:00:00" >      
        <file name="stream_list.atmosphere.output"/>
</stream>
```
We think having new attributes `start_time_on`, `start_time_after`, `end_time` and `output_interval`, achieve the same result of your proposed syntax, but are easier to read and understand, allow any number of streams and will also eliminate the need to have multiple alarms per stream. This will help cut down on complexity, while keeping the familiarity to stream creation.

The ability to specify either a `MPAS_Time_type` or a `MPAS_TimeInterval_type` is a nice idea and I can definitely see being useful for real time forecasts and probably other use cases. I think we should keep that syntax.


Understandably, this will be more work. However, @mgduda and I are both willing to help out and get this into the model. We'd both like this in the model and we're more then willing to help contribute and create tests. Please, let us know if you have any additional comments, questions or points of discussion!

> Does MPAS have the equivalent of WRFs WTF?

MPAS currently does not have a testing framework like WTF that we use for testing, though we are slowly working on one. Currently, we mainly run unit tests and we also put tests for framework changes like this one into the `test_core`, which can be compiled and ran just as any other core.

mgduda:
@mark-petersen @matthewhoffman @akturner It seems appropriate to bring you all into this discussion. In summary, we'd like to add the capability to MPAS to write output streams (and to read input streams?) only during some window of time given as a starting time and a stopping time. The question that seems most important to discuss at this moment is how this should look from a user interface perspective. After we've hashed out a few quick ideas here, we could probably put together a requirements and design document if there's interest in that.

mgduda:
To offer my input on the stream interface, as @MiCurry said, using separate XML attributes for the starting and stopping time of the stream seems much easier to explain to users, and it fits better with the existing design of the streams interface: we have one attribute that controls one aspect of a stream. The only alternative I'd offer from what @MiCurry has already said is that the new attributes could be called something like `output_start` and `output_stop`, which could be set to an absolute time or a relative time. The following seems readable to me with almost no explanation needed:
```
<stream name="diagnostics_1hr"
        type="output"
        filename_template="diag.$Y-$M-$D_$h.$m.$s.nc"
        output_start="+0:00:00"
        output_interval="1:00:00"
        output_stop="+36:00:00" >

        <file name="stream_list.atmosphere.diagnostics"/>
</stream>

<stream name="diagnostics_6hr"
        type="output"
        filename_template="diag.$Y-$M-$D_$h.$m.$s.nc"
        output_start="+36:00:00"
        output_interval="6:00:00"
        output_stop="+120:00:00" >

        <file name="stream_list.atmosphere.diagnostics"/>
</stream>
```
I do appreciate the concern of @t-brown that with many intervals for a stream, having separate `<stream>` tags for each interval could lead to long streams definition files. However, I'd be interested to understand how often a user would likely use more than two intervals for an output stream? Sacrificing some duplication for the sake of clarity seems like a reasonable tradeoff to me.

I'd propose that we don't worry about implementation details until we have a design for the interface that we can all live with.

xylar:
@t-brown, thank you for suggesting this functionality and giving it a first go!  This will be really useful for cases where there's a  spin-up phase and a more equilibrated phase.

For what it's worth, I prefer @mgduda's syntax.  I find it much more intuitive than combining all of this into a single string, even at the expense of redundancy.  (In any case, I think there's a way to indicate that a stream has the same contents as another stream, right?)

I think a short design doc would be useful for preserving this thinking for posterity.

mark-petersen:
I agree this would be useful, and that `output_start` and `output_stop` are clear, no explanation needed. Thanks.

t-brown:
Hi all,

I have two concerns with duplicating the stream definitions, which I unfortunately do not have the free cycles to address.

- The fact that the stream ID seems to be hard coded in [places](https://github.com/MPAS-Dev/MPAS-Model/blob/ce78c320d2f73a2a9f4b4ab1448bb3061c8c5844/src/core_atmosphere/mpas_atm_core.F#L508).
- They will fail the [filename uniqueness check](https://github.com/MPAS-Dev/MPAS-Model/blob/ce78c320d2f73a2a9f4b4ab1448bb3061c8c5844/src/framework/xml_stream_parser.c#L469).

Maybe instead of duplicating the stream, have the interval as a XML element (not an attribute), then one could have multiple elements. For example:

```
<stream name="diagnostics"
  type="output"
  filename_template="diag.$Y-$M-$D_$h.$m.$s.nc" >
  <file name="stream_list.atmosphere.diagnostics"/>
  <output>
    <start>mode_start</start>
    <interval>1:00:00</interval>
    <end>+35:00:00</end>         
  </output>
  <output>
    <start>+36:00:00</start>
    <interval>6:00:00</interval>
    <end>model_end</end>         
  </output>
</stream>
```

So I'm going to close this PR and look forward to seeing your solution in a future version of MPAS.

MiCurry:
Thanks again @t-brown for sharing your changes with us.

For a future design documents @mgduda and have been discussing updating the filename uniqueness check to determine if two streams overlap or not and failing the check if they do and passing the check if they don’t overlap, even if they have the same filename. But likewise, we’ve got a few other things on our plate at the current moment.

As for the hard coded stream ID, I’m not 100% sure on that. I’m sure @mgduda knows more about this.

Another point of future discussion and decision is how inclusivity or exclusivity is specified. For example:
```
ouput_start_on=“+0:00:00”
output_end=“+12:00:00”
…
ouput_start_after=“+12:00:00”
output_end=“+24:00:00”
```

PR #219, opened 2019-04-24
Add MPAS-A "init case" for producing regional LBCs

mgduda:
This merge adds the capability to the init_atmosphere core to produce lateral boundary conditions
for regional MPAS-Atmosphere. In order to produce a set of LBC files using the init_atmosphere
core, a user must do the following:
1) Select config_init_case = 9 in the namelist.init_atmosphere file
2) Set config_start_time, config_stop_time, and config_fg_interval for the range of times and interval at which LBCs will be interpolated
3) Ensure that the output_interval for the 'lbc' stream in the streams.init_atmosphere file matches config_fg_interval
4) Ensure that "intermediate" files exist in the run directory for the range of dates specified in the namelist

Note that the LBCs produced by the code here are interpolated from datasets in
the WRF Pre-processing System's "intermediate" file format; this code does not handle
the interpolation of fields from one MPAS mesh to another.

### DIFF
--- a/src/core_init_atmosphere/Registry.xml
+++ b/src/core_init_atmosphere/Registry.xml
@@ -310,6 +310,7 @@
         <packages>
                 <package name="initial_conds" description="Used by test cases that produce initial conditions."/>
                 <package name="sfc_update" description="Used by test cases that produce surface updates."/>
+                <package name="lbcs" description="Active when lateral boundary conditions are interpolated and written."/>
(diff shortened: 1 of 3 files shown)


mgduda:
The changes in this PR have been tested with CFSv2 data. More testing is probably needed with datasets on native model levels (e.g., model-level ERA-I data).

mgduda:
Testing with model-level ERA-I data for initial and boundary conditions revealed no issues, so I think this PR should be ready for review.

PR #221, opened 2019-04-25
Enable blending of terrain field with first-guess terrain along domain boundary

mgduda:
This merge adds the ability to blend the MPAS-A terrain field with first-guess terrain
along the lateral boundaries of a regional domain. This capability is activated by
a new namelist option, config_blend_bdy_terrain, in the &vertical_grid namelist group.

Terrain blending happens before terrain smoothing and vertical grid generation,
but after static interpolation. The first-guess terrain is taken to be
the SOILGHT field found in an intermediate file described by config_met_prefix
and config_start_time.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -2498,6 +2498,7 @@ module init_atm_cases
       real (kind=RKIND), dimension(:), pointer :: dvEdge, dcEdge, areaCell 
       real (kind=RKIND), dimension(:,:), pointer :: v
       real (kind=RKIND), dimension(:,:), pointer :: sorted_arr
+      integer, dimension(:), pointer :: bdyMaskCell
 
       type (field1DReal), pointer :: tempField
       type (field1DReal), pointer :: ter_field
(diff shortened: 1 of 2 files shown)


src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -5196,4 +5122,187 @@ subroutine decouple_variables(mesh, nCells, nVertLevels, state, diag)
...
+      character(len=*), intent(in) :: config_start_time
+      integer, intent(in) :: nCells
+      real (kind=RKIND), dimension(:), intent(in) :: latCell  ! These four variables (latCell, lonCell, bdyMaskCell, and ter)
+      real (kind=RKIND), dimension(:), intent(in) :: lonCell  ! may actually have more than nCells elements, for example,
+      integer, dimension(:), intent(in) :: bdyMaskCell        ! if the arrays include a "garbage cell".
+      real (kind=RKIND), dimension(:), intent(inout) :: ter   ! 
+      integer, intent(out) :: ierr
+
+      integer, parameter :: nBdyLayers = 7   ! The number of relaxation layers plus the number of specified layers
+      integer, parameter :: nSpecLayers = 2  ! The number of specified layers
```
MiCurry: Maybe this can be done in a future PR, but it might be worth determine a place where to define both `nBdyLayers` and `nSepcLayers`. As currently, another nBdyLayers is also defined in `mpas_init_atm_static.F` to avoid. I don't think there is currently a good place to put it. Perhaps as a configuration option?

mgduda: I agree that it would be better to obtain these from a single location, ideally from the input mesh itself. At present, however, we aren't encoding these in the limited-area meshes, and the values of 7 and 2 are hard-coded in the `limited-area` program. Once `limited-area` has been updated or rewritten, I think we could save these as global attributes in the mesh file, and pull the values from there.

PR #177, opened 2019-03-07
Update COMPASS scripts to support python 3

xylar:
The changes mostly involve fixing `print` statements.

The list command has also been modified to produce a sorted list of test cases (as opposed to the random order that was previously produced).

PR #222, opened 2019-04-29
Fix decoding issue of git version for py3

xylar:
The `utility_scripts` were not included in #177 

Fixes a bug by decoding the `git_version` into a unicode string in `manage_regression_test_suite.py`, required for python 3 compatibility.

### DIFF
--- a/testing_and_setup/compass/utility_scripts/compare_fields.py
+++ b/testing_and_setup/compass/utility_scripts/compare_fields.py
@@ -1,8 +1,11 @@
 #!/usr/bin/env python
-import sys, os, glob, shutil
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import sys
+import os
 import numpy as np
 
-from netCDF4 import *
 from netCDF4 import Dataset as NetCDFFile
 import argparse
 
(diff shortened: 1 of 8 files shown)


xylar:
## Testing

I used this fix as part of testing my new COMPASS conda environment for python 3 (see https://github.com/MPAS-Dev/geometric_features/pull/102#issuecomment-487390567).

testing_and_setup/compass/utility_scripts/compare_fields.py
```diff
@@ -18,153 +21,153 @@
 args = parser.parse_args()
 
 if not args.filename1:
-	parser.error("Two filenames are required inputs.")
+    parser.error("Two filenames are required inputs.")
```
xylar: Lots of tabs have been replaced with spaces.  It's a bad practice to have a mix of tabs and spaces in a python file, and generally tabs should be avoided.

xylar:
@mark-petersen, this was included in #229 and was therefore part of your testing and mine.  However, it needs to get merged to `develop` and then we need to merge `develop` to `ocean/develop` before #229 can go in.

testing_and_setup/compass/utility_scripts/compare_fields.py
mark-petersen: Thanks. I appreciate that.

mark-petersen:
@mgduda and @matthewhoffman: FYI this is COMPASS only, but a merge into develop. It does not effect atmosphere core.  I will be merging this today.

xylar:
Thanks!

Issue #215, opened 2019-04-22
Mid-point averaging vs area averaging for forcing


pwolfram:
In specification of the edge-normal wind at https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F#L204 a factor of 0.5 is used for the average:
```fortran
      ! Convert coupled climate model wind stress to MPAS-O wind stress
      !$omp do schedule(runtime) private(cell1, cell2, zonalAverage, meridionalAverage)
      do iEdge = 1, nEdges
        cell1 = cellsOnEdge(1, iEdge)
        cell2 = cellsOnEdge(2, iEdge)

        zonalAverage = 0.5_RKIND * (windStressZonal(cell1) + windStressZonal(cell2))
        meridionalAverage = 0.5_RKIND * (windStressMeridional(cell1) + windStressMeridional(cell2))

        surfaceStress(iEdge) = surfaceStress(iEdge) + cos(angleEdge(iEdge)) * zonalAverage + sin(angleEdge(iEdge)) &
                             * meridionalAverage
      end do
!$omp end do
```

For a multi-resolution mesh, the average should really be area-weighted, not premultiplied by 0.5, which assumes an equilateral mesh.  This will introduce wind-forcing errors on meshes that are changing resolution quickly.

pwolfram:
cc @sbrus89

PR #227, opened 2019-05-03
Use area-weighting when converting wind stress from centers to edges

sbrus89:
This PR addresses issue #215. When computing the wind stress on an edge from the adjacent cell centered values, an area-weighted average (based on edge kite area) is used instead of the mean value.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F
+++ b/src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F
@@ -168,14 +168,21 @@ contains
       !
       !-----------------------------------------------------------------
 
-      integer :: iEdge, cell1, cell2, iCell, nCells, nEdges
+      integer :: iEdge, cell1, cell2, iCell, nCells, nEdges, vertex1, vertex2, cell
       integer, dimension(:), pointer :: nCellsArray, nEdgesArray
 
       integer, dimension(:,:), pointer :: cellsOnEdge
-
+      integer, dimension(:,:), pointer :: verticesOnEdge
+      
+      real (kind=RKIND) :: x1, y1, z1
+      real (kind=RKIND) :: x2, y2, z2
+      real (kind=RKIND) :: x3, y3, z3
+      real (kind=RKIND) :: areaWeights(2)
       real (kind=RKIND) :: meridionalAverage, zonalAverage
       real (kind=RKIND), dimension(:), pointer :: angleEdge
       real (kind=RKIND), dimension(:), pointer :: windStressZonal, windStressMeridional
+      real (kind=RKIND), dimension(:), pointer :: xVertex, yVertex, zVertex 
+      real (kind=RKIND), dimension(:), pointer :: xCell, yCell, zCell 
 
       err = 0
 
@@ -188,6 +195,13 @@ contains
 
       call mpas_pool_get_array(meshPool, 'angleEdge', angleEdge)
       call mpas_pool_get_array(meshPool, 'cellsOnEdge', cellsOnEdge)
+      call mpas_pool_get_array(meshPool, 'verticesOnEdge', verticesOnEdge)
+      call mpas_pool_get_array(meshPool, 'xVertex', xVertex)
+      call mpas_pool_get_array(meshPool, 'yVertex', yVertex)
+      call mpas_pool_get_array(meshPool, 'zVertex', zVertex)
+      call mpas_pool_get_array(meshPool, 'xCell', xCell)
+      call mpas_pool_get_array(meshPool, 'yCell', yCell)
+      call mpas_pool_get_array(meshPool, 'zCell', zCell)
 
       call mpas_pool_get_array(forcingPool, 'windStressZonal', windStressZonal)
       call mpas_pool_get_array(forcingPool, 'windStressMeridional', windStressMeridional)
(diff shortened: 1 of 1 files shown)


src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F
```diff
@@ -198,11 +212,36 @@ subroutine ocn_surface_bulk_forcing_vel(meshPool, forcingPool, surfaceStress, su
...
+        z3 = zVertex(vertex2)
+
+        do iCell = 1,2
+          cell = cellsOnEdge(iCell, iEdge)
+          
+          x1 = xCell(cell)
+          y1 = yCell(cell)
+          z1 = zCell(cell)
+
+          areaWeights(iCell) = 0.5_RKIND*(sqrt(((y2-y1)*(z3-z1)-(y3-y1)*(z2-z1))**2+ &
```
xylar: @sbrus89, the area of the "kite" between edges is `dcEdge*dvEdge`.  I think you could use this (or half of this) as your area and save a *lot* of computation.

sbrus89:
@pwolfram and I discussed this a bit and decided to open it up to anyone else who has feedback. The current approach requires a kiteAreasOnEdge array for the area-weights, which is currently not computed in MPAS (as far as I know). Right now I'm recomputing this each time and it should eventually be pre-computed. The question is where pre-computing this would fit best in the code.

We also discussed using the cellArea values as weights, which is a simpler approach.

Here's a diagram of the implementation at present:

wind_stress_averaging

sbrus89:
I also realize this won't be BFB in E3SM, so I understand any hesitation to adopt this.

xylar:
Believe edges are required to be half way between cell centers, so the red triangles you have drawn above are required to be identical, each with an area 0.5*dcEdge*dvEdge.

xylar:
I could also be wrong, since I think Jigsaw relaxes some of these requirements.  @dengwirda, do you have thoughts?

xylar:
Here's the mesh spec for reference: https://mpas-dev.github.io/files/documents/MPAS-MeshSpec.pdf

src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F
pwolfram: Thanks @xylar, good call.

sbrus89:
If that's the case, this might not even be needed.

xylar:
> If that's the case, this might not even be needed.

Yes, actually, I think you're right.  The areas are supposed to be the same from either side of an edge so no weighting should be needed.  Again, Jigsaw might relax this a bit but I doubt it's a major source of error even so.  And this would almost certainly not be the only place in the code that makes this assumption.

src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F
pwolfram: But, dcEdge is between cell centers.  So, this implicitly is averaging for the area computation, correct?

pwolfram:
The areas of the red triangles cannot be the same, the height of them from the edge to cell mid-points differs.  So, dcEdge*dvEdge is already doing some averaging and in that case the 0.5 factor emerges.

src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F
pwolfram: I don't think we can simplify like this via `dcEdge*dvEdge`; we need the full triangle contained in the hexagon to be computed.

xylar:
@pwolfram, for the Voronoi mesh, the segment between vertices is defined to be perpendicular to the segment between cell centers.  The edge is exactly half way between cell centers.  That means that the two red triangles are of equal area: each is 0.5*dcEdge*dvEdge.

pwolfram:
@xylar, I don't believe this is true but is except for constant resolution meshes.  We can make the cell on one side bigger, which moves the cell center, but the edge remains fixed.  This means that the distance from the edge mid-point to the cell center on one side has now increased or decreased relative to the other side that remained fixed.

> Believe edges are required to be half way between cell centers, so the red triangles you have drawn above are required to be identical, each with an area 0.5dcEdgedvEdge.

xylar:
Note: we had a call about this and @pwolfram and @sbrus89 will look into this further with regard to their coastal meshes.  @pwolfram is still not completely convinced so further discussion may be needed.

pwolfram:
Thanks @xylar, you are correct and thank you for clearing this up-- we conflated triangles with hexagons based on coming from triangular meshes originally.  You can see this property is an issue on the triangular mesh, but not on the hexagonal mesh (see below for illustration).

Screen Shot 2019-05-03 at 9 33 35 AM

The outstanding question for @dengwirda is whether this property will be relaxed under a Hodge Optimized Triangulation as he has been exploring.

src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F
pwolfram: Thanks @xylar!

xylar:
I suggest closing this at least for now.

sbrus89:
Agreed

dengwirda:
Okay, I think there's a few things worth considering here!

Firstly, I believe that @xylar is correct -- by definition each Voronoi edge is guaranteed to bisect the Delaunay edge it's paired with, so the area of the triangles you mention should be the same.

The opposite is not true though -- Delaunay edges don't necessarily bisect Voronoi edges, except when the mesh consists of all regular hexagons. I think this is essentially what @pwolfram is getting at, and can be seen in the figure.

A particularly bad case can occur when obtuse triangles are involved -- in these cases the Voronoi and Delaunay edges actually don't intersect at all (though the projection of the Voronoi edge would still hit the Delaunay midpoint).

Generally speaking `JIGSAW` is reasonably good at generating meshes without any obtuse triangles, even in cases where the resolution is changing, so this is hopefully not too much of an issue in practice, though probably worth keeping in mind.

**

The weighted case (i.e. Hodge optimised) is different though -- here constraints are relaxed, and neither edge is guaranteed to bisect the other (they are still orthogonal though).

I won't go on too much about this, but the general idea is that by distributing the offset between both the triangle and cell edges, it seems to be possible to generate meshes that are more regular overall. For example, it's possible to have obtuse triangles in weighted meshes and for all of the cell and triangle edges to intersect properly.

`MPAS` is currently only using unweighted meshes (i.e. Voronoi / Delaunay grids), so dealing with the weighted geometry isn't currently a live issue. 

To support Hodge-type meshes in the future though, I believe the full set of calculations (as per @sbrus89) would be needed.

sbrus89:
Thanks @dengwirda! We can leave this on the back-burner and keep it in mind for future reference.

PR #237, opened 2019-05-08
Only access config_pio_stride and config_pio_num_iotasks when necessary

mgduda:
This merge ensures that the namelist options config_pio_stride and config_pio_num_iotasks
are accessed only when necessary, enabling the namelists for coupled MPAS configurations
to be minimized by omitting these options.

In the mpas_framework_init_phase2 routine, the namelist options config_pio_stride
and config_pio_num_iotasks were accessed regardless of whether MPAS was being
run in a configuration that used an external PIO io_system or not.

For stand-alone configurations, an external PIO io_system will generally not be
available, in which case config_pio_stride and config_pio_num_iotasks must be
accessible so that their values can be passed to mpas_io_init, where PIO is
initialized.

In coupled configurations, an external PIO io_system will often be provided by
the driver, and a minimal MPAS Registry.xml file may omit config_pio_stride and
config_pio_num_iotasks. Accordingly, the values of these non-existent namelist
options must not be accessed in mpas_framework_init_phase2.

Besides accessing config_pio_stride and config_pio_num_iotasks only when
necessary, this commit also moves the access of config_calendar_type inside of
an if-test on the presence of an external calendar.

### DIFF
--- a/src/framework/mpas_framework.F
+++ b/src/framework/mpas_framework.F
@@ -85,22 +85,36 @@ module mpas_framework
       call mpas_pool_set_error_level(MPAS_POOL_WARN)
 #endif
 
-      call mpas_pool_get_config(domain % configs, 'config_calendar_type', config_calendar_type)
-      call mpas_pool_get_config(domain % configs, 'config_pio_num_iotasks', config_pio_num_iotasks)
-      call mpas_pool_get_config(domain % configs, 'config_pio_stride', config_pio_stride)
-
       if (present(calendar)) then
          call mpas_timekeeping_init(calendar)
       else
+         call mpas_pool_get_config(domain % configs, 'config_calendar_type', config_calendar_type)
          call mpas_timekeeping_init(config_calendar_type)
       end if
 
-      pio_num_iotasks = config_pio_num_iotasks
-      pio_stride = config_pio_stride
+      !
+      ! Note: pio_num_iotasks and pio_stride are only used in MPAS_io_init if io_system is
+      !       not present. In stand-alone configurations, we expect that io_system will not
+      !       be present and that pio_num_iotasks and pio_stride will be available from
+      !       the namelist; in other systems, a PIO io_system may be provided.
+      !
+      if (.not. present(io_system)) then
+         call mpas_pool_get_config(domain % configs, 'config_pio_num_iotasks', config_pio_num_iotasks)
+         call mpas_pool_get_config(domain % configs, 'config_pio_stride', config_pio_stride)
+         pio_num_iotasks = config_pio_num_iotasks
+         pio_stride = config_pio_stride
+      else
+         pio_num_iotasks = -1    ! Not used when external io_system is provided
+         pio_stride      = -1    ! Not used when external io_system is provided
+      end if
+
+      ! A value of 0 for pio_num_iotasks actually indicates that every task should be an I/O task
       if (pio_num_iotasks == 0) then
          pio_num_iotasks = domain % dminfo % nprocs
       end if
+
       domain % ioContext % dminfo => domain % dminfo
+
       call MPAS_io_init(domain % ioContext, pio_num_iotasks, pio_stride, io_system)
 
    end subroutine mpas_framework_init_phase2!}}}


matthewhoffman:
I agree this looks good on inspection.  However, @mark-petersen , perhaps we should test this in E3SM if you haven't already.  I expect it would not hurt anything, but it is also a bit mysterious to me why we did not experience whatever problems the current functionality has when run in an ESM.

mgduda:
@matthewhoffman If the MPAS components in E3SM include the `config_pio_num_iotasks` and `config_pio_stride` namelist options in their `Registry.xml` files, I don't think the issue addressed in this PR would have come up. Because MPAS-A is just the dycore in CAM, we're trying to minimize the set of namelist options needed by the MPAS-A dycore. Since the PIO I/O system is supplied external to the dycore, we've chosen to omit `config_pio_num_iotasks` and `config_pio_stride`, which leads to problems when we query and attempt to use these in `mpas_framework_init_phase2`.

matthewhoffman:
@mgduda , I'm trying to understand this better.  In looking at the MPASO driver code in E3SM, it appears that MPASO is given a PIO I/O system from E3SM, and therefore does not set up its own PIO I/O system.  So if I understand it correctly, in our case `config_pio_num_iotasks` and `config_pio_stride` are being available through the MPASO namelist/Registry-default, but are not being used to actually set up the PIO I/O system.

So you are saying that in your case you want to be able to remove those two namelist options from Registry?  (Removing them from your input namelist is obviously irrelevant, because the Registry defaults would kick in instead.)  Does that mean you are using different versions of Registry for standalone and CESM configurations of MPAS-A?  We have not considered that - is that a challenge to maintain?

src/framework/mpas_framework.F
```diff
@@ -85,22 +85,31 @@ subroutine mpas_framework_init_phase2(domain, io_system, calendar)!{{{
...
+         pio_num_iotasks = config_pio_num_iotasks
+         pio_stride = config_pio_stride
+         if (pio_num_iotasks == 0) then
+            pio_num_iotasks = domain % dminfo % nprocs
+         end if
       end if
+
       domain % ioContext % dminfo => domain % dminfo
+
       call MPAS_io_init(domain % ioContext, pio_num_iotasks, pio_stride, io_system)
```
matthewhoffman: For clarity, it would be good to initialize `pio_num_iotasks` and `pio_stride` to something before the call to `MPAS_io_init` with an `else` clause to the if-construct added above.  I realize that if `io_system` is provided then those two variables are unused in `MPAS_io_init`, but that's not obvious in this routine. and it still seems potentially dangerous (if that changes in the future), or at least confusing, to leave them uninitialized.

mgduda:
@matthewhoffman You're correct that if E3SM is providing a PIO I/O system to MPAS-O, then the values of `config_pio_stride` and `config_pio_num_iotasks` are ignored. Note that the if-test [in mpas_io_init](https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/framework/mpas_io.F#L116) will set a pointer to the provided I/O system and skip most of the rest of the routine.

When MPAS-A is used as a dycore in CAM, we've so far avoided the need to have separate Registry files by adding a new attribute, `used_by`, to namelist, streams, and var tags that are needed by CAM; in our script to generate the CAM namelist reading code, we simply ignore all XML tags that don't have `used_by="cam"` (or more specifically, that don't include `cam` in the `used_by` attribute value. XML has proved to be extensible in this regard. This all came about because the CAM developers really want us to read namelists using CAM's facilities as is done in other dycores, rather than using MPAS's native namelist reading routine. And, there was a (very reasonable) insistence that confusing references to PIO settings not be present in the MPAS-A dycore even if the don't appear in the namelist, since the dycore in CAM should have nothing to do with setting up PIO.

src/framework/mpas_framework.F
mgduda: That's a good point -- that we should never pass uninitialized values to non-optional subroutine arguments -- though I would probably disagree that what we're doing is confusing given the [prominent comment block](https://github.com/MPAS-Dev/MPAS-Model/blob/aa5d8d92c2a5a5cc85c0f7a63a967b8f803be7c8/src/framework/mpas_framework.F#L95-L100) explaining what is happening. I'll push another commit adding an else clause that initializes `pionum_iotasks` and `pio_stride`.

matthewhoffman: @mgduda , thanks for making my suggestion.  It looks like the else clause initializes them to unassigned pointers now, though.  How about 0 or -1 or something?

mgduda: Ugh -- that's really bad of me! I had copy-pasted the assignment from the `if` case to the `else` case, got momentarily distracted, and forgot to actually clean up the assignments before committing and pushing. What do you think about setting `pio_stride` to 1 and `pio_num_iotasks` to 0 (so that after the if-else block, `pio_num_iotasks` would be set to the total number of MPI tasks)? This would correspond to what I believe is the default in all core's `Registry.xml` files.

matthewhoffman: Ok, I thought it might be something like that. :)  I like the idea of something like -1 or -999 to hint that these values are not actually going to be used.  But I'm fine with whatever you prefer, I was primarily interested in seeing some initialized value in there.

mgduda: Using `-1` sounds reasonable to me. I'll also add line comments to state that these values are never expected to be used when an external `io_system` is provided. I should be able to push an amended commit in about 30 minutes.

matthewhoffman: @mgduda , thanks for dealing with my picky requests. :)  Looks good.

PR #242, opened 2019-05-09
Add optional argument 'unitNumbers' to core % setup_log(...) routine

mgduda:
This merge adds a new optional argument, unitNumbers, to the core%setup_log routine.

The mpas_log_init routine takes an optional argument, unitNumbers, to specify
which Fortran unit numbers to use for the output and error log files written by
a core. The mpas_log_init routine is called by the core % setup_log routine,
which previously did not have an argument for passing in logging unit numbers.

This commit adds an optional argument, unitNumbers, to the core % setup_log
routine, and it modifies the implementation of each core's setup_log routine
to pass the optional unitNumbers argument to the mpas_log_init routine.

The core % setup_log routine is called by the MPAS subdriver's mpas_init
routine, and at present no changes are made to the mpas_init routine to pass
optional unitNumbers to the core % setup_log routine. However, the changes in
this commit do enable alternate implementations of mpas_init to supply unit
numbers to be used for logging, e.g., in coupled systems where the coupler
has already defined logging units for each component.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core_interface.F
+++ b/src/core_atmosphere/mpas_atm_core_interface.F
@@ -195,15 +195,16 @@ module atm_core_interface
    !>  and allow the core to specify details of the configuration.
    !
    !-----------------------------------------------------------------------
-   function atm_setup_log(logInfo, domain) result(iErr)!{{{
+   function atm_setup_log(logInfo, domain, unitNumbers) result(iErr)!{{{
 
       use mpas_derived_types, only : mpas_log_type, domain_type
       use mpas_log, only : mpas_log_init, mpas_log_open
 
       implicit none
 
-      type (mpas_log_type), intent(inout), pointer :: logInfo  !< logging information object to set up
-      type (domain_type), intent(in), pointer :: domain  !< domain object to provide info for setting up log manager
+      type (mpas_log_type), intent(inout), pointer :: logInfo    !< logging information object to set up
+      type (domain_type), intent(in), pointer :: domain          !< domain object to provide info for setting up log manager
+      integer, dimension(2), intent(in), optional :: unitNumbers !< Fortran unit numbers to use for output and error logs
       integer :: iErr
 
       ! Local variables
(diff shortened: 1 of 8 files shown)


mgduda:
In implementing the MPAS-Atmosphere dycore in CESM's CAM, I've found that I can reuse much of the MPAS subdriver's `mpas_init` routine, but I do need to be able to feed Fortran unit numbers to the `mpas_log_init` routine via the `core % setup_log` function. The changes in this PR seem like a useful generalization.

mgduda:
I've tested that all cores compile with the GNU compilers, and to verify that I can indeed supply unit numbers -- which may or may not be opened -- from a modified version of `mpas_init`, I've added code like this to the `mpas_init` routine:
```
       unitNumbers(1) = 80
       unitNumbers(2) = 81

       write(logname, fmt='(a,i4.4,a)') 'stdout.', domain_ptr % dminfo % my_proc_id, '.log'
       open(unit=unitNumbers(1), file=trim(logname), form='formatted')

       write(logname, fmt='(a,i4.4,a)') 'stderr.', domain_ptr % dminfo % my_proc_id, '.log'
       open(unit=unitNumbers(2), file=trim(logname), form='formatted')

       ierr = domain_ptr % core % setup_log(domain_ptr % logInfo, domain_ptr, unitNumbers)
       if ( ierr /= 0 ) then
          call mpas_dmpar_global_abort('ERROR: Log setup failed for core ' // trim(domain_ptr % core % coreName))
       end if

```

matthewhoffman:
@mgduda . our drivers for MPAS components in E3SM don't use `mpas_init` directly, but do obviously include all of the same steps.  Out of curiosity, are you able to call `mpas_init` directly?  Or are you simply copying that routine into CESM code but wanting them to be as similar as possible?

(In our case, I don't think we could use `mpas_init` directly, because we have three cores compiled into a single executable, and there are core-specific ifdefs in `mpas_init`.)

mgduda:
@matthewhoffman We're not able to use `mpas_init` directly for the MPAS-A dycore implementation in CAM, either. But, as you guessed, we would like the CAM-specific initialization to follow as closely as possible the stand-alone `mpas_init` code. For any deviations from the stand-alone `mpas_init` code that seem like useful generalizations, I thought it might be nice to feed those back into the stand-alone subdriver.

mark-petersen:
I'm out until June 13. Glancing over this it looks fine. Please proceed without me.

PR #243, opened 2019-05-11
Re-compute initial RH w.r.t. ice where temperature is below freezing 

weiwangncar:
This PR re-computes RH with respect to ice above freezing level in init_atmosphere.

The RH from init_atmosphere is currently with respect to water (coming from WPS/ungrib), which is not consistent with RH from model output, which is computed wrt to ice. When RH from init_atmosphere is used for comparison, an unrealistic bias would be seen.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -2552,6 +2552,7 @@ module init_atm_cases
 
       real (kind=RKIND), dimension(nVertLevels, nCells) :: rel_hum, temperature, qv
       real (kind=RKIND) :: ptmp, es, rs, rgas_moist, qvs, xnutr, znut, ptemp, rcv
+      real (kind=RKIND) :: eis, ews, r1
       integer :: iter
 
       real (kind=RKIND), dimension(nVertLevels + 1) :: hyai, hybi, znu, znw, znwc, znwv, hyam, hybm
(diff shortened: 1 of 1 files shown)


mgduda:
@weiwangncar Would it be preferable to compute relative humidity using the same methodology as in the [compute_relhum routine used by the model](https://github.com/MPAS-Dev/MPAS-Model/blob/v6.0/src/core_atmosphere/physics/mpas_atmphys_driver_microphysics.F#L726-L781)?

mgduda:
@weiwangncar Will the changed proposed here handle the case where we initialize with specific humidity, but the intermediate file contains RH? I think this would be the case, e.g., if we initialize with ERA-Interim model-level data, where the `calc_ecmwf_p.exe` program would have produced an `RH` field in the intermediate file, but we might still prefer to use `config_use_spechumd = true` when running `init_atmosphere_model`?

mgduda:
The changes in this PR don't compile with gfortran:
```
mpif90 -D_MPI -DUNDERSCORE -DCORE_INIT_ATMOSPHERE -DMPAS_NAMELIST_SUFFIX=init_atmosphere -DMPAS_EXE_NAME=init_atmosphere_model -DMPAS_NATIVE_TIMERS -DMPAS_GIT_VERSION=v6.2-1-g9451455 -O3 -m64 -ffree-line-length-none -fconvert=big-endian -ffree-form -fdefault-real-8 -fdefault-double-8 -c mpas_init_atm_cases.F -DUSE_PIO2 -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -DUSE_PIO2 -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -I/sysdisk1/duda/mpas-libs-gcc-8.3.0/include -I../framework -I../operators  -I../external/esmf_time_f90
mpas_init_atm_cases.F:4660:21:

                   eis = .01 * exp (9.550426 - (5723.265 / t(k,iCell)) + (3.53068 * alog(t(k,iCell))) &
                     1
Error: Symbol ‘eis’ at (1) has no IMPLICIT type; did you mean ‘es’?
mpas_init_atm_cases.F:4662:21:

                   ews = 6.112 * exp(17.67 * (t(k,iCell)-273.15) / ((t(k,iCell)-273.15)+243.5))
                     1
Error: Symbol ‘ews’ at (1) has no IMPLICIT type; did you mean ‘es’?
mpas_init_atm_cases.F:4665:23:

                      r1 = ((273.15_RKIND - t(k,iCell)) / 20._RKIND)
                       1
Error: Symbol ‘r1’ at (1) has no IMPLICIT type; did you mean ‘d1’?
```

weiwangncar:
@mgduda Just pushed up one commit, and it allows the code to compile.

weiwangncar:
@mgduda Regarding your first question, I think it is not clear which way is a better way. Using the ungrib formula probably keeps the RH as close as to that from the input data. Using the model formula perhaps will make it more consistent. If there is agreement on what to do, we can change it.

Regarding comment #2, no, where RH is recomputed now will not cover the EC case. It can be addressed by moving the re-computation code outside the current conditional block. Do we have any case where RH is not available?

weiwangncar:
@mgduda Moved the code outside the use_spechumd block so that it will now cover cases where both specific humidity and RH are present in the input file. Tested with a EC dataset as such and it produced expected outcome, e.g. increased RH in cold regions.

mgduda:
@weiwangncar This looks reasonable to me. Before merging, I would like to move the RH adjustment code to its own function and to squash the commit history into a single commit. Would that be alright with you?

weiwangncar:
Yes it is fine by me.

On May 12, 2019, at 11:31 AM, Michael Duda <notifications@github.com> wrote:

@weiwangncar <https://github.com/weiwangncar> This looks reasonable to me.
Before merging, I would like to move the RH adjustment code to its own
function and to squash the commit history into a single commit. Would that
be alright with you?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/MPAS-Dev/MPAS-Model/pull/243#issuecomment-491613984>,
or mute the thread

weiwangncar:
@mgduda Thanks for the code modification. It looks good to me. Just one suggestion to add something that looks like the following to the comment section of the function: 'This uses the formula in Ungrib to re-compute RH with respect to ice in an attempt to re-construct RH that is more like the incoming field.'

mgduda:
@weiwangncar Thanks for checking. I've updated the commit at the top of the new function.

PR #244, opened 2019-05-11
Add ocean nightly test suite back to COMPASS

mark-petersen:
We removed the nightly test suite on the last merge, and here we add it back it with the QU240 cases created by jigsaw.


mark-petersen:
This works on grizzly with the python compass module. 
<details>
<summary> Output from the nightly regression </summary> <p>

```
cd testing_and_setup/compass/
gr-fe2:compass$ ./manage_regression_suite.py -t ocean/regression_suites/nightly.xml -f general.config.ocean_turq -s --work_dir /lustre/scratch4/turquoise/mpeterse/runs/190511d_test -b /lustre/scratch4/turquoise/mpeterse/runs/190511c_test

On a compute node:
cd /lustre/scratch4/turquoise/mpeterse/runs/190511d_test
module unload python; source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_compass.sh
module list

Currently Loaded Modules:
  1) git/2.19.1   2) gcc/5.3.0   3) openmpi/1.10.5   4) netcdf/4.4.1   5) parallel-netcdf/1.5.0   6) pio/1.7.2

./nightly_ocean_test_suite.py
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - SE Blocks Test
      PASS
 ** Running case Global Ocean 240km - RK4 Blocks Test
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
TEST RUNTIMES:
00:12 Baroclinic_Channel_10km_-_Decomp_Test
00:12 Baroclinic_Channel_10km_-_Restart_Test
00:14 Baroclinic_Channel_10km_-_Thread_Test
00:21 Global_Ocean_240km_-_Analysis_Test
01:01 Global_Ocean_240km_-_Init_Test
00:04 Global_Ocean_240km_-_Performance_Test
00:21 Global_Ocean_240km_-_RK4_Blocks_Test
00:15 Global_Ocean_240km_-_Restart_Test
00:14 Global_Ocean_240km_-_SE_Blocks_Test
00:20 Periodic_Planar_20km_-_LIGHT_particle_region_reset_test
00:28 Periodic_Planar_20km_-_LIGHT_particle_time_reset_test
00:27 ZISO_20km_-_Smoke_Test
00:15 ZISO_20km_-_Smoke_Test_with_frazil
00:31 sub-ice-shelf_2D_-_restart_test
Total runtime 04:55
```
 </p></details>
That example is from a comparison case, where `PASS` means it passed the run and the bfb comparison.

mark-petersen:
@xylar I added all test cases in the previous version except `with_land_ice` on the QU240. I suspect that will take a bit longer, and I prefer to get the rest of them in sooner, as I'm out of time and out on travel this week.

The other major difference is that we created the QU240 mesh and ICs for every test before. On this new one I run an `init` case for QU240 only once, and then link to that one for all QU240 tests. So the whole regression suite goes faster too.

testing_and_setup/compass/manage_regression_suite.py
```diff
@@ -262,7 +262,7 @@ def setup_suite(suite_tag, work_dir, model_runtime, config_file, baseline_dir,
     regression_script.write("        runtime = np.ceil(float(subprocess."
                             "check_output(\n"
                             "                ['grep', 'real', outputfile])."
-                            "split('\\n')[-2].split()[1]))\n")
+                            "decode().split('\\n')[-2].split()[1]))\n")
```
xylar: @mark-petersen, this will need to be in a separate PR to `develop`.  We're not really allowed to change the basic COMPASS scripts in `ocean/develop`.

xylar:
Other than the one script that can't be part of this PR, this looks good.  I see some hooks for a QU240 case with land ice, but I would probably put that in a separate folder from the QU240 (e.g. QU240wLI).

It's 1:30 am so I won't test this until tomorrow...

testing_and_setup/compass/manage_regression_suite.py
xylar: Also, I really prefer to be explicit about what format is being decoded, i.e. `decode('utf-8')`.

mark-petersen:
Rebased on ocean/develop with #245 merged in. Changed 16 cores to 4. On 
`init/config_initial_state.xml`
that is a link to a common init file for all resolutions, so I prefer to leave it as 36. Then all resolutions can create an IC with no changes  Even though 36 is overkill for QU240, it still works fine.

I removed the `QU240/with_land_ice` case. In this new setup it makes more sense to have every distinct mesh be made at this level:
```
testing_and_setup/compass/ocean/global_ocean/QU240
testing_and_setup/compass/ocean/global_ocean/QU240wLI
```
with an init directory in each. We probably want to change `wLI` as well, as that implies that MALI is running.

Anyway, the ocean cavity cases will come in a later PR.

mark-petersen:
@xylar I just retested. Please approve when ready, I'll merge in.

mark-petersen:
On second thought, changed init cores from 36 to 8. We use lower resolutions much more often than high resolutions, so that is a better default.

xylar:
> On second thought, changed init cores from 36 to 8. We use lower resolutions much more often than high resolutions, so that is a better default.

Thank you for this.  I really do think the 36 core requirement would be problematic.  I think it's more useful to have this lower default and have a specific config file for any test cases that really need more cores (which I don't believe any do).  Even the very big test case likely work with 8 cores.

xylar:
> I removed the QU240/with_land_ice case. In this new setup it makes more sense to have every distinct mesh be made at this level:

Yes, thanks for that.  I completely agree that they need to be handled at the "resolution" level.

> We probably want to change wLI as well, as that implies that MALI is running.

Yes, that's something to address before adding any new test cases with land-ice cavities.  I haven't come up with a TLA I like for them but I definitely agree that wLI has to go.  Something for another PR...

PR #245, opened 2019-05-12
Fix decoding in script generated for regression suites

xylar:
This is required for python 3 compatibility of the resulting script.


xylar:
@mark-petersen, this is the piece of #244 that needs to go into `develop` (and then be merged back into `ocean/develop`) before you can merge #244.

xylar:
Tested as part of testing #244

mark-petersen:
FYI @mgduda and @matthewhoffman merging this into develop, it is a compass fix only.

PR #247, opened 2019-05-14
Check for a valid vertical grid in MPAS-A LBC setup case

mgduda:
This merge adds a check in the MPAS-A LBC setup case to look a valid
vertical grid, which is needed in order to correctly produce LBC update files.
The check in init_atm_case_lbc considers the maximum value of
the zgrid field. If the global maximum of zgrid is zero, it is very likely
that the input stream did not contain valid vertical grid information.


mgduda:
@MiCurry No additional tests are necessary. Thanks for taking a look!

Issue #235, opened 2019-05-07
New COMPASS EC60to30, NAEC60to30cr8 and QU60 test cases don't spin up


xylar:
As part of #229, I tried to run all `global_ocean` (but not `global_ocean_V1`) test cases.  Most test cases ran to completion or could be fixed with small tweaks to the configuration that I will include in #229.  The cases that are broken beyond seemingly easy repair are:
The following is a list of tests that failed (but for reasons I believe are unrelated to this PR):
* `-o ocean -c global_ocean -r EC60to30 -t spin_up` during `simulation` because of CFL violation: 
```
CFLNumberGlobal = 0.0917271835580922, 0.122637831833784, 0.150533786272102, 
    0.17555059144412, 0.197662099329266, 0.217073518294376, 
    0.234323872726046, 0.250136386921142, 0.264790486294835, 
    0.27825253635325, 0.290323188086416, 0.301051958089178, 
    0.310209832190098, 0.3176288821503, 0.32330034866235, 0.328730874921286, 
    0.340932315339754, 0.39929006348568, 0.818147632488227, 1.93089241771095, 
    3.78167130632891, 9.49790843196592, 0.358590340863955
```
* `-o ocean -c global_ocean -r NAEC60to30cr8 -t spin_up`: the final (`test_final_settings`) phase seems to be misconfigured in various ways.  Links to `restart` and `Restart_timestamp`  are missing.  `timeSeriesStatsMonthly` is enabled but there isn't a corresponding stream.  It attempts to change to monthly restarts but the last restart is at `0001-01-07`, which isn't an allowable combination in MPAS.
* `-o ocean -c global_ocean -r QU60 -t spin_up` during `simulation` for unknown reasons during the first time step

I don't believe any of these issues are related to #229 but haven't had time to run the tests with `ocean/develop` to be completely sure.

xylar:
@mark-petersen, I think you are likely to be the best person to handle these but @sbrus89 was also involved in creating some of these cases so I've assigned you both.

mark-petersen:
@kristin-hoch and I spun up the jigsaw EC60to30 a month ago, we will make sure the spin-up sequence matches what we did for a successful run. Thanks.

kristin-hoch:
[...]

PR #248, opened 2019-05-14
Fix tracer depth conversion in EC60to30 init

xylar:
With this change, this test case spins up successfully.


xylar:
This partially addresses #235.  Thanks to @kristin-hoch (https://github.com/MPAS-Dev/MPAS-Model/issues/235#issuecomment-492012490) for figuring this out!

With this fix, a test on Grizzly running with the default setup is successful.  Here's the CFL as a function of time during spin-up:

image

PR #249, opened 2019-05-14
Reduce time btr and full time step in QU60

xylar:
In my testing, this was necessary to get through the simulation phase.

I tried time steps of 50 minutes/ btr of 2.5 minutes and still got immediate crashes.


xylar:
Partially addresses #235

xylar:
The need for a shorter time step may be explained by the higher vertical resolution (80 layers, 1.5 m at the surface) compared with the same setup in V1 (which I think had 60 vertical layers, but I could be mistaken).

PR #255, opened 2019-05-16
Enable soundings to handle non-existence of config_sounding_interval option

mgduda:
This merge enables the sounding diagnostics module to handle the non-existence
of the config_sounding_interval namelist option.

When MPAS-Atmosphere is built as a dycore in other modeling systems
(e.g, CAM/CESM), only a subset of the stand-alone MPAS-Atmosphere namelist
options may be available. To avoid segfaults in the soundings diagnostic
module due to the non-existence of the config_sounding_interval option in
such cases, add extra logic in the soundings_setup routine to detect that
this namelist option is not available and deactivate soundings just as we
would if config_soundings_interval = 'none'.

### DIFF
--- a/src/core_atmosphere/diagnostics/soundings.F
+++ b/src/core_atmosphere/diagnostics/soundings.F
@@ -50,8 +50,9 @@ module soundings
     !-----------------------------------------------------------------------
     subroutine soundings_setup(configs, all_pools, simulation_clock, dminfo)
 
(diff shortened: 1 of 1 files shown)


mgduda:
@MiCurry Thanks for the quick review. I just rebased the branch onto v6.0, since I also want to merge this branch onto code that is based on MPAS v6.0. The soundings code hasn't been modified since v6.0, so this should have no impact on the code changes.

PR #258, opened 2019-05-20
Move LIGHT scripts to ocean COMPASS directory

mark-petersen:
LIGHT scripts were previously in the general directory for COMPASS scripts, but are only used for ocean simulations, so the ocean subdirectory is more appropriate.


mark-petersen:
@pwolfram and @bradyrx take a quick look. Do any other scripts list the path for these files and need to be modified?

bradyrx:
I'm unaware of anything that calls these scripts directly within MPAS-Model. This should be fine just moving them to the new directory.

mark-petersen:
@pwolfram this is a LIGHT directory, go ahead and merge when you want.

mark-petersen:
@pwolfram I'm doing a python script update in E3SM. I'm guessing I can merge this in now?

pwolfram:
Yes please

pwolfram:
Thanks

PR #266, opened 2019-05-28
Fix race condition in file tests when clobber_mode = 'append' or 'overwrite'

mgduda:
This merge fixes a race condition in file tests when clobber_mode = 'overwrite'
or 'append'.

The stream manager's write_stream(...) routine tests on the existence of files
to be written when clobber_mode = 'append' or 'overwrite' for a stream.
Each MPI task was independently testing whether the output file existed using
calls to Fortran's inquire intrinsic, and the results of these inquiries could
disagree among MPI tasks, apparently due to race conditions between the creation
of output files and the testing of files on different tasks.

The solution adopted in this commit is essentially the same approach as is used
in the low-level mpas_io_open(...) routine: we test for the existence of the file
only on the IO_NODE and broadcast the result to all other MPI tasks.

There are two places in write_stream(...) where file inquiries are modified.
There is also one file inquiry in read_stream(...) that has been modified, too,
not out of concerns for correctness, but simply because it seemed more efficient
to make filesystem inquiries only from one MPI task and to broadcast the result
than to have a potentially large number of MPI tasks all making filesystem
inquiries.

### DIFF
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -3047,6 +3047,8 @@ module mpas_stream_manager
     !-----------------------------------------------------------------------
     subroutine write_stream(manager, stream, blockID, timeLevel, mgLevel, forceWritenow, writeTime, ierr) !{{{
 
+        use mpas_dmpar, only : IO_NODE
+
         implicit none
 
         type (MPAS_streamManager_type), intent(inout) :: manager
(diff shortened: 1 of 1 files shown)


mgduda:
@matthewhoffman and @mark-petersen The race conditions addressed by this PR are not always easy to reproduce. In order to (mostly) reliably trigger the error, I introduced random delays for each MPI tasks in the `write_stream` routine. If you're interested, I can share with you my modifications to introduce these random delays for your own testing.

mgduda:
@mark-petersen This issue actually caused the code to crash. It's not a completely satisfying explanation of the underlying cause, but what is apparently happening is that some MPI tasks get to the `inquire` check before others, determine that the file does not exist, then get down to calls to PIO that create the file; other MPI tasks that lag behind then see the file and take a different code path, leading to problems in other MPI collective calls. I had assumed that all PIO calls were collective and that output files should not be created until all MPI tasks had made the PIO call, but it seems that I could be mistaken in that assumption.

I expect that this would be seen in MPAS-Ocean if an output stream had `clobber_mode="overwrite"` or `clobber_mode="append"` and the MPI tasks were sufficiently unbalanced in terms of where they each were in code execution in the stream manager.

This does appear to be a rarely encountered race condition. As I mentioned, I had to insert code to add random delays to tasks in order to reliably reproduce the issue.

PR #268, opened 2019-05-31
ocean tracer advection CPU optimizations

philipwjones:
ocean tracer advection CPU optimizations, including:
  - moving options to module private variable to reduce config calls
  - added contiguous keyword to pointers
  - passed pointers to automatic arrays
  - replaced diagnostic string argument with logical flag
  - replaced scratch fields with local thread-shared allocatable arrays
  - changed loop order in provisional thickness calc to reduce flops
  - introduced temps in horz flux calculation to reduce flops
  - moved min, max calculations to separate loops (both vert, horz)
  - modified high-order vert flux to move conditionals outside loops
  - inlined high-order vert flux functions
  - reordered a few other calculations for efficiency
  - general cleanup to remove stale variables, reduce long lines, etc

Changes are not b4b, but are at round-off. I've been very careful, but due to the large number of changes, probably should be evaluated in a transport test case or longer sim.

Tested mostly with a QU240 test case both after a single step (to confirm roundoff) and in 2-day runs. Most tests on Cori. Changes resulted in 2-3x speedup for tracer adv timers, depending on configuration.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_advection.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_advection.F
@@ -5,22 +5,23 @@
 ! Additional copyright and license information can be found in the LICENSE file
 ! distributed with this code, or at http://mpas-dev.github.com/license.html
 !
-!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
 !
 !  ocn_tracer_advection
 !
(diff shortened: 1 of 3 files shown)


mark-petersen:
Rebased. We will merge this in when E3SM is open for a non-bfb PR.

mark-petersen:
On nightly regression suite, errors after ~5 steps are all 1e-13 or less in average difference, 1e-10 or less in max difference. I don't see any cause for concern here.

mark-petersen:
Merged into ocean/develop locally, ran through tests again. All looks good.

PR #271, opened 2019-06-06
Communicate effective radii for cloud, ice, and snow between MPAS and WSM6

mgduda:
This merge corrects a bug, wherein the effective radii for cloud, ice, and
snow were not being communicated between the WSM6 microphysics
scheme and MPAS. Specifically, this fix affects simulations that use
the WSM6 microphysics scheme and the RRTMG radiation schemes,
and that have config_microp_re = true in the namelist. Without this fix,
the RRTMG schemes would not use the effective radii from the microphysics,
but would instead use zero values (which are then set according to a lower
bound in the RRTMG schemes) everywhere for these fields.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_interface.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_interface.F
@@ -572,6 +572,9 @@
           qi_p(i,k,j) = qi(k,i)
           qs_p(i,k,j) = qs(k,i)
           qg_p(i,k,j) = qg(k,i)
+          recloud_p(i,k,j) = re_cloud(k,i)
+          reice_p(i,k,j)   = re_ice(k,i)
+          resnow_p(i,k,j)  = re_snow(k,i)
        enddo
        enddo
        enddo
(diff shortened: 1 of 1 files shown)


mgduda:
@weiwangncar and @ldfowler58 If you have any suggestions for how to improve the PR description, just let me know. I'm not sure that my description is as accurate as it could be. Otherwise, I think this PR is ready for review.

ldfowler58:
I approve the PR to correct the communication of re_cloud, re_ice, and re_snow between the WSM6 cloud microphysics scheme where they are calculated and the other MPAS parameterizations.

weiwangncar:
Do we need to say something like it affected radiation calculation if config_micro_re is used for WSM6?

mgduda:
@weiwangncar Is radiation the only other place where effective radii are used right now?

weiwangncar:
@mgduda Yes.

mgduda:
@weiwangncar This would impact just the RRTMG schemes, and not the CAM scheme, right?

weiwangncar:
@mgduda That is correct. CAM radiation does not use these effective radii.

mgduda:
@weiwangncar Thanks! I've updated the PR description to try to capture these details.

PR #276, opened 2019-06-07
Add log message for lbc updates

MiCurry:
This PR is of a commit that produces a log message each time that the lbc's for a regional MPAS
simulation are updated.


mgduda:
@MiCurry Can you change the number of `-` symbols around the LBC update message so that these messages match those printed for SST updates? Otherwise, we get something like the following in the log files:
```
 ------------------------------------------------------
 Updated lateral boundary conditions. LBCs are now valid
 from 2019-01-28_00:00:00 to 2019-01-28_03:00:00
 ------------------------------------------------------
 ----------------------------------------------------------------------
   Read 'surface' input stream valid at 2019-01-28_00:00:00
    Timing for stream input: 0.576355E-01 s
 ----------------------------------------------------------------------
```
While you're updating the code, I'd also propose using single quotes around the strings that are passed to `mpas_log_write` to match the style in the rest of the module.

MiCurry:
Sounds good to me! The lengths have been updated!

PR #277, opened 2019-06-07
Add a CMakeLists.txt for mpas/src

jgfouca:
This will allow E3SM, which is being converted to a CMake-based build
system, to build MPAS.

This is only for mpas-ocean for now. Subsequent PRs will be needed for other components.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,0 +1,391 @@
+cmake_minimum_required(VERSION 3.9)
+cmake_policy(SET CMP0057 NEW)
+
+project(MPAS C CXX Fortran)
+
+if (SMP_PRESENT)
+  set(THREADDIR "threads")
+  set(compile_threaded TRUE)
+else()
+  set(THREADDIR "nothreads")
+  set(compile_threaded FALSE)
+endif()
+
+include(${CASEROOT}/Macros.cmake)
+
+# JGF TODO Check for build_options.mk
+set(EXE_NAME ${CORE}_model)
+set(NAMELIST_SUFFIX ${CORE})
+
+# Map the ESM component corresponding to each MPAS core
+if (CORE STREQUAL "ocean")
+  set(COMPONENT ocn)
+elseif(CORE STREQUAL "landice")
+  set(COMPONENT "glc")
+elseif(CORE STREQUAL "seaice")
+  set(COMPONENT "ice")
+endif()
+
+if (USE_ESMF_LIB)
+  set(ESMFDIR "esmf")
+else()
+  set(ESMFDIR "noesmf")
+endif()
+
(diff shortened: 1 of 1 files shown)


jgfouca:
@jonbob , assign reviewers etc. as appropriate.

jonbob:
@jgfouca - can you not delete Makefile_in.E3SM? I know our developers will want to be able to build inside E3SM with something they're used to

jgfouca:
@jonbob , yes. And, now that I think about it, it will still be needed for the other components until they are converted too.

jonbob:
I think you'll find that some people will still want the option to use the old system. For most of us, it's easier to understand and modify

jgfouca:
@jonbob , we may be able to keep the old system if developers want to do standalone E3SM-like MPAS builds. I don't think we'll be able to continue support for driving the MPAS builds within E3SM via Makefiles.

jonbob:
@jgfouca - that may be a serious issue then. We have several people who need to modify the build system to develop new features, and I don't think they're fluent in cmake

jonbob:
Is there a reason we can't support the current build system as well as cmake?

jgfouca:
@jonbob , it should be as easy as adding files to RAW_SOURCES.

> Is there a reason we can't support the current build system as well as cmake?

Right now, at a very high level, CIME builds components as follows:
```
build.py
   for each component:
       run shell command component/cime_config/buildlib
```
Right now, the logic for what build system to invoke is in buildlib and would be easy to switch back to old Makefile system.

However, the goal of the CMake refactor is:
```
build.py
   cmake components
   make
```
In which case, it becomes much more difficult to have MPAS doing it's own thing.

jonbob:
The problem is when we need to do something more complicated than just add new files. And we often do

jgfouca:
@jonbob , I'd be happy to help with that.

I'm assuming that the really tricky stuff (cvmix, BGC, and the registry processing stuff) is pretty stable and that the most likely changes are adding/removing files and change CPP flags.

mgduda:
@jgfouca I'm assuming there are no plans to modify the stand-alone builds of any of the cores, right? Also, would you mind squashing the two commits in this PR together? There doesn't seem to be a good reason to remove a file and to then commit it again in the next commit. Ultimately it doesn't matter, but I think it's nice to keep the commit history tidy.

jgfouca:
@mgduda , that's correct, no plans to modify standalone. I do think it would be better, longterm, for both E3SM and MPAS if MPAS had a single CMake build system.

I will squash my commits.

jgfouca:
I'm seeing a DIFF in my testing, I may need to make additional commits.

jgfouca:
OK, the issue was just an out-of-date baseline. This PR is ready from my point of view.

jgfouca:
Replaced by #281

PR #279, opened 2019-06-12
fix ZISO python quotes

mark-petersen:
Nightly test suite on grizzly currently fails on `ziso/20km/default` case. Needs this correction, otherwise text parsing is incorrect within python.


mark-petersen:
@pwolfram this might be a python 2 to 3 fix-up. We now are using python 3 on LANL IC with:
```
module unload python; source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_compass.sh
```
You can approve visually, or test with your laptop setup.

PR #281, opened 2019-06-14
All E3SM builds of MPAS now go through CMake

jgfouca:


jgfouca:
Posting this so you guys can see what a more-finalized version of this work will look like.

jgfouca:
Replaced by #386

PR #275, opened 2019-06-07
Update ISOMIP+ test case visualization and instructions

xylar:
This merge updates the ISOMIP+ viz to make it a python 3 package.  It also adds detailed instructions for how to setup and run the test case on LANL IC.

I added ffmpeg as a new package to be included in the COMPASS conda environment.  ISOMIP+ viz uses this to make movies.

PR #282, opened 2019-06-16
Add time-varying land-ice forcing to MPAS-Ocean

xylar:
This merge adds time-varying land-ice forcing via the following fields:
* `landIceDraft`
* `landIcePressure`
* `landIceFraction`

A test case `isomip_plus/2km/time_varying_Ocean0` is added as a (relatively trivial) test of this forcing.  In this test case, `landIcePressure` and `landIceDraft` are initially 10% of their "correct" values and increase linearly over 1 year to their full values, then remain constant for an additional year.  However, the test case crashes sometime during the second year because layers become too thin and/or tilted, or velocities become too large.  This is an intentional feature of the test case, showing why further work is needed to handle coupled ice dynamics.

Modifications are made to the ISOMIP+ viz (#275) to correctly handle repeated running of the viz while the simulation is in progress.  A few other tweaks (e.g. the format of dates in the figure titles) are also included.

Modifications were made to how sea-level is maintained in ISOMIP+ test cases despite freshwater fluxes and changes in land-ice pressure.  A fictitious evaporation was already used at the northern boundary, but this has been modified to adjust to sea-level itself rather than melt fluxes to accommodate changing land-ice pressure.

Finally, a division-by-zero error in timeSeriesStatsGlobal was fixed.  If melt fluxes are turned on but there are no locations where the melt mask nonzero (which happened in my testing), global stats was trying to compute mean fluxes and other stats with zero area in the denominator.  This has been fixed.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -758,6 +758,30 @@
 			description="Number of days to ramp up time varying forcing"
 			possible_values="Any positive real number"
 		/>
+		<nml_option name="config_use_time_varying_land_ice_forcing" type="logical" default_value=".false." units="unitless"
+			description="If true calculate input forcing fields."
+			possible_values=".true. or .false."
+		/>
+		<nml_option name="config_time_varying_land_ice_forcing_start_time" type="character" default_value="0001-01-01_00:00:00" units="unitless"
+			description="Forcing time to use at the simulation start time"
+			possible_values="'YYYY-MM-DD_HH:MM:SS"
+		/>
+		<nml_option name="config_time_varying_land_ice_forcing_reference_time" type="character" default_value="0001-01-01_00:00:00" units="unitless"
+			description="Reference time for the forcing"
+			possible_values="'YYYY-MM-DD_HH:MM:SS"
+		/>
+		<nml_option name="config_time_varying_land_ice_forcing_cycle_start" type="character" default_value="0001-01-01_00:00:00" units="unitless"
+			description="Start time for the forcing cycle."
+			possible_values="'YYYY-MM-DD_HH:MM:SS"
+		/>
+		<nml_option name="config_time_varying_land_ice_forcing_cycle_duration" type="character" default_value="2-00-00_00:00:00" units="unitless"
+			description="Duration of the forcing cycle."
+			possible_values="'YYYY-MM-DD_HH:MM:SS"
+		/>
+		<nml_option name="config_time_varying_land_ice_forcing_interval" type="character" default_value="01:00:00" units="unitless"
+			description="Time between forcing inputs"
+			possible_values="'YYYY-MM-DD_HH:MM:SS"
+		/>
 	</nml_record>
 	<nml_record name="coupling" mode="init;forward">
 		<nml_option name="config_ssh_grad_relax_timescale" type="real" default_value="0.0" units="seconds"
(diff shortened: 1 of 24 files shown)


xylar:
@mark-petersen, this is not at all pressing.  It is for SciDAC work and is a starting point for other testing.  But if another good chance comes up soon to merge changes that don't affect E3SM, this would be one.

Please let me know if the change to `timeSeriesStatsGlobal` concerns you and I can move it to another PR.

xylar:
@sbrus89, I tacked `land_ice_forcing` onto your work in `mpas_ocn_time_varying_forcing.F` but I do still feel like it would make sense to have separate modules for atmosphere and land-ice forcing.  My feeling is that this is not the place to do that and we should do a separate clean-up PR after this one to take care of that so there isn't confusion.  Do you agree?

xylar:
Here are some examples of the output from this test case:
https://www.dropbox.com/sh/yv9bfexmww8qkyw/AAA3-PUyhHDVIavIVVEyhFg8a?dl=0

sectionTemp_0001

sectionTemp_0007

sectionTemp_0013

xylar:
My initial test was with a linear EOS and the corresponding pressure-gradient.  But even after switching to the GM equation of state, the Haney Number is able to reach unexpectedly large values in this experiment before causing the code to crash.  Typically, we restrict the Haney number to stay below 5 but it exceeds 8 near the grounding line in this configuration for several months before crashing:

sectionHaney_0016
topHaney_0016
botHaney_0016

sbrus89:
> @sbrus89, I tacked `land_ice_forcing` onto your work in `mpas_ocn_time_varying_forcing.F` but I do still feel like it would make sense to have separate modules for atmosphere and land-ice forcing. My feeling is that this is not the place to do that and we should do a separate clean-up PR after this one to take care of that so there isn't confusion. Do you agree?

@xylar, I think this is a good idea, but that as a part of the clean-up we should do something to try to unify all the features that use the common time-varying forcing framework. It seems like there are getting to be a lot of calls in `ocn_forward_mode_run` that update the fields for various time-varying data fields. Maybe each forcing should have it's own separate module for initializing, getting and processing it's own data, but all of the calls to the subroutines (surrounded by the various `if config_use_feature_XXXX` statements) get consolidated into a common time varying forcing module. I think this kind of combines you and I have done with what has been done previously.

Hopefully that makes sense. I could be way off here. Let me know what you think.

xylar:
@sbrus89 wrote:
> Hopefully that makes sense. I could be way off here. Let me know what you think.

Yes, I think that makes sense.  Some routine would be called for each type of forcing currently involved in this unappetizing construct:
```
            ! write forcing restart if any type(s) time varying forcing is used
            if (trim(config_sw_absorption_type)=='ohlmann00' .or. &
                config_use_ecosysTracers .or. &
                (config_use_activeTracers_surface_restoring .and. config_use_surface_salinity_monthly_restoring) .or. &
                config_use_time_varying_atmospheric_forcing) then
                config_use_time_varying_atmospheric_forcing .or. &
                config_use_time_varying_land_ice_forcing) then

                call ocn_time_varying_forcing_write_restart_times(domain)
            end if
```
where we would essentially find out if any of them are active.  The logic for determining what it means for them to be "active" would be housed internally to the module.

I think that's what we had discussed in your PR as well but we postponed that clean up.  I'm afraid I'm proposing the postpone yet again....

sbrus89:
@xylar, exactly. It's totally fine to postpone it since it's mostly a readability issue. Maybe next time you're back in town we could work together on creating the clean-up PR.

xylar:
@mark-petersen, does it make sense to include this with other bit-for-bit merges you're making to E3SM?

mark-petersen:
It looked like there were enough forward model changes to put it by itself. I'll put it on the list though.

xylar:
That's fine.  It shouldn't affect anything outside of test cases that explicitly set:
```
config_use_time_varying_land_ice_forcing = .true.
```
but it certainly doesn't hurt from my perspective to keep it on its own.

xylar:
@mark-petersen, thanks a lot!  This will be fun to have as a test case to play with on a regular basis.

Issue #217, opened 2019-04-23
Split explicit time integration challenges


pwolfram:
Following #206, tested split-explicit time stepping was not able to provide a boundary condition with a reasonable time scale for relaxation to the coastal boundary, unlike RK4, which suggests the time integrator will not be suitable for application of tidally-forced boundary conditions via use of restoring over a short time scale. 

cc @mark-petersen

xylar:
@pwolfram, could you rephrase this is more of an issue to be solved?  First, the title is so broad that I wouldn't know what it is useful for just scrolling through the MPAS-Model issues.  Second, I know that restoring at boundaries does work in certain conditions with split-explicit because I use boundary restoring in ISOMIP+ test cases.  I think issues should be phrased as something that could be addressed with a pull request (or at most a few), rather than either a place to list challenges one is facing or an open-ended and vague feature request.

pwolfram:
@xylar, the issue is related to the performance of the split explicit time integrator, as I've discussed with @mark-petersen.  

In #206, the tau restoring timescale needs to be roughly an order of magnitude larger for split explicit than for RK4 (just for stability), which means that the tidal restoring boundary condition effectively does not work with split explicit.  @mark-petersen has mentioned that time-step based convergence of the integrator is also not as expected, although I have not explored this personally.  

Thus, following improvements to split explicit, #206 can be used as one of the test beds or metrics to make sure its performance is improved by switching the time integrator from 'RK4' to 'split_explicit' (which is unstable if you were to try this out now).

xylar:
Great, thanks you for clarifying the title and the scope of the issue.

pwolfram:
Thanks for asking a key question to clarify @xylar!

PR #284, opened 2019-06-17
Adds option to directly force tidal boundary

pwolfram:
This adds an additional option that allows direct forcing of the tidal boundary, instead
of use of a thickness tendency to apply the tidal boundary.  

The test case has been distinguished into these two separate options:

 * `thickness_source` that uses restoring to apply the tidal boundary condition via a thickness tendency
* `direct` that sets the layer thicknesses to obtain a desired sea surface height


pwolfram:
# Testing
## `thickness_source`
tidalcomparison
## `direct`
tidalcomparison

Note that the `direct` approach has a clamped boundary that perfectly respects the applied tidal forcing, whereas the `thickness_source` approach has the lag as noted in #206

pwolfram:
@mark-petersen, note I only implemented this for RK4 because at present because of #217, although we could probably extend this to split explicit without too much trouble.  Given that a primary need for this is to support testing of wetting and drying, which is only implemented for RK4 at present, I'm not sure we need to support split explicit to start.

pwolfram:
Hm, this is interesting @mark-petersen.  When I rebased onto ocean/develop the thickness_source approach improved in quality...

tidalcomparison

Did you update the thickness tendency terms anytime recently?  The phase lag is decreased, which is great!

pwolfram:
@mark-petersen, this also is needed by the drying slope cases too and it works when incorporated.  I'm ready for it to be merged pending your checks (using the drying slope prototype as a half-verification).  Note the tidal amplitude is forced on the left side and results are as expected too using the `direct` approach, much better than the `thickness_source`:

dryingslopecomparison
tidalcomparison

pwolfram:
Thanks @mark-petersen!

PR #285, opened 2019-06-19
Fixes Rayleigh bottom drag that wasn't applied unless Rayleigh drag applied

pwolfram:
This one line fix ensures that Rayleigh bottom drag is applied even if entire-water column Rayleigh drag isn't applied, e.g.,
```
-      if ( .not. rayleighFrictionOn ) return
+      if ( .not. (rayleighFrictionOn .or. rayleighBottomFrictionOn)) return
```

cc @sbrus89- thanks for the fruitful discussion yesterday!

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vel_forcing_rayleigh.F
+++ b/src/core_ocean/shared/mpas_ocn_vel_forcing_rayleigh.F
@@ -122,7 +122,7 @@ contains
 
       err = 0
 
-      if ( .not. rayleighFrictionOn ) return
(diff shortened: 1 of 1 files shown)


pwolfram:
@mark-petersen, given we don't need this for E3SM, I'm thinking a quick merge into `ocean/coastal` is ok instead of `ocean/develop` for now.  But, it should be tagged as ocean and coastal for record-keeping.

pwolfram:
Thanks @mark-petersen!

PR #288, opened 2019-06-20
Adds options for 3d varying GM bolus and 2d varying phase speed

vanroekel:
In the GM bolus calculation, there is a specified diffusivity and phase
speed (see Ferrari et al. 2010).  Currently MPAS-O assumes these values
are fixed in time and space.  This PR allows separate flags to enable
2D varying (+time) phase speed (based on the first baroclinic mode phase
speed) and a separate flag for 3D (+time) varying diffusivity.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -375,6 +375,22 @@
 					description="If true, the standard GM for the tracer advection and mixing is turned on. This includes the redi portion which is captured in hmix."
 					possible_values=".true. or .false."
 		/>
+		<nml_option name="config_gm_lat_variable_c2" type="logical" default_value=".false." units="NA"
+					description="If true, spatially variable phase speed is used Following Ferrari et al (2010)"
+					possible_values=".true. or .false."
+		/>
+		<nml_option name="config_gm_kappa_lat_depth_variable" type="logical" default_value=".false." units="NA"
+					description="If true, spatitally and deph varying Bolus Kappa is used, following Danabasoglu et al (2007)"
+					possible_values=".true. or .false."
+		/>
+		<nml_option name="config_gm_min_stratification_ratio" type="real" default_value="0.1" units="NA"
+					description="minimum value for N2/Nmax2 ratio"
+					possible_values="small positive reals"
+		/>
+		<nml_option name="config_gm_min_phase_speed" type="real" default_value="0.1" units="m s^{-1}"
+					description="minimum allowed phase speed for spatially variable computation"
+					possible_values="small positive reals"
+		/>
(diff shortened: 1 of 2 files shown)


vanroekel:
This has currently been tested in 5 year normal forced G-cases.  Results are here

https://portal.nersc.gov/project/acme/lvroekel/GM_20190605_GMtest_Kappa3d_vary_c2.anvil_yr5/

and a baseline is here

https://portal.nersc.gov/project/acme/lvroekel/20190420.GMPAS-IAF.T62_oEC60to30v3_gradientRI_0.6.edison_yr5/

A full core cycle test is in progress, currently at year 25

vanroekel:
@mark-petersen here is the 3d varying bolus kappa pr.

mark-petersen:
Design document is here:
https://acme-climate.atlassian.net/wiki/spaces/OCNICE/pages/980713736/ocean+GM+Parameter+function+of+depth

From the July 10 meeting notes:
The core cycle of the new GM bolus and control are done.  You are welcome to discuss them at the cryo meeting tomorrow, I'll be out of town.  The links are:
__CONTROL__
https://portal.nersc.gov/project/acme/lvroekel/GM600_20190624_GMtest_constant_intel.compy_yr_50_60
__Test of variable GM Bolus__
https://portal.nersc.gov/project/acme/lvroekel/GM600_20190617_GMtest_Kappa3d_vary_c2_intel.compy_yr_50_60
A few notes from @vanroekel 
1) Surface salinity biases are pretty nicely improved, continues to subsurface as well (pretty clear in sea floor maps near antarctica)
2) MLD is better, perhaps a bit too deep.
3) Sea ice is better as well.

From @milenaveneziani:
4) Also note that the vertical structure of temperature, salinity and PD in the SO is much improved (look at the transects), especially for transects at 187, 253, 280 and 318 longitude.
5) SO bottom temperature is much better, while bottom salinity is only marginally improved.

vanroekel:
removing the don't merge label, this is tested in a full core cycle and is flag protected -- default off.

xylar:
> Design document is here: (Confluence link)

Having the design doc on Confluence isn't a good idea.  It defeats the purpose of open development.  Could this be copied as pdf or something to the local directory with design docs?

pwolfram:
Ideally we would have design documents with markdown or rst text so that we can readily render them via sphinx to a webpage documentation for the code some day, but I know this is extra work so it may not be worth it in the short term.

xylar:
@pwolfram, I agree.  I think new design docs should be written that way (like @sbrus89 did for the one design doc we currently have) but, like you, I didn't think that was a reasonable thing to ask for in this case.

PR #289, opened 2019-06-20
Migrates Rayleigh drag from explicit to implicit

pwolfram:
Previously, Rayleigh drag was implemented explicitly, which
meant it was not coupled with vertical mixing.  Now,
Rayleigh drag is implemented implicitly so that it is directly
coupled to implicit vertical mixing.


src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -271,7 +272,7 @@ subroutine ocn_vel_vmix_tend_implicit(meshPool, dt, kineticEnergyCell, vertViscT
       nEdges = nEdgesArray( 1 )
 
       allocate(A(nVertLevels),B(nVertLevels),C(nVertLevels),velTemp(nVertLevels))
-      A(1)=0
+      A(1)=0.0_RKIND
```
mark-petersen: Thanks for correcting this.

src/core_ocean/shared/mpas_ocn_vel_forcing_rayleigh.F
```diff
@@ -34,15 +34,16 @@ module ocn_vel_forcing_rayleigh
...
 
    !--------------------------------------------------------------------
    !
    ! Public member functions
    !
    !--------------------------------------------------------------------
 
-   public :: ocn_vel_forcing_rayleigh_tend, &
-             ocn_vel_forcing_rayleigh_init
+   public :: ocn_vel_forcing_rayleigh_init
```
mark-petersen: I think it makes more sense to remove this module completely, rather than have a module file just for an init routine to obtain two variables. Could you just delete it, and just grab `rayleighDampingCoef,  rayleighBottomDampingCoef` in the vmix init routine? I can do it if you don't have time.

pwolfram:
Good call @mark-petersen -- if this is done by today or Monday does that work?

mark-petersen:
No rush. Anytime next week.

src/core_ocean/shared/mpas_ocn_vmix.F
pwolfram: :+1:

src/core_ocean/shared/mpas_ocn_vel_forcing_rayleigh.F
pwolfram: Agreed, this makes the PR much cleaner and simplifies the code base.

pwolfram:
Pushed the fixes @mark-petersen, will update the drying slope as a way to test.

pwolfram:
Results are as expected and as shown to you yesterday (recognizing this is a draft figure that needs to be cleaned up for a new PR for the drying slope as discussed):

dryingslopecomparison

pwolfram:
@mark-petersen, I think this is ready to merge if it passes your tests.

pwolfram:
Thanks @mark-petersen!

PR #293, opened 2019-06-21
Change name coriolis to hadv_coriolis

### LINKED ISSUE #291: `config_disable_vel_coriolis` option also turns off momentum advection
The option `config_disable_vel_coriolis=.false.` disables more than Coriolis forcing, but also the momentum advection term via absolute vorticity and kinetic energy, e.g., Ringler et al (2010) for `\eta` and `K`, e.g., terms in equation 4.  Ideally, we would have separate flags to distinguish between turning off Coriolis itself and momentum advection.

The place in the code where they are mixed via the `config_disable_vel_coriolis` flag, which turns into a `coriolisOn` flag (e.g., at https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_vel_coriolis.F#L130) is at https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_vel_coriolis.F#L174

mark-petersen:
The coriolis module was improperly named, as it also includes horizontal
advection of momentum.

fixes #291

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -1198,7 +1198,7 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_disable_vel_coriolis" type="logical" default_value=".false." units="unitless"
-					description="Diables tendencies on the velocity field from the Coriolis force."
+					description="Diables tendencies on the velocity field from the Coriolis force and momentum advection."
 					possible_values=".true. or .false."
 		/>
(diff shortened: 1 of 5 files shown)


mark-petersen:
@pwolfram I decided to go all in and change the module name. Compiled with both gnu and intel, debug and optimized on both, on grizzly. Passes nightly regression bfb against ocean/develop.

A visual inspection by you is sufficient, since it passed all those tests.

PR #297, opened 2019-06-24
Fixes byte to string conversion python3 error

pwolfram:
Python3 treats `supprocess.check_output` as a byte,
which is no longer strictly equivalent to a string
as in python2.  Consequently, conversion is needed
for equivalence between the strings for the python3
migration.

This fixes errors like

```
└─▪ ./clean_testcase.py -n 47 
  File "./clean_testcase.py", line 106, in <module>
    git_version = git_version.strip('\n')
TypeError: a bytes-like object is required, not 'str'
```

### DIFF
--- a/testing_and_setup/compass/clean_testcase.py
+++ b/testing_and_setup/compass/clean_testcase.py
@@ -84,7 +84,7 @@ if __name__ == "__main__":
 
         regex = re.compile('(\d):')
         core_configuration = subprocess.check_output(['./list_testcases.py'])
-        for line in core_configuration.split('\n'):
+        for line in core_configuration.decode('utf-8').split('\n'):
             if regex.search(line) is not None:
                 conf_arr = line.replace(":", " ").split()
                 case_num = int(conf_arr[0])
@@ -103,7 +103,7 @@ if __name__ == "__main__":
     os.chdir(os.path.dirname(os.path.realpath(__file__)))
     git_version = subprocess.check_output(['git', 'describe', '--tags',
                                            '--dirty'])
-    git_version = git_version.strip('\n')
+    git_version = git_version.decode('utf-8').strip('\n')
     os.chdir(old_dir)
     calling_command = ""
     write_history = False
@@ -119,7 +119,7 @@ if __name__ == "__main__":
             core_configuration = subprocess.check_output(
                     ['./list_testcases.py', '-n',
                      '{:d}'.format(int(case_num))])
-            config_options = core_configuration.strip('\n').split(' ')
+            config_options = core_configuration.decode('utf-8').strip('\n').split(' ')
             args.core = config_options[1]
             args.configuration = config_options[3]
             args.resolution = config_options[5]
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, thank you for making this pull request.  COMPASS scripts belong to MPAS framework, so could you please rebase onto `develop` and modify the destination of the pull request to be `develop` instead of `ocean/develop`?

In the meantime, I'll take a more careful look at the changes you're proposing.

pwolfram:
@xylar, sounds good.  I think this was just something that fell through the cracks given other lines like https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/manage_regression_suite.py#L516... no biggie -- at least we are close to python3 compatibility!  Will rebase on `develop` in a little bit.

xylar:
Yep, the changes look fine to me.

To be honest, it's possible I forgot to test `clean_testcase.py`.  I never use it.  I always just delete the directory and start over or overwrite the existing test case without cleaning.

pwolfram:
@xylar, it is a key part of my workflow.  But, as far as I can tell, it is fixed now.  Please let me know what else we need to do here to merge.

xylar:
I'm testing now.  It looks good to me but I'd like to make sure it works before I approve.

xylar:
Also, we need to coordinate a merge and then a merge of `develop` to `ocean/develop` with @mark-petersen so my guess is this fix won't be in for a little while.

mark-petersen:
@pwolfram and @xylar thanks. I'll give it a quick test, and merge into develop and ocean/develop this week.

mark-petersen:
@matthewhoffman this is just FYI, a PR into the COMPASS shared framework.

mark-petersen:
@mgduda FYI, this is COMPASS only, will merge into develop branch.

mark-petersen:
@pwolfram you can also merge this one commit into `ocean/coastal` if you like. I'll probably update framework and merge coastal for E3SM at the same time, so it's fine.

pwolfram:
Thanks @mark-petersen, I'll just cherry-pick that onto coastal and then force push-- would that work?

xylar:
> I'll just cherry-pick that onto coastal and then force push-- would that work?

I really don't think that's a good idea.  You should merge develop into `ocean/coastal`.  If you cherry-pick, you end up with a different commit hash and then when `ocean/coastal` gets merged back, we can't tell which is the "true" source of that commit.

I think you just need to merge `develop` into `ocean/coastal`.

mark-petersen:
My order of merging will be: (1) merge develop into ocean/develop. (2) merge coastal into ocean/develop. @xylar's recommendation is better because it keeps the commit history. The other stuff on develop won't affect coastal.

pwolfram:
Thanks @mark-petersen and @xylar!  I've merged `ocean/develop` into `ocean/coastal` and `develop` into `ocean/coastal`, so this should be better prepared for @mark-petersen's merges.  Thanks for all the help!

PR #298, opened 2019-06-25
Add optional equation of state: TEOS10

mark-petersen:
Currently MPAS-Ocean only has the old Jackett-McDougall EOS and linear EOS. This adds the modern TEOS10 as a non-default option.


xylar:
I tried looking through the code to see what's being done here.  It seems like TEOS-10 is being added, but still with a reference pressure?  If so, that's fine for now.  But I think we cannot (and should never) claim to be using TEOS-10 until we switch to non-Boussinesq equations, pressure coordinates, and using the proper pressure in the EOS.  @dengwirda, please let me know if that's consistent with our discussions both at the All-hands Meeting and in the last couple of weks.

dengwirda:
@xylar I think it's possible to deploy `TEOS10` in a Boussinesq model, but `TEOS10` does seem more consistent to me with a full non-Boussinesq model, where layer pressure coordinates, etc are available directly.

@mark-petersen Xylar and I discussed related ideas recently that I'll quickly summarise:

- Perhaps taking `p_effective = rho0 * g * z` and `rho = EOS(T,S,p_effective)` might be a 1st-order way of incorporating better time-varying depth dependence to the EOS in the current Boussinesq framework. Obviously this is still an approx., but at least it responds to time- and state-changes.

- The Boussinesq vs non-Boussinesq question dictates what the model itself conserves (volume vs mass). Having proper mass conservation is obviously desirable long-term, espec. for questions like sea-level rise, etc. In a non-Boussinesq model the vertical coordinate is really `p` -- layers have a "pressure-thickness" rather than a geometric one.

In terms of `TEOS10`, also keep in mind changes to the `T, S` variables -- `T` = conservative temperature and `S` = absolute salinity in `TEOS10`. You cannot call the `TEOS10` `rho()` functions with potential temperature and practical salinity. 

While there are `T, S` conversion routines available, the recommended approach with `TEOS10` (as I understand it) is to formulate the full model in terms of conservative temperature, etc directly. The [`TEOS10` guide](http://www.teos-10.org/pubs/Getting_Started.pdf) contains more detail in Sections 8, 9.

mark-petersen:
@xylar and @dengwirda, thanks for the comments. I'm afraid this branch is not yet in a reviewable state. I made the PR early so that @alicebarthel and I have a place to easily work through some issues. The first is to get the modules hooked up. After that we will look at issues like non-Boussinesq, conservative temperature, and absolute salinity more carefully. TEOS10 will remain default off throughout, but I at least wanted to get started.

philipwjones:
Note that John Dukowicz came up with a non-boussinesq correction to a fixed profile that we used to use in POP. It seemed to capture most (but obviously not all) the effects on the EOS. That could be an intermediate fix.

I'm more concerned about the point-wise functional form of the standard TEOS-10 package. That will not be performant on our GPU architectures and is going to need a rewrite or perhaps an automated script to extend to multi-dim arrays.

xylar:
@mark-petersen, thanks for the info.  I just wanted to make sure this wasn't going to be a rush job.  I think it's big enough that it needs a design doc and discussion.

mark-petersen:
Agreed. Adding a new EOS involves more parts of the model than I expected when I started this, so certainly deserves a design document.

mark-petersen:
Besides, I think a 6 to 12 month expiration date on PRs makes us better developers, so we don't let code just flounder.

PR #300, opened 2019-06-26
Prevent divide-by-zero in Langmuir number

mark-petersen:
Without this fix, code will die with zero division if there is a zero in the 10m wind speed.


mark-petersen:
@qingli411 and @vanroekel, a visual inspection is enough. I tested it, and this fixes the problem.

PR #301, opened 2019-06-26
Update to jigsawpy

sbrus89:
Replaces the jigsaw-geo-matlab library in the mesh generation workflow with jigsawpy which is now a part of a new conda package. This conda environment can be activated on IC with the commands:

```
conda activate compass_jigsawpy_py3.7
```

This PR is mostly @xylar's work. I just made a few tweaks to get it working.


sbrus89:
There's one thing I'd like other people's opinion on:

Do we still want to save the `cellWidth.mat` file? This used to be necessary to pass the results of the `define_base_mesh.cellWidthVsLatLon()` function to `jigsaw_driver.m`. Now that we've gotten rid of MATLAB, the actual arrays are being passed as calling arguments in `jigsaw_driver.py`. However, in its current form, `inject_meshDensity.py` still uses `cellWidth.mat` to read in the `cellWidth`, `lon`, and `lat` arrays.

It could be cleaner to pass the `cellWidth`, `lon`, and `lat` arrays in the `inject_meshDensity` function call, but to me that kind of makes `inject_meshDensity` less flexible because `define_base_mesh.cellWidthVsLatLon()` always has to be run beforehand. Maybe that's okay though?

sbrus89:
This isn't BFB with the meshes I had made previously. I guess I shouldn't expect them to be given that I'm now using the @xylar's conda package instead of my own conda installation and complied versions of jigsaw/mpas_tools executables. 

Having a standard conda package from here on out should fix this issue.

xylar:
> Do we still want to save the cellWidth.mat file? This used to be necessary to pass the results of the define_base_mesh.cellWidthVsLatLon() function to jigsaw_driver.m. Now that we've gotten rid of MATLAB, the actual arrays are being passed as calling arguments in jigsaw_driver.py. However, in its current form, inject_meshDensity.py still uses cellWidth.mat to read in the cellWidth, lon, and lat arrays.

It's okay to write these out to a file.  But I really think a `.mat` file is inappropriate.  I think we don't want anything to do with Matlab.  They should be written to NetCDF either using xarray or NetCDF4, or if absolutely necessary to a standard format for numpy arrays (http://lagrange.univ-lyon1.fr/docs/numpy/1.11.0/reference/generated/numpy.save.html).

> This isn't BFB with the meshes I had made previously.  I guess I shouldn't expect them to be given that I'm now using the @xylar's conda package instead of my own conda installation and complied versions of jigsaw/mpas_tools executables.

I wouldn't count on this if the version of `jigsaw` changes or such.  I don't think we have any reliable way of reconstructing an identical conda environment to what was used to build a given mesh (reconstructing a conda environment exactly is currently not an option with conda/conda-forge at the moment).  So I think we have to live with bit-for-bit recreation of meshes not being possible once the conda environment that created them is no longer available.

sbrus89:
@xylar, writing it to netCDF is a good idea. I went ahead and did that in the commits above.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
```diff
@@ -37,15 +36,15 @@ def build_mesh(preserve_floodplain=False, floodplain_elevation=20.0,
...
+    da.to_netcdf('cellWidthVsLatLon.nc')
+    
 
-    print('Step 2. Build mesh using JIGSAW')
-    args = ["octave", "--silent", "--eval",
-            "jigsaw_driver"]
-    print("running", ' '.join(args))
-    subprocess.check_call(args, env=os.environ.copy())
+    print('Step 2. Generate mesh with JIGSAW')
+    jigsaw_driver(cellWidth, lon, lat)
```
pwolfram: Where is this function specified?

pwolfram: The reason I ask is because I don't see an import statement at the top of this function.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
```diff
@@ -21,7 +20,7 @@
 from mpas_tools.conversion import convert
 from mpas_tools.io import write_netcdf
 
-
+from jigsaw_to_MPAS.jigsaw_driver import jigsaw_driver
```
sbrus89: @pwolfram, the import is here

pwolfram: Thanks @sbrus89! I blanked it out because of the white space line above.  I'd recommend remove line 23/22 for clarity because typically all imports are lumped together.

pwolfram: Some folks leave the space-- this is ok.

xylar: I like to separate external packages from local ones in imports. @sbrus89 was using my code, so that was my call.

sbrus89: I like that approach @xylar, so I'll leave it

testing_and_setup/compass/ocean/global_ocean/config_files/config_base_mesh.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0"?>
 <config case="base_mesh">
+	<get_file dest_path="work_case_dir" file_name="topo.msh">
```
sbrus89: @xylar, I don't think `topo.msh` is actually being used in the workflow right now. It might be a good idea to keep it around for visualization purposes. Was that the intention?

xylar: i couldn't tell why this was part of so many test cases...

sbrus89:
@mark-petersen, let me know if you want me to squash this down before you merge.

mark-petersen:
I removed those wget lines for topo.msh. It is not needed, and I don't want our COMPASS process to download a file every time we run an init case.

mark-petersen:
Merged locally, tested nightly regression suite once more, and EC60to30. All looks fine. Thanks @dengwirda, @xylar, and @sbrus89!

PR #304, opened 2019-06-27
BGC tag to short version

xylar:
Intended to fix https://github.com/E3SM-Project/E3SM/issues/3022


mark-petersen:
Tested download and compile on grizzly, works fine.

PR #305, opened 2019-06-28
bug fixes for the nonlocal source term in KPP

maltrud:
This PR fixes several bugs in the calculation of the nonlocal source term in KPP.  for temperature, we need to subtract off the contribution from thickness fluxes that don't change buoyancy.  for salinity, we need to add virtual salinity fluxes that do change buoyancy.

these fixes have been tested in E3SM G-cases and behave as expected.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -1367,7 +1367,7 @@ contains
       real (kind=RKIND), dimension(:), pointer :: dcEdge, dvEdge, areaCell
       real (kind=RKIND), dimension(:), pointer :: penetrativeTemperatureFlux, surfaceThicknessFlux, &
            surfaceBuoyancyForcing, surfaceFrictionVelocity, penetrativeTemperatureFluxOBL, &
-           normalVelocitySurfaceLayer, surfaceThicknessFluxRunoff
+           normalVelocitySurfaceLayer, surfaceThicknessFluxRunoff, rainFlux, evaporationFlux
       real (kind=RKIND), pointer :: config_flux_attenuation_coefficient, config_flux_attenuation_coefficient_runoff
 
       real (kind=RKIND), dimension(:), pointer :: surfaceStress, surfaceStressMagnitude
@@ -1441,6 +1441,8 @@ contains
       call mpas_pool_get_array(forcingPool, 'penetrativeTemperatureFlux', penetrativeTemperatureFlux)
       call mpas_pool_get_array(forcingPool, 'surfaceStress', surfaceStress)
       call mpas_pool_get_array(forcingPool, 'surfaceStressMagnitude', surfaceStressMagnitude)
+      call mpas_pool_get_array(forcingPool, 'rainFlux', rainFlux)
+      call mpas_pool_get_array(forcingPool, 'evaporationFlux', evaporationFlux)
 
       call mpas_pool_get_array(tracersPool, 'activeTracers', activeTracers, 1)
       call mpas_pool_get_array(tracersSurfaceFluxPool, 'activeTracersSurfaceFlux', activeTracersSurfaceFlux)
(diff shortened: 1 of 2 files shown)


src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1496,12 +1498,15 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
...
 
        nonLocalSurfaceTracerFlux(indexTempFlux, iCell) = activeTracersSurfaceFlux(indexTempFlux,iCell) &
-               + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell) - fracAbsorbed * &
-               surfaceThicknessFlux(iCell) * activeTracers(indexTempFlux,1,iCell) + &
-               activeTracersSurfaceFluxRunoff(indexTempFlux,iCell) * fracAbsorbedRunoff
+               + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell)  &
+               - fracAbsorbed * (rainFlux(iCell) + evaporationFlux(iCell)) * activeTracers(indexTempFlux,1,iCell)/rho_sw  &
+               - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell) * (  &
+                   max(activeTracers(indexTempFlux,1,iCell),0.0_RKIND) &
+                 + min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)    )/rho_sw
```
mark-petersen: @maltrud Please check this:
```
  max(activeTracers(indexTempFlux,1,iCell),0.0_RKIND) &
+ min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND) 
```
is equal to:
```
activeTracers(indexTempFlux,1,iCell)
```
so that must not be what you meant.

maltrud: @mark-petersen , i actually did this on purpose to denote that there are 2 distinct cases depending on whether T>0 or T<0.  if T>0, then we need to subtract off the runoff surface T flux because it doesn't change the surface buoyancy.  if T<0, then no "effective flux" was needed, but it does change buoyancy so we need to add it in.

but you're right, it is confusing and should probably be in a comment instead.

maltrud: @mark-petersen , @vanroekel discovered an error.  the line with the 'max' needs to be removed:

max(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)
then the code should be ready.  can you modify the code, or should i?

mark-petersen: I just did it. It compiles. Check the revised code - is that what you want?

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1496,12 +1498,14 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
        ! transport code.  This includes tracer forcing due to thickness
 
        nonLocalSurfaceTracerFlux(indexTempFlux, iCell) = activeTracersSurfaceFlux(indexTempFlux,iCell) &
-               + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell) - fracAbsorbed * &
-               surfaceThicknessFlux(iCell) * activeTracers(indexTempFlux,1,iCell) + &
-               activeTracersSurfaceFluxRunoff(indexTempFlux,iCell) * fracAbsorbedRunoff
+               + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell)  &
+               - fracAbsorbed * (rainFlux(iCell) + evaporationFlux(iCell)) * activeTracers(indexTempFlux,1,iCell)/rho_sw  &
+               - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell) &
+                * min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)/rho_sw
```
mark-petersen: @maltrud check this revised version.

maltrud: @mark-petersen this looks right to me.  thanks.

mark-petersen:
Rebased and retested.

PR #307, opened 2019-07-01
Enable MPAS to be compiled into position-independent code

mgduda:
This merge enables MPAS to be compiled into positiion-independent code.

In order to allow MPAS code to be called from other systems with more
flexibility, e.g., as a shared library, this commit adds a new compile option,
`SHAREDLIB=true`, that causes all source files to be compiled with appropriate
flags to generate position-independent code.

The appropriate flags to use are not known for all compilers with build
targets in the top-level Makefile, and if SHAREDLIB=true is specified for
such a build target, an error like the following will be generated:

```
*** Position-independent code was requested but PIC flags are not available. Please add PIC flags for the 'ftn' target.  Stop.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,8 @@ xlf:
 	"LDFLAGS_DEBUG = -O0 -g" \
 	"FFLAGS_OMP = -qsmp=omp" \
 	"CFLAGS_OMP = -qsmp=omp" \
+	"PICFLAG = -qpic" \
+	"BUILD_TARGET = $(@)" \
 	"CORE = $(CORE)" \
 	"DEBUG = $(DEBUG)" \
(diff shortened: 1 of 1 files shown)


mgduda:

mgduda:
@mark-petersen I'm not sure this is the best explanation, but [position-independent code](https://en.wikipedia.org/wiki/Position-independent_code) is code that is compiled such that it makes no assumptions about the absolute locations of symbols (e.g., subroutines). Position-independent code is necessary if one wishes to create a shared library, which might be loaded anywhere in memory at run-time. As you pointed out, the changes in this PR only add additional compiler flags if `SHAREDLIB=true` is specified when invoking the top-level `Makefile`.

PR #308, opened 2019-07-01
Support multiple instances of domains in MPAS infrastructure

mgduda:
This merge introduces several changes in the MPAS infrastructure to better support
multiple concurrent instances of domains.
1) The `domain_ptr` and `corelist` module variables in `mpas_subdriver` are now
subroutine arguments to `mpas_init`, `mpas_run`, and `mpas_finalize`. See commit ea13ea2.
2) The pointer to `mpas_log_info` is now reset at the beginning of each call to
`mpas_run` and `mpas_finalize`. See commit db97841.
3) The domain type now includes a unique global ID, which is now used in log filenames
when multiple domains are active. See commit 9597b6f.
4) When MPAS is using an externally initialized MPI communicator, calls to `MPI_Abort`
are avoided whenever possible. See commit ffe6b7b.
5) In order to avoid eventually running out of unit numbers, logging units are explicity
freed with `mpas_release_unit` after the associated log unit has been closed.
See commit 4a78140.

### DIFF
--- a/src/framework/mpas_dmpar.F
+++ b/src/framework/mpas_dmpar.F
@@ -349,7 +349,13 @@ include 'mpif.h'
 #ifdef _MPI
       integer :: mpi_ierr, mpi_errcode
 
-      call MPI_Abort(dminfo % comm, mpi_errcode, mpi_ierr)
+      !
+      ! Only if MPI was initialized by MPAS should we call MPI_Abort
+      ! Otherwise, MPI was initialized outside of MPAS and we should not abort on MPI_COMM_WORLD
+      !
+      if ( dminfo % initialized_mpi ) then
+         call MPI_Abort(dminfo % comm, mpi_errcode, mpi_ierr)
+      end if
 #endif
 
       stop
--- a/src/framework/mpas_log.F
+++ b/src/framework/mpas_log.F
@@ -120,6 +120,7 @@ module mpas_log
       ! local variables
       !-----------------------------------------------------------------
       character(len=16) :: taskString  !< variable to build the task number as a string with appropriate zero padding
+      character(len=4) :: domainString  !< variable to store the domain number as a string with appropriate zero padding
       integer :: unitNumber !< local variable used to get a unit number
       character(len=strKind) :: proposedLogFileName, proposedErrFileName
       logical :: isOpen
@@ -176,6 +177,10 @@ module mpas_log
       mpas_log_info % taskID = domain % dminfo % my_proc_id
       mpas_log_info % nTasks = domain % dminfo % nprocs
 
+      ! Store the domain number
+      !   This will be used to number the log files
+      mpas_log_info % domainID = domain % domainID
+
       ! Set log file to be active or not based on master/nonmaster task and optimized/debug build
       !   * Optimized build: Only master task log is active
       !   * Debug build: All tasks active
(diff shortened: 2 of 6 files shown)


mgduda:

mgduda:
@mark-petersen @matthewhoffman Firstly, there's no particular rush to get this merged, so please take your time in reviewing the changes in this PR for compatibility with E3SM.

In order to make the full set of changes easier to understand, they're broken down into a sequence of commits that should each be more-or-less digestible.

I think the only controversial changes are those in 62d1bb6. I need to do more checking to convince myself that not calling `MPI_Abort` won't allow MPAS to continue executing with ill effects rather than either stopping execution (the current approach with `MPI_Abort`) or returning an error code (which probably doesn't happen in the required places at the moment).

mark-petersen:
@mgduda does merging this PR require any corresponding changes in each core?

jjguerrette:
@matthewhoffman, where is the comment on the variable declaration about being initialized externally?  I only see the actual behavior in `mpas_dmpar_init`, and not a comment.  `initialized_mpi` is false when MPI is initialized externally (`mpi_comm` is present).  If there is a comment that says otherwise, then it needs to be changed.  Or possibly another comment is needed to avoid confusion.

jjguerrette:
Forget my previous comment. I see that @mgduda added a code comment in `mpas_log_types.inc`.  The comment should be, `Whether MPI was initialized internally in MPAS`

mgduda:
@mark-petersen Merging this PR should not require any changes to cores *as stand-alone models*. However, I don't know whether any of the changes here might impact MPAS cores within E3SM. Do the MPAS E3SM components use the code from `src/driver/mpas_subdriver.F` directly , or is the driver-level code rewritten for MPAS cores within E3SM?

mgduda:
@mark-petersen @matthewhoffman I've just amended the last commit in this PR to correct the comments surrounding the `initialized_mpi` variable. What was commit 62d1bb6 has been amended as ffe6b7b.

src/framework/mpas_log.F
```diff
@@ -823,13 +843,16 @@ subroutine log_abort()
...
       close(mpas_log_info % errorLog % unitNum)
+      call mpas_release_unit(mpas_log_info % errorLog % unitNum)
 
       deallocate(mpas_log_info % errorLog)
       deallocate(mpas_log_info % outputLog)
       deallocate(mpas_log_info)
 
 #ifdef _MPI
-      call MPI_Abort(MPI_COMM_WORLD, mpi_errcode, mpi_ierr)
+      if ( mpas_log_info % initialized_mpi ) then
```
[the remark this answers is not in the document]
mgduda: @mark-petersen I've just updated the commit that introduced these changes with some explanatory comments. Essentially, if MPAS is using an externally-supplied communicator, we don't want to call `MPI_Abort` on `MPI_COMM_WORLD`. The MPI communicator would be initialized, but not by MPAS.

mgduda:
@jjguerrette I can understand that during development having MPAS call `MPI_Abort` might have been problematic. But, is it the case that you still need the `MPI_Abort` changes for JEDI? Under what sorts of circumstances was MPAS aborting, but JEDI could continue running? Are there changes that could be made in JEDI to ensure that MPAS is not run in a way that will cause it to abort?

jjguerrette:
@mgduda, one problem is that MPI_COMM_WORLD as defined in MPAS is not available in a JEDI  simulation.  We could start by trying to make the `MPI_Abort` call in `log_abort` refer to, e.g., `dminfo%comm`.  Do you know a way to get `dminfo%comm` or an equivalent piped down into `log_abort`?  Or is such a behavior undesirable?

If we can get that behavior enabled then I can test in a JEDI simulation for which we expect an MPAS forecast error to exercise that bit of code.

mgduda:
@jjguerrette I'll see if we can instead call `MPI_Abort` on `dminfo%comm`. There was some discussion yesterday (30 Sept 2019) with @matthewhoffman and @xylar regarding the abort changes, and the general consensus was that changing the fundamental way in which critical errors function is probably too risky. So, I'm eager to work with you to find an alternative to not calling `MPI_Abort` when MPAS is initialized with an external communicator.

src/framework/mpas_dmpar.F
```diff
@@ -349,7 +349,13 @@ subroutine mpas_dmpar_abort(dminfo)!{{{
...
       integer :: mpi_ierr, mpi_errcode
 
-      call MPI_Abort(dminfo % comm, mpi_errcode, mpi_ierr)
+      !
+      ! Only if MPI was initialized by MPAS should we call MPI_Abort
+      ! Otherwise, MPI was initialized outside of MPAS and we should not abort on MPI_COMM_WORLD
+      !
+      if ( dminfo % initialized_mpi ) then
+         call MPI_Abort(dminfo % comm, mpi_errcode, mpi_ierr)
+      end if
```
mgduda: @jjguerrette These changes seem like they wouldn't be useful to JEDI, since even if we avoided calling `MPI_Abort` here, line 361 (just a few lines below) would execute a Fortran `stop` statement. I suppose these changes were made on the path to the changes in the `mpas_log.F` abort behavior. If that's the case, could you confirm that reverting these changes in `mpas_dmpar.F` would not be problematic for your use case?

jjguerrette: As discussed in person, this mpas_dmpar_abort may no longer be called.  We may need to revisit `mpas_dmpar_global_abort` when I have an appropriate use-case, but for now the changes to this subroutine can be reverted.

mgduda:
Referring to parts listed in the PR description, the following closed PRs handle parts 1, 2, 3, and 5 from this PR:

- PR #744 handles part 1
- PR #746 handles part 2
- PR #749 handles part 3
- PR #748 handles part 5

Since we've decided against including changes from the fourth part (which avoids calls to `MPI_Abort`), all parts of this PR have been accounted for, and we can now close this PR.

PR #309, opened 2019-07-01
Reserve some unit numbers for unformatted I/O

mgduda:
This merge reserves some unit numbers for unformatted I/O in mpas_io_units.

The mpas_io_units module now reserves the upper part of the range of possible
unit numbers for unformatted I/O. An unformatted I/O unit can be requested
with the optional 'unformatted' argument to mpas_new_unit.

Having specific unit numbers reserved for unformatted I/O is helpful in cases
where the compiler is able to use environment variables to set the endianness
or other properties based on unit number. For example, one could request that
units 101 through 200 perform unformatted I/O using big-endian byte order.

### DIFF
--- a/src/framework/mpas_io_units.F
+++ b/src/framework/mpas_io_units.F
@@ -21,8 +21,19 @@ module mpas_io_units
 
    use mpas_kind_types
 
-   integer, parameter, private :: maxUnits = 99
-   logical, dimension(0:maxUnits), private, save :: unitsInUse
+   implicit none
+
+   private
+
+   integer, parameter :: maxUnits = 99
+   logical, dimension(0:maxUnits), save :: unitsInUse
+
+   ! Units reserved for unformatted I/O
+   integer, parameter :: unformatted_min = 95
+   integer, parameter :: unformatted_max = maxUnits
+
+   public :: mpas_new_unit, &
+             mpas_release_unit
 
    contains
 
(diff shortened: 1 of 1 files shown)


mgduda:

mgduda:
@matthewhoffman I think the specific use case for these changes was to enable setting the endianness of unformatted output and input files at run-time using [GFORTRAN_CONVERT_UNIT](https://gcc.gnu.org/onlinedocs/gcc-5.2.0/gfortran/GFORTRAN_005fCONVERT_005fUNIT.html) with the GNU compilers. I'm not really sure of the details, but somehow it was not feasible or possible to compile the code, e.g., with `-fconvert=big-endian`. So, we can now request a unit number within a pre-set range with
```fortran
call mpas_new_unit(unit, unformatted=.true.)
```
and control the endianness of this unit with, e.g.,
```
export GFORTRAN_CONVERT_UNIT=big_endian:95-99
```

jjguerrette:
When MPAS is built as a shared library, using `-fconvert=big-endian` during the MPAS build was not working in the dependent program.  In our group's particular case, the encompassing program is based in C++ where `fconvert` is not an accepted flag.

In #108, I change the unformatted unit numbers to 101-200, because there our application required more then 5 unformatted units.  Perhaps some of those changes to src/framework/mpas_io_units.F could be merged here @mgduda.  Sorry for the confusion.

mgduda:
@mark-petersen and @matthewhoffman Sorry for the additional work, but I just added one new commit, 7aacdd1, that changes the range of unit numbers reserved for unformatted I/O to [101,200]. Do you see any problems with this additional commit?

mark-petersen:
Fine with me. Looking at these PR changes, if we never call `mpas_new_unit` from our own core with the `unformatted=.true.` logical, then it wouldn't use any new units in the unformatted range anyway, so there is no potential for problems.

matthewhoffman:
@mgduda , that's not a problem - thanks for asking.

PR #311, opened 2019-07-02
Adds option for an explicit barotropic (ssh-only) pressure gradient

pwolfram:
This PR ensures that the applied pressure gradient is strictly from
the barotropic term, e.g., - g d eta / dx.


pwolfram:
@mark-petersen, this is the "2nd" PR I recommended be merged back into `ocean/develop`.

pwolfram:
Thanks @mark-petersen!

PR #312, opened 2019-07-02
Adds depth-averaged Rayleigh drag

pwolfram:
Insteady of a term like

du
-- = -r U + ...
dt

this applies

du       
-- = - (r/H) U
dt      

where r is the Rayleigh drag parameter.  Note, this is done
on a total depth basis; H is not the layer thickness.

This option allows for a direct comparison against idealized, quasi-analytical test cases as found in Warner et al (2013).

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix.F
@@ -189,18 +189,18 @@ contains
 
 !***********************************************************************
 !
-!  routine ocn_vel_vmix_tend_implicit
+!  routine ocn_vel_vmix_tend_implicit_rayleigh
 !
 !> \brief   Computes tendencies for implicit momentum vertical mixing
-!> \author  Mark Petersen
-!> \date    September 2011
+!> \author  Mark Petersen, Phillip J. Wolfram
+!> \date    September 2011, July 2019
 !> \details
 !>  This routine computes the tendencies for implicit vertical mixing for momentum
-!>  using computed coefficients.
+!>  using computed coefficients and includes implicit rayleigh drag.
 !
 !-----------------------------------------------------------------------
 
-   subroutine ocn_vel_vmix_tend_implicit(meshPool, dt, kineticEnergyCell, vertViscTopOfEdge, layerThickness, & !{{{
+   subroutine ocn_vel_vmix_tend_implicit_rayleigh(meshPool, dt, kineticEnergyCell, vertViscTopOfEdge, layerThickness, & !{{{
                                          layerThicknessEdge, normalVelocity, err)
 
       !-----------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


pwolfram:
@mark-petersen, my guess is that this PR may break BFB but I'm not 100% sure.  Regardless, it does add a few flops to computation of vertically implicit coefficients.

pwolfram:
Following our conversation today, this PR will also include an explicit separation of the implicit Rayleigh terms from the standard vertical implicit solve-- stay tuned for the commit doing this!

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -572,8 +732,15 @@ subroutine ocn_vmix_implicit(dt, meshPool, diagnosticsPool, statePool, forcingPo
       !  Implicit vertical solve for momentum
       !
       call mpas_timer_start('vmix solve momentum', .false.)
-      call ocn_vel_vmix_tend_implicit(meshPool, dt, kineticEnergyCell, vertViscTopOfEdge, layerThickness, layerThicknessEdge, &
```
pwolfram: @mark-petersen, does this seem reasonable to you?  I'm going to be of course testing this on the drying slope test case next.

pwolfram: I'm thinking I'll merge it and then if there is a bug do another PR to keep moving quickly.

pwolfram:
Thanks @mark-petersen!

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -663,7 +830,8 @@ subroutine ocn_vmix_init(domain, err)!{{{
       logical, pointer :: config_disable_vel_vmix, config_disable_tr_vmix
       logical, pointer :: config_use_implicit_bottom_drag
       real (kind=RKIND), pointer :: config_implicit_bottom_drag_coeff
-      logical, pointer :: config_Rayleigh_friction, config_Rayleigh_bottom_friction
+      logical, pointer :: config_Rayleigh_friction, config_Rayleigh_bottom_friction, &
+                          config_Rayleigh_damping_coeff_depth_variable
```
mark-petersen: @pwolfram There is a mismatch here. It is supposed to be `config_Rayleigh_damping_depth_variable`, right? You have coeff in there by mistake.  I merged coastal into ocean/develop locally, this breaks testing using debug.

I can fix it. Just checking that you meant this variable.

mark-petersen: @pwolfram I ran out of time. If you get a chance, please correct config_Rayleigh_damping_coeff_depth_variable to config_Rayleigh_damping_depth_variable. If you don't have time, we can do this later.

PR #317, opened 2019-07-03
Fix vertical coordinate weights

### LINKED ISSUE #316: bottomDepth of zero causes divide by zero
For wetting/drying, we need to allow bottomDepth to be zero, which can easily occur on a beach. The model currently divides by zero in the z-star ALE layer thickness weighting.

mark-petersen:
This PR does two things:
- Add optional new weighting in ALE SSH thickness distribution
- Add tapered weights options for z-star

The first is used for wetting/drying, to allow bottomDepth to be zero, which can easily occur on a beach. It changes the weighting for the z-star distribution from 
`vertCoordMovementWeights(k) * restingThickness(k, iCell)`
to 
`vertCoordMovementWeights(k)`.
The first formulation can be seen in eqn 6 of Petersen et al 2015.

The new formulation stretches in proportion to the defined weights. The standard formulation stretched in proportion to the weights*restingThickness. That stretches deep layers _more than shallow layers_ with surface gravity waves.

These changes are protected by flags, so are bfb in MPAS-Ocean and E3SM.

Fixes #316

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -2206,7 +2206,7 @@
 			 description="Reference depth of ocean for each vertical interface. Used in 'z-level' type runs."
 		/>
 		<var name="vertCoordMovementWeights" type="real" dimensions="nVertLevels" units="unitless"
-			 description="Weights used for distribution of sea surface heigh purturbations through multiple vertical levels."
+			 description="Weights used for distribution of sea surface height perturbations through multiple vertical levels."
 		/>
 		<var name="boundaryEdge" type="integer" dimensions="nVertLevels nEdges" units="unitless"
 			 description="Mask for determining boundary edges. A boundary edge has only one active ocean cell neighboring it."
@@ -2218,13 +2218,13 @@
 			 description="Mask for determining boundary cells. A boundary cell has at least one inactive cell neighboring it."
 		/>
 		<var name="edgeMask" type="integer" dimensions="nVertLevels nEdges" units="unitless"
-			 description="Mask on edges that determines if computations should be done on edge."
+			 description="Mask on edges that determines if computations should be done on edges."
 		/>
 		<var name="vertexMask" type="integer" dimensions="nVertLevels nVertices" units="unitless"
-			 description="Mask on vertices that determines if computations should be done on vertice."
+			 description="Mask on vertices that determines if computations should be done on vertices."
 		/>
 		<var name="cellMask" type="integer" dimensions="nVertLevels nCells" units="unitless"
-			 description="Mask on cells that determines if computations should be done on cell."
+			 description="Mask on cells that determines if computations should be done on cells."
 		/>
 		<var name="edgeSignOnCell" type="integer" dimensions="maxEdges nCells" units="unitless"
 			 description="Sign of edge contributions to a cell for each edge on cell. Used for bit-reproducible loops. Represents directionality of vector connecting cells."
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -239,7 +239,15 @@
 	<nml_record name="ALE_vertical_grid" mode="forward">
 		<nml_option name="config_vert_coord_movement" type="character" default_value="uniform_stretching" units="unitless"
 					description="Determines the vertical coordinate movement type. 'uniform_stretching' distributes SSH perturbations through all vertical levels (z-star vertical coordinate); 'fixed' places them all in the top level (z-level vertical coordinate); 'user_specified' allows the input file to determine the distribution using the variable vertCoordMovementWeights (weighted z-star vertical coordinate); and 'impermeable_interfaces' makes the vertical transport between layers zero, i.e. $w^t=0$ (idealized isopycnal)."
-					possible_values="'uniform_stretching', 'fixed', 'user_specified', 'impermeable_interfaces', 'tapered_250m', 'tapered_500m', tapered_1000m'"
+					possible_values="'uniform_stretching', 'fixed', 'user_specified', 'impermeable_interfaces', 'tapered'"
+		/>
+		<nml_option name="config_vert_taper_weight_depth_1" type="real" default_value="250.0" units="m"
+					description="Vertical coordinate taper weight is one above this depth, linearly decreases to zero below."
+					possible_values="any positive real value, but typically 100 to 1000 m."
+		/>
+		<nml_option name="config_vert_taper_weight_depth_2" type="real" default_value="500.0" units="m"
+					description="Vertical coordinate taper weight is zero below this depth, linearly increases to one above."
+					possible_values="any positive real value, but typically 100 to 1000 m and greater than config_vert_taper_weight_depth_1."
 		/>
 		<nml_option name="config_use_min_max_thickness" type="logical" default_value=".false." units="unitless"
 					description="If true, a minimum and maximum thicknesses are enforced to prevent massless and very thick layers."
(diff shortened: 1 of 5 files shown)


mark-petersen:
@pwolfram Please test this on your wetting/drying case.

src/core_ocean/shared/mpas_ocn_init_routines.F
```diff
@@ -459,6 +460,37 @@ subroutine ocn_init_routines_vert_coord(domain)!{{{
...
             vertCoordMovementWeights = 1.0_RKIND
 
+         ! Set weight tapering:
+         !  1.0 shallower than depth1
+         !  linear in between
+         !  0.0 deeper than depth2
+         elseif (config_vert_coord_movement.eq.'tapered_250m') then
+
+            depth1 = 250.0
+            depth2 = 500.0
```
xylar: Is there a compelling reason not to make these 2 depths namelist parameters and just have `'tapered'` rather than 3 hard-coded choices?

xylar:
@mark-petersen, could you make sure to test this with `isomip_plus/2km/Ocean0` or another test case with an ice shelve cavity?  I don't see any reason it won't work but this definitely could affect ice-shelf test cases.

src/core_ocean/shared/mpas_ocn_init_routines.F
mark-petersen: Thanks for looking through it. I went back and forth on that, but I think you're right. I originally thought that 'user_defined' would catch anything else, but for the ramp function your suggestion makes more sense.

src/core_ocean/shared/mpas_ocn_init_routines.F
```diff
@@ -392,7 +392,8 @@ subroutine ocn_init_routines_vert_coord(domain)!{{{
       type (block_type), pointer :: block
 
       integer :: iTracer, cell, cell1, cell2
-      real (kind=RKIND) :: normalThicknessFluxSum, thicknessSum, hEdge1, zMidPBC
+      real (kind=RKIND) :: normalThicknessFluxSum, thicknessSum, hEdge1, zMidPBC, &
+         depth1, depth2
```
pwolfram: What happens if `depth1` is junk data, e.g., for `'uniform stretching'`?  Should it be set to zero below somewhere before the if statements (which could be a case construct)?

src/core_ocean/shared/mpas_ocn_init_routines.F
```diff
@@ -459,6 +460,37 @@ subroutine ocn_init_routines_vert_coord(domain)!{{{
 
             vertCoordMovementWeights = 1.0_RKIND
 
+         ! Set weight tapering:
+         !  1.0 shallower than depth1
+         !  linear in between
+         !  0.0 deeper than depth2
+         elseif (config_vert_coord_movement.eq.'tapered_250m') then
```
pwolfram: Should we make these case constructs?

src/core_ocean/shared/mpas_ocn_thick_ale.F
```diff
@@ -151,16 +151,19 @@ subroutine ocn_ALE_thickness(meshPool, verticalMeshPool, SSH, ALE_thickness, err
...
-         ! Note that restingThickness is nonzero, and remaining terms are perturbations about zero.
          do k = 1, kMax
-            SSH_ALE_Thickness(k) = SSH_ALE_Thickness(k) / thicknessSum
-            ALE_Thickness(k, iCell) = restingThickness(k, iCell) + SSH_ALE_Thickness(k)
+            ! Using this, we must require that sum(restingThickness(k, iCell))
+            ! summed over k is equal to bottomDepth.
+            ! This is equation 4 and 6 in Petersen et al 2015, but with eqn 6
+            ! altered so only the W_k weights are used. The resting
+            ! thickness shown in eqn 6 is not included here.
+            ALE_Thickness(k, iCell) = restingThickness(k, iCell) + ssh(iCell) * vertCoordMovementWeights(k) / weightSum
```
pwolfram: Perfect, thanks!

src/core_ocean/shared/mpas_ocn_init_routines.F
xylar: If `depth1` is a junk value, it doesn't get used by anything.  There's no reason to set it to zero.

src/core_ocean/shared/mpas_ocn_init_routines.F
xylar: I find a case statement harder to read.  I personally prefer if/elseif.

src/core_ocean/shared/mpas_ocn_init_routines.F
mark-petersen: Fixed.

src/core_ocean/shared/mpas_ocn_init_routines.F
mark-petersen: These are specified with a flag now anyway.

src/core_ocean/shared/mpas_ocn_init_routines.F
mark-petersen: I think if/elseif is more readable too. This is in init, so not a performance issue.

pwolfram: Phil Jones' seminar last week suggested it was better from a performance standpoint, hence my recommendation.  Its irrelevant now anyway.

src/core_ocean/shared/mpas_ocn_init_routines.F
```diff
@@ -460,34 +463,16 @@ subroutine ocn_init_routines_vert_coord(domain)!{{{
...
-
-         elseif (config_vert_coord_movement.eq.'tapered_1000m') then
-
-            depth1 = 1000.0
-            depth2 = 1500.0
-            do k = 1, nVertLevels
-               vertCoordMovementWeights(k) = 1.0_RKIND + (refZMid(k) + depth1)/(depth2 - depth1)
+               vertCoordMovementWeights(k) = 1.0_RKIND + &
+                 (refZMid(k) + config_vert_taper_weight_depth_1) / &
+                 (config_vert_taper_weight_depth_2 - config_vert_taper_weight_depth_1)
```
pwolfram: Much cleaner, I like it!

src/core_ocean/shared/mpas_ocn_init_routines.F
xylar: Any self-respecting compiler should be able to see that and optimize it on its own.  I think it's a fool's errand to worry about such things unless they're deep in a loop or something.

pwolfram:
@mark-petersen, this works with the drying slope case for sigma layers-- thank you!

image

pwolfram:
@mark-petersen, this is great and I'm happy for a merge.  Thanks!

mark-petersen:
@vanroekel is out of town, removing him from reviewer list.

src/core_ocean/Registry.xml
```diff
@@ -239,7 +239,19 @@
...
-					possible_values="'uniform_stretching', 'fixed', 'user_specified', 'impermeable_interfaces'"
+					possible_values="'uniform_stretching', 'fixed', 'user_specified', 'impermeable_interfaces', 'tapered'"
+		/>
+		<nml_option name="config_ALE_thickness_proportionality" type="character" default_value="restingThickness_times_weights" units="unitless"
+					description="When config_vert_coord_movement='uniform_stretching' (z-star type coordinate), determines whether ALE layer thickness is proportional to the resting thickness times weights, or just the weights. The first is standard for global runs and is what is specified in Petersen et al 2015 eqns 4 and 6. The second is useful for wetting/drying test cases where resting thickness may be zero at the coastlines."
+					possible_values="'restingThickness_times_weights' or 'weights_only'"
```
mark-petersen: @pwolfram I realized that we don't want this alteration in general, so I am protecting it with this flag. In general in the ocean, we need to have the thickness vary in proportion to the resting thickness (i.e. original distribution), not the weights. You can run wetting/drying test cases with the 
`config_ALE_thickness_proportionality = weights_only` 
setting. But this breaks ice shelf cavities, as @xylar feared, because you need to 'squeeze the accordion' in proportion to the original thicknesses, especially for extreme compression, like we have under ice shelves.

This commit is now bit-for-bit with previous for MPAS-O and E3SM. We will need to think about this more carefully to put wetting/drying along coastlines and severe SSH alterations below ice shelves in the same domain.

mark-petersen:
BFB against ocean/develop for nightly test suite with gnu on grizzly. Also tested debug using gnu and intel.

mark-petersen:
rebased.

PR #270, opened 2019-06-06
Add option to seed LIGHT particles at cell-centers with gaussian noise added

bradyrx:
This PR adds a flag to the LIGHT particle initialization script to seed 3 particles at the cell-center with a Gaussian noise perturbation. It also allows a user to seed both the cell-center and vertices simultaneously.

PR #319, opened 2019-07-04
Convert LIGHT particle init script to python3

bradyrx:
This converts the LIGHT `make_particle_file` init script from python2 to python3. It cleans up syntax using `flake8`, `isort`, and `black`, as well as some manual changes (e.g., constants are now in CAPS, and type casting of variables is handled in `argparse`).

### DIFF
--- a/testing_and_setup/compass/ocean/scripts/LIGHTparticles/make_particle_file.py
+++ b/testing_and_setup/compass/ocean/scripts/LIGHTparticles/make_particle_file.py
@@ -1,15 +1,49 @@
 #!/usr/bin/env python
 """
-File writes a particle input dataset for use in MPAS-O / E3SM.
+Creates a particle input dataset for use of LIGHT in MPAS-O.
 
-Example usage is
+Base usage (required fields):
 
-    python make_particle_file.py -i init.nc -g graph.info.part.6 \
-            -o particles.nc -p 6 --spatialfilter SouthernOceanXYZ \
-            --downsample=1
+    ./make_particle_file.py -i init.nc -g graph.info.part.6 \
+            -o particles.nc -p 6
 
-Phillip J. Wolfram
-Last Modified: 08/03/2018
+By default, surface, isopycnal, and passive floats are all seeded. One can select
+particle modes by passing a list:
+
+    ./make_particle_file.py -i init.nc -g graph.info.part.6 \
+            -o particles.nc -p 6 -t surface,passive
+
+If passive floats are being used, the default seeding mode is 10 vertical particles
+linearly distributed through depth at each grid cell. The number of layers can be
+modified with `--nvertlevels INT`. Current seeding modes supported are 'linear',
+'log', and 'denseCenter'. One of the three can be passed using `--vertseedtype`.
+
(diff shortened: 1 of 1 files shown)


bradyrx:
@pwolfram , this has no breaking changes to my knowledge. I did some light testing and things seem to work as they did before, given these are just syntax changes.

This is branched off of https://github.com/MPAS-Dev/MPAS-Model/pull/270, so it's taking some of those commits along for the ride. After https://github.com/MPAS-Dev/MPAS-Model/pull/270 is merged, I'll rebase onto `ocean/develop` and cherrypick here to drop those commits. Currently getting merge conflicts from it.

mark-petersen:
@bradyrx go ahead and rebase on ocean/develop, and see if those are the changes you want in this PR. That will take care of any merge conflicts.

mark-petersen:
Coincidentally, @amrapallig is running some test cases with particles. @amrapallig please run the make_particle_file.py script in this PR and make sure it runs the same way as before. You can use the `ncdiff file1.nc file2.nc fileout.nc` command to see if it produces the same output as before.

bradyrx:
Okay @mark-petersen, this is rebased onto ocean/develop with the proper commits now. @amrapallig can go ahead and check if files are BFB between this and the older script. Thanks!

amrapallig:
Both #270  and #319 result in different nParticle size compared to the older script when downsample is anything other than the default value (--downsample=0).  For example, if --downsample=1, nParticles=1211540 for #270, nParticles=1212530 for #319 and nParticles=1212772 for #older script. With --downsample=0, these are exactly the same. Though this might not result in any drastic change in the particle statistics, better to check why the nParticles are different.

mark-petersen:
@bradyrx when using `--downsample=1`,  we got a different number of particles before and after this PR. Do you think that is any cause for concern? I don't care if this PR changes the number sightly. I just want to make sure we are not introducing an error by mistake, if there is no reason to expect this.

bradyrx:
@mark-petersen, @amrapallig, there's no reason to expect this since most of this PR is just cosmetic. I'll give this a look Wed or Thurs to see what's going on. Although I'd be more concerned if this was without downsampling, i.e., that particles were being added somewhere in the vertical or horizontal. It seems more like it's changing the number of particle columns being stripped while coarsening in the horizontal.

pwolfram:
@bradyrx, I'm not 100% sure that pyamg is BFB depending upon runs.  If not, we wouldn't expect the same downsampled result each time.  But, arguably, it would be good to ensure that downsample is BFB each time it is run.  

@amrapallig, if you run the same script twice (exactly same commands with downsample), does the output file have the same number of particles?

bradyrx:
@pwolfram , that was going to be my first test. I imagine `pyamg` might be introducing a different coarsening solution each time. I.e., this problem (if it is one) is outside of the changes made here. Curious to see the results from @amrapallig .

amrapallig:
Yes, the same script with the same commands with downsample (other than default zero) gives a different number of particles each time (varying by few hundreds ). The oldest script also does the same.  It has nothing to do with the recent changes.

mark-petersen:
@pwolfram we've tested this and are satisfied. Can you approve?

Issue #315, opened 2019-07-03
densitySurfaceDisplaced not used


xylar:
A lot of code and presumably computing time goes into computing `densitySurfaceDisplaced` for KPP, but then this field is never used:
```
$ grep -r densitySurfaceDisplaced
src/core_ocean/shared/mpas_ocn_diagnostics.F:           layerThickness, zMid, zTop, densitySurfaceDisplaced, density, &
src/core_ocean/shared/mpas_ocn_diagnostics.F:      type (field2DReal), pointer :: densitySurfaceDisplacedField, thermalExpansionCoeffField, salineContractionCoeffField
src/core_ocean/shared/mpas_ocn_diagnostics.F:      call mpas_pool_get_field(scratchPool, 'densitySurfaceDisplaced', densitySurfaceDisplacedField)
src/core_ocean/shared/mpas_ocn_diagnostics.F:      call mpas_allocate_scratch_field(densitySurfaceDisplacedField, .true., .false.)
src/core_ocean/shared/mpas_ocn_diagnostics.F:      densitySurfaceDisplaced => densitySurfaceDisplacedField % array
src/core_ocean/shared/mpas_ocn_diagnostics.F:                                         densitySurfaceDisplaced, err, thermalExpansionCoeff, salineContractionCoeff, &
src/core_ocean/shared/mpas_ocn_diagnostics.F:      call mpas_deallocate_scratch_field(densitySurfaceDisplacedField, .true.)
src/core_ocean/Registry.xml:		<var name="densitySurfaceDisplaced"
```
This field and all associated code for computing the EOS in `'surfaceDisplaced'` mode should be removed.

xylar:
@vanroekel, you had brought this to my attention.  Feel free to do this clean up if you have time.  I'll try to get to it at some point in the not too distant future if you don't.  It's dumb to be implementing functionality for new EOSes that we don't use...

philipwjones:
In older versions of KPP, this was used for surface buoyancy forcing and for the BL depth, but if it's not used anymore for those, would certainly be great to remove another bit of complexity from EOS.

PR #320, opened 2019-07-07
Removes References to surfaceDensityDisplaced

vanroekel:
SurfaceDensityDisplaced is not used but computed.  All references and
associated computations of this variable are removed.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -1372,7 +1372,7 @@ contains
 
       real (kind=RKIND), dimension(:), pointer :: surfaceStress, surfaceStressMagnitude
       real (kind=RKIND), dimension(:,:), pointer ::  &
-           layerThickness, zMid, zTop, densitySurfaceDisplaced, density, &
+           layerThickness, zMid, zTop, density, &
            normalVelocity, activeTracersSurfaceFlux, thermalExpansionCoeff, salineContractionCoeff, &
            activeTracersSurfaceFluxRunoff, nonLocalSurfaceTracerFlux
 
@@ -1391,7 +1391,7 @@ contains
       real (kind=RKIND) :: numerator, denominator, turbulentVelocitySquared, fracAbsorbed, fracAbsorbedRunoff
       real (kind=RKIND) :: factor, deltaVelocitySquared, delU2, invAreaCell
 
-      type (field2DReal), pointer :: densitySurfaceDisplacedField, thermalExpansionCoeffField, salineContractionCoeffField
+      type (field2DReal), pointer :: thermalExpansionCoeffField, salineContractionCoeffField
 
       call mpas_timer_start('KPP input fields')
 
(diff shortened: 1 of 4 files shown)


vanroekel:
@xylar here is the PR to address #315 

Confirmed this complies cleanly, but have not tested if BFB yet.  Working on this now.

xylar:
...and I suppose this got stuck because of #321.  Thanks for taking care of this.

xylar:
@vanroekel, I think there's one additional piece of clean-up needed here.  Since `tracersSurfaceLayerValue` is no longer used by any EOS, it should be removed as an argument to each EOS and should be removed entirely from `mpas_ocn_equation_of_state.F`.

vanroekel:
confirmed code is BFB using gfortran on grizzly.  Removing in progress.  @mark-petersen would you like me to run ifort too?  Or would you rather just run your standard testing?

mark-petersen:
Thanks for the cleanup. BFB with gfortran is enough, thanks. I'll test and merge with the next bfb bunch.

mark-petersen:
FYI, I was going to merge this, but merged locally and it failed some MPAS-O regression testing. I didn't have time to investigate. When it is time to merge, we should rebase and test again. This is a simple clean-up PR, we can do this when we have time.

mark-petersen:
Rebased. Fails bfb comparison with ocean/develop for global test cases.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1448,25 +1342,17 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
...
-      densitySurfaceDisplaced => densitySurfaceDisplacedField % array
       thermalExpansionCoeff => thermalExpansionCoeffField % array
       salineContractionCoeff => salineContractionCoeffField % array
 
       ! Only need to compute over the 0 and 1 halos
       nCells = nCellsArray( 3 )
 
-      ! compute EOS by displacing SST/SSS to every vertical layer in column
-      call ocn_equation_of_state_density(statePool, diagnosticsPool, meshPool, scratchPool, nCells, 0, 'surfaceDisplaced', &
-                                         densitySurfaceDisplaced, err, thermalExpansionCoeff, salineContractionCoeff, &
```
mark-petersen: @vanroekel This EOS call was removed, but thermalExpansionCoeff and salineContractionCoeff are used below. Was that an oversight? I added it back in with `'absolute', density` but global runs still differ from ocean/develop.

vanroekel: yes, this was overzealous removal of ```densitySurfaceDisplaced```. and it is not surprising after looking at this more that the PR is non bfb. we are changing the alpha and beta values used in KPP.  My preferred long term solution is to change this call to compute potentialDensity and return thermalExpansionCoeff and salineContractionCoeff associated with that displacement.  I expect that change to be non-BFB but is unlikely to be climate changing.  I would suggest we hold off on this PR until I can fully test a change to potential density.

xylar:
@mark-petersen, in my view, there's no point in doing this PR until `surfaceDisplaced` is eliminated as a flag to the EOS.  So rather than adding back a scratch variable, let's just hold off until @vanroekel can test.

mark-petersen:
@vanroekel, the eos call I added back in still has `'surfaceDisplaced'` as an argument, which now doesn't make sense. But putting the EOS call back in does give me BFB match with ocean/develop, so that was the problem. I'll let you take it from here.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1447,26 +1344,12 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
...
-      call mpas_pool_get_field(scratchPool, 'salineContractionCoeff', salineContractionCoeffField)
-      call mpas_allocate_scratch_field(densitySurfaceDisplacedField, .true., .false.)
-      call mpas_allocate_scratch_field(thermalExpansionCoeffField, .true., .false.)
-      call mpas_allocate_scratch_field(salineContractionCoeffField, .true., .false.)
-
-      densitySurfaceDisplaced => densitySurfaceDisplacedField % array
-      thermalExpansionCoeff => thermalExpansionCoeffField % array
-      salineContractionCoeff => salineContractionCoeffField % array
+      call mpas_pool_get_array(diagnosticsPool, 'inSituThermalExpansionCoeff',thermalExpansionCoeff)
+      call mpas_pool_get_array(diagnosticsPool, 'inSituSalineContractionCoeff',salineContractionCoeff)
```
mark-petersen: @vanroekel I just retested, and this matches bfb with ocean/develop for all cases now. I see the EOS call below is deleted. Does this mean that `inSituThermalExpansionCoeff` and `inSituSalineContractionCoeff` can just be obtained because they were computed previously, presumably in the EOS call in diagnostics, so there is no need to calculate them again here?

vanroekel:
@mark-petersen I've finished cleaning up this PR.  I see BFB with ocean/develop in the EC60to30 test case.  However, as I mentioned off line E3SM simulations are not at all BFB.  It's very confusing as to why.  Let me know what you find in E3SM testing.  I don't have any other ideas to try in E3SM.

mark-petersen:
Rebased on `ocean/develop`. Testing again.

xylar:
@mark-petersen and @vanroekel, I believe this needs to be revived from the grave when time allows.

mark-petersen:
I tried to rebase this, but became clear that there are enough changes that it needs to be redone. I think it is still worth doing because it removes extra variables that are floating around, and reduces code complexity.

I think this could be simplified by splitting it into three PRs:
1. Remove `inSituEOS` package
2. Remove tracersSurfaceLayerValue, normalVelocitySurfaceLayer
3. Remove thermalExpansionCoeff, salineContractionCoeff

mgduda:

PR #323, opened 2019-07-10
Surface thickness flux spreading in MPAS-O

maltrud:
This PR allows ocean surface thickness fluxes into/out of the ocean due to precipitation, evaporation, ice formation/melt, and ice runoff to be spread out vertically in a manner analogous to what is done currently for just liquid runoff.  there is a new namelist variable that sets the e-folding scale of the vertical spreading profile.

this capability was added to reduce the sometimes unrealistically large vertical stability due to dumping large amounts of fresh water only in the top layer of the model.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -1506,7 +1506,7 @@ contains
                             -100.0_RKIND) )
        fracAbsorbedPrecEvapIce = 1.0_RKIND - exp( max(-layerThickness(1, iCell) / &
           config_flux_attenuation_coefficient_prec_evap_ice, -100.0_RKIND) )
-                            
+
        ! Store the total tracer flux below in nonLocalSurfaceTemperatureFlux for use in the CVMix nonlocal
        ! transport code.  This includes tracer forcing due to thickness
 
@@ -1515,7 +1515,6 @@ contains
                - fracAbsorbedPrecEvapIce * (rainFlux(iCell) + evaporationFlux(iCell)) * &
                    activeTracers(indexTempFlux,1,iCell)/rho_sw &
                - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell) * (  &
-                   max(activeTracers(indexTempFlux,1,iCell),0.0_RKIND) &
                  + min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)    )/rho_sw
 
        nonLocalSurfaceTracerFlux(indexSaltFlux,iCell) = activeTracersSurfaceFlux(indexSaltFlux,iCell) &
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -153,8 +153,10 @@ contains
 
       real (kind=RKIND), pointer :: config_flux_attenuation_coefficient
       real (kind=RKIND), pointer :: config_flux_attenuation_coefficient_runoff
+      real (kind=RKIND), pointer :: config_flux_attenuation_coefficient_prec_evap_ice
       real (kind=RKIND), dimension(:), pointer :: surfaceFluxAttenuationCoefficient
       real (kind=RKIND), dimension(:), pointer :: surfaceFluxAttenuationCoefficientRunoff
+      real (kind=RKIND), dimension(:), pointer :: surfaceFluxAttenuationCoefficientPrecEvapIce
 
       real (kind=RKIND) :: areaTri1, layerThicknessVertexInv, tempRiVal, dcEdge_temp, dvEdge_temp, weightsOnEdge_temp
       real (kind=RKIND), dimension(:), allocatable:: layerThicknessVertexVec
@@ -176,6 +178,8 @@ contains
       call mpas_pool_get_config(ocnConfigs, 'config_flux_attenuation_coefficient', config_flux_attenuation_coefficient)
       call mpas_pool_get_config(ocnConfigs, 'config_flux_attenuation_coefficient_runoff',  &
                                              config_flux_attenuation_coefficient_runoff)
+      call mpas_pool_get_config(ocnConfigs, 'config_flux_attenuation_coefficient_prec_evap_ice',  &
+                                             config_flux_attenuation_coefficient_prec_evap_ice)
       call mpas_pool_get_config(ocnConfigs, 'config_use_wetting_drying', config_use_wetting_drying)
       call mpas_pool_get_config(ocnConfigs, 'config_thickness_flux_type', config_thickness_flux_type)
 
@@ -261,6 +265,8 @@ contains
 
       call mpas_pool_get_array(diagnosticsPool, 'surfaceFluxAttenuationCoefficient', surfaceFluxAttenuationCoefficient)
       call mpas_pool_get_array(diagnosticsPool, 'surfaceFluxAttenuationCoefficientRunoff', surfaceFluxAttenuationCoefficientRunoff)
+      call mpas_pool_get_array(diagnosticsPool, 'surfaceFluxAttenuationCoefficientPrecEvapIce',  &
+           surfaceFluxAttenuationCoefficientPrecEvapIce)
       !
       ! Compute height on cell edges at velocity locations
       !   Namelist options control the order of accuracy of the reconstructed layerThicknessEdge value
@@ -852,6 +858,7 @@ contains
       do iCell = 1, nCells
          surfaceFluxAttenuationCoefficient(iCell) = config_flux_attenuation_coefficient
          surfaceFluxAttenuationCoefficientRunoff(iCell) = config_flux_attenuation_coefficient_runoff
+         surfaceFluxAttenuationCoefficientPrecEvapIce(iCell) = config_flux_attenuation_coefficient_prec_evap_ice
       end do
       !$omp end do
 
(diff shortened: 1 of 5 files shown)


mark-petersen:
I rebased on `ocean/develop`, compiled, and this passes global ocean test cases. 

@maltrud please look at the `files changed` tab and see if these look like the right code changes.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1489,19 +1502,24 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
...
 
        nonLocalSurfaceTracerFlux(indexTempFlux, iCell) = activeTracersSurfaceFlux(indexTempFlux,iCell) &
-               + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell) - fracAbsorbed * &
-               surfaceThicknessFlux(iCell) * activeTracers(indexTempFlux,1,iCell) + &
-               activeTracersSurfaceFluxRunoff(indexTempFlux,iCell) * fracAbsorbedRunoff
+               + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell)  &
+               - fracAbsorbedPrecEvapIce * (rainFlux(iCell) + evaporationFlux(iCell)) * &
+                   activeTracers(indexTempFlux,1,iCell)/rho_sw &
+               - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell) * (  &
+                 + min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)    )/rho_sw
```
xylar: Shouldn't freshwater fluxes from land ice also be included here?

xylar: I guess sea-ice fluxes aren't included so maybe not, but what's special about rain and evap?

xylar:
An extra line needs to be added here:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_diagnostics.F#L1832
to modify `surfaceFluxAttenuationCoefficientPrecEvapIce` in the same way as `surfaceFluxAttenuationCoefficient` under land ice.  Or we would need to add yet another `surfaceFluxAttenuationCoefficientLandIce` and another set of loops...

src/core_ocean/shared/mpas_ocn_diagnostics.F
xylar: I'm pretty sure we want to include all these other `surfaceThicknessFlux` terms that this change removes:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F#L299-L300
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_surface_land_ice_fluxes.F#L269

maltrud:
@xylar , for the first question, the effect of land ice is already included in activeTracersSurfaceFlux(indexTempFlux,iCell) so doesn't need to be handled explicitly.  

for the second question, evap and precip are forced to have water entering/leaving at the local SST (instead of at 0C as it was before). we account for that added heat with an additional surface heat flux in activeTracersSurfaceFlux(indexTempFlux,iCell).  but this flux doesn't change buoyancy (only heat content) so we have to remove it from the buoyancy flux.

src/core_ocean/shared/mpas_ocn_diagnostics.F
xylar: @vanroekel?

xylar:
@maltrud, your comment refers to KPP?  Okay, I'll defer to you and desist on that.

xylar:
The issue remains that `surfaceFluxAttenuationCoefficientPrecEvapIce` needs to be modified under land ice in the same way that we modify `surfaceFluxAttenuationCoefficient`.  Otherwise, land ice freshwater fluxes will no longer be distributed as expected.

maltrud:
@xylar , for the third question, i chose not to touch land ice fluxes.  i don't think that what i've done would change the land ice flux treatment at all.  if you believe that it would be beneficial to treat land ice fluxes differently than they are now following a similar procedure, that's up to you.

maltrud:
@xylar ok i just saw your message about needing to modify stuff under land ice--i'll check that.

xylar:
>  i chose not to touch land ice fluxes.

Because `landIceFreshwaterFluxes` are added to `surfaceThicknessFluxes`, which were prerviously distributed with `surfaceFluxAttenuationCoefficient`, and now are distributed with `surfaceFluxAttenuationCoefficientPrecEvapIce`, I very much do expect land-ice fluxes to be affected.

src/core_ocean/shared/mpas_ocn_diagnostics.F
vanroekel: In looking at this a bit, I agree with @xylar here.  I think the problem is that @maltrud is talking about the thickness flux spreading, but Xylar raises an important point that there is still a bug in the non local flux for KPP, which is more appropriate to #305 .  Landice and seaice freshwater do get added to the heat flux and thickness flux such that the heat content doesn't change, but we need to correct for that additional flux to retain a net zero flux for KPP.  As the code is right now a seaice or landice freshwater flux will create an artificially strong surface buoyancy flux which would reduce mixing.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1489,19 +1502,24 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
        !
 
        ! Compute fraction of thickness flux that is in the top model layer
-       fracAbsorbed = 1.0_RKIND - exp( max(-layerThickness(1, iCell) / config_flux_attenuation_coefficient, -100.0_RKIND) )
        fracAbsorbedRunoff = 1.0_RKIND - exp( max(-layerThickness(1, iCell) / config_flux_attenuation_coefficient_runoff, &
                             -100.0_RKIND) )
+       fracAbsorbedPrecEvapIce = 1.0_RKIND - exp( max(-layerThickness(1, iCell) / &
+          config_flux_attenuation_coefficient_prec_evap_ice, -100.0_RKIND) )
+
```
vanroekel: I think we need a third fracAbsorbed for all other thickness fluxes that are not being spread, otherwise we will not have a balance of heat fluxes and the corresponding cooling from the layer expansion from the thickness flux.  Similar to my comment below.

maltrud:
yes, i was incorrect before.  instead of saying i chose not to modify land ice treatment, i should have said i chose not to _exempt_ land ice from being treated the same as other (non-runoff) thickness fluxes.  if you'd like to add another attenuation scale for land ice fluxes, go ahead and do that.  this PR is for water cycle, so ice cavities won't be enabled.  for testing with ice cavities enabled, one can set the attenuation scale back to 0.001 to get the previous behaviour.

xylar:
@maltrud, if you look carefully at the line I posted before, which I will share again:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_diagnostics.F#L1832
we currently modify `surfaceFluxAttenuationCoefficient` to have a different value under land ice than elsewhere.  The current code (before this PR) thus distributes land ice fluxes (including thickness fluxes) over 10 m even if they are distributed over 0.001 m elsewhere.

With this PR, that functionality is broken unless the same treatment is added for the new `surfaceFluxAttenuationCoefficientPrecEvapIce`.  I really must insist that some careful testing be done to make sure this doesn't break ice-shelf cases before this can go in, regardless of whether it is intended for water cycle runs.  It isn't acceptable to me to break the cryosphere configuration in haste to get new features in for water cycle.

maltrud:
ok, there are too many issues to be solved before the freeze.  maybe this will get in later, maybe not.

xylar:
@maltrud, I'm sorry for my tone last night.  As @mark-petersen kindly pointed out to me, I probably shouldn't be responding to pull requests at 1 am.

It seems like @vanroekel was fine with the KPP changes made here (the remaining concerns about fluxes from land ice and sea ice seem unrelated to this PR).

The only remaining change I needed was the addition of this one line:
```Fortran
            surfaceFluxAttenuationCoefficientPrecEvapIce(iCell) = config_land_ice_flux_attenuation_coefficient
```
below the line:
```
            surfaceFluxAttenuationCoefficient(iCell) = config_land_ice_flux_attenuation_coefficient
```
at
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_diagnostics.F#L1832

I got frustrated because it didn't seem like I was making myself very clear on this and that the runs with ice shelves might get broken in the haste to get the PR in before the code freeze.  But the fix is simple enough that a quick nightly test could confirm that no harm is done once the above line is added.

If you wan to push for this to happen before the code freeze, I'd support that.  I'll do what I can to make sure ice-shelf runs aren't negatively affected.

mark-petersen:
It was still worth having a PR with the code and issues raised for later reference. No harm done.  Mat, don't worry about it now, but after you return we should discuss what is worth merging versus adding to a development list for later.

@maltrud, could you please comment here on what simulations you ran with the thickness flux spreading, what parameter values you used, and what differences you saw?  Were there improvements that would indicate that thickness flux spreading should be on by default in V2 simulations?

maltrud:
@mark-petersen i did simulations with an attenuation scale 0.001, 10., and 50. meters.  there were significant improvements in the 50m case, but i'm no longer fully confident in the results due to possible energy conservation issues.  they're likely quite small errors, but i need to fix that before definitively saying it should be implemented in v2, especially since it won't be ready by july 15.

PR #328, opened 2019-07-19
MPASA - Discard soundings that are not within a Limited-Area domain

MiCurry:
This PR contains changes and a bug fix for src/diagnostics/soundings.F soundings_setup routine.

First, it implements a KD-Tree to fix a bug that would place soundings in the incorrect grid cell if within a concave decomposition. Using the KD-Tree the correct cell will be located even within concave decomposition.

Next, it adds logic to discard specified soundings that are outside of regional domains. Using the in_cell(...) function, to determine if the station is actually within the closet cell (found by the KD-Tree).

@mgduda, this PR is still missing its last commit (The MPI communication about owned and not owned stations). I noticed there are a few functions for calculating sphere_angle (As well as arc_length). There is a sphere_angle within [src/operators/mpas_geometry_utils.F](https://github.com/MPAS-Dev/MPAS-Model/blob/51d5624709f75182130a2fd94715e1bb13310172/src/operators/mpas_geometry_utils.F#L21-L71). Should I be using that one? Or copy it into the file, like I have done here.

### DIFF
--- a/src/core_atmosphere/diagnostics/Makefile
+++ b/src/core_atmosphere/diagnostics/Makefile
@@ -9,6 +9,7 @@ DIAGNOSTIC_MODULES = \
 	convective_diagnostics.o \
 	pv_diagnostics.o \
 	soundings.o \
+	mpas_kd_tree.o
 
 isobaric_diagnostics.o: mpas_atm_diagnostics_utils.o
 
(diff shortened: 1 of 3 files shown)


MiCurry:
@mgduda, I just pushed new changes that use `mpas_sphere_angle` that is within in operators/mpas_geometry_utils. I figured that its better to use that function rather then copy and paste a new one. I'm guessing that is also more thoroughly tested.

MiCurry:
@mgduda do you also have any good examples within MPAS for MPI communications? I've tried to do a bit of searching, but I haven't found anything yet!

PR #330, opened 2019-07-23
SST input and output interval check

MiCurry:
This commit checks to see that `config_fg_interval` in the
namelist.init_atmosphere and the output interval for the `surface` stream are
the same for init_case 8.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -74,7 +74,7 @@ module init_atm_cases
       character(len=StrKIND), pointer :: xtime
 
       type (MPAS_Time_type)  :: curr_time, stop_time
-      type (MPAS_TimeInterval_type)  :: clock_interval, lbc_stream_interval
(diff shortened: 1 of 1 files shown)


mgduda:
@MiCurry Can you modify the calls to `mpas_log_write` so that the `messageType` arguments are lined up? I think this might make the code a bit cleaner looking:
```fortran
            call mpas_log_write('****************************************************************',       messageType=MPAS_LOG_ERR)
            call mpas_log_write('The intermediate SST file interval specified by ''config_fg_interval''', messageType=MPAS_LOG_ERR)
            call mpas_log_write('does not match the output_interval for the ''surface'' stream.',         messageType=MPAS_LOG_ERR)
            call mpas_log_write('Please correct the namelist.init_atmosphere and/or',                     messageType=MPAS_LOG_ERR)
            call mpas_log_write('streams.init_atmosphere files.',                                         messageType=MPAS_LOG_ERR)
            call mpas_log_write('****************************************************************',       messageType=MPAS_LOG_CRIT)
```

MiCurry:
@mgduda Sorry about the delay on this one. I just modified the `mpas_log_write` calls as you requested!

PR #334, opened 2019-08-08
 Update COMPASS ocean README with new geometric_features

xylar:
Updated to v.0.1.3


xylar:
@mark-petersen, the compass environments on IC and blues are now updated with these changes.  (There wasn't on on cori, as I thought.)

PR #339, opened 2019-08-15
Add GM bolus eddy stats

lconlon:
Added monthly mean zonal and meridional bolus velocities as well as bolus velocity times temperature to eddy stats for Labrador Sea bias/ GM testing
bolus_velocities


vanroekel:
@lconlon thanks the code looks good.  I've also successfully tested in e3sm and get reasonable output.  A few more requests for testing.  Could you try build with gfortran and ifort including with debug?  I've already verified this is BFB with master.

lconlon:
@vanroekel Thanks- builds successfully with debug on with both gfortran and ifort

PR #341, opened 2019-08-19
COMPASS ocean nightly regression: Add GM and cleanup

mark-petersen:
Previously, GM was not tested in the nightly regression. This turns on GM for global cases, including the new depth varying piece. 

Also, I moved init flags to the end of the default namelist, and changed the suggested forward and init namelist to the full version (namelist.ocean). Now test cases no longer spew the warning
```
*** Encountered an issue while attempting to read namelist record ...
```
into the log file for every single missing namelist record.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -3720,9 +3720,8 @@
 			/>
 	</var_struct>
 
-<!-- Include init modes registry file -->
-#include "mode_init/Registry.xml"
-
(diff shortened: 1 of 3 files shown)


mark-petersen:
@vanroekel, if I were quicker I would have added this to your PR #288. I realized today we need to add the flags to COMPASS to test GM for BFB and performance. The other stuff is to just make the output files more readable.

I tested the nightly regression suite on this PR and everything works. I think a visual inspection is sufficient here.

PR #345, opened 2019-08-26
add mpas_dmpar_bcast_real4 and mpas_dmpar_bcast_real4s

ldfowler58:
In mpas_dmpar.F, added the subroutines mpas_dmpar_bcast_real4 and mpas_dmpar_bcast_real4s. These subroutines are used to read input binary data in the Thompson cloud microphysics scheme with the aerosol aware option.


mgduda:
I think we can avoid adding two new routines by changing the type of the arguments to `mpas_dmpar_bcast_real` and `mpas_dmpar_bcast_reals` from `RKIND` to `R4KIND`, but I need to do some investigation first to see whether compilers will consider `double precision` and `real (kind=R8KIND)` to be the same.

PR #354, opened 2019-09-03
Fixes typo for depth-variable Rayleigh drag

pwolfram:
Previously, Registry and code was inconsistent with the variable
`config_Rayleigh_damping_coeff_depth_variable` used, which was
inconsistent with the registry value of `config_Rayleigh_damping_depth_variable`.


pwolfram:
This is for consistency with https://github.com/pwolfram/MPAS-Model/blob/coastal/fix_Rayleigh_damping_depth_variable_typo/src/core_ocean/Registry.xml#L471

PR #357, opened 2019-09-04
Do not apply Redi cross-terms to bottom layer

mark-petersen:
comments here

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
@@ -323,7 +323,7 @@ contains
                invAreaCell = 1.0_RKIND / areaCell(iCell)
                do i = 1, nEdgesOnCell(iCell)
                   iEdge = edgesOnCell(i, iCell)
-                  do k = 1, maxLevelEdgeTop(iEdge)
+                  do k = 1, maxLevelEdgeTop(iEdge)-1
                      s_tmpU = relativeSlopeTapering(k, iEdge) * relativeSlopeTopOfEdge(k, iEdge) * dTracerdZTopOfEdge(k, iEdge)
                      s_tmpD = relativeSlopeTapering(k+1, iEdge) * relativeSlopeTopOfEdge(k+1, iEdge) &
                             * dTracerdZTopOfEdge(k+1, iEdge)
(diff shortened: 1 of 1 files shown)


mark-petersen:
I posted this problem as an E3SM issue: https://github.com/E3SM-Project/E3SM/issues/3172

Using a debug tracer of 1 in layers 1-10 and 0 elsewhere, the tracer exceeds the min/max bounds to 1.2 and -0.2. So this fix is not perfect, but will allow us to test Redi with minimal changes.

xylar:
> the tracer exceeds the min/max bounds to 1.2 and -0.2.

That doesn't sound great.  In POP when I used 2nd-order rather than monotonic advection, I've seen cases where temperature going outside physical bounds causes pretty bad melt rates in ice-shelf cavities.  This is partly why I feel like a test with ice-shelf cavities is quite important with this change.

mark-petersen:
For some background, the Redi mixing term is:
image

(see the [original GM design document](https://github.com/MPAS-Dev/MPAS-Documents/tree/master/ocean/gm)) where:
- K^r is the 2x2 tensor
- kappa_R is `config_Redi_kappa`
- h is layer thickness
- S is isopycnal slope vector
- K33 = S^2, a scalar
- phi is the tracer

I call the terms on the right 1-4, where 
- term 1 is strictly a horizontal (del2) operator - strictly speaking, along coordinate surface r.
- terms 2 and 3 are the cross-terms (horizontal/vertical)
- term 4 is the vertical mixing term.

mark-petersen:
>That doesn't sound great.

@xylar, I agree that deviating outside global bounds is worrisome, and will require careful testing. That said, I've been stuck on this problem for six months because of poor results in unrealistic, contrived conditions. The simple example of a step function from 0 to 1 is very problematic because term 2 has very high dphi/dz, and then fluxes tracers in the horizontal. A realistic tracer would almost never step from a global min to global max from one cell to the next. But in some situations it could, like iron fertilization from surface deposition - we need to think through those cases.

Anyway, I've decided to proceed with testing despite these concerns.

xylar:
By all means, proceed with testing.  But I think BGC cases, which require accurate monotonic advection, are precisely some of the cases where we hope to use Redi.  So the fact that monotonicity is lost with Redi might be a show stopper until we can get a better handle on why.  "Step functions" (or at least steep gradients) aren't as rare in the ocean as one might like to think.  The sub-ice-sshelf boundary layer has strong gradients in T and S, and overshooting like here can mean the difference between melting (with positive buoyancy) and freezing (with negative buoyancy), with qualitatively different resulting dynamics (e.g. stable vs. unstable stratification).

I'm advocating proceeding with caution.

mark-petersen:
Meeting notes for reference:
https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/1030947002/2019-09-05+Testing+Redi+Mixing+note+1+30pm+time

PR #358, opened 2019-09-04
Adds drying safety factor needed for real-world cases

pwolfram:
Adds a safety factor to ensure a layer doesn't drying within a specified neighborhood of the minimum thickness.


pwolfram:
@mark-petersen, this is a safety factor that was identified in the Sandy runs as needed to prevent artificial drying of a cell.  I'd like to merge this in before we merge in the full Sandy case.  I'm currently testing on the drying slope to prove it doesn't do anything deleterious to the wetting and drying behavior.

pwolfram:
Thanks @mark-petersen!  The suggestion is a great one and I've pushed the change.

pwolfram:
# Testing
Results are as expected using the drying slope case with 1km sigma coordinate:
dryingslopecomparison

pwolfram:
Thanks @mark-petersen!

PR #359, opened 2019-09-05
Adds wetting and drying Gaussian hump test cases

pwolfram:
Adds a test case for Gaussian hump idealized "hurricane" for real-world bathymetry to demonstrate stability of wetting and drying with flooding on land in a more realistic real-world case than the drying slope test case.

### DIFF
--- a/src/core_ocean/mode_init/Registry_hurricane.xml
+++ b/src/core_ocean/mode_init/Registry_hurricane.xml
@@ -1,14 +1,42 @@
 	<nml_record name="hurricane" mode="init" configuration="hurricane">
-		<nml_option name="config_hurricane_vert_levels" type="integer" default_value="3" units="unitless"
+		<nml_option name="config_hurricane_vert_levels" type="integer" default_value="3" units="m"
 			description="Number of vertical levels for hurricane."
 			possible_values="Any positive integer number greater than 2."
 		/>
-		<nml_option name="config_hurricane_min_depth" type="real" default_value="10.0" units="meters"
+		<nml_option name="config_hurricane_min_depth" type="real" default_value="10.0" units="m"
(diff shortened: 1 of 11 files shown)


pwolfram:
# Testing with Gaussian hump with amplitude of 4m:
## Initial at t=0 hrs:
image
image

## At 4 hrs:

image
image

pwolfram:
image

Note, bathymetry and vertical coordinate system is shifted to avoid instability with `bottomDepth < 0`, as discussed earlier @mark-petersen.

pwolfram:
@sbrus89, can you please test and verify that this works for you?  I'm also running your Sandy test case to ensure that it works as expected following these changes.

pwolfram:
@mark-petersen, we will verify that Sandy case is intact and then merge.  Thanks!

pwolfram:
@sbrus89, here is the result from the sandy case, which is stable, but doesn't appear to be as good as some of the recent runs we've done based on plots we've made.  All plots are out of QuickViz on the pointStats.  



Are the parameters in the namelist.ocean file in the repo the best ones we have?  I don't think they are because `config_use_const_visc=.false.`, for instance.  

Thus, because this is stable I'm inclined to think it is working and we should have a separate PR to deal with tuning the Sandy case.

Please let me know what you think but I'm inclined to think this is ready to merge and you should do a separate PR to submit the tuned results that we anticipate to be better for Sandy.

pwolfram:
Using `ocean/coastal' yield the following result, which is qualitatively equivalent to the above (noting I've removed the superflous modification of Bahama bathymetry so results aren't expected to be BFB.



Given these results I'll merge this PR and we can fine tune the existing Sandy simulation in a follow-up PR (e.g., using the depth-averaging approach that produces a better match against the data).

PR #361, opened 2019-09-05
Fixes to mpas_io

### LINKED ISSUE #352: Nested string in src/framework/mpas_io.F breaks NEC compiler
I believe these lines in `src/framework/mpas_io.F` are legal w.r.t. the Fortran standard:
```4607          call mpas_log_write('Attribute ''' // trim(attName) // ''' has a value longer than StrKIND. ' // &
4608              'It will be cut to a length of $i', MPAS_LOG_WARN, intArgs=(/StrKIND/) )```
The NEC 2.3.1 compiler is reporting it as a failure in the line continuation.
I suggested that the NEC people fix their compiler, but on the other hand, it's more readable when you nest different kinds of quotes
```4599    call mpas_log_write("Attribute '" // trim(attName) // "' has a value longer than StrKIND. " // &
4600            "It will be cut to a length of $i", MPAS_LOG_WARN, intArgs=(/StrKIND/) )```
and the compiler takes this input without any problem.
Would you consider changing the code, just to make it more generally compatible?
The above source snippet is from MPAS-A v7.0.

matthewhoffman:
This merge includes some small fixes to the mpas_io module.  

The primary one modifies the reading of char data types to ignore null characters.  That behavior already exists for char_1 types, but it was not previously included for char types.  Without this fix, any null characters that exist at the end of a char data type get read into the MPAS variable.  In addition to looking ugly, that behavior can lead to unexpected results, notably errors in parsing xtime.  That particular situation happens when a restart file written using PIO2 (which uses C under the hood and terminates strings with null characters) is read back into a build of MPAS using PIO1 (which is based on Fortran and does not expect null-terminated strings).

The second fix is a formatting change that fixes #352

### DIFF
--- a/src/framework/mpas_io.F
+++ b/src/framework/mpas_io.F
@@ -1384,6 +1384,7 @@ module mpas_io
       integer, dimension(6) :: start6
       integer, dimension(6) :: count6
       character (len=StrKIND), dimension(1) :: tempchar
+      integer :: charLen
       type (fieldlist_type), pointer :: field_cursor
 
       integer i, j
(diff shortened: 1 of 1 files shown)


matthewhoffman:
@mgduda , no rush on this PR.  This obscure change was prompted after a couple days of trying to understand timekeeping errors that suddenly appeared after switching from PIO2 to PIO1 in my MPAS build halfway through a run.  The problem would generally occur if someone using PIO1 was given a restart file generated from a PIO2-enabled MPAS run.

mark-petersen:
@matthewhoffman would it be best to merge these in before the framework merge for E3SM components next week? If so, perhaps @mgduda could review this little PR early next week?

matthewhoffman:
@mark-petersen , thanks for looking this over.  I don't think this is a critical bug fix.  That said, I think it will be relevant to any situation where you move from PIO2 to PIO1 between runs.  This is a pretty unusual workflow, but it seems possible it could be encountered if, e.g., a spinup was done on one machine using PIO2 and then, say, the result was scp'ed to a different machine using PIO1 for a forward simulation.

mgduda:
@matthewhoffman Could you explain a little more why the changes in PR #161 don't address any issues you were seeing with, e.g., the `xtime` variable? The changes from #161 were put in place specifically to address these issues, and from what I can recall (and looking a the [changes in that PR](https://github.com/MPAS-Dev/MPAS-Model/pull/161/files), it does appear that we handle both 0-d strings and 1-d strings).

matthewhoffman:
@mgduda , hmm, it looks like the problem was simply that I have not merged develop into landice/develop since last year, and therefore the changes from #161 were not present in the code I was using.  I had forgotten about #161, so I guess I reinvented the wheel.  I'm thinking the best course of action is to close this PR without merging, but let me know if you see any value in merging it anyway.

mgduda:
@matthewhoffman I think it would be fine to close this PR.

matthewhoffman:
This issue was addressed by #161 , so I'm closing it.

PR #365, opened 2019-09-10
Updates to hurricane test case

sbrus89:
Most important change is turning on constant vertical mixing and setting it equal to 1e4. This depth averages the water column, resulting in improved stability and a better validation results against water level gauge data from Hurricane Sandy.

The split-explicit time-step is also increased to 40/4 seconds from 30/3.


sbrus89:
Orange line is the previous results, green is the current results with this PR.

pwolfram:
@mark-petersen, note this is only for COMPASS.  Please let me know if you'd like to review.  I would like to get this merged today.  Thanks!

PR #367, opened 2019-09-13
Pool memory leak

### LINKED ISSUE #350: Memory leak with RK4 in hurricane simulation
This memory leak error was identified on the branch at https://github.com/pwolfram/MPAS-Model/tree/coastal/Delaware_gaussian_hump with hash ba4bbdab91052b9ff3474a0540436a0e55fed6e5 for the Sandy Hurricane case and produces an insufficient virtual memory error.
 
@hrmoncada, I have given you the tgz file with the example case on LANL IC, although you can build it from COMPASS for the `ocean/hurricane/USDEQU120at30cr10rr2WD/sandy/` case.  Modules and run instructions for IC are detailed below.
 
Error messages occur after some time (depending upon total number of time steps) and are of the form
```
forrtl: severe (41): insufficient virtual memory                                
ocean_model        0000000000E6FE10  for_alloc_allocat     Unknown  Unknown
ocean_model        0000000000749E66  Unknown               Unknown  Unknown
ocean_model        00000000006C577E  Unknown               Unknown  Unknown
ocean_model        00000000007382D0  Unknown               Unknown  Unknown
ocean_model        000000000094972C  Unknown               Unknown  Unknown
ocean_model        00000000006ABD6B  Unknown               Unknown  Unknown
ocean_model        0000000000699590  Unknown               Unknown  Unknown
ocean_model        000000000063EE31  Unknown               Unknown  Unknown
ocean_model        000000000040F277  Unknown               Unknown  Unknown
ocean_model        000000000040F1FE  Unknown               Unknown  Unknown
ocean_model        000000000040F1AE  Unknown               Unknown  Unknown
libc-2.17.so       00002ADC24D273D5  __libc_start_main     Unknown  Unknown
ocean_model        000000000040F0A9  Unknown               Unknown  Unknown
...

```

This has occurred for a variety of runs (shown below): 

dt | last stable time stamp |
-- | -- | 
10s | 2012-10-05_22:18:50 |
5s | 2012-10-03_11:09:05 |
25s | 2012-10-13_07:43:45 | 
25s | 2012-10-13_08:22:55 |
25s restart (from above) | 2012-10-25_08:30:25  |

This was experienced on 360 processors for 51536 cells, or ~143 cells / proc.

As you can see, this is a memory leak because the failure time depends on the time step and it is possible to advance the simulation via a restart.

I'm running with the following approach and modules:

```
#!/bin/bash
#SBATCH --account=XYZ
#SBATCH --job-name=XYZ
#SBATCH --qos=interactive
#SBATCH --time=4:00:00
#SBATCH --output=run.o%j
#SBATCH --nodes=10
#SBATCH --error=run.e%j
#SBATCH --mail-user=XYZ@com.com
#SBATCH --mail-type=all

# Modules for grizzly
source ~/module_source_gr.sh

gpmetis graph.info 360
srun -n 360 ./ocean_model -n namelist.ocean -s streams.ocean
```
where `module_source_gr.sh` is 
```
#!/bin/bash
module purge

echo "loading modules anaconda, intel, openmpi, netcdf, pnetcdf, pio for grizzly"
module load intel/17.0.1 openmpi/1.10.5 netcdf/4.4.1 parallel-netcdf/1.5.0 pio/1.7.2
conda activate compass_py3.7

alias metis=gpmetis
```

@hrmoncada, posting images of your output from valgrind will help us see where the memory leak is and that will help us trace it to the exact line of code where memory is allocated.

cc @mark-petersen and @sbrus89

mark-petersen:
The `destroy_pool` subroutine does not deallocate the memory associated with a field, because it uses a simple fortran `deallocate` rather than the more thorough `mpas_deallocate_field` routine. This has become a problem when using RK4 time-stepping in the ocean model, as we hit an out-of-memory error.

fixes #350

### DIFF
--- a/src/framework/mpas_field_routines.F
+++ b/src/framework/mpas_field_routines.F
@@ -1388,43 +1388,65 @@ module mpas_field_routines
 
 !***********************************************************************
 !
-!  routine mpas_deallocate_field0d_logical
+!  routine mpas_deallocate_field0D_logical
 !
-!> \brief   MPAS 0D logical deallocation routine.
-!> \author  Doug Jacobsen
+!> \brief   MPAS 0D int deallocation routine.
+!> \author  Doug Jacobsen, Michael G. Duda
 !> \date    04/02/13
 !> \details 
-!> This routine deallocates a 0D logical field.
+!>  This routine deallocates a 0D int field. If the optional argument
+!>  keep_head_ptr is .true., the field type pointed to by f will not be
+!>  deallocated -- only the contents (array, attLists, constituentNames)
+!>  will be deallocated. If the field contains multiple blocks, all memory
+!>  for all blocks other than the first (including the field types themselves)
+!>  will be deallocated.
+!>
+!>  The principal use fo rthe keep_head_ptr argument is for situations in
+!>  which elemements of an array of fields need to be deallocated.
 !
 !-----------------------------------------------------------------------
-   subroutine mpas_deallocate_field0d_logical(f)!{{{
-       type (field0dLogical), pointer :: f !< Input: Field to deallocate
-       type (field0dLogical), pointer :: f_cursor
+   subroutine mpas_deallocate_field0d_logical(f, keep_head_ptr)!{{{
+
+       implicit none
+
+       type (field0dLogical), target :: f !< Input: Field to deallocate
+       logical, intent(in), optional :: keep_head_ptr
+
+       type (field0dLogical), pointer :: f_cursor, f_next
+       logical :: local_keep_head_ptr
        integer :: threadNum
        integer :: i, iErr
 
        threadNum = mpas_threading_get_thread_num()
-   
-       if ( threadNum == 0 ) then
-          f_cursor => f
-   
-          do while(associated(f_cursor))
-            if(associated(f % next)) then
-              f => f % next
-            else
-              nullify(f)
-            end if
-   
-            if ( associated(f_cursor % attLists) ) then
-               do i = 1, size(f_cursor % attLists, dim=1)
-                  call mpas_deallocate_attlist(f_cursor % attLists(i) % attList, iErr)
-               end do
-               deallocate(f_cursor % attLists)
-            end if
 
-            deallocate(f_cursor)
-            f_cursor => f
-          end do
+       if (present(keep_head_ptr)) then
+           local_keep_head_ptr = keep_head_ptr
+       else
+           local_keep_head_ptr = .false.
+       end if
+
+       if ( threadNum == 0 ) then
+           f_cursor => f
+           do while(associated(f_cursor))
+               f_next => f_cursor % next
+
+               if (associated(f_cursor % attLists)) then
+                   do i = 1, size(f_cursor % attLists, dim=1)
+                       call mpas_deallocate_attlist(f_cursor % attLists(i) % attList, iErr)
+                   end do
+                   deallocate(f_cursor % attLists)
+               end if
+
+               if (associated(f_cursor % constituentNames)) then
+                   deallocate(f_cursor % constituentNames)
+               end if
+
+               if ((.not. local_keep_head_ptr) .or. (.not. associated(f_cursor, f))) then
+                   deallocate(f_cursor)
+               end if
+
+               f_cursor => f_next
+           end do
        end if
 
    end subroutine mpas_deallocate_field0d_logical!}}}
@@ -1432,43 +1454,65 @@ module mpas_field_routines
 
 !***********************************************************************
 !
-!  routine mpas_deallocate_field0d_integer
+!  routine mpas_deallocate_field0D_integer
 !
-!> \brief   MPAS 0D integer deallocation routine.
-!> \author  Doug Jacobsen
+!> \brief   MPAS 0D int deallocation routine.
+!> \author  Doug Jacobsen, Michael G. Duda
 !> \date    04/02/13
 !> \details 
-!> This routine deallocates a 0D integer field.
+!>  This routine deallocates a 0D int field. If the optional argument
+!>  keep_head_ptr is .true., the field type pointed to by f will not be
+!>  deallocated -- only the contents (array, attLists, constituentNames)
+!>  will be deallocated. If the field contains multiple blocks, all memory
+!>  for all blocks other than the first (including the field types themselves)
+!>  will be deallocated.
+!>
+!>  The principal use fo rthe keep_head_ptr argument is for situations in
+!>  which elemements of an array of fields need to be deallocated.
 !
 !-----------------------------------------------------------------------
-   subroutine mpas_deallocate_field0d_integer(f)!{{{
-       type (field0dInteger), pointer :: f !< Input: Field to deallocate
-       type (field0dInteger), pointer :: f_cursor
+   subroutine mpas_deallocate_field0d_integer(f, keep_head_ptr)!{{{
+
+       implicit none
+
+       type (field0dInteger), target :: f !< Input: Field to deallocate
+       logical, intent(in), optional :: keep_head_ptr
+
+       type (field0dInteger), pointer :: f_cursor, f_next
+       logical :: local_keep_head_ptr
        integer :: threadNum
        integer :: i, iErr
 
        threadNum = mpas_threading_get_thread_num()
-   
-       if ( threadNum == 0 ) then
-          f_cursor => f
-   
-          do while(associated(f_cursor))
-            if(associated(f % next)) then
-              f => f % next
-            else
-              nullify(f)
-            end if
-   
-            if ( associated(f_cursor % attLists) ) then
-               do i = 1, size(f_cursor % attLists, dim=1)
-                  call mpas_deallocate_attlist(f_cursor % attLists(i) % attList, iErr)
-               end do
-               deallocate(f_cursor % attLists)
-            end if
 
-            deallocate(f_cursor)
-            f_cursor => f
-          end do
+       if (present(keep_head_ptr)) then
+           local_keep_head_ptr = keep_head_ptr
+       else
+           local_keep_head_ptr = .false.
+       end if
+
+       if ( threadNum == 0 ) then
+           f_cursor => f
+           do while(associated(f_cursor))
+               f_next => f_cursor % next
+
+               if (associated(f_cursor % attLists)) then
+                   do i = 1, size(f_cursor % attLists, dim=1)
+                       call mpas_deallocate_attlist(f_cursor % attLists(i) % attList, iErr)
+                   end do
+                   deallocate(f_cursor % attLists)
+               end if
+
+               if (associated(f_cursor % constituentNames)) then
+                   deallocate(f_cursor % constituentNames)
+               end if
+
+               if ((.not. local_keep_head_ptr) .or. (.not. associated(f_cursor, f))) then
+                   deallocate(f_cursor)
+               end if
+
+               f_cursor => f_next
+           end do
        end if
 
    end subroutine mpas_deallocate_field0d_integer!}}}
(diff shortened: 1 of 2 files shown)


mark-petersen:
@pwolfram, please cherry-pick 1670ba45 and test your hurricane case. In the simple test case you provided here: https://github.com/MPAS-Dev/MPAS-Model/issues/350#issuecomment-530976119 it previously ran 26k iterations before a memory error. With this fix it runs 79k iterations, so a factor of three on that test. Obviously, there is still a leak, but this might get you through a queue cycle. I spent several more hours replacing other deallocate statements, but others would cause an error.  I also looked for other memory leaks without luck.

mark-petersen:
@mgduda this was a minor memory leak that came up last March that we didn't fix. It is now stopping a set of simulations, so is a priority.  You give a potential solution here: https://github.com/MPAS-Dev/MPAS-Model/pull/185#issuecomment-473466314 which I implemented in this PR, but it only partially fixes the memory leak. A few days after that, here: https://github.com/MPAS-Dev/MPAS-Model/pull/185#issuecomment-480415072 you said it needs a different solution than you previously thought.

I've been testing alterations "experimentally", but I honestly don't understand the pool clone and destroy routines very well. Any help or advice you could provide would be greatly appreciated.

mark-petersen:
There are other deallocate lines, like this set:
```
 203    recursive subroutine mpas_pool_destroy_pool(inPool)!{{{
...
 253                   else if (associated(dptr % r1)) then
 254                      if (associated(dptr % r1 % array)) then
 255                         deallocate(dptr % r1 % array, stat=local_err)
 256                      end if
 258                      deallocate(dptr % r1, stat=local_err)
```
and I tried to alter it to this:
```
 253                   else if (associated(dptr % r1)) then
 254                      if (associated(dptr % r1 % array)) then
 255                         nullify(dptr % r1 % array)
 256                      end if
 257                      call mpas_deallocate_field(dptr % r1)
```
I made that change to all data types in `mpas_pool_destroy_pool` and it compiles. But it has an immediate run-time error, trying to access out-of-bounds memory, so that didn't work.  

A second set just after that looks like this:
```
 306                   else if (associated(dptr % r1a)) then
 307                      do j=1,dptr % contentsTimeLevs
 308                         if (associated(dptr % r1a(j) % array)) then
 309                            deallocate(dptr % r1a(j) % array)
 310                         end if
 311                      end do
 312                      deallocate(dptr % r1a)
```
that is also a candidate for a memory leak due to the extra field information that remains. But the alteration to:
```
 312                      call mpas_deallocate_field(dptr % r1a)
```
gets a compiler error, presumably because `r1a` is an array of arrays, and `mpas_deallocate_field` does not have a subroutine for that type.

pwolfram:
@mark-petersen, thanks for the help.  I've cherry-picked the commit and am running the job now and can report back later about its status.

mgduda:
@mark-petersen I should be able to look more closely at this by the end of the week. From the discussion on PR #185 I think this is my point of concern regarding the approach taken at present:

> Since we do a shallow copy of the source fields to the destination fields in a pool clone, if we ever access the attLists or constituentNames of members of the clone, we'd run into problems by calling mpas_deallocate_field in the mpas_pool_clone_pool routine.

mark-petersen:
@mgduda thanks for looking. The changes I made in this PR only extended the simulations by @pwolfram by ~20%, and it still then dies with an 'out of memory' error.  I don't know how to completely purge the memory allocated in the mpas_pool_clone_pool routine, so any help you can provide is greatly appreciated.

mark-petersen:
@mgduda any thoughts on this? Any help you can provide would be greatly appreciated.

mark-petersen:
@mgduda, we are stuck on this, and it's limiting for some of our simulations. If you have any suggestions, that would be extremely helpful. I don't know how to find further memory leaks by simply looking for missing deallocates or nullifies.

pwolfram:
@mgduda, can you please provide some advice on this issue when you get a chance?  I'm sure you are very busy but if you don't have any time to engage at all, can you please just let us all know so we can move forward?  Thanks!

xylar:
@pwolfram and @mgduda, it might be worth having an MPAS developer telecon about this one (and whatever other pressing MPAS issues).

pwolfram:
@xylar, do you know when the next one is scheduled?

xylar:
@pwolfram, they aren't scheduled anymore.  You have to ask for one as topics come up that need to be discussed, like this one.  If posting here doesn't do the trick, maybe a quick email to the core MPAS team would do it.

mgduda:
@pwolfram @mark-petersen @xylar Sorry, all, for not taking a closer look long ago! I'll start in this afternoon and let you all know what I'm able to find by tomorrow afternoon at the latest.

mark-petersen:
@mgduda thanks. Let us know if you need a test case to exemplify the behavior.

mgduda:
@mark-petersen @pwolfram I have a tentative set of changes that I think will work. I've pushed these to a branch named 'framework/destroy_pool_mem_leak' in my fork of the repository. Would you all be willing to try merging this branch (which is based on the v7.0 tag) to your working branch to see if it addresses the memory leaks you're seeing? My approach was slightly different than what was done in this PR, so I created a separate branch.

sbrus89:
@mgduda, Thanks for taking a look at this! I'm testing your branch now on a test case that previously failed due to this issue. Just for my own future reference, the simulation began at 2012-10-10_00:00:00 and crashed due to insufficient virtual memory at 2012-10-21_22:13:45. I'll report back on whether or not the changes allow the run to complete.

cc @pwolfram, @mark-petersen

pwolfram:
@mgduda, thank you *so VERY* **MUCH**!  Am testing it now and it absolutely is solid under my test case for a simplified case  at `pwolfram/coastal/memory_leak_test` using `compass/ocean/hurricane/USDEQU120at30cr10rr2/test_memory`.  This was really easy to get it to crash before.  

I'll defer to @sbrus89 for the real-world case next to confirm but strongly expect to not see an issue.  We can let you know.

cc @mark-petersen

pwolfram:
After discussion with @mark-petersen and just to document -- it ran on the order of 130,000 clone / destroy pool cycles and checking out the memory footprint with `htop` it was rock solid with no apparent long term average memory creep. 

@mark-petersen will test more broadly tomorrow.  Note the functionality this affects is probably limited only to ocean RK4 time integration and the shallow water pool because the issue was with the `mpas_destroy_pool` routine if I understand correctly.

sbrus89:
My run with @mgduda's changes went from 2012-10-10_00:00:00 to  2012-10-31_22:02:55 (when it timed out after reaching the 4hr interactive queue time limit) as compared the previous memory leak fail at 2012-10-21_22:13:45. This was only a little over 2 simulation days short of the stop time for the test case. Based on this results and @pwolfram and @mark-petersen's testing, I would suspect that these changes solve the problem. I'll go ahead and run on standard to make sure the run completes, but I don't anticipate any issues.

mark-petersen:
I rebased and added the three commits by @mgduda.  

@mgduda, was it your intention for me to keep my commit, b1d7384, or does your work replace that?

src/framework/mpas_field_routines.F
```diff
@@ -1716,44 +1869,66 @@ end subroutine mpas_deallocate_field1d_real!}}}
...
+                   deallocate(f_cursor % attLists)
+               end if
+
+               if (associated(f_cursor % constituentNames)) then
+                   deallocate(f_cursor % constituentNames)
+               end if
 
-            deallocate(f_cursor)
+               if ((.not. local_keep_head_ptr) .or. (.not. associated(f_cursor, f))) then
+                   deallocate(f_cursor)
```
mark-petersen: @mgduda I did a local test merge into ocean/develop. I get a seg fault here when I compile with `intel/17.0.1` in debug mode, also with gnu, but only if I run multiple blocks per core. That is, if the flag `config_number_of_blocks` is set to a different number from the `mpirun -n` number of cores.  For example,
```
config_number_of_blocks=8
mpirun -n 4 ./ocean_model
```
causes a seg fault, but 
```
config_number_of_blocks=8
mpirun -n 8 ./ocean_model
```
runs fine. Using `config_number_of_blocks=0` is also fine.

mgduda:
@mark-petersen I think the changes in my branch can completely replace the changes in your pool_memory_leak branch. There are just a couple of points about my changes that might be worth discussing on Monday.

mgduda:
@mark-petersen I can try to track down the segfault with multiple blocks per MPI task. We can't really run with anything other than 1 block per task with MPAS-Atmosphere, but if you could provide a simple test case for MPAS-Ocean, I imagine it shouldn't be too much work to address the issue with multiple blocks.

mark-petersen:
@mgduda I removed my previous commit, and get the identical error. You can use this test:
https://drive.google.com/open?id=1Fc9kVl_MTEeVNANXAzrCn72BaxJIGFrG
you only need to add the executable, then use `run.py` or `mpirun -n 4 ocean_model`.  This was run with this PR branch merged into the `ocean/develop` branch.

It runs through the whole test, and only dies on the finalize routines. Here is the sequence, from bottom to top, with lines interleaved:
```
forrtl: severe (174): SIGSEGV, segmentation fault occurred
...
ocean_model        0000000002E77E2D  mpas_field_routin        1927  mpas_field_routines.F
1927                    deallocate(f_cursor)

ocean_model        0000000002ECD11B  mpas_pool_routine         289  mpas_pool_routines.F
 289                         call mpas_deallocate_field(dptr % r2a(j), keep_head_ptr=.true.)

ocean_model        0000000002ED043A  mpas_pool_routine         348  mpas_pool_routines.F
 348                   call mpas_pool_destroy_pool(ptr % data % p)

ocean_model        0000000002E66C6B  mpas_domain_routi         167  mpas_domain_routines.F
167       call mpas_pool_destroy_pool(b % structs)

ocean_model        0000000002E669CB  mpas_domain_routi         126  mpas_domain_routines.F
124       block_ptr => dom % blocklist
125       do while (associated(block_ptr))
126          call mpas_deallocate_block(block_ptr)
127          block_ptr => block_ptr % next
128       end do

ocean_model        0000000002B0A4CD  mpas_framework_mp         143  mpas_framework.F
  143       call mpas_deallocate_domain(domain)

ocean_model        0000000000415BA3  mpas_subdriver_mp         391  mpas_subdriver.F
391       call mpas_framework_finalize(domain_ptr % dminfo, domain_ptr)

ocean_model        0000000000410249  MAIN__                     18  mpas.F
```
It appears to be caused by `mpas_deallocate_block` getting called twice by one processor.

Thanks for you help on this, we really appreciate it!

mgduda:
@mark-petersen I just pushed three new commits to the `framework/destroy_pool_mem_leak` branch in my fork. I think these changes should address the deallocation errors you're seeing with multiple blocks. If you do notice any other odd behavior with the recent changes, just let me know and I'll continue to investigate. The commit messages will hopefully explain what I've done, but if there are any questions at all I'll be glad to clarify!

mark-petersen:
@mgduda I added your three new commits, and they worked great! It solved the problem we saw before on finalize. Thank you!

@sbrus89 or @pwolfram if either of you have time to test these new commits on your previous case, that would be great. 

I would like to merge this into `develop` soon. @mgduda please approve if you are happy with this final version. @matthewhoffman, do you want to review?

mgduda:
@mark-petersen The `framework/destroy_pool_mem_leak` branch in my fork is based off of the `v7.0` tag, while the changes in your `pool_mem_leak` branch are based off a recent `develop` branch. Ideally, we'd base the changes on `v7.0`, which is an ancestor of the current `develop` so that anyone working off of the MPAS v7.0 release code could pick up these fixes without introducing unrelated code changes from `develop`. Would you all be open to the idea of force-pushing `framework/destroy_pool_mem_leak` onto `pool_mem_leak` before we merge this PR?

mark-petersen:
@mgduda no problem. I did that and retested, and it all works.

src/framework/mpas_field_routines.F
```diff
@@ -1860,139 +1707,961 @@ end subroutine mpas_deallocate_field4d_real!}}}
...
+               end if
+
+               if (associated(f_cursor % constituentNames)) then
+                   deallocate(f_cursor % constituentNames)
+               end if
 
-            deallocate(f_cursor)
+               if (.not. local_keep_head_ptr) then
+                   deallocate(f_cursor)
+               end if
```
amametjanov: Please note that IBM compiler (`xlf2008_r v.16.1.1-3`) stops on this `deallocate` with run-time error:
```
1525-109 Error encountered while attempting to deallocate a data object.  The program will stop.
```
After re-running with additional `STAT` and `ERRMSG` arguments, deallocate returns status value of `2: An invalid data object has been specified for the DEALLOCATE statement.`. It appears that the standard wants the object to be either an allocatable array or an object with `target` attribute. That's not the case here with `f_cursor` declared with `pointer` attribute.

Grepping for calls to this subroutine, all callers either 
1. call with `keep_head_ptr=.true.` and deallocate the object immediately after the call, or
2. call with `keep_head_ptr=.false.` and nullify the object right after the call.

If I comment out the `if` statement, the run proceeds with no errors. It appears that all calls need to pass in `keep_head_ptr=.true.` (or just remove this logic and the corresponding call to `deallocate` the head pointer). Would that be OK, @mgduda ?

mgduda: I'll take a closer look. I'm hesitant to remove a `DEALLOCATE` statement without understanding whether this might re-introduce a memory leak under certain circumstances. From what I recall of the 2003 standard, objects with the `POINTER` attribute implicitly have the `TARGET` attribute.

amametjanov: Here is a branch that lets me build and run a sea-ice case: [diff](https://github.com/MPAS-Dev/MPAS-Model/compare/develop...amametjanov:framework/nullify-field-pointers). I can try valgrind on the case that had mem-leaks. How to configure one?

When I try `SMS.T62_oQU120.CMPASO-NYF` with `config_time_integrator = 'RK4'` the case runs into `SIGFPE: erroneous arithmetic operation`. Does an RK4 case run in E3SM?

mgduda: Sorry for not following-up on this issue for so long. I just encountered the same issue on Summit, and I'll see what I can find.

mgduda: I just opened Issue #481 to track progress on this.

PR #368, opened 2019-09-17
Remove unused variables hEddyFlux,hKappa,hKappaQ

mark-petersen:
These three 3D variables are in the Registry, and one is obtained, but none are used. Git blame shows they were added by Todd and Doug in 2013, but then Qingshan Chen added his own naming scheme for GM.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -2565,19 +2565,6 @@
 			 description="kinetic energy of horizonal velocity on cells"
 			 packages="forwardMode;analysisMode"
 		/>
-		<var name="hEddyFlux" type="real" dimensions="nVertLevels nEdges Time" units=""
-			 description="Eddy flux in Gent-McWilliams eddy parameterization"
-			 packages="forwardMode;analysisMode"
-		/>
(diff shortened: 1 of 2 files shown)


mark-petersen:
This is part of a larger effort to reduce the memory footprint of MPAS-Ocean, as we have egregious overuse of memory due to lack of care (myself included). This is an obvious one. 

I started looking at memory use when an 18to6 run with LIGHT had 'out of memory' errors until we went up to 16k cores, and we had to reduce it.

mark-petersen:
@hrmoncada, please practice your bit-for-bit tests on this PR. Also, do a memory test to report the total memory footprint before/after this PR. Possible tools are intel VTune, perf, and intel advisor. Thanks.

hrmoncada:
I will, thanks

hrmoncada:
@mark-petersen I can not take your configuration ocena file

gr-fe1:compass$ take /usr/projects/climate/mpeterse/repos/model/ocean_develop/testing_and_setup/compass/general.config.ocean_hrmoncada
Cannot find giver's account information.

hrmoncada:
@mark-petersen  never mind,  I got the file :)

PR #371, opened 2019-09-18
Add tidal potential forcing for 8 major constituents

sbrus89:
Includes diurnal constituents: K1, O1, P1, Q1 and semi-diurnal constituents: K2, N2, S2, M2. A tanh ramp in time is used to gradually apply the forcing (similar to time-varying atmospheric forcing). It has been implemented for both split_explicit and RK4 time-stepping.

Also, there are several updates to the Hurricane Sandy test case:
 - Increase maximum depth to 6000m (from 2000m)
 - Decrease minimum depth to 10m (from 30m)
 - Use 100layerE3SMv1 vertical grid (was 60layerPHC) 
 - Decrease forcing ramps to 10 days and shorten simulation duration to 10/10-11/3 (was 10/1-11/8)
 
Any feedback on the implementation is welcome.


sbrus89:
Here are some results compared to our previous best solution without tides.

- These runs use a mesh with 2km resolution in Delaware Bay and 5km in the Chesapeake Bay and Long Island regions. 
- Finer resolution is going to be required to improve these results, especially inside the bays.
 - Also, these runs do not use wetting and drying, which should also help improve the comparison to observed data.

src/core_ocean/Registry.xml
```diff
@@ -3171,6 +3218,52 @@
...
 		/>
+
+		<!-- Fields for tidal potential forcing -->
+		<var name="tidalPotentialEta" type="real" dimensions="nCells Time" units="m"
+			description="Equilibrium tidal potential"
+			packages="forwardMode"
+		/>
+		<var name="tidalPotentialZMid" type="real" dimensions="nVertLevels nCells Time" units="m"
+			description="zMid - Equilibrium tidal potential"
+			packages="tidalPotentialForcingPKG"
```
mark-petersen: Thanks for putting all your 3D variables in packages. This is the right way to do it.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -817,11 +820,14 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
                                           * normalBarotropicVelocitySubcycleCur(eoe) * fEdge(eoe)
                           end do
 
+                          sshCell1 = sshSubcycleCur(cell1) - tidalPotentialEta(cell1)
+                          sshCell2 = sshSubcycleCur(cell2) - tidalPotentialEta(cell2)
+
```
mark-petersen: Let me think about this. There is a danger in adding extra computation and (even more so) extra memory access for special types of simulations, that are then used by all simulations, even if they compute with zeros. The barotropic subcycling does not scale well and takes a large fraction of the simulation time. These lines are being added to the barotropic subcycling.

Alternatives are to make a separate loop for including tides, or using CPP pragmas. Both come with their own issues of code upkeep and readability.

sbrus89: I suppose one option would be to have a ``sshScratch`` variable and do something like this:

```
if (config_use_tidal_potential_forcing) then
  do iCell = 1,nCell
     sshScratch(iCell) = sshSubcycleCur(iCell) - tidalPotentialEta(iCell)
  end do
else
  do iCell = 1,nCell
     sshScratch(iCell) = sshSubcycleCur(iCell)
  end do
end if
.
.
.
do iEdge = 1, nEdges

  cell1 = cellsOnEdge(1,iEdge)
  cell2 = cellsOnEdge(2,iEdge)
 
  normalBarotropicVelocitySubcycleNew(iEdge) =  ...
   * (sshScratch(cell2) - sshScratch(cell1) ) / dcEdge(iEdge) + ...

end do

```
That at least gets rid of the extra indirect memory accesses inside the edge loop, but it still involves another array.

sbrus89: This would avoid the extra copy...

```
if (config_use_tidal_potential_forcing) then
  call mpas_pool_get_array(statePool, 'sshSubcycle', sshSubcycleCur, oldBtrSubcycleTime)
  call mpas_pool_get_array(scratchPool, 'sshScratch, sshScratch)
  call mpas_pool_get_array(forcingPool, 'tidalPotentialEta', tidalPotentialEta)
  do iCell = 1,nCell
     sshScratch(iCell) = sshSubcycleCur(iCell) - tidalPotentialEta(iCell)
  end do
else
  call mpas_pool_get_array(statePool, 'sshSubcycle', sshScratch, oldBtrSubcycleTime)
end if
```

sbrus89: @mark-petersen, I think this is resolved now based on what we discussed. Let me know if you have other thoughts.

src/core_ocean/Registry.xml
sbrus89: Sure thing!

sbrus89:
Results of the finer mesh USDEQU60at15cr5rr1 compared with USDEQU120at30cr10rr2. The pre-storm tidal amplitudes are too high, but the phase errors are better and the surge is also improved.

pwolfram:
Will want to add simple SAL in before we merge this week in conversations with Brian Arbic and Nairita Pal.

pwolfram:
@vanroekel and @xylar, if you'd like to review this week could you please take a look over the next day or so?  In conversations with @mark-petersen we need to get this in this week. @proteanplanet
 has offered to take a look in a G case with this merged branch.  Thanks!

xylar:
@pwolfram, I wasn't aware a review had been requested.  I'll take a look at this tomorrow.

vanroekel:
I don't feel qualified to review these changes.  I'll remove my name from the list

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -794,10 +799,24 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
...
                      call mpas_pool_get_array(statePool, 'normalBarotropicVelocitySubcycle', normalBarotropicVelocitySubcycleNew, &
                                               newBtrSubcycleTime)
-                     call mpas_pool_get_array(statePool, 'sshSubcycle', sshSubcycleCur, oldBtrSubcycleTime)
 
                      call mpas_pool_get_array(diagnosticsPool, 'barotropicForcing', barotropicForcing)
 
+                     ! Subtract tidal potential from ssh, if needed
+                     if (config_use_tidal_potential_forcing) then
+                       call mpas_pool_get_array(statePool, 'sshSubcycle', sshSubcycleCurNoTides, oldBtrSubcycleTime)
+                       call mpas_pool_get_array(forcingPool, 'sshSubcycleCurWithTides', sshSubcycleCur)
```
xylar: I find the discrepancy between the names of the local pointers and the variables in the pool pretty confusing.  I grepped the code for sshSubcycleNoTides trying to see where that got updated only to realize that's not the name of it elsewhere.  In general, it helps to keep the names of local pointers as consistent as possible with the names in pools.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -1000,6 +1018,25 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
                    call mpas_pool_get_field(scratchPool, 'btrvel_temp', btrvel_tempField)
                    btrvel_temp => btrvel_tempField % array
 
+                   ! Subtract tidal potential from ssh, if needed
+                   if (config_use_tidal_potential_forcing) then
+                     call mpas_pool_get_array(statePool, 'sshSubcycle', sshSubcycleCurNoTides, oldBtrSubcycleTime)
+                     call mpas_pool_get_array(statePool, 'sshSubcycle', sshSubcycleNewNoTides, newBtrSubcycleTime)
+                     call mpas_pool_get_array(forcingPool,'sshSubcycleCurWithTides', sshSubcycleCur)
+                     call mpas_pool_get_array(forcingPool,'sshSubcycleNewWithTides', sshSubcycleNew)
```
xylar: Same comment here, please consider making the local variable names the same as those in the pool.

xylar:
@sbrus89 or @pwolfram, could you point to the design doc here if it's publicly available?

xylar:
If I have found the correct design doc via confluence, I see the following:
> For instance, in a shallow water model, we replace the term $\nabla \eta$ in the momentum equation with a gradient of $\eta$ referenced to the equilibrium tide and self-attraction and loading terms...In essence, the equilibrium tide and self-attraction and loading terms reset the reference against which pressure gradients are computed.

Is this the approach that is being taken here?  If so, I'm concerned that just resetting the SSHH will have adverse consequences such as lack of mass conservation in a fully 3D model.  In other words, the SSH is an important dynamical variable beyond its influence on the pressure gradient.  Instead, it seems to me that tides (and eventually changes in equilibrium sea level) would need to be incorporated as change the reference gravitational potential, which is currently assumed to correspond to z=0.  @sbrus89 and @pwolfram (and maybe @mark-petersen), could we discuss this over the phone sometime today?

pwolfram:
@xylar, the design doc is at https://www.overleaf.com/read/dxgprfyjnnnw Is this the one you were looking at for your review?

xylar:
@pwolfram, yes, that's the document I was looking at.  However, I don't think it's a good idea to share an editable link to it on a public forum (though that ship has sailed).

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -794,10 +799,24 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
...
+                     ! Subtract tidal potential from ssh, if needed
+                     if (config_use_tidal_potential_forcing) then
+                       call mpas_pool_get_array(statePool, 'sshSubcycle', sshSubcycleCurNoTides, oldBtrSubcycleTime)
+                       call mpas_pool_get_array(forcingPool, 'sshSubcycleCurWithTides', sshSubcycleCur)
+                       call mpas_pool_get_array(forcingPool, 'tidalPotentialEta', tidalPotentialEta)
+                       call mpas_pool_get_dimension(block % dimensions, 'nCells', nCellsPtr)
+
+                       nCells = nCellsPtr
+                       do iCell = 1, nCells
+                         sshSubcycleCur(iCell) = sshSubcycleCurNoTides(iCell) - tidalPotentialEta(iCell)
```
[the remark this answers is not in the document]
sbrus89: @xylar, if `config_use_tidal_potential_forcing` is true, the `sshSubcycleCur` variable name is aliased to the temporary array `sshSubcycleCurWithTides` and 'sshSubcycleCurNoTides' is assigned to the actual SSH. This way, the SSH is not modified. I did this so the subcycle computation could remain unchanged and that no additional calculations would be done there.

I agree it's a little confusing, but @mark-petersen was (rightly) concerned about adding FLOPS to the subcycle loop for a feature that is not going to be used for E3SM at this point. This is what I came up with to get around that. I'm definitely open to other suggestions.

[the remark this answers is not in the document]
xylar: @sbrus89, my small concern is the naming convention for the variables in the code being inconsistent with the names in the pools.  I'd change that if I were you but it's not critical.

My big concern is the implementation of tides as a modification to SSH.  I do not believe that is a correct approach, and they should instead be a modification of the gravitational potential, which would result in the modification of the SSH as an emergent property (as @pwolfram says above).  The design doc gives the incorrect impression that modifying SSH is the way to go because this is equivalent to modifying the gravitational potential for a shallow water model.  But the same is not true of a 3D, primitive-equation ocean model.  Thus, my comment pointing to the 2 chapters of the MOM5 manual that discuss modification of the gravitational potential in general and then specifically for tides.  I think that's the (only?) way to go.

pwolfram:
@xylar, can you please be more specific about your concern?  We are effectively using 8.2 from the MOM link (with SAL needing to be added as noted above).  But, we are just going to use the very simple scalar approximation for SAL for this time with a more advanced SAL added later.  Are you concerned about using this form itself?

Maybe this is the issue as implied in design doc that we replace ssh (we don't):

∇η→∇(η−η_EQ−η_SAL) (3.4)

We are not modifying the free surface as this implies but rather adding on the gravitational potential terms to account for the tide (reflected in the momentum equation via the gradient of the potentials to represent the tidal forcing). This results in the free surface as an emergent property.  Thus, η ≠ η−η_EQ−η_SAL, but we are adding the following forcing terms to the momentum equation RHS: -∇(η_EQ+η_SAL).  These terms aren't there currently but need to be added as tendencies to have the tidal solution computed as an emergent property of the free surface.

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -326,6 +330,17 @@ subroutine ocn_tend_vel(tendPool, statePool, forcingPool, diagnosticsPool, meshP
...
+      ! Compute tidal potential forcing, if needed 
+      if (config_use_tidal_potential_forcing) then
+        call ocn_compute_tidal_potential_forcing(meshPool, forcingPool, diagnosticsPool, err)
+      end if
+
+      if (config_use_tidal_potential_forcing .and. config_time_integrator == 'RK4') then
+        call mpas_pool_get_array(forcingPool, 'tidalPotentialZMid', zMid)
+      else
+        call mpas_pool_get_array(diagnosticsPool, 'zMid', zMid)
+      end if
```
xylar: @pwolfram and @sbrus89, okay, so his is the modification to the code that actually changes the physics?  That wasn't clear to me at all.

xylar: Now that I understand, it seems like a clever things to do.  I'll give @pwolfram a call soon to make sure I understand fully.

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -326,6 +330,17 @@ subroutine ocn_tend_vel(tendPool, statePool, forcingPool, diagnosticsPool, meshP
       !
       call ocn_vel_vadv_tend(meshPool, normalVelocity, layerThicknessEdge, vertAleTransportTop, tend_normalVelocity, err)
 
+      ! Compute tidal potential forcing, if needed 
+      if (config_use_tidal_potential_forcing) then
+        call ocn_compute_tidal_potential_forcing(meshPool, forcingPool, diagnosticsPool, err)
+      end if
+
+      if (config_use_tidal_potential_forcing .and. config_time_integrator == 'RK4') then
```
xylar: Should this work in split explicit, too?  Is the RK4 flag needed?

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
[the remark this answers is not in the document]
xylar: Talking with @pwolfram on the phone, I got assurances that @mark-petersen and he would verify that these modifications to the subcycled SSH do not feed into the SSH at the end of a time step, and are just used to compute the gravitational potential in the barotropic subcycle.

sbrus89:
@xylar, thanks so much for reviewing so carefully! I really appreciate it. In light of your comments, I made some changes which (I hope) help clarify what is being done and make things a bit cleaner. Thanks again.

src/core_ocean/shared/mpas_ocn_tendency.F
sbrus89: For split explicit, we only want to add the tidal potential to the barotropic equations.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
sbrus89: I took care of this the best I could in my latest commits.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
sbrus89: I took care of this the best I could in my latest commits.

src/core_ocean/shared/mpas_ocn_tendency.F
sbrus89: I tried to make things more explicit in my recent changes.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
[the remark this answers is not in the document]
sbrus89: Since each of the subcycle calculations is being done inside blocks, the pointer association to the working array that contains ssh-tidal_potential will not effect anything outside of that particular block.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -1000,6 +1023,26 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
...
+                   if (config_use_tidal_potential_forcing) then
+                     call mpas_pool_get_array(forcingPool,'sshSubcycleCurWithTides', sshSubcycleCurWithTides)
+                     call mpas_pool_get_array(forcingPool,'sshSubcycleNewWithTides', sshSubcycleNewWithTides)
+                     call mpas_pool_get_array(forcingPool, 'tidalPotentialEta', tidalPotentialEta)
+                     call mpas_pool_get_dimension(block % dimensions, 'nCells', nCellsPtr)
+
+                     nCells = nCellsPtr
+                     do iCell = 1, nCells
+                       sshSubcycleCurWithTides(iCell) = sshSubcycleCur(iCell) - tidalPotentialEta(iCell)
+                       sshSubcycleNewWithTides(iCell) = sshSubcycleNew(iCell) - tidalPotentialEta(iCell)
```
xylar: Where are `sshSubcycleCurWithTides` and `sshSubcycleNewWithTides` actually used in the code?  I can only find them being computed here but not used to update any prognostic variables.

src/core_ocean/shared/mpas_ocn_tendency.F
xylar: Really?  I'm surprised by that.

xylar: @mark-petersen and @pwolfram, can you confirm that?  I don't see why the pressure gradient should be computed differently between the 2 time integration approaches.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
xylar: Sorry, I get it now.  I'm still getting tripped up by pulling in a variable with one name from a pool and using it with another name in the code, but I understand that that's practical here.

mark-petersen:
@sbrus89 please confirm if you are done with alterations.
@xylar and @pwolfram please check approved when you are ready - can you do that today or early tomorrow? (already past COB for Xylar).
After that, I'll test locally to confirm bfb for our standard configurations, then approve and merge. 

After meeting with @sbrus89 yesterday, I'm satisfied with this PR overall, and that Steve addressed Xylar's concerns with today's changes.

sbrus89:
@mark-petersen, As Phil mentioned a while back, I'm going to add the scalar self attraction and loading term in before the merge. I'm testing it right now. The additions are very minor.

sbrus89:
Just for the sake of documentation, here are results with the scalar approximation for SAL on the lower-resolution USDEQU120at30cr10rr2 mesh:

sbrus89:
@mark-petersen, I just need to test RK4 now and then this should be ready to go.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -798,6 +805,24 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
...
+                       call mpas_pool_get_array(forcingPool, 'tidalPotentialEta', tidalPotentialEta)
+                       call mpas_pool_get_dimension(block % dimensions, 'nCells', nCellsPtr)
+
+                       nCells = nCellsPtr
+                       do iCell = 1, nCells
+                         sshSubcycleCurWithTides(iCell) = sshSubcycleCur(iCell) - tidalPotentialEta(iCell) &
+                                                        - config_self_attraction_and_loading_beta * sshSubcycleCur(iCell)
+                       end do
+
+                       call mpas_pool_get_array(forcingPool, 'sshSubcycleCurWithTides', sshSubcycleCur)
```
pwolfram: Much clearer now!

src/core_ocean/shared/Makefile
```diff
@@ -66,13 +66,14 @@ OBJS = mpas_ocn_init_routines.o \
 	   mpas_ocn_sea_ice.o \
 	   mpas_ocn_framework_forcing.o \
 	   mpas_ocn_time_varying_forcing.o \
-	   mpas_ocn_wetting_drying.o
+	   mpas_ocn_wetting_drying.o \
+           mpas_ocn_tidal_potential_forcing.o
```
pwolfram: Seems like there is a strange issue here with indenting...

src/core_ocean/shared/mpas_ocn_tendency.F
pwolfram: I thought I understood clearly yesterday when we discussed @mark-petersen and @sbrus89 but not I'm not so sure.  Can we please clarify here for posterity?  

My recollection is that for split explicit the pressure gradient term already has the barotropic contribution removed, but for split explicit the baroclinic and barotropic are combined, hence for split explicit the calls to `ocn_vel_pressure_grad_tend` only worry about the baroclinic terms (no need for adjustment of the tidal potential) whereas for RK4 they include both baroclinic and barotropic (and hence the need for adjustment of the tidal potential).  Thus in the split explicit routine there is the tidal potential, and because there is no equivalent change in the RK4 routine, it is made in the tendency and hence the need for the RK4 flag.  Thus, the application of the tidal potential forcing is different in each approach and hence the code doesn't "overlap" for each case.

Would you both concur this is the correct interpretation or have I missed something?  Just want to make sure we have this written down clearly so we are all on the same page.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
```diff
@@ -0,0 +1,663 @@
...
+!
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+!
+!  ocn_tidal_potential_forcing
+!
+!> \brief MPAS ocean tidal potential forcing module
+!> \author Steven Brus 
+!> \date   September 2019
+!> \details
+!>  This module contains routines for the tidal potential forcing.
```
pwolfram: @sbrus89, can we please provide a **read-only** link to the design document here?

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
```diff
@@ -0,0 +1,663 @@
...
+                                       err)
+
+        do iCell = 1,nCells
+          lat = latCell(iCell)
+          latitudeFunction(iCell,1) = 3.0_RKIND*sin(lat)**2 - 1.0_RKIND 
+          latitudeFunction(iCell,2) = sin(2.0_RKIND*lat)
+          latitudeFunction(iCell,3) = cos(lat)**2
+        end do
+        
+        do iCon = 1,nTidalConstituents
```
pwolfram: Nice, I think this will be helpful for debugging and documentation and since it is on init having this level of output should be ok.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
```diff
@@ -0,0 +1,663 @@
...
+
+      call mpas_get_time(refTime, YYYY=refYear,DoY=refDay,H=refHour)
+      refDay = refDay + 1
+
+      ! Calculate locations of sun and moon
+      call orbit(refYear,refDay,refHour, &
+                 s,p,h,p1,N,I,nu,xi,nup,nup2,pc)
+
+
+      !! From tide design doc
```
pwolfram: Do we need to have these comments currently if we link to the tide design doc in the header?

pwolfram:
@mark-petersen, the plan is to have this merged before #366, which can use some work here to simplify the number of lines of code committed.

src/core_ocean/shared/mpas_ocn_tendency.F
sbrus89: @pwolfram, that is correct as far as I understand it.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
sbrus89: Thanks!

src/core_ocean/shared/Makefile
sbrus89: Thanks for catching that. I'll fix it.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
sbrus89: @pwolfram, Do I have to have some kind of permissions to do that? I don't see where a can get a link under the share tab.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
sbrus89: Thanks, I think it's worth keeping in there.

sbrus89:
One more set of results showing that RK4 works. This is on the coarse USDEQU120at30cr10rr2 mesh with a 25 second RK4 timestep as opposed to the 4/40 second split explicit timestep used in all the previous results.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
xylar: @pwolfram and @sbrus89, I think the right way to do this is to put a pdf (and maybe a tex file) of the design document here: 
https://github.com/MPAS-Dev/MPAS-Model/tree/ocean/develop/docs/ocean/design_docs

I don't believe Overleaf an appropriate space to try to share design docs from -- it is only an authoring tool, not meant for read-only sharing.  Projects typically get archived when they are no longer used.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
sbrus89: Do you think we should keep them in there? I'm fine either way. I can also put in a more specific comment.

pwolfram: Let's keep them for now, I don't think this will hurt and it is more vs less information.  We don't compile it so minimally it will help jog memories or make connections to the design doc.

pwolfram:
@sbrus89, I checked the "verify RK4" box.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
sbrus89: @xylar, that's a good suggestion.

@pwolfram , I went ahead and added pdf and tex files of the design doc to the location Xylar pointed out. I also made a note in the header.

sbrus89:
@pwolfram, is there anything else you'd like me to address? Once I've taken care of everything, let me know and I'll squash.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
pwolfram: @sbrus89, I'm going to move the overleaf to the repo @xylar recommended.  We just want to link there or have some comment about where to find the design doc.  Ideally, this would include the PR number for the design_doc.  Please feel free to begin squashing and we can just update that reference at the end.  I probably need 20min to get the design doc migrated over, so working in parallel is probably good...

pwolfram: If I can push to your fork this could be easy and you could just rebase from your remote or cherry-pick the commit and then squash it with the rest.

pwolfram: Looks like you already grabbed it, good work!

pwolfram: Please feel free to squash down and we can move forward.

pwolfram: I made a note on the overleaf of 

MPAS-O Ocean Tides Design Document Snapshot on Oct 18, 2019 placed at https://github.com/MPAS-Dev/MPAS-Model/tree/ocean/develop/docs/ocean/designdocs

So we know we moved it over to the MPAS today.

pwolfram:
@sbrus89, please let me know when you do the final squash and then I'll review.  Thanks!

pwolfram:
Once @mark-petersen approves and we squash I can merge to coastal.

sbrus89:
@pwolfram, I just squashed.

pwolfram:
@sbrus89, had to update the design doc part to include the other tex files and remove the pdf.

pwolfram:
Just to confirm no damage was done:

```
design_docs pwolfram$ git diff 2aae280 34698ba --stat
 docs/ocean/design_docs/tides.pdf             |  Bin 204589 -> 0 bytes
 docs/ocean/design_docs/tides/ametsoc.bst     | 1741 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 docs/ocean/design_docs/tides/tides.bib       |   35 +++++
 docs/ocean/design_docs/{ => tides}/tides.tex |    0
 4 files changed, 1776 insertions(+)
```

pwolfram:
@mark-petersen, can you please approve so that I can merge?

sbrus89:
Oops. Sorry I forgot to include the other tex files

pwolfram:
@sbrus89, it is all good.  Please enjoy your day off and many thanks for the help!

pwolfram:
Thanks everyone!

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -326,6 +336,31 @@ subroutine ocn_tend_vel(tendPool, statePool, forcingPool, diagnosticsPool, meshP
...
+        call mpas_pool_get_array(forcingPool, 'tidalPotentialZMid', tidalPotentialZMid)
+        call mpas_pool_get_array(forcingPool, 'tidalPotentialEta', tidalPotentialEta)
+        call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
+
+        do iCell = 1, nCells
+          do k = 1, maxLevelCell(iCell)
+            tidalPotentialZMid(k,iCell) = zMid(k,iCell) - tidalPotentialEta(iCell) &
```
pwolfram: @sbrus89, potential issue may be here with #366 including tidal forcing.  `zMid` is a strong negative with the spurious datum shift we are employing in #366 until the vertical datum issue is resolved.

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -326,6 +336,31 @@ subroutine ocn_tend_vel(tendPool, statePool, forcingPool, diagnosticsPool, meshP
...
+                                        - config_self_attraction_and_loading_beta * zMid(k,iCell)
```
pwolfram: Will test `config_self_attraction_and_loading_beta=0` and report back because offset at `zMid` is likely cause of problem to start beside offset.

pwolfram: That fixed the issue, so having a water column shifted away from 0 for `zMid` will cause a pretty spurious artificial tidal flow.

pwolfram: Fixed means code was stable, but results are not correct with tides on (but are ok with tides off), which means we still have a bug here.

pwolfram: # SE no wetting / drying


# RK4 no wetting / drying


# no tides with wetting / drying


# tides with wetting / drying but no SAL

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
```diff
@@ -0,0 +1,665 @@
...
+      real (kind=RKIND), dimension(:), pointer :: lonCell
+      integer, dimension(:), pointer ::  maxLevelCell
+
+      integer, pointer :: nTidalConstituents
+      real (kind=RKIND), dimension(:), pointer :: amplitude, frequency, loveNumbers
+      real (kind=RKIND), dimension(:), pointer :: nodalFactorAmplitude, nodalFactorPhase, astronomicalArgument
+      real (kind=RKIND), dimension(:,:), pointer :: latitudeFunction
+      integer, dimension(:), pointer :: constituentType
+      real (kind=RKIND), dimension(:), pointer :: eta
+      real (kind=RKIND), dimension(:,:), pointer :: zMid, tidalPotentialZMid
```
pwolfram: We don't need these here, do we?

pwolfram: cc @sbrus89

src/core_ocean/shared/mpas_ocn_tendency.F
pwolfram: These are all for `USDEQU120at30cr10rr2*` meshes.

src/core_ocean/shared/mpas_ocn_tidal_potential_forcing.F
sbrus89: Correct, the `zMid` and `tidalPotentialZMid` could be removed. At one point I was using them here, but moved those calculations elsewhere and forgot to get rid of the declarations. Thanks for catching this.

src/core_ocean/shared/mpas_ocn_tendency.F
pwolfram: Using the approach of adding back out the vertical shift produces:



The tidal signal is contained in the output now, but the result is not as good as say just the standard RK4 without wetting / drying, so something is still not completely right.



Del4 differed between the runs, am reruning with matching del4 values.

pwolfram: Note, hard-coded code change was `tidalPotentialZMid(k,iCell) = (zMid(k,iCell) + 977.502_RKIND) - tidalPotentialEta(iCell) &...`

pwolfram: Station 8519483 looks better though, especially compared with the previous results.

pwolfram: # Possible correction
@sbrus89, if we allow for zMid offset from zero, this means that this 
`- config_self_attraction_and_loading_beta * zMid(k,iCell)` may not be correct.
Shouldn't it be something more like this?
`- config_self_attraction_and_loading_beta * (ssh(iCell)-ssh_dry(iCell))`

Given it is a small correction I'm wondering if we shouldn't just turn it off.

Given the approximation really is only deep water as the text implies, maybe we should do something like this:
`- config_self_attraction_and_loading_beta * (bottomDepth > 100)*ssh(iCell)`

So, we wouldn't do any SAL for shallow waters with less than 100m depth.

# Reference
https://www.sciencedirect.com/science/article/pii/S0967064504002048
image

PR #373, opened 2019-09-18
Remove compile-time switch between PIO1 and PIO2

jayeshkrishna:
Earlier versions of PIO1 and PIO2 had incompatibilities between
the interfaces and types provided to the user. These issues were
fixed and there is no longer a requirement to compile MPAS
differently for PIO1 and PIO2.

[BFB]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,13 +426,6 @@ FCINCLUDES =
 LIBS = 
 
 #
-# If user has indicated a PIO2 library, define USE_PIO2 pre-processor macro
-#
-ifeq "$(USE_PIO2)" "true"
-	override CPPFLAGS += -DUSE_PIO2
-endif
-
-#
 # Regardless of PIO library version, look for a lib subdirectory of PIO path
 # NB: PIO_LIB is used later, so we don't just set LIBS directly
 #
@@ -577,12 +570,6 @@ ifeq "$(USE_PAPI)" "true"
 else # USE_PAPI IF
 	PAPI_MESSAGE="Papi libraries are off."
 endif # USE_PAPI IF
-
-ifeq "$(USE_PIO2)" "true"
-	PIO_MESSAGE="Using the PIO 2 library."
-else # USE_PIO2 IF
-	PIO_MESSAGE="Using the PIO 1.x library."
-endif # USE_PIO2 IF
 
 ifdef TIMER_LIB
 ifeq "$(TIMER_LIB)" "tau"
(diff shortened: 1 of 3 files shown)


jayeshkrishna:
I tested this change with the latest E3SM master (v1.1.0-beta-2249-ga9ec682) using the following developer test - ERS_Ld5.T62_oQU120.CMPASO-NYF

mgduda:
So long as there are users with specific version of PIO (i.e., those versions that contain interfaces that are incompatible with PIO1), I don't think we can remove this switch.

mgduda:
@jayeshkrishna I just noticed that this PR is targeting `e3sm/develop` and not `develop`. However, if there is ever any intent of merging `e3sm/develop` into `develop`, I think my objection still stands.

jonbob:
Thanks for weighing in @mgduda. I'm sure we want to keep framework shared between E3SM and NCAR, so we'll have to find a different way of handling this.

mgduda:
@jonbob Out of curiosity, what needs to be "handled" in E3SM? If newer versions of PIO2 are compatible with the PIO1 interfaces, does the code not compile if `USE_PIO2` is omitted entirely from the build options? We have considered the idea of just detecting which PIO interfaces to use, rather than forcing users to specify `USE_PIO2=true` or not. Would this be of any help in E3SM?

jonbob:
@mgduda - thanks. I have to admit I hadn't really thought much about the details, as much as I wanted to make sure we don't do anything that would diverge from the shared version of framework

mark-petersen:
@jayeshkrishna there could be users that are using older versions of PIO1, and it seems unnecessary to remove this switch. What if E3SM just always compiles with `USE_PIO2=true`, regardless of whether E3SM uses PIO1 or PIO2, since we are presumably using newer versions of PIO1?

i.e., add `USE_PIO2=true` to these three lines in E3SM:
```
components/mpas-albany-landice/cime_config/buildlib:55:my $bldcmd = "cd $root_path; ${GMAKE} all ${MAKE_ARGS} CORE=$core ESM=$esm DRIVER=$component_driver GEN_F90=true";
components/mpas-ocean/cime_config/buildlib:54:my $bldcmd = "cd $root_path; ${GMAKE} all ${MAKE_ARGS} CORE=$core ESM=$esm DRIVER=$component_driver GEN_F90=true";
components/mpas-seaice/cime_config/buildlib:54:my $bldcmd = "cd $root_path; ${GMAKE} all ${MAKE_ARGS} CORE=$core ESM=$esm DRIVER=$component_driver GEN_F90=true";
```

jayeshkrishna:
ok, since backward compatibility is important to MPAS users I will close this PR and look for alternatives for E3SM.

PR #376, opened 2019-09-24
Fix threading issue in MPAS-O GM routine

jonbob:
This PR fixes a threading issue that was introduced in [E3SM PR #3057](https://github.com/E3SM-Project/E3SM/pull/3057) that brought in the initial implementation of 3D varying GM bolus. After that commit, some E3SM threading tests were observed to fail with different results.

Tested with:
* SMS_P12x2.T62_oQU240.CMPASO-NYF.compy_intel - BFB results 10 times
* SMS_P12x2.ne4_oQU240.A_WCYCL1850.compy_intel.allactive-mach_mods - BFB results 10 times

This last test is one that was failing on sandiatoss3.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -200,16 +200,23 @@ contains
       ! Assign a huge value to the scratch variables which may manifest itself when
       ! there is a bug.
       !$omp do schedule(runtime) private(k)
-      do iEdge = 1, nEdges + 1
+      do iEdge = 1, nEdges
          do k = 1, nVertLevels
             gradDensityEdge(k, iEdge) = huge(0D0)
+            gradZMidEdge(k, iEdge) = huge(0D0)
+            normalGMBolusVelocity(k, iEdge) = 0.0_RKIND
+         end do
+      end do
+      !$omp end do
+
+      !$omp do schedule(runtime) private(k)
+      do iEdge = 1, nEdges + 1
+         do k = 1, nVertLevels
             gradDensityTopOfEdge(k, iEdge) = huge(0D0)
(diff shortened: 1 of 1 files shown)


jonbob:
Just a quick note -- I had intel inspector working and did some checking on memory as well. The tools points at invalid memory access in the gm code as well, particularly this loop:
````
      do iEdge = 1, nEdges + 1
         do k = 1, nVertLevels
            gradDensityEdge(k, iEdge) = huge(0D0)
            gradDensityTopOfEdge(k, iEdge) = huge(0D0)
            dDensityDzTopOfEdge(k, iEdge) = huge(0D0)
            gradZMidEdge(k, iEdge) = huge(0D0)
            gradZMidTopOfEdge(k, iEdge) = huge(0D0)
            relativeSlopeTopOfEdge(k, iEdge) = 0.0_RKIND
            relativeSlopeTapering(k, iEdge) = 0.0_RKIND
            normalGMBolusVelocity(k, iEdge) = 0.0_RKIND
         end do
      end do
      !$omp end do
````
I believe this is because some of the arrays are dimensioned nEdges and the loop goes over nEdges+1. Also the "private(k)" is missing in the code -- I'll commit it to this PR.

mark-petersen:
@amametjanov or @philipwjones could you please look at the openMP pragmas and confirm that they are done correctly?

philipwjones:
@mark-petersen OpenMP pragmas look ok to me - I know @amametjanov has been removing the private clauses since there is a default private clause on the top-level parallel region, so they're redundant but do no harm.

amametjanov:
Yes, this looks good. I think the last two commits, which make loop extents match array sizes, fix the real culprits. Formatting is much better too.

mark-petersen:
I moved the commits over to ocean/develop, otherwise we will bring extra things over with the merge.

PR #382, opened 2019-10-15
fix layerthickness call

mark-petersen:
layerThickness pointer was passed into a subroutine but it is unassociated, and never used in that subroutine. This causes an error with Intel 19 compilers. This PR simply removes the unused variable.


mark-petersen:
Fixes https://github.com/E3SM-Project/E3SM/issues/3190#issuecomment-541798084
Compiled and tested with nightly regression suite.

PR #384, opened 2019-10-18
Merge coastal features into ocean/develop

mark-petersen:
This PR is an accumulation of PRs into the ocean/coastal branch. We are bringing them in at once for efficiency:
#285,  #289, #284, #295, #310, #311, #312, #335, #354, #356, #358, #359, #365, #371

PR #385, opened 2019-10-23
Ocean/coastal update (force push from ocean/develop)

pwolfram:
This is just a reset of ocean/coastal using ocean/develop following #384 as recommended by @mark-petersen.  This is not strictly needed but is to document the force-push of `ocean/develop` onto `ocean/coastal`.


xylar:
How do you do a pull request of a force push? I thought a PR was always a merge.

pwolfram:
@xylar, it was just a direct push, not a force-push.  I'll edit the above description accordingly because the force-push wasn't needed.

pwolfram:
Thanks!

xylar:
I think @mark-petersen has in mind to do a hard reset so ocean/coastal points to the exact same hash as ocean/develop.  I would think that would require a force push but I could be wrong.

pwolfram:
I think we have that, unless I've missed something.  Are you seeing something different @xylar?

```bash
└─▪ git fetch --all
...
Fetching pwolfram
...
   6fe9af3..7053845  ocean/coastal -> MPAS-Dev/ocean/coastal
   559c823..7053845  ocean/develop -> MPAS-Dev/ocean/develop
...

mark-petersen:
Yes, this all looks right. `ocean/coastal` is now identical to `ocean/develop`. This is a little cleaner than doing a back-merge.

pwolfram:
We also have a record of what we did, which is key.  It is possible this has broken something, but at least we have a landing page and record of when it went wrong (and potentially how in terms of a commit range).

xylar:
Okay, like I said, I just didn't know that was possible.

PR #388, opened 2019-11-05
Add Coastal US Plus (CUSP) case to compass

mark-petersen:
This adds the CUSP "revised" mesh with the wider Gulf Stream region. I am adding a single case that is 8km resolution and 80 layers, but it is easy to change those two settings as needed. It is a revision of the previous case `NAECto60to30cr8`, but I decided to call this case simply `CUSP8` to match the papers.


mark-petersen:
@kristin-hoch I think this all works. Please test by creating and spinning up two 12-km CUSP meshes, one with 60 layers and one with 80 layers. I ran the base_mesh last night on grizzly and it works, it just takes several hours. These are the current libraries for base_mesh:
```
module purge
conda activate e3sm_unified_1.2.6
```
because the latest (e3sm_unified_1.3.0) does not include jigsawpy. Please post your results on this page as part of your review.

mark-petersen:
Note, to change between 80 layers and 60 layers, redo the `initial_state` step but with the COMPASS file 
`testing_and_setup/compass/ocean/global_ocean/CUSP8/init/config_initial_state.xml`
linked to
`../../config_files/config_initial_state_60_layers.xml`
and recreate the case. You can copy in the base_mesh and cull_mesh directories from another directory if you already made it.

mark-petersen:
@sbrus89 and @pwolfram, I'm trying to trim unused cases as I go. I did this:
- rename `global_ocean/NAEC60to30cr8` to `global_ocean/CUSP8`
- remove three cases from global_ocean: NAEC60to30cr8L60, USDEQU120cr10rr1, USDEQU300cr20rr3

Please confirm if that is OK. @sbrus89 indicated that your global cases are all now in the `hurricane` subdirectory, including `USDEQU120at30cr10rr2` and `USDEQU60at15cr5rr1`.

vanroekel:
@mark-petersen by final design do you just mean of CUSP8?  This can't be the final design for water cycle v2 simulations in E3SM.  It is much too expensive.  I think we'll probably want to test ~12km in the HR zone first before finalizing meshes for v2 water cycle.

pwolfram:
@mark-petersen, I'm ok with removal of `USDEQU120cr10rr1, USDEQU300cr20rr3` from `global_ocean`, but can't recall who is concerned with the `NAEC60to30cr8L60` simulation.  I think this may be @maltrud but don't know for sure, so I can't give approval to remove `NAEC60to30cr8L60`.

mark-petersen:
> by final design do you just mean of CUSP8?

@vanroekel "final" was a poor choice of words. I should change to "current". I wanted to merge this in so we have a record of it. We will start with the CUSP8 layout to test a CUSP12 this week.

PR #389, opened 2019-11-06
Adds variable Cd and Manning's n values 

pwolfram:
Adds variable bottom drag capability as well as Manning's n values
for variable drag in coastal areas.

Note, equation for Manning's n conversion to Cd is `Cd=g*n^2*h^(-1/3)`, where `n` is Manning's coefficient, `g` is gravity, and `h` is the total water depth.

Testing is against the drying slope case.

### DIFF
--- a/src/core_ocean/mode_init/mpas_ocn_init_tidal_boundary.F
+++ b/src/core_ocean/mode_init/mpas_ocn_init_tidal_boundary.F
@@ -102,6 +102,7 @@ contains
     integer, dimension(:), pointer :: maxLevelCell
     real (kind=RKIND), dimension(:), pointer :: yCell, refBottomDepth, bottomDepth, vertCoordMovementWeights, dcEdge
     real (kind=RKIND), dimension(:), pointer :: tidalInputMask
+    real (kind=RKIND), dimension(:), pointer :: bottomDrag
     real (kind=RKIND), dimension(:), pointer :: ssh
     real (kind=RKIND), dimension(:,:), pointer :: layerThickness, restingThickness, zMid
     real (kind=RKIND), dimension(:,:,:), pointer :: activeTracers, debugTracers
@@ -112,6 +113,7 @@ contains
                                         config_tidal_boundary_layer_type
     logical, pointer :: config_tidal_boundary_use_distances, config_alter_ICs_for_pbcs
     logical, pointer :: config_use_wetting_drying, config_tidal_start_dry
+    logical, pointer :: config_use_variable_drag
     real (kind=RKIND), pointer :: config_tidal_boundary_right_bottom_depth, &
                                   config_tidal_boundary_left_bottom_depth, &
                                   config_tidal_boundary_plug_width_dist, config_tidal_boundary_plug_width_frac, &
@@ -119,7 +121,10 @@ contains
                                   config_tidal_boundary_salinity, config_tidal_boundary_isopycnal_min_thickness, &
                                   config_tidal_boundary_left_value, &
                                   config_tidal_boundary_right_value, &
-                                  config_drying_min_cell_height
+                                  config_drying_min_cell_height, &
+                                  config_tidal_boundary_water_thickness, &
+                                  config_tidal_forcing_left_Cd_or_n, &
+                                  config_tidal_forcing_right_Cd_or_n
     integer, pointer :: config_tidal_boundary_min_vert_levels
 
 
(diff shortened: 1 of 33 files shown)


pwolfram:
Note, need to remove RK4 fixes by Michael and need to squash commits.  Can do this after folks have a chance to review the code.

pwolfram:
# Testing

## variable Cd: `ocean/drying_slope/zstar_variableCd/1km`
dryingslopecomparison

## variable Manning's n: `ocean/drying_slope/zstar_variableManningn/1km`
dryingslopecomparisonManningn

pwolfram:
Could potentially use a function pointer to make things simpler @mark-petersen but kept standard approach we've been using and duplicated function but with need high-level key terms to keep functionality with switch outside any main loops for performance reasons.

pwolfram:
@mark-petersen, can you please review?  The "Don't Merge" label is just to point out that I have Michael's RK4 fix (needs removed since that will be merged into dev) and need to squash commits prior to the merge.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -481,6 +482,322 @@ subroutine ocn_vel_vmix_tend_implicit(meshPool, dt, kineticEnergyCell, vertViscT
...
+         ! so recompute layerThicknessEdge here.
+         cell1 = cellsOnEdge(1,iEdge)
+         cell2 = cellsOnEdge(2,iEdge)
+         do k = 1, N
+            layerThicknessEdge(k,iEdge) = 0.5_RKIND * (layerThickness(k,cell1) + layerThickness(k,cell2))
+         end do
+
+         ! average cell-based implicit bottom drag to edges and convert Mannings n to Cd
+         implicitCd = gravity*(0.5_RKIND*(bottomDrag(cell1) + bottomDrag(cell2)))**2.0 * &
+           0.5_RKIND * (ssh(cell1) + ssh(cell2) + bottomDepth(cell1) + bottomDepth(cell2))**(-1.0_RKIND/3.0_RKIND)
```
mark-petersen: It does seem unfortunate to add two new subroutines here to alter two lines, the calculation of implicitCd and for B(n) below.  This goes back to design choices of efficiency versus versatility. Choices are:
1. Extra subroutines with redundant code (your choice here).
2. If statements around smaller loops.
3. Include extra calculations with 0/1 coefficients for flag choices.
4. C pre-processor branching

None are perfect, but I agree with your choice here.

mark-petersen:
@pwolfram Please run your standard cases to exercise this code with gnu and intel compilers, using DEBUG=true.

src/core_ocean/shared/mpas_ocn_vmix.F
pwolfram: Thanks Mark!

src/core_ocean/mode_init/mpas_ocn_init_tidal_boundary.F
```diff
@@ -244,6 +252,15 @@ subroutine ocn_init_setup_tidal_boundary(domain, iErr)!{{{
           (config_tidal_boundary_right_bottom_depth - config_tidal_boundary_left_bottom_depth)
       end do
 
+      if (config_use_variable_drag) then
```
pwolfram: @caozd999, here is the selection of different types of bottom drag you can use as an example to build your own transect and variable drag.

pwolfram: https://github.com/MPAS-Dev/MPAS-Model/pull/389/files#diff-6270014bc9ae9b8554f8af50127caf5fL242 above shows bottom depth that needs modified.

I recommend building your own new `src/core_ocean/mode_init/mpas_ocn_init_tidal_transect.F` case.

pwolfram:
@mark-petersen, here are the tests you've requested.

# Testing

## intel-debug
dryingslopecomparisonManningn

## intel-release
dryingslopecomparisonManningn

Gnu to follow later...

pwolfram:
# Testing

## gnu-release
dryingslopecomparisonManningn

## gnu-debug
dryingslopecomparisonManningn

pwolfram:
@mark-petersen and @sbrus89, I'm planning to merge because everything checks out.  If you both could approve before I do this I would appreciate it.

sbrus89:
@pwolfram, the only comment I have would be that the existing constant bottom drag could be fit into this framework by just setting the `bottomDrag` variable to that constant. It adds an additional array for runs that don't require the spatially variable functionality, but would keep things simpler because there would be one fewer `ocn_vel_vmix_tend_implicit` routines. 

Just a thought and definitely not an issue that should be addressed by this PR.

pwolfram:
Thanks @sbrus89!  I'll let @mark-petersen comment here, but believe we should skip that because it would modify the E3SM runtime and not necessarily be BFB.  Minimally, it would change performance for large-scale runs with constant Cd.  But, I agree, we are adding complexity via leaving that additional option in, which your approach would remove.

pwolfram:
Thanks @mark-petersen, double-checked it worked on the merge commit I just pushed.  Thanks also @sbrus89

pwolfram:
@caozd999, PR is now merged.

caozd999:
@pwolfram great! Thank you.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -481,6 +482,322 @@ subroutine ocn_vel_vmix_tend_implicit(meshPool, dt, kineticEnergyCell, vertViscT
...
+         ! layerThicknessEdge is computed in compute_solve_diag, and is not available yet,
+         ! so recompute layerThicknessEdge here.
+         cell1 = cellsOnEdge(1,iEdge)
+         cell2 = cellsOnEdge(2,iEdge)
+         do k = 1, N
+            layerThicknessEdge(k,iEdge) = 0.5_RKIND * (layerThickness(k,cell1) + layerThickness(k,cell2))
+         end do
+
+         ! average cell-based implicit bottom drag to edges and convert Mannings n to Cd
+         implicitCd = gravity*(0.5_RKIND*(bottomDrag(cell1) + bottomDrag(cell2)))**2.0 * &
```
pwolfram: @caozd999, here is the example of how I converted Manning's n into Cd for computation in the momentum equations.

PR #391, opened 2019-11-07
Changes to COMPASS jigsaw tools to work for planar meshes too

matthewhoffman:
This merge modified the jigsaw_to_MPAS scripts in COMPASS to support running the Jigsaw workflow for planar meshes in addition to the current support for spherical meshes.  This is done by adding new flags to necessary functions that enable logic for spherical or planar coordinates to be used.  The default is always spherical, so existing ocean test cases do not need to be modified.

### DIFF
--- a/testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
+++ b/testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
@@ -31,32 +31,49 @@ from define_base_mesh import define_base_mesh
 
 
 def build_mesh(preserve_floodplain=False, floodplain_elevation=20.0,
-               do_inject_bathymetry=False):
+               do_inject_bathymetry=False, geometry='sphere'):
 
-    print('Step 1. Build cellWidth array as function of latitude and '
-          'longitude')
-    cellWidth, lon, lat = define_base_mesh.cellWidthVsLatLon()
-    da = xarray.DataArray(cellWidth,
+    if geometry=='sphere':
+        on_sphere=True
+    else:
+        on_sphere=False
+
+    print('Step 1. Build cellWidth array as function of horizontal coordinates')
+    if on_sphere:
+       cellWidth, lon, lat = define_base_mesh.cellWidthVsLatLon()
+       da = xarray.DataArray(cellWidth,
                           dims=['lat','lon'],
                           coords={'lat':lat,'lon':lon},
                           name='cellWidth')
-    da.to_netcdf('cellWidthVsLatLon.nc')
-    
+       cw_filename='cellWidthVsLatLon.nc'
+       da.to_netcdf(cw_filename)
+    else:
+       cellWidth, x, y, geom_points, geom_edges = define_base_mesh.cellWidthVsXY()
+       da = xarray.DataArray(cellWidth,
+                          dims=['y','x'],
+                          coords={'y':y,'x':x},
+                          name='cellWidth')
+       cw_filename='cellWidthVsXY.nc'
+       da.to_netcdf(cw_filename)
+
 
     print('Step 2. Generate mesh with JIGSAW')
-    jigsaw_driver(cellWidth, lon, lat)
+    if on_sphere:
+       jigsaw_driver(cellWidth, lon, lat)
+    else:
+       jigsaw_driver(cellWidth, x, y, on_sphere=False, geom_points=geom_points, geom_edges=geom_edges)
 
     print('Step 3. Convert triangles from jigsaw format to netcdf')
     jigsaw_to_netcdf(msh_filename='mesh-MESH.msh',
-                     output_name='mesh_triangles.nc', on_sphere=True)
+                     output_name='mesh_triangles.nc', on_sphere=on_sphere)
 
     print('Step 4. Convert from triangles to MPAS mesh')
     write_netcdf(convert(xarray.open_dataset('mesh_triangles.nc')),
                  'base_mesh.nc')
 
     print('Step 5. Inject correct meshDensity variable into base mesh file')
-    inject_meshDensity(cw_filename='cellWidthVsLatLon.nc',
-                       mesh_filename='base_mesh.nc')
+    inject_meshDensity(cw_filename=cw_filename,
+                       mesh_filename='base_mesh.nc', on_sphere=on_sphere)
 
     if do_inject_bathymetry:
         print('Step 6. Injecting bathymetry')
@@ -89,6 +106,7 @@ if __name__ == '__main__':
     parser.add_argument('--floodplain_elevation', action='store',
                         type=float, default=20.0)
     parser.add_argument('--inject_bathymetry', action='store_true')
+    parser.add_argument('--geometry', default='sphere')
     cl_args = parser.parse_args()
     build_mesh(cl_args.preserve_floodplain, cl_args.floodplain_elevation,
-               cl_args.inject_bathymetry)
+               cl_args.inject_bathymetry, cl_args.geometry)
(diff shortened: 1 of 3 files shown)


matthewhoffman:
@xylar , @sbrus89 , these are changes I made to allow me to use the existing ocean jigsaw_to_mpas tools from within the landice side of compass for making landice meshes.  I've tested that the changes work for my use, but I have not tested existing ocean test cases.  

If there is a different way you would prefer to achieve the same functionality, we can discuss it.

cc: @pwolfram

xylar:
@matthewhoffman, at a glance, this looks good to me.  We'll need to re-run several ocean test cases that use this to make sure nothing changes, as you say.  I'll be happy to help with that when I get back.

Does it make sense to move `jigsaw_to_MPAS` outside of `ocean` if it is used across components?  If so, how do we propagate that change to `develop` and `landice/develop`?  Currently, we have `utility_scripts` and this infrastructure doesn't quite seem to qualify but we could rename that folder to `shared` and put `jigsaw_to_MPAS` inside that.

matthewhoffman:
@xylar , thanks for the feedback.  I think we might have discussed moving it out of ocean at one point.  I agree it would be slightly cleaner to do so, but I'm also fine with leaving it as it is for simplicity.  I found it was easy enough to make use of it from the landice side like:
```
<add_link source_path="mpas_model" source="testing_and_setup/compass/ocean/jigsaw_to_MPAS" dest="jigsaw_to_MPAS"/>
```
So it's fine with me if we let the ocean own it.

mark-petersen:
@matthewhoffman Thanks for this contribution to the ocean COMPASS cases. I can test with our nightly test suite, and merge it with the next round of COMPASS PRs. Is it waiting for anything now, other than a such a test?

xylar:
Maybe waiting for me to rebuild COMPASS envs?

matthewhoffman:
@mark-petersen & @xylar , at a minimum this is waiting on testing all the ocean COMPASS cases to make sure nothing breaks.  I agree with Xylar, that that might as well wait until his new COMPASS envs tomorrow.  Beyond that, you all are welcome to review or test further as you see fit.  But I think it's pretty straightforward and transparent.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
```diff
@@ -31,32 +31,49 @@
...
+       cw_filename='cellWidthVsLatLon.nc'
+       da.to_netcdf(cw_filename)
+    else:
+       cellWidth, x, y, geom_points, geom_edges = define_base_mesh.cellWidthVsXY()
+       da = xarray.DataArray(cellWidth,
+                          dims=['y','x'],
+                          coords={'y':y,'x':x},
+                          name='cellWidth')
+       cw_filename='cellWidthVsXY.nc'
+       da.to_netcdf(cw_filename)
```
sbrus89: `da.to_netcdf(cw_filename)` could probably be moved outside the if structure here so it's only in one place, but it's also totally fine the way it is.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_meshDensity.py
```diff
@@ -16,48 +16,63 @@
...
+       meshDensity[:] = meshDensityInterp(
+           np.vstack(
+            (ds.variables['lonCell'][:],
              np.mod(
-                ds.variables['lonCell'][:] +
+                ds.variables['latCell'][:] +
                 np.pi,
                 2 *
                 np.pi) -
                 np.pi)).T)
```
sbrus89: It seems like there's an ocean-related change here. I feel like the order should have been `np.vstack((ds.variables['lonCell'][:], ds.variables['lonCell'][:]))` like it is now, but the `np.mod()` should still operate on `ds.variables['lonCell'][:]` (I think). @mark-petersen, do you have an opinion here?

testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_meshDensity.py
```diff
@@ -16,48 +16,63 @@
...
 
     print('Open unstructured MPAS mesh file...')
     ds = nc4.Dataset(mesh_filename, 'r+')
     meshDensity = ds.variables['meshDensity']
 
-    print('Preparing interpolation of meshDensity from lat/lon to mesh...')
+    print('Preparing interpolation of meshDensity from native coordinates to mesh...')
     meshDensityInterp = interpolate.LinearNDInterpolator(
-        np.vstack((LAT.ravel(), LON.ravel())).T, meshDensityVsLatLon.ravel())
+        np.vstack((X.ravel(), Y.ravel())).T, meshDensityVsXY.ravel())
```
sbrus89: I like the generalization here.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_meshDensity.py
matthewhoffman: @sbrus89 , to be clear, I don't think I changed the meaning of this operation.  The only thing that changed from the original version was the order of latCell and lonCell was flipped to match the new convention I introduced of X,Y for both planar and sphere coordinates.  Here is what the two versions of the code look like without the confusing diff format:

Old:
```
    meshDensity[:] = meshDensityInterp(
        np.vstack(
            (ds.variables['latCell'][:],
             np.mod(
                ds.variables['lonCell'][:] +
                np.pi,
                2 *
                np.pi) -
                np.pi)).T)
```

New:
```
       meshDensity[:] = meshDensityInterp(
           np.vstack(
            (ds.variables['lonCell'][:],
             np.mod(
                ds.variables['latCell'][:] +
                np.pi,
                2 *
                np.pi) -
                np.pi)).T)
```

Ok, now that I write it out, it looks you are right that the mod operation and subsequent subtraction of pi needs to stay with lonCell.  An alternative would be to use Y,X format for both sphere and planar meshes (and keep this confusing operation unchanged).  Let me know what you prefer and I can update the PR.

sbrus89: @matthewhoffman, I probably prefer the X,Y convention as you have it (as long as the mod() goes with lonCell like you mentioned). However, I know @mark-petersen wrote this so it's possible he has a reason to keep it Y,X. I think as long as the order of the coordinates used in setting up the interpolation object are consistent with the order with which it is called, it shouldn't matter. Personally, X,Y is more intuitive.

matthewhoffman: @sbrus89 , I've pushed a commit that corrects the arguments in the call to meshDensityInterp.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
matthewhoffman: I considered that but thought it might look slightly less readable if that line was dangling after the if-construct.  I figured this way each branch is fully self-contained.

sbrus89: Perfect, I agree. I hesitated to even bring it up.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_meshDensity.py
sbrus89: Looks good to me!

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
xylar: It's not a great idea to create a variable (`da` in this case) within an if/else and then use it outside unless it's really necessary so I agree that it's better to leave as it is.

matthewhoffman:
I expect changes to be BFB for the ocean test cases.  If not, I would like to understand why they differ before this is merged.

xylar:
> The nightly regression suite includes both spherical and planar cases

To be clear, I don't think any planar ocean test cases currently use jigsaw, right?  That wouldn't have been possible before @matthewhoffman's changes.

mark-petersen:
Correct. Planar cases use pre-made mesh files.

xylar:
I just wanted to make clear that the planar capability was not tested by the ocean regression suite.

PR #392, opened 2019-11-10
Fix python 3 in BGC and CVMix makedep.py

xylar:
This is needed for E3SM builds of `maint-1.0` with python 3.

PR #395, opened 2019-11-12
Switch to getting Ocean-BGC and CVMix form hashes instead of branches

xylar:
In my last PR (#392), I was downloading Ocean-BGC and CVMix from their respective `maint-1.0` branches.  In the discussion of https://github.com/E3SM-Project/E3SM/pull/3308, it became clear there was concern about this approach, so I have switched to commit hashes instead.

### DIFF
--- a/src/core_ocean/get_BGC.sh
+++ b/src/core_ocean/get_BGC.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ## BGC Tag for build
-BGC_TAG=maint-1.0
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I was able to download the source for Ocean-BGC and CVMix and build successfully on my laptop.  I did not do any test runs.

xylar:
@mark-petersen and @jonbob, could you review at your earliest convenience?

PR #398, opened 2019-11-12
Correctly assign/discard soundings in regional MPAS-A and on concave decompositions

MiCurry:
This pull requests contains changes for fixing two bugs related to sounding locations in MPAS-A. 

The first fix, fixes soundings being wrongly assigned to the wrong cell on a concave decomposition using the nearest_cell function. The first commit adds a K-Dimensional Tree which circumvents nearest_cell's flaw and correctly assigns the sounding to the correct cell, regardless of the decomposition.

The next fix, discards soundings that lie outside of a simulation's region. Using the KD-Tree and the in_cell routine. The last commit adds logic to communicate to the user the sounding discard using MPI. WIthout this MPI communication, a sounding may be 'discarded' when it is in fact in the region, but owned by another task.

@mgduda I didn't realize this was two bug fixes in one. We could easily remove the KD-Tree and use nearest_cell and the in_cell routine to discard soundings outside of the region and then make another PR using the KD fixing the concave decomposition for the next release.

### DIFF
--- a/src/core_atmosphere/diagnostics/Makefile
+++ b/src/core_atmosphere/diagnostics/Makefile
@@ -9,6 +9,7 @@ DIAGNOSTIC_MODULES = \
 	convective_diagnostics.o \
 	pv_diagnostics.o \
 	soundings.o \
+	mpas_kd_tree.o
 
 isobaric_diagnostics.o: mpas_atm_diagnostics_utils.o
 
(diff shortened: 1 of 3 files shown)


mgduda:
Since we use the `in_cell`, `mirror_point`, and `rotate_about_vector` routines in the `init_atmosphere` core as well, perhaps I should make a PR to the `operators` to add these three routines there. Then, we can update this PR to use those routines from `mpas_geometry_utils`.

MiCurry:
The KD Tree should either be updated with the tie-break code from #630, symlinked with `init_atmosphere/mpas_kd_tree.F90` or the `mpas_kd_tree.F90` should be placed somewhere inside of the framework directory.

PR #401, opened 2019-11-14
Update README for COMPASS ocean with latest packages

xylar:
Update versions of `geometric_features` and `mpas_tools` following recent changes.


xylar:
@mark-petersen and @matthewhoffman, this is just do document the COMPASS environments I plan to create tomorrow to replace the ones on cori, anvil and IC that we currently use.

xylar:
@mark-petersen, you'll merge this according to E3SM scheduling, I guess?

mark-petersen:
yes, later today or tomorrow. I need to do a few things with framework merges first.

PR #402, opened 2019-11-15
MALI: Several Improvements related to handling Albany velocity solver

mperego:
Several modifications to improve MPAS/Albany interface 

1. Modifications to interface to simplify the code, reduce communications and improve the partitioning FE mesh to avoid critical situations for the solvers:
   - use MPAS global ids for FE elements entities (years ago we created contiguous global IDs for the FE meshes to avoid some issues with solvers that have been overcome)
   - change ownership of FE triangles (Voronoi vertices) to better distribute triangles along the interface between processors
   - change ownership of FE nodes (Voronoi cells) to avoid corner situations like having a processor that owns a node but no triangles containing that node

2.  Allow building shared libraries for Albany and Trilinos

3.  Enable use of Prismatic FE elements (in alternative to tets).

4. Remove obsolete code related to LIFEV library

5. Modify writing of ascii mesh so that it can be printed in parallel

6. Fix logic to determine whether the finite element mesh needs to be recomputed. Note: in order to compare the cell mask with the previous time steps, we would need the cell mask to have 2 time levels. However, if we set the cell mask to have 2 time levels we get different results. This needs to be fixed. At the moment the cell mask is set to have 1 time level, so it might happen that the mesh is not recomputed when it should.


matthewhoffman:
@mperego , I'm finally starting to review this now.  One thing I am concerned about is commit b514e40150100e499cd7ae0c7ed7df9270a38e7c where there still is some work to do with the time levels of the cellMask field.  I propose that we remove that commit from this PR and deal with it later.  That way trying to resolve that issue does not hold up merging the rest of this PR.  If you are ok with that, I can modify the branch to have that commit removed, and then open a new PR with just that commit that we can finish working on later.

mperego:
Sure, It makes sense to me.

matthewhoffman:
Note: Closing and reopening as #452 in order to fix branch names and update commit history at the same time.

matthewhoffman:
Also opening #453 separately for item 6 in the list above.

PR #403, opened 2019-11-17
Fix bug in landice.cmake

jgfouca:
Fix mistake in landice.cmake that was causing Albany-enabled compsets to fail to build in E3SM.

I am unsure how this obvious mistake made it through my initial rounds of testing.

### DIFF
--- a/src/core_landice/landice.cmake
+++ b/src/core_landice/landice.cmake
@@ -71,7 +71,7 @@ list(APPEND RAW_SOURCES
   core_landice/mode_forward/mpas_li_subglacial_hydro.F
 )
 
-if (CPPFLAGS MATCHES ".*MPAS_LI_BUILD_INTERFACE.*")
+if (CPPDEFS MATCHES ".*MPAS_LI_BUILD_INTERFACE.*")
   list(APPEND RAW_SOURCES core_landice/mode_forward/Interface_velocity_solver.cpp)
 endif()
 


jgfouca:
Note that SMS.f09_g16_a.MALI will fail on the E3SM dashboard until this gets merged, but I don't think that is urgent.

matthewhoffman:
@mgduda , I am thinking the best workflow would be to merge this branch into develop and then I merge develop into e3sm/develop.  Github doesn't let you change the destination branch for a PR, so I'd have to close and open a new PR, or just merge without a PR.  Let me know if you have a preference.  I was going to skip code review for this, because it is an obvious one-line change in the landice dir and only affects the cmake build system.

xylar:
> Github doesn't let you change the destination branch for a PR, so I'd have to close and open a new PR, or just merge without a PR.

You can just hit the edit button and change the destination.

mark-petersen:
@matthewhoffman I agree, and just changed it to merge into develop.

mgduda:
@mark-petersen Agreed -- this only affects the CMake build in the landice directory.

mark-petersen:
@jgfouca this is now on e3sm/develop as well. I will make an E3SM PR to update mpas-source in the next few days.

jgfouca:
@mark-petersen thanks!

Issue #394, opened 2019-11-12
Do we need local versions of config flags?


mark-petersen:
In almost every MPAS-Ocean module, config flags are transferred to private module variables. For example, in the ocean core, for del2, we have:
```
   ! Private module variables
   logical :: del2On
   real (kind=RKIND) :: eddyDiff2
```
and on init:
```
   subroutine ocn_tracer_hmix_del2_init(err)!{{{
...
      logical, pointer :: config_use_tracer_del2
      real (kind=RKIND), pointer :: config_tracer_del2
...
      call mpas_pool_get_config(ocnConfigs, 'config_use_tracer_del2', config_use_tracer_del2)
      call mpas_pool_get_config(ocnConfigs, 'config_tracer_del2', config_tracer_del2)

      del2on = .false.

      if ( config_use_tracer_del2 ) then
         if ( config_tracer_del2 > 0.0_RKIND ) then
            del2On = .true.
            eddyDiff2 = config_tracer_del2
         endif
      endif
```
Is there any performance reason to do this? It seems like a lot of extra bookkeeping, and extra variables obscure code. Options are:
1. Do it this way (use extra variables with different names)
2. Call `mpas_pool_get_config` for config flags every time the main routine is called (i.e. every time step)
3. Define those `config_*` variables as private module variables and fill them in on init
4. Obtain ALL config_* variables ONCE on init in some new module (say ocn_configs) and then have every module `use ocn_configs` at the top and use config variables by name. Then we would never declare config flags or call `mpas_pool_get_config` within these normal modules.

I'm asking now because I'm adding new flags for Redi mixing, but it is relevant to all modules. Any of 1-3 we could alter piecemeal. Number 4 (which I prefer) would need a larger coordinated change.

I remember from @philipwjones talk that evaluating strings in if statements is expensive, so there may be two cases here, one for config flags that are strings, and one for flags that are already logicals, integers, or real.

amametjanov:
+1 on doing it once with option 4. Not sure if there's a reason for the current method: module-private vars still depend on run-time input namelists: i.e. they's not compile-time constant `parameter` vars.

philipwjones:
We do not want options 1,2 as they are likely causing some unnecessary performance overhead doing the character matching every time step.  We definitely want to set them on init and not keep retrieving them from config. 

[...]

PR #406, opened 2019-12-03
Diagnostics: Move mpas_pool_get_config to init

mattdturner:
This PR moves all calls in `mpas_ocn_diagnostics.F` to `mpas_pool_get_config` from individual subroutines to the `ocn_diagnostics_init` subroutine.  This somewhat cleans up the individual subroutines, and prevents the calls to `mpas_pool_get_config` from unnecessarily being called every timestep.  This required moving the declarations of `config_*` variables to module-level variables

The PR also replaces all calls to `mpas_allocate_scratch_field` with straight allocates.  This should enhance the performance by reducing the amount of pointer checks.  This also necessitated moving certain variables from subroutine-level to module-level variables.  These variables can/should be moved back to subroutine level if the OpenMP implementation is modified to be more local (i.e., open the OpenMP parallel region around loops instead of before the timestep loop).

By opening this PR, I am hoping to get feedback on the modifications prior to making similar changes in other files.

This begins to address issue #394

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -1955,8 +1955,6 @@
 			<var name="iceFraction"/>
 			<var name="windSpeed10m"/>
 			<var name="nAccumulatedCoupled"/>
-			<var name="thermalExpansionCoeff"/>
-			<var name="salineContractionCoeff"/>
 			<var name="landIceFraction"/>
 			<var name="landIceMask"/>
 			<var_array name="landIceInterfaceTracers"/>
@@ -3332,53 +3330,6 @@
 			 units="m"
 			 description="total thickness at edge, sum(h)"
 		/>
-		<var name="vorticityGradientTangentialComponent"
-			persistence="scratch"
-			type="real" dimensions="nVertLevels nEdges Time"
-			units="s^{-1} m^{-1}"
-			description="gradient of vorticity in the tangent direction (positive points from vertex1 to vertex2)"
-		/>
-		<var name="vorticityGradientNormalComponent"
-			persistence="scratch"
-			type="real" dimensions="nVertLevels nEdges Time"
-			units="s^{-1} m^{-1}"
-			description="gradient of vorticity in the normal direction (positive points from cell1 to cell2)"
-		/>
-		<var name="normalizedRelativeVorticityVertex"
-			persistence="scratch"
-			type="real" dimensions="nVertLevels nVertices Time" units="s^{-1}"
-			description="curl of horizontal velocity divided by layer thickness, defined at vertices"
-		/>
-		<var name="normalizedPlanetaryVorticityVertex"
-			persistence="scratch"
-			type="real" dimensions="nVertLevels nVertices Time" units="s^{-1}"
-			description="earth's rotational rate (Coriolis parameter, f) divided by layer thickness, defined at vertices"
-		/>
-		<var name="kineticEnergyVertex"
-			persistence="scratch"
-			type="real" dimensions="nVertLevels nVertices Time" units="m^2 s^{-2}"
-			description="kinetic energy of horizonal velocity defined at vertices"
-		/>
-		<var name="kineticEnergyVertexOnCells"
-			persistence="scratch"
-			type="real" dimensions="nVertLevels nCells Time" units="m^2 s^{-2}"
-			description="kinetic energy of horizonal velocity defined at vertices"
-		/>
-		<var name="densitySurfaceDisplaced"
-			persistence="scratch"
-			type="real" dimensions="nVertLevels nCells Time" units="kg m^{-3}"
-			description="Density computed by displacing SST and SSS to every vertical layer within the column"
-		/>
-		<var name="thermalExpansionCoeff"
-			 persistence="persistent"
-			 type="real" dimensions="nVertLevels nCells Time" units="C^{-1}"
-			 description="Thermal expansion coefficient (alpha), defined as $-1/\rho d\rho/dT$ (note negative sign)."
-		/>
-		<var name="salineContractionCoeff"
-			 persistence="persistent"
-			 type="real" dimensions="nVertLevels nCells Time" units="PSU^{-1}"
-			 description="Saline contraction coefficient (beta), defined as $1/\rho d\rho/dS$. This is also called the haline contraction coefficient."
-		/>
 
 		<!-- FIELDS FOR TENSOR OPERATIONS -->
 		<var name="normalVelocityTest"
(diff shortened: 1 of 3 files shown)


mattdturner:
Regarding testing...
Moving the calls to `mpas_pool_get_config` from subroutines to initialization resulted in no performance change and bit-for-bit results (for 10-day QU240 simulations on 1 node on Cori-haswell).

Replacing the calls to `mpas_allocate_scratch_field` with straight allocates is not a bit-for-bit change.  Tests were run on 1 node with 10-day QU240 simulations and 2-day EC60to30 simulations.  Simulations run without optimization (i.e., replacing `-O3` with `-O0` in the Makefile) on Cori-haswell are bit-for-bit.  The modifications, though, allow the compiler to have additional information about the arrays, which allows the compiler to perform additional optimizations.  As a result, it is highly likely that the differences are all related to rounding differences.

For a 2-day EC60to30 simulation, the differences (maximum difference) for a select number of variables for a 16-core, 1-thread simulation are shown below:

```
layerThickness:
- Base: 9.74852
- Test: 9.74754
normalVelocity:
- Base: 1.39803359917e-06
- Test: -2.34341186148e-05
ssh:
- Base: -0.502952
- Test: -0.504905
temperature:
- Base: 13.63883
- Test: 13.96296
```

The performance difference for these changes were mostly negligible for a 16-core EC60to30 simulation (0.7%, `time integration` timer, average of 3 simulations).  At higher core counts, I would expect the performance difference to be larger since the reduction in pointer checks would remain constant, but would become a larger percentage of the overall simulation time.

philipwjones:
@mattdturner There is one bug - the blTempScratch and blSaltScratch have a an allocate where there should be a deallocate (looks like a cut-and-paste error). Also, the tracer index for T,S are also static so could be moved up into init too.  

We probably don't need to do this now, but I would like to replace the scalar pointers with just scalar copies. Some compilers have choked on these in the past and in many cases, there is little saving on storage. But a better solution would be to have the config retrievals return a copy instead of a pointer so this could be a framework change later. Similarly, I think the character variables need to be converted to an enum choice at some point, but again that belongs in the relevant module, not in this higher level routine.

mattdturner:
> There is one bug - the blTempScratch and blSaltScratch have a an allocate where there should be a deallocate (looks like a cut-and-paste error).

Thanks for catching that.  It is indeed a cut-paste error, but I'm surprised the compiler didn't throw an error.  I'll push a fix shortly.

mattdturner:
> the blTempScratch and blSaltScratch have a an allocate where there should be a deallocate (looks like a cut-and-paste error). Also, the tracer index for T,S are also static so could be moved up into init too.

@philipwjones Both of these issues should now be resolved.

mark-petersen:
@mattdturner Thanks for your work on this. Today is packed, but I'll review and get back to you soon.

mark-petersen:
@mattdturner Some notes on our conventions. We use a standard git commit log message format with a short first line, and optional longer description below, like this:
```
commit message, max 50 characters
<blank line>
longer text, optional
```
That is so this wonderful command:
```
git log --graph --oneline -n 12
```
is readable, and `git log` provides more details. Don't worry about revising this PR, it's FYI. Note that `git commit -m "message"`  just creates the first line, while no `-m` sends you to the text editor.

mattdturner:
> Note that `git commit -m "message"` just creates the first line, while no `-m` sends you to the text editor.

Ah, ok.  I guess I'll stop using `git commit -m` for commits.

mark-petersen:
Tested on grizzly LANL IC for all four: gnu/intel and debug/optimized and openMP. Used nightly regression suite, all tests pass and are bfb with previous on both gnu and intel using optimized, but with OMP_NUM_THREADS=1.

mark-petersen:
Also matching bfb with previous (ocean/develop) using `OMP_NUM_THREADS=2`. Must be a machine and optimize-specific thing on cori. Let's not worry about it, since your non-optimized match.

mattdturner:
>  Feel free to make PRs along the way as you see fit.

@mark-petersen Sounds good.  I'll go ahead and replicate these changes throughout the rest of the MPAS-O code.

mark-petersen:
@mattdturner before we merge, please search for the scratch variables that you've removed like this:
```
cd src/core_ocean
git grep -in boundaryLayerTemperature
Registry.xml:2867:                      <var name="landIceBoundaryLayerTemperature" array_group="landIceBoundaryLayerValues" units="C"
Registry.xml:3182:                      <var name="avgLandIceBoundaryLayerTemperature" array_group="landIceBoundaryLayerValues" units="C"
Registry.xml:3630:              <var name="boundaryLayerTemperatureScratch"
shared/mpas_ocn_diagnostics.F:1673:      type (field1DReal), pointer :: boundaryLayerTemperatureField, boundaryLayerSalinityField
shared/mpas_ocn_diagnostics.F:1739:      call mpas_pool_get_dimension(diagnosticsPool, 'index_landIceBoundaryLayerTemperature', indexBLT)
shared/mpas_ocn_surface_land_ice_fluxes.F:427:      type (field1DReal), pointer :: boundaryLayerTemperatureField, boundaryLayerSalinityField, &
shared/mpas_ocn_surface_land_ice_fluxes.F:447:      call mpas_pool_get_dimension(diagnosticsPool, 'index_landIceBoundaryLayerTemperature', indexBLT)
```
ignore the land ice ones, but this helps you delete all the lines associated with these removed variables. The 'Field' variables should be deleted as well, e.g. `boundaryLayerTemperatureField`. It looks like `shared/mpas_ocn_surface_land_ice_fluxes.F:427` got left there by mistake, and can be deleted too.

In the Registry, delete the whole entry:
```
3630         <var name="boundaryLayerTemperatureScratch"
3631             persistence="scratch"
3632             type="real" dimensions="nCells" units="^\circ C"
3633             description="temperature averaged vertically over the sub-ice-shelf boundary layer"
3634         />
```
But it is helpful to keep the description. So now that it is a local scratch variable, add that description and units to the first place you declare that scratch variable. Thanks!

mark-petersen:
@philipwjones please approve when you are ready. This can be merged after a few others, probably next week.

mattdturner:
I removed the scratch variables from the Registry, and deleted the now-unused variables in `mpas_ocn_diagnostics.F`.  I also added the description and units as comments when each scratch variable is declared.

In the Registry.xml file, `thermalExpansionCoeff` and `salineContractionCoeff` were both defined as `persistent` instead of `scratch` variables.  I'm not sure if that would make it a problem to delete them, but the model compiles and runs successfully with DEBUG enabled.

mattdturner:
Superseded by #457

PR #409, opened 2019-12-11
Write two new include files for config flags

mark-petersen:
These files are now created on compile:
- config_declare.inc: Declaration line for every variable in config pool.
- config_get.inc: mpas_pool_get_config line for every variable in config pool.

These additions require no changes for the cores, but it provides the option to include these files in a config module upon model initialization, so that one never needs to declare or get config variables anywhere else.

With this addition, all cores will create these files upon compile. Increase in compile time is extremely minimal, and there is no impact on run time.

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -530,7 +530,7 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 
 	int in_subpool;
 
-	FILE *fd, *fd2;
+	FILE *fd, *fd2, *fcd, *fcg;
 
 	const_core = ezxml_attr(registry, "core_abbrev");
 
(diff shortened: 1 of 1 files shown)


mark-petersen:
See example of how to use this in the ocean core at #410. These changes will simplify all the bookkeeping required in writing MPAS code, as any config flag would now be available in any module without individual declarations.

mark-petersen:
@mgduda @matthewhoffman @akturner 
I merged this into `develop` locally and compiled with seaice, landice, and atmosphere cores. It produces the two include file as expected and causes no problems for compilation.

If you have time to review this week, that would be helpful.

mgduda:
@mark-petersen Sorry for the delay on my end. I'll review this more thoroughly in the next day or two. My initial impression is that this seems fine, and that we might just want to consider how to handle these sorts of extensions to the code generation in future.

mgduda:
For future reference, here's a minimal Fortran module that makes use of the two new files:
```fortran
module core_configs

    use mpas_kind_types, only : RKIND, StrKIND

    implicit none

#include "config_declare.inc"

contains

    subroutine core_configs_setup(configPool)

        use mpas_derived_types, only : mpas_pool_type
        use mpas_pool_routines, only : mpas_pool_get_config

        type (mpas_pool_type), intent(in) :: configPool

#include "config_get.inc"
    
    end subroutine core_configs_setup

end module core_configs
```

mark-petersen:
@matthewhoffman and @akturner do you mind if I merge? Then I can start cleaning up config flags in the ocean core.

mark-petersen:
Thanks @mgduda for your review and concise module above. Since this is a 'no harm' addition, and hearing no objections, I'll go ahead and merge.

Issue #228, opened 2019-05-04
Parallel MPAS-O build


pwolfram:
Previously, I believe we could have built in parallel with a `-j 6` option (thanks to @douglasjacobsen `b277e03d17c94bb320138a5061b019f935a4b4df`).  However, this is no longer the case and a failed build occurs when attempting to build in parallel.  There are many parts of our build process that are independent and it may be worth ensuring parallel builds again for more rapid development.

cc @mark-petersen, @xylar and @philipwjones

PR #410, opened 2019-12-11
Autogenerate ocean config declarations and get commands

mark-petersen:
Add a new module ocn_config that has public declarations of all config flag variables. These are autogenerated by the framework (see #409) and added to mpas_ocn_config.F using:
```
#include "../inc/config_declare.inc"
#include "../inc/config_get.inc"
```

With these changes, we may delete config lines such as:
```
real (kind=RKIND), pointer :: config_tracer_del2
call mpas_pool_get_config(ocnConfigs, 'config_tracer_del2', config_tracer_del2)
```
from all other modules. All variables in the ocnConfig pool will be available by simply adding
```
use ocn_config
```
to the top of the module. Importantly, this change can be done to the modules one-by-one. This PR includes the changes to `mpas_ocn_tracer_hmix_del2.F` as an example.

### DIFF
--- a/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
+++ b/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
@@ -37,6 +37,7 @@ module ocn_analysis_mode
    use ocn_diagnostics
    use ocn_equation_of_state
    use ocn_constants
+   use ocn_config
 
    private
 
@@ -87,8 +88,11 @@ module ocn_analysis_mode
       call mpas_get_time(startTime, dateTimeString=startTimeStamp)
       ierr = ior(ierr, err_tmp)
 
-      ! Setup ocean config pool
+      ! Set up ocean constants
       call ocn_constants_init(domain % configs, domain % packages)
+
+      ! Setup ocean config pool
+      call ocn_config_init(domain % configs)
 
       !
       ! Read input data for model
--- a/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
@@ -81,6 +81,7 @@ module ocn_forward_mode
    use ocn_time_varying_forcing
 
    use ocn_constants
+   use ocn_config
 
    implicit none
    private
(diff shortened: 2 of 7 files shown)


mark-petersen:
After #409 is merged I will remove the framework changes from this PR and we can proceed.

mark-petersen:
Note the example in this PR:
https://github.com/MPAS-Dev/MPAS-Model/blob/b1e7065ad09505d1191580a85c34442aa3483681/src/core_ocean/shared/mpas_ocn_tracer_hmix.F#L29
That's all it takes, and then all config declarations and 'get' lines can be deleted in that module.

This PR was tested with both gnu and intel, debug and optimized. It passes bfb with ocean/develop, and that includes obtaining the del2 config flags this new way.

mattdturner:
Looks good.  I like the simplicity of only needing to add the line `use ocn_config`.  Once #409 and this PR are merged I can modify the rest of the MPAS-O subroutines to use the new module file instead of calling `mpas_pool_get_config`.  Unless you disagree, I think it would be fine to include those modifications in the same PR as my work replacing `mpas_allocate_scratch_field`.

mattdturner:
As an aside regarding `src/core_ocean/shared/Makefile`, is there a reason why the dependencies are manually defined?  It seems like this occasionally causes problems were dependencies aren't updated (see #228 ).  Should these eventually be replaced with auto-generated dependencies (obviously not as part of this PR)?

mark-petersen:
>I think it would be fine to include those modifications in the same PR as my work replacing mpas_allocate_scratch_field

Yes, I agree.

mark-petersen:
> is there a reason why the dependencies are manually defined? 

To my knowledge, a simple Makefile requires us to specify dependancies. I agree that is a headache. 

@jgfouca added cmake capability in #386. It only works for E3SM right now, but we hope to get cmake to work for stand-alone soon. 

If you look at his new ocean.cmake file:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/ocean.cmake
it appears that cmake auto-detects the dependancies from the `use` lines in each module, as they are not specified. @jgfouca is that correct?

mattdturner:
>  but we hope to get cmake to work for stand-alone soon.

Ah, ok.  I wasn't aware that there were plans (or desire) to get cmake to work for stand-alone MPAS-O.  I know cmake is capable of automatically generating the dependencies, so that will be good once cmake is working for stand-alone.

jgfouca:
> cmake auto-detects the dependencies from the use lines in each module

Correct.

mark-petersen:
@mattdturner, framework changes are now in MPAS, and staged for E3SM. I rebased this, and will merge soon. The MPAS PRs have to wait in line for E3SM merges, so they don't pile up in case there is a problem.

mark-petersen:
Rebased, and retested with intel debug and gnu on grizzly. bfb with previous.

PR #414, opened 2019-12-11
Change output to single precision in stand-alone default streams files

mark-petersen:
This adds the flag
```
precision="single"
```
to the default streams file for output files. It does not alter restart streams. Bit-for-bit testing in the nightly regression remains 8-byte reals because that is specified in compass.

The purpose of adding single precision to the default streams is to remind users that single precision is sufficient and would typically be used for output. This does not affect E3SM, as the streams files are created from a script there. We are changing to single precision analysis output there as well, see https://github.com/E3SM-Project/E3SM/pull/3360

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -1632,6 +1632,7 @@
 
 		<stream name="output"
 				type="output"
+				precision="single"
 				filename_template="output/output.$Y-$M-$D_$h.$m.$s.nc"
 				filename_interval="01-00-00_00:00:00"
 				reference_time="0001-01-01_00:00:00"
@@ -1747,6 +1748,7 @@
 		<!-- Diagnostics streams for forward mode -->
 		<stream name="additional_output"
 				type="none"
+				precision="single"
 				filename_template="output/additional_output.$Y-$M-$D_$h.$m.$s.nc"
 				filename_interval="01-00-00_00:00:00"
 				reference_time="0001-01-01_00:00:00"
--- a/src/core_ocean/analysis_members/Registry_TEMPLATE.xml
+++ b/src/core_ocean/analysis_members/Registry_TEMPLATE.xml
@@ -43,6 +43,7 @@
 	<streams>
 		<stream name="temPlateOutput" type="output"
 				mode="forward;analysis"
+				precision="single"
 				filename_template="analysis_members/temPlate.$Y-$M-$D.nc"
 				filename_interval="01-00-00_00:00:00"
 				output_interval="00-00-01_00:00:00"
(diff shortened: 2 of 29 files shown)


mark-petersen:
@vanroekel I think a quick look is sufficient. This does not alter any simulations, but is more of a reminder for our reference. It passes the nightly regression suite.

mark-petersen:
Compiled, ran nightly regression suite with gnu. This does not change any COMPASS streams files, because those are specified by compass, see files like:
```
ocean/templates/analysis_members/*
ocean/global_ocean/QU240/daily_output_test/config_test.xml
```

This PR changes the auto-generated `streams.ocean` and `streams.ocean.forward` files.

xylar:
@mark-petersen, this will affect `compass` 1.0, which does not use the templates you point to.  That may be okay, and we could explicitly override the defaults for regression testing.  But in general we would need to be careful that regression testing doesn't become single precision because of this.

It looks like you left global stats at double precision?  If so, I think that is necessary for accurately computing budgets and such.

mark-petersen:
Yes, globalStats remains at double precision, as i/o size is not a concern, and it is useful for a quick view of loss of accuracy and simulation differences.

That is good to know about compass 1.0. My preference is still to add single precision as the default. For the nightly regression, all bfb comparisons should use double precision. We could either compare restart files, or have an output template that compares only prognostic variables at double precision. In the current compass that is here:
```
ocean/templates/streams/minimal_output.xml
<template>
	<streams>
		<stream name="output">
			<attribute name="type">output</attribute>
			<attribute name="filename_template">output.nc</attribute>
			<attribute name="output_interval">0000_00:00:01</attribute>
			<attribute name="clobber_mode">truncate</attribute>
			<add_contents>
				<member name="tracers" type="var_struct"/>
				<member name="xtime" type="var"/>
				<member name="normalVelocity" type="var"/>
				<member name="layerThickness" type="var"/>
			</add_contents>
		</stream>
	</streams>
</template>
```

xylar:
Regression testing uses bunches of different kinds of output depending on the test.  But it would be easy to make sure there's some sort of sanity check that `precision="double"` is part of any streams used for validation.  For example, we could check the type of the variables to make sure they're not single as part of validation (perhaps with a flag to allow single precision in unusual circumstances).

Are you sure that this won't mess up regression testing in legacy COMPASS?  For example will the frazil output, land-ice fluxes, etc., etc. still be double precision in COMPASS?  I'm not so sure that we use templates like you list above for many test cases in the regression suites.

I will do some testing to make sure this doesn't have any adverse consequences as part of my review.

mark-petersen:
Using compass, I compiled this branch and created a nightly regression suite. In there,
```
grep precision ocean/*/*/*/*/streams.ocean
```
does not find anything. It appears that current compass only uses the compass templates, not the auto-generated one.

xylar:
> It appears that current compass only uses the compass templates, not the auto-generated one.

Oh, interesting!  I somehow assumed otherwise.  I think we need to check that for the other test suites (but we don't need to run them).  I can do that.

xylar (approved):
I crated all test cases in all ocean test suites and none picks us the `precition="single"` attribute.  I think we're good.

PR #415, opened 2019-12-12
Create small patch domain on a sphere

mark-petersen:
In COMPASS, create a domain on the sphere that is simply all the cells within a lat/lon square.

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/patch/init/cull_mesh_patch.py
+++ b/testing_and_setup/compass/ocean/global_ocean/patch/init/cull_mesh_patch.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python
+"""
+This script performs the first step of initializing the global ocean.  This
+includes:
+1. combining Natural Earth land coverage north of 60S with Antarctic
+   ice coverage or grounded ice coverage from Bedmap2.
+2. combining transects defining critical passages.*
+3. combining points used to seed a flood fill of the global ocean.
+4. create masks from land coverage.
+5. add land-locked cells to land coverage mask.
+6. create masks from transects.*
+7. cull cells based on land coverage but with transects present
+8. create flood-fill mask based on seeds
+9. cull cells based on flood-fill mask
+10. create masks from transects on the final culled mesh*
+* skipped if flag --with_critical_passages not present
+
+Optionally, the -p flag provides the path to the geometric_data
+directory from the geometric_features repository, which is assumed
+to be the current directory by default. Also, the optional
+--with_cavities flag indicates that ice-shelf cavities are present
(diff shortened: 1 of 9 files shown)


testing_and_setup/compass/ocean/global_ocean/patch/init/cull_mesh_patch.py
```diff
@@ -0,0 +1,64 @@
...
+landMask = numpy.logical_and(numpy.logical_and(lat >= minlat, lat <= maxlat),
+                             numpy.logical_and(lon >= minlon, lon <= maxlon))
+
+# Convert logical to int
+landMask = landMask.astype(int)
+
+dsLandMask = xarray.Dataset()
+dsLandMask['cullCell'] = landMask
+
+dsCulledMesh = conversion.cull(dsBaseMesh, dsMask=dsLandMask,
```
mark-petersen: @xylar, when I run this case: `-o ocean -c global_ocean -r patch -t init` it runs through successfully and creates `culled_mesh.nc`, but no cells are culled. Perhaps you can see why. Is this dsMask argument looking for an array called `cullCell` within the Dataset called dsLandMask? I'm not very familiar with xarray.

testing_and_setup/compass/ocean/global_ocean/patch/init/cull_mesh_patch.py
```diff
@@ -0,0 +1,63 @@
...
+oceanMask = numpy.logical_and(numpy.logical_and(lat >= minlat, lat <= maxlat),
+                              numpy.logical_and(lon >= minlon, lon <= maxlon))
+
+# Convert logical to int
+landMask = numpy.logical_not(oceanMask).astype(int)
+
+dsBaseMesh['cullCell'] = landMask
+
+dsCulledMesh = conversion.cull(dsBaseMesh,
+                               graphInfoFileName='culled_graph.info')
```
xylar: @mark-petersen, `cullCell` is added to the base mesh, not passed as a separate mask dataset.  (If you want to do the latter, you need to call the field `regionMasks` and it needs to have an extra dimension that we don't want to bother with in this case.)

Also, `cullCells` is 1 where we get rid of cells and 0 where we keep them, so I made some changes accordingly.

xylar:
@mark-petersen, I made some changes so cell culling works properly.  Happy to explain more if needed when you get back.  I ran the full test case and it didn't crash...  I didn't look at the output beyond the `cull_mesh` stage so I'll leave it to you to double check before merging.  Let me know if you need more testing from me.

xylar:
@mark-petersen, please close this and optionally move it to https://github.com/MPAS-Dev/compass

mark-petersen:
Closing. This PR was just something we tried, and is a low priority.

PR #418, opened 2019-12-20
First commit for modification of python scripts plotting four test cases

hyungyukang:
Modification of python scripts plotting four test cases


hyungyukang:
Thanks for your effort on this, @mark-petersen!

PR #420, opened 2020-01-07
Change cvmix and BGC to submodules

jgfouca:
This is a significant change to the structure of the MPAS repository. Instead of external code dependencies being downloaded by home-grown scripts, we instead express the dependency as a git submodule. This will impact the cvmix and BGC dependencies for core_ocean.

I tested building the E3SM case ERS_Ld5.T62_oQU120.CMPASO-NYF with both the old (Makefile-based) and new (CMake-based) build systems.

For additional discussion, see https://github.com/E3SM-Project/E3SM/issues/3377


xylar:
@mark-petersen, do we want this to go to `ocean/develop` instead of `e3sm/develop`?

xylar:
Since calling:
```
```
is definitely not part of the workflow for folks that build standalone MPAS-Ocean, we'll have to make sure our various documentation includes this new set of instructions.  Might also be important to mention at an E3SM LANL meeting or something like that.

mark-petersen:
Yes, I can move the PR over to merge into `ocean/develop` tomorrow. I also compiled as-is, and it passed the nightly regression suite.

mark-petersen:
@jgfouca: After 
`make gfortran CORE=ocean`
the command:
`make clean CORE=ocean`
does not work. It looks like the directory has moved because we get the full cvmix and BGC repos now, but this doesn't work either:
```
git diff
diff --git a/src/core_ocean/Makefile b/src/core_ocean/Makefile
index 24ae63199..480a99285 100644
--- a/src/core_ocean/Makefile
+++ b/src/core_ocean/Makefile
@@ -75,13 +75,13 @@ analysis_members: libcvmix shared

 clean:
        if [ -d cvmix ]; then \
-               (cd cvmix; make clean) \
+               (cd cvmix/src/shared; make clean) \
        fi

        if [ -d BGC ]; then \
-               (cd BGC; make clean) \
+               (cd BGC/src/shared; make clean) \
        fi
```
It gets stuck in an endless make clean loop. Do you see the problem?

xylar:
@mark-petersen, I was able to do `make clean` just fine with gnu under linux.  What are you using?

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "src/core_ocean/cvmix"]
+	path = src/core_ocean/cvmix
+	url = git@github.com:CVMix/CVMix-src.git
+[submodule "src/core_ocean/BGC"]
+	path = src/core_ocean/BGC
+	url = git@github.com:E3SM-Project/Ocean-BGC.git
```
mark-petersen: @mgduda @matthewhoffman @akturner, we would like to add two submodules for use by the ocean core. This file needs to be added to the top level MPAS directory. It will not affect any other cores, but I wanted to ask your approval. If you want to add submodules to your core at a later time, you would add them to this file. On a new clone or branch, use
```
git submodule update --init
```
and it will fill in the submodules for ALL cores, not just yours. But I think that behavior is fine. That is what we do in the E3SM repo.

xylar:
> Yes, I can move the PR over to merge into ocean/develop tomorrow.

Can't @jgfouca just rebase onto `ocean/develop` and edit the PR to make that the new base?

.gitmodules
xylar: Hmm, I hadn't thought of that.  This could become irritating if the submodules become large but I guess we'll cross that bridge when we come to it.

Cores other that the ocean should still work fine without `git submodule update --init`, right?

xylar: Also, maybe this means these changes need to go to `develop`, not `ocean/develop`.

mark-petersen:
@jgfouca never mind about that `make clean` comment. My mistake, it works. @xylar thanks for checking.

.gitmodules
mark-petersen: Since this is a new file and an essential part of this PR, it's OK to merge into `ocean/develop`. I think having submodules for many cores (potentially) in the future is not problematic. The user can either get them all with one command:
```
git submodule update --init
```
or just get the ones relevant to their own core, i.e.
```
git submodule update --init src/core_ocean/BGC
git submodule update --init src/core_ocean/cvmix
```
Either way is not much trouble.

mark-petersen:
Moved commit and PR over to ocean/develop.  Everything works for me, but I'll wait until we hear from the other core developers.

jgfouca:
Thanks @mark-petersen !

.gitmodules
mark-petersen: @mgduda do you approve the addition of this `.gitmodules` file?

matthewhoffman: The addition of a .gitmodules file in the root directory makes sense to me.

mgduda:
Sorry for joining the discussion late. Let me give this some more careful though tomorrow. What is driving the move to git submodules? Were any other options considered?

jgfouca:
> Were any other options considered?

@mgduda , other potential options would be to snapshot/git-subtree the cvmix and BGC code directly into the MPAS repo. That removes the submodule complexity but can cause confusion about the origin/ownership of the code since it would appear to the causal developer to just be a subdirectory within MPAS.

> What is driving the move to git submodules?
A couple things
1) Having the build system manage the downloading of external packages resulted in hung builds for some people (git fetch operations in get_cvmix.sh/get_BGC.sh were waiting for password).
2) get_cvmix.sh/get_BGC.sh were home-rolled implementations of functionality that already exists in many flavors in the broader community.

jgfouca:
@xylar , @mark-petersen ,

@mgduda raised a good point. The E3SM repo has already committed to submodules, so we (I) reflexively treat submodules as the single solution to managing external source dependencies. MPAS has not taken this step, so this may catch some MPAS developers by surprise.

From a developers perspective, snapshotting (adding the external dependency directly into the repo) is the simplest approach. They never even have to know that cvmix etc. come from outside MPAS, and they don't have to do any extra steps in Git to work with the repo. This potentially introduces confusion if a developer thinks they can modify the external code.

I guess I just want to double-check that you guys are happy with submodules.

mgduda:
@jgfouca Thanks very much for the replies. I agree that subtrees would be the easiest, but I also agree that it could lead to confusion if someone tried to edit and commit the external code.

If a user of stand-alone MPAS-Ocean were to download a tar file of a release of MPAS (from the [Releases tab](https://github.com/MPAS-Dev/MPAS-Model/releases)) that uses git submodules, do you know whether the `.gitmodules` file would be present? I seem to recall that GitHub removes the `.git` subdirectory from tar files, and I don't know if this would affect git submodules as well.

As @mark-petersen said, if the total set of submodules becomes unwieldy, we always have the option of updating just the submodules that are needed by a specific core.

jgfouca:
@mgduda,

I have no experience with doing GitHub releases on a repo with submodules. A quick google revealed this:
https://github.community/t5/How-to-use-Git-and-GitHub/Create-Release-that-contains-submodule/td-p/12073

which suggests that releases ignore submodules, which is definitely not ideal.

jgfouca:
@mgduda , @mark-petersen , any updates on this?

mgduda:
@jgfouca Given that the submodules are not required by any shared MPAS infrastructure (and therefore won't affect MPAS-Atmosphere), I have no objections. I'd only like to voice my concern that anyone who downloads MPAS as a `.tar.gz` through the GitHub `Releases` tab will potentially not be able to compile MPAS-Ocean.

jgfouca:
Just for the record, I'm not at all opposed at all to snapshotting these repos since they both are quite small. The only concern is legal/human issues since it would look to the casual observer that cvmix/bgc would be subdirectories of mpas-ocean.

mark-petersen:
I'm back from two weeks of travel and working through PRs. This one is next, but still behind the E3SM Redi PR.

jgfouca:
@mark-petersen , update?

mark-petersen:
We postponed the Redi PR for further testing. I'm working through the accumulated pile of PRs now. Sorry for the delay.

mark-petersen:
Rebase, recompiled, tested with stand-alone MPAS. Everything works.

@jgfouca this is next in line for MPAS-Ocean, and will have it's own E3SM PR later this week. Thanks for your patience.

mark-petersen:
Hearing no objections, I'm merging this in. Thanks @jgfouca.

PR #422, opened 2020-01-09
Codes for a semi implicit barotropic mode solver

hyungyukang:
Adds semi-implicit barotropic mode solver.


mark-petersen:
@hyungyukang Thank you for your effort on this. Please also link or attach any design documents or testing results you already have. For example, you could link to the talk you gave at the last E3SM meeting.

hyungyukang:
> @hyungyukang Thank you for your effort on this. Please also link or attach any design documents or testing results you already have. For example, you could link to the talk you gave at the last E3SM meeting.

I'll do that soon. Thanks, @mark-petersen!

hyungyukang:
Uploading a slide file summarizing this work. A bit revised from my AGU talk last December.
[Kang_A Semi-implicit Barotropic Mode Solver for the MPAS-Ocean.pdf](https://github.com/MPAS-Dev/MPAS-Model/files/4377125/Kang_A.Semi-implicit.Barotropic.Mode.Solver.for.the.MPAS-Ocean.pdf)

mark-petersen:
@hyungyukang I am able to work on this now. My apologies for the delay.

It looks like commit e89de64ad removed all the files in the repo, and 687f11ac3 added them back again. At the command line, `git diff --stat` and `wc` are handy for this:
```
git diff --stat 3979482ff e89de64ad | wc -l
1764
```
What were you trying to do with those two commits? We need to reset back to 3979482ff and add just the changes you intended. One way to do that is to use a unix `cp` or `tar` on the files you want, then 
`git reset --hard 3979482ff`
and copy them back into place again, commit, and `git push --force` back to this branch.

mark-petersen:
Rebased. For my own reference, I compiled lapack (https://github.com/Reference-LAPACK/lapack) on LANL IC as follows:
```
git clone git@github.com:Reference-LAPACK/lapack.git
cd lapack
mkdir build/lapack_gnu/lib
cd build
module load gcc/5.3.0 openmpi/1.10.5 cmake
cmake -DCMAKE_INSTALL_LIBDIR=/usr/projects/climate/mpeterse/repos/lapack/build/lapack_gnu/lib ..
cmake --build . -j --target install
ls lapack_gnu/lib
    cmake  libblas.a  liblapack.a  pkgconfig
```
Then to compile:
```
export LAPACK=/usr/projects/climate/mpeterse/repos/lapack/build/lapack_gnu
make $COMPILER CORE=ocean OPENMP=false DEBUG=true GEN_F90=true
```
Some test cases run using semi-implicit, but some result in a seg fault with an `invalid memory reference`. Looking into it now...

mark-petersen:
Fixed `invalid memory reference`. A flag name had changed. Not passing restart tests yet.

mark-petersen:
@hyungyukang, did you test the semi-implicit solver with restarts? Most of my tests work when `config_do_restart = .false.` but in restart tests with `config_do_restart = .true.`, where the initial condition is read in from a restart file, it dies with
```
CRITICAL ERROR: Iteration number exceeds Max. #iteration: PROGRAM STOP
```
In this example, the full run goes 4 timesteps from time 0 to 8 hours (2hr/time step) and completes successfully. The restart run reads in the restart file at time step 2 (4 hours) and goes to 8 hours, and should be bfb identical to the full run.

Is there a variable that you need added to the restart file? Or some other condition that could be missing for a restart? Perhaps on restart init we have to populate a variable used for the initial guess for the implicit solver?

Makefile
```diff
@@ -494,6 +494,12 @@ ifneq ($(wildcard $(PIO_LIB)/libgptl\.*), )
 	LIBS += -lgptl
 endif
 
+ifneq "$(LAPACK)" ""
+        LIBS += -L$(LAPACK)/lib
+        LIBS += -llapack
+        LIBS += -lblas
```
mark-petersen, line 500: @hyungyukang I had to add this like to connect to the blas library as well. It appears that you did not need this flag when you compiled?

hyungyukang:
@mark-petersen, thanks for reviewing the codes. Actually I've never tested the solver using `config_do_restart = .false.`. I just checked the same problem. 
But when I run the model with `config_time_integrator = 'split_explicit'`, the model also showed error:
`ERROR:  Warning: Sea surface height is outside of acceptable physical range, i.e. abs(sum(h)-bottomDepth)>20m.
  iCell: 1, maxLevelCell(iCell): 55, bottomDepth(iCell): 4222.79159790825, sum(h): 0.00000000000000`

I'm currently running these cases - QU240, EC60to80, and RRS30to10. Are these test cases working with `config_do_restart = .false.`? If not, can you please recommend some other test cases?

Makefile
hyungyukang: @mark-petersen, without `LIBS += -lblas`, the model was compiled for some platforms (NERSC, MacOS, Ubuntu) for me. Did you fail to compile before you add it? If so, we should add it. Thanks!

mark-petersen:
You need to run the code forward and write restart files, and then restart from there. I can send a test case. That SSH warning is because the model is reading in all zeros, because no restart file is available.

hyungyukang:
> You need to run the code forward and write restart files, and then restart from there. I can send a test case. That SSH warning is because the model is reading in all zeros, because no restart file is available.

@mark-petersen, got it. I'll run it today. I'll let you know when I find problems as soon as possible.

mark-petersen:
@hyungyukang  I just sent you a zip file of our nightly regression suite, with all the correct flags for your rebased PR. unzip it, then
```
cd ocean/global_ocean/QU240/restart_test/
```
and point ocean_model to your executable in each subdirectory.

You can use run.py or cd into full_run and then restart_run and run each of those separately. The run.py in the restart_test directory compares if you get bit-for-bit matching solutions at the end of the full run and restart run.

hyungyukang:
> @hyungyukang I just sent you a zip file of our nightly regression suite, with all the correct flags for your rebased PR. unzip it, then
> 
> ```
> cd ocean/global_ocean/QU240/restart_test/
> ```
> 
> and point ocean_model to your executable in each subdirectory.
> 
> You can use run.py or cd into full_run and then restart_run and run each of those separately. The run.py in the restart_test directory compares if you get bit-for-bit matching solutions at the end of the full run and restart run.

@mark-petersen, just received it. Thanks!

mark-petersen:
I'll let you test it. When the restart run is not identical to the original run, the most likely cause is that you have a variable or setting in memory that is not saved for the restart. In that case, you need to add that variable to
```
Registry.xml
1503         <stream name="restart"
```
list just below that.

Another possibility is that you are missing something in your init routine, like initializing variables for your implicit method. The original run (no restart) initializes the `normalVelocity` to zero everywhere, so it can hide an error of something you might need to initialize on every simulation start, not just the restart ones.

hyungyukang:
@mark-petersen , I just fixed the bugs for restart run. Can you please test it?

hyungyukang:
@mark-petersen , I've made some changes for the SI code based on what I've tested in other branch.
One thing you have to change before run the model is `namelist.ocean`. I create a new section and changed some in `Registry.xml`.

Since explicit-subcycling and semi-implicit schemes are sharing these options (`config_n_ts_iter`, `config_n_bcl_iter_beg`, `config_n_bcl_iter_mid`, `config_n_bcl_iter_end`), so I take them out from `&split_explicit_ts` and put them in `&split_timestep_share` which is the new namelist section.

```
/
&split_timestep_share
    config_n_ts_iter = 2
    config_n_bcl_iter_beg = 1
    config_n_bcl_iter_mid = 2
    config_n_bcl_iter_end = 2
/
&split_explicit_ts
    config_btr_dt = '0000_00:00:15'
    config_n_btr_cor_iter = 2
    config_vel_correction = .true.
    config_btr_subcycle_loop_factor = 2
    config_btr_gam1_velWt1 = 0.5
    config_btr_gam2_SSHWt1 = 1.0
    config_btr_gam3_velWt2 = 1.0
    config_btr_solve_SSH2 = .false.
/
&semi_implicit_ts
    config_btr_si_preconditioner = 'ras'
    config_btr_si_tolerance = 1.0e-9
    config_n_btr_si_outer_iter = 2
/
```

And now the SI code passed the restart run test:
```
Beginning variable comparisons for all time levels of field 'temperature'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
 ** PASS Comparison of temperature between full_run/output.nc and
    restart_run/output.nc
Beginning variable comparisons for all time levels of field 'salinity'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
 ** PASS Comparison of salinity between full_run/output.nc and
    restart_run/output.nc
Beginning variable comparisons for all time levels of field 'layerThickness'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
 ** PASS Comparison of layerThickness between full_run/output.nc and
    restart_run/output.nc
Beginning variable comparisons for all time levels of field 'normalVelocity'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
 ** PASS Comparison of normalVelocity between full_run/output.nc and
```

mark-petersen:
@hyungyukang this is great progress. I just ran our nightly regression suite with gnu debug and optimized, but with openmp off. I compile with
```
make gfortran CORE=ocean OPENMP=false DEBUG=true
```
and get
```
./nightly_ocean_test_suite.py
 ** Running case Baroclinic Channel 10km - Default Test
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - BGC Ecosys Test
      PASS
      PASS
      PASS
      PASS
   ** FAIL (See case_outputs/Baroclinic_Channel_10km_-_Decomp_Test for more information)
      PASS
   ** FAIL (See case_outputs/sub-ice-shelf_2D_-_restart_test for more information)
      PASS
      PASS
```
First, thank you for fixing the restart, it now works! That is great. 
Let's not worry about the sub-ice-shelf yet. 
We do need to worry about the fail on the `Baroclinic Channel 10km - Decomp Test`. That means that if you run with 4 versus 8 cores (say), that they are not bfb with each other. There are two possible causes:
1. There is place where we need to run a calculation on more halo layers (`nCells = nCellsArray( 2 )`, `nEdges = nEdgesArray( size(nEdgesArray) )`, etc).
2. a halo update needs to be added somewhere

I will make a QU240 that tests partitions as well and send it to you. I don't have time to debug it much this week, but you could try. The general method is, make sure you also get a non-bfb match between 4 vs 8 cores (say), but a bfb match with split explicit. 

Then in your semi-implicit code, set your nCells and nEdges variables are maxed out, i.e.
`nCells = nCellsArray( size(nCellsArray) )`, `nEdges = nEdgesArray( size(nEdgesArray) )`. Making them large is safer. Once we have bfb match across partitions, we can make them smaller for efficiency. You may get an error with it all maxed out, and need to reduce it by one somewhere.

If that does not produce a bfb match, then look through and guess where you might need a halo update, and add it. Again, you can put in lots extra and there is no harm done. Once you get a bfb match, we need to take out all the extra ones, because halo updates are communication and are very expensive.

hyungyukang:
Thanks Mark for the results. I'll be waiting for the partition test and look into the baroclinic channel test as well!

mark-petersen:
I just sent a tar file with a new global test case, comparing 4 vs 8 processors. To run it:
```
cd ocean/global_ocean/QU240/decomp_test/
./run.py
 * Running 4proc_run
  - Complete
 * Running 8proc_run
  - Complete
Beginning variable comparisons for all time levels of field 'temperature'. Note any time levels reported are 0-based.
```
and it shows a mismatch between the two. Run with split explicit and it should match. Give it a try, and you can explore the halo strategies I have above.

philipwjones:
Note that global sums are not bfb on different processor counts unless you write your own reproducible sum (MPI reductions will not guarantee the sum is done in the same order). So for the new solver, you will not get bfb on different numbers of MPI tasks. This was also true for POP. If you want, we can write a reproducible sum as a debug option for testing, but reproducibility will come at a performance penalty and we would not want it on full time. (Though I noticed there's a seminar next week on just this topic at LANL...)

hyungyukang:
> Note that global sums are not bfb on different processor counts unless you write your own reproducible sum (MPI reductions will not guarantee the sum is done in the same order). So for the new solver, you will not get bfb on different numbers of MPI tasks. This was also true for POP. If you want, we can write a reproducible sum as a debug option for testing, but reproducibility will come at a performance penalty and we would not want it on full time. (Though I noticed there's a seminar next week on just this topic at LANL...)

Thanks @philipwjones . I agree with your comment. Mark and I got machine precision error (not bfb) without any preconditioner. Also, if we use preconditioners such as block-Jacobi and RAS, then the error is much larger since the structure of the preconditioners is changing based on the number of processors. So we'll never obtain bfb result when using the new solver except for some specially-designed debugging tests as you said.

mark-petersen:
@hyungyukang, on the decomp test that I sent you recently, I'm getting differences of 1e-2 to 1e-4 in the prognostic variables after 3 time steps with the QU240. I know our conversation above indicates that we don't expect a bfb match, but this difference seems much higher than that. What do you think?

hyungyukang:
> @hyungyukang, on the decomp test that I sent you recently, I'm getting differences of 1e-2 to 1e-4 in the prognostic variables after 3 time steps with the QU240. I know our conversation above indicates that we don't expect a bfb match, but this difference seems much higher than that. What do you think?

@mark-petersen , I'm going to test more about this. Can you please wait more? I'll update you with some test results.

mark-petersen:
@mattdturner or @philipwjones could one of you help me update and review this implicit time-stepping routine? A check-list of items is:
1. config flags: we now use a module and remove declarations (after #457)
2. Change scratch arrays to local (after #457)
3. Add OpenMP pragmas (after #513)
4. Put all the semi-implicit variables in a package, so memory is not allocated if not running with semi-implicit timestepping.
5. Test for performance and memory usage within MPAS-Ocean stand-alone. (@hyungyukang has done this, but we need to test ourselves as well)
6. Make sure E3SM compiles, with semi-implicit off, on our standard platforms. Note semi-implicit requires linking to LAPACK and BLAS libraries.
7. Test in E3SM for performance and any problems on our various platforms
8. Compare decadal E3SM climate simulations between old split explicit and new semi-implicit codes (With help from @vanroekel and others, not by performance team).
9. Decide if semi-implicit should be the default time-stepping method (need consensus of simulation teams).

This PR can be merged after steps 1-6, and steps 7-9 would be additional MPAS and E3SM PRs. I think we need to dedicate several weeks to this effort, after #457, #513, and others are merged.

mattdturner:
> 1. config flags: we now use a module and remove declarations (after #457)
> 
> 2. Change scratch arrays to local (after #457)
> 
> 3. Add OpenMP pragmas (after #513)
 
Since #457 and #513 are my PRs, I can take care of these after I finalize debugging #457 

> 5. Test for performance and memory usage within MPAS-Ocean stand-alone. (@hyungyukang has done this, but we need to test ourselves as well)

Should I just compare performance and peak memory usage between this PR and the `split` time-stepping?

> 6. Make sure E3SM compiles, with semi-implicit off, on our standard platforms. Note semi-implicit requires linking to LAPACK and BLAS libraries.

I can confirm this while working through 1-3

> 7. Test in E3SM for performance and any problems on our various platforms
> 
> 8. Compare decadal E3SM climate simulations between old split explicit and new semi-implicit codes (With help from @vanroekel and others, not by performance team).
> 
> 9. Decide if semi-implicit should be the default time-stepping method (need consensus of simulation teams).

I'll leave these last 3 for a wider group of people...

hyungyukang:
> @hyungyukang, on the decomp test that I sent you recently, I'm getting differences of 1e-2 to 1e-4 in the prognostic variables after 3 time steps with the QU240. I know our conversation above indicates that we don't expect a bfb match, but this difference seems much higher than that. What do you think?

@mark-petersen , can you please check `config_dt` used for the test? Is it 5min, right? And where did you get the values '1e-2 to 1e-4' from? The values are calculated from `compared_fields.py`?

mark-petersen:
@hyungyukang, if you look at the tar file I sent associated with https://github.com/MPAS-Dev/MPAS-Model/pull/422#issuecomment-653080548, there are directories for the 4 and 8 processor runs. Namelists are in there. config_dt should be 2 hours. These have a bit-for-bit match on different partitions with our current timestepping methods. I understand that an implicit method could loose that, but I thought the differences were much higher than expected. Try to run the test directly from that tar file, and ask if you have questions.

hyungyukang:
> @hyungyukang, if you look at the tar file I sent associated with [#422 (comment)](https://github.com/MPAS-Dev/MPAS-Model/pull/422#issuecomment-653080548), there are directories for the 4 and 8 processor runs. Namelists are in there. config_dt should be 2 hours. These have a bit-for-bit match on different partitions with our current timestepping methods. I understand that an implicit method could loose that, but I thought the differences were much higher than expected. Try to run the test directly from that tar file, and ask if you have questions.

@mark-petersen , Oh... I've been used the baroclinic channel decomp test. That's the why there was a pretty big difference between your error values and mine. After some modifications of the code, I reduced the error for the baroclinic channel decomp test (up to two orders). But I realized that it was not very helpful for the global ocean decomp test. Anyway, I'll look into it more and also run long term simulations to compare some fields.

hyungyukang:
@mark-petersen , I compared some long-term (11,000 days) simulation results for several options using QU240 test case. I wanted to check this just in case that I can't reduce the error anymore. The normalized L1 norm using the kineticEnergyCell (3D) was calculated and compared for each run. Here's brief description of options for each run.

> P068_RAS : SI with RAS preconditioner using 68 cores
> P136_RAS : SI with RAS preconditioner using 136 cores
> P068_NON : SI without preconditioner using 68 cores
> P136_NON : SI without preconditioner using 136 cores
> P068_EXP : Explicit-subcycling using 68 cores
> P136_EXP : Explicit-subcycling using 136 cores

As expected, the error growth rapidly until 5,000 days and then saturated after that. Even for P068_NON vs P136_NON case which we had near-machine precision error for the first 3 steps, trend of time evolution of the L1 norm looks pretty similar to the others. And Salil Mahajan is going to check the climate reproducibility by using his test method whether each run is producing the same climate. Also, I have a plan to perform a non-BFB solution reproducibility test based on time step convergence [(Wan et al., 2017)](https://gmd.copernicus.org/articles/10/537/2017/gmd-10-537-2017.pdf). If all tests are passed, is it acceptable? How do you think about this strategy?
I'm still thinking a way to reduce the error for the semi-implicit method. But.. I'm narrowing my thoughts to the impossible unless there is no sacrifice of computational performance and accuracy.

fig_l1norm_KE

mark-petersen:
@hyungyukang I think this is a great strategy for non-bfb merges, and is quite convincing for this resolution. It is certainly enough for this PR into MPAS-Ocean. When we test for a later E3SM pull request, we will run at our standard resolutions and look at side-by-side comparisons.

mark-petersen:
@hyungyukang, I thought you would like to know, we have made progress on the list on the above comment,
https://github.com/MPAS-Dev/MPAS-Model/pull/422#issuecomment-656158430
#457 and #513 are merged, and the performance team is helping us with these steps - @mattdturner is assisting with work on this PR, with some input from @philipwjones and myself.

mattdturner:
I just rebased this branch, and updated it to reflect #457 and #513.

After rebase, I also ran the nightly test suite.  All tests PASS, except for the `decomposition` tests (expected per https://github.com/MPAS-Dev/MPAS-Model/pull/422#issuecomment-653809430) and the sub-ice-shelf test.

I'll move on to 5-7 from https://github.com/MPAS-Dev/MPAS-Model/pull/422#issuecomment-656158430

hyungyukang:
@mark-petersen and @mattdturner , thank you so much for your contributions to this PR. And I reported the progress of this PR in the monthly meeting last week. Also, I've got some results using the non-BFB solution reproducibility test based on time step convergence (Wan et al., 2017). According to the results (the figure), the semi-implicit solver using different cores is producing non-BFB results (P16 vs P08 and P04 vs P08 - the red lines), but their solution reproducibility is passed the test. The thick gray dashed line indicates the criteria for Pass (if higher) / Fail (if lower). Furthermore, comparison between the semi-implicit and the explicit-subcycling scheme (the blue line) passed the test, which means that they are solving the same set of equations.
fig_TSC_test_result

mattdturner:
Rebased to include recent bugfixes for MPASO code, to allow for easier E3SM testing.  

Currently seeing `T62_oQU120_ais20.MPAS_LISIO_TEST` tests PASS, but `T62_oQU240.GMPAS-IAF` and `T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF` tests fail with a "state invalid" error in E3SM.

hyungyukang:
> Rebased to include recent bugfixes for MPASO code, to allow for easier E3SM testing.
> 
> Currently seeing `T62_oQU120_ais20.MPAS_LISIO_TEST` tests PASS, but `T62_oQU240.GMPAS-IAF` and `T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF` tests fail with a "state invalid" error in E3SM.

@mattdturner , can I update some modifications for the SI code? I think the modifications (not much) can solve the problem.

mattdturner:
> @mattdturner , can I update some modifications for the SI code? I think the modifications (not much) can solve the problem.

@hyungyukang Please do.  Once you push your updates, I'll re-merge into E3SM and re-run the tests.

hyungyukang:
> > @mattdturner , can I update some modifications for the SI code? I think the modifications (not much) can solve the problem.
> 
> @hyungyukang Please do. Once you push your updates, I'll re-merge into E3SM and re-run the tests.

@mattdturner , I just updated the code. Can you please test it again?

mattdturner:
I should mention that I am not yet sure if the failures are a result of a bug in the actual code, or in the implementation in E3SM.  Hopefully my testing today (if I can gets debug runs to work on Cori) will provide some additional insight.

hyungyukang:
> I should mention that I am not yet sure if the failures are a result of a bug in the actual code, or in the implementation in E3SM. Hopefully my testing today (if I can gets debug runs to work on Cori) will provide some additional insight.

@mattdturner , Thank you so much for letting me know the results. Could you please let me know how can I run the tests that you are running? It would be faster that we can solve the problem together. Please just let me know some webpages or manuals.

mattdturner:
Unfortunately, right now testing this branch in E3SM isn't a simple process.  Here is what I have had to do:

- Clone E3SM on Cori `git clone git@github.com:E3SM-Project/E3SM.git`
- `cd E3SM`
- `git checkout mattdturner/ocean/semi_implicit`. 
- `git submodule update --init --recursive`
- `cd components/mpas-source`
- `git remote add hyungyukang https://github.com/hyungyukang/MPAS-Model.git`
- `git fetch hyungyukang`
- `git merge hyungyukang/semi_implicit`

At this point, you should have my E3SM branch w/ the semi-implicit solver code included in MPAS-O.  The `mattdturner/ocean/semi_implicit` branch modifies some of the namelist files within E3SM.  However, it still has `split-explicit` as the default time integrator.

To set `semi-implicit` as the time integrator, edit `components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml` (line 47 I think):
```
-<config_time_integrator>'split_explicit'</config_time_integrator>
+<config_time_integrator>'semi_implicit'</config_time_integrator>
```

You can then run a test by `cd cime/scripts` (from E3SM root directory).  Since $SCRATCH is currently read-only, you'll have to set a test directory in your $HOME.  I've done this w/ the following command:
```
./create_test -q test -r $HOME/e3sm_scratch_tmp --output-root $HOME/e3sm_scratch_tmp --walltime 00:30:00 SMS.T62_oQU240.GMPAS-IAF.cori-haswell_intel
```

For some reason, `create_test` appears to be ignoring the `-q test` flag.  So, I had to make the following edit prior to running `create_test`.  This is unnecessary once Cori is back online at full capacity.:
```
diff --git a/cime_config/machines/config_batch.xml b/cime_config/machines/config_batch.xml
index dde1ecee8..69943c68c 100644
--- a/cime_config/machines/config_batch.xml
+++ b/cime_config/machines/config_batch.xml
@@ -310,7 +310,7 @@
      </directives>
      <queues>
        <queue walltimemax="00:30:00" nodemax="64" strict="true">debug</queue>
-       <queue walltimemax="01:00:00" default="true">regular</queue>
+       <queue walltimemax="01:00:00" default="true">test</queue>
      </queues>
   </batch_system>
```

---

Tests should run when using `intel` compiler since `-mkl` is included in the link line by default.  If using `GNU`, you'll need to make the following edits to have E3SM load the lapack module:
```
diff --git a/cime_config/machines/config_machines.xml b/cime_config/machines/config_machines.xml
index 962c7a8ce..f28c95285 100644
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -151,6 +151,7 @@
         <command name="load">gcc/8.3.0</command>
         <command name="rm">cray-libsci</command>
         <command name="load">cray-libsci/19.06.1</command>
+        <command name="load">/global/homes/m/mturner/.local/modulefiles/lapack/3.9.0/gnu8.3.0</command>
       </modules>
 
       <modules>
@@ -303,6 +304,7 @@
         <command name="load">gcc/8.3.0</command>
         <command name="rm">cray-libsci</command>
         <command name="load">cray-libsci/19.06.1</command>
+        <command name="load">/global/homes/m/mturner/.local/modulefiles/lapack/3.9.0/gnu8.3.0</command>
       </modules>
 
       <modules>
```

Permissions should be correct so that you can load the lapack module that is in my home directory.  If not, let me know.



There is probably an easier way to set this up then what I have detailed, but this is what I have done based on what I know about the E3SM build / test framework.

If you run into any issues running, let me know.  Cori's queue throughput is pretty slow right now, so I haven't been able to run E3SM through DDT to get a better idea of why NaNs are showing up.

hyungyukang:
> Unfortunately, right now testing this branch in E3SM isn't a simple process. Here is what I have had to do:
@mattdturner , Thank you so much for the detailed descriptions. I'll follow your guidance to run the models after Cori is really back to service! Thanks again.

mattdturner:
For 1 of the tests that I am running (`SMS_D_PMx1.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel`), the crash is a result of a divide by zero.

Line 2467 in `mpas_ocn_time_integration_si.F90`:
```
               CGcst_beta0 = (CGcst_alpha0/CGcst_omega0) * CGcst_r00r1_global / CGcst_r00r0_global
```

`CGcst_r00r0_global` has a value of 0 here.  This is a result of line 2138
```
             call mpas_dmpar_sum_real_array(dminfo, 3, CGcst_allreduce3, CGcst_allreduce_global3)
```

After that line at some point in the simulation, I print the value of `CGcst_allreduce_global3` on each rank.  The log shows the following for 1 of the ranks:
```
159:  debug, mdt :: global reduce =   1.946873136433337E-030  1.340746811130887E-027
159:   0.000000000000000E+000
```

I'm not sure what range of values the `CGcst_allreduce_global3` variable should have.  As a result, I'm not sure if this is the actual problem resulting from a very rare value showing up, or if something else is causing this variable to have values that it shouldn't have.

hyungyukang:
> I'm not sure what range of values the `CGcst_allreduce_global3` variable should have. As a result, I'm not sure if this is the actual problem resulting from a very rare value showing up, or if something else is causing this variable to have values that it shouldn't have.

@mattdturner , I was reproducing the error, but Cori is now stopped. Cori will be back this morning, so I'll dive into the problem today.

hyungyukang:
> For 1 of the tests that I am running (`SMS_D_PMx1.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel`), the crash is a result of a divide by zero. ...

@mattdturner , I found there are some problems in namelist building. In `components/mpas-ocean/bld/build-namelist`, the group lists for the semi-implicit solver were not updated. Here's updates:
```
diff --git a/components/mpas-ocean/bld/build-namelist b/components/mpas-ocean/bld/build-namelist
index adecfc67c..aba6506c0 100755
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -1596,7 +1596,9 @@ my @groups = qw(run_modes
                 pressure_gradient
                 eos
                 eos_linear
+                split_timestep_share
                 split_explicit_ts
+                semi_implicit_ts
                 ale_vertical_grid
                 ale_frequency_filtered_thickness
                 debug
```

After I made the above edits, I got this results:
```
2020-10-07 07:03:26: case.setup starting
 ---------------------------------------------------
2020-10-07 07:03:30: case.setup success
 ---------------------------------------------------
2020-10-07 07:03:31: case.build (SHAREDLIB_BUILD) starting
 ---------------------------------------------------
2020-10-07 07:05:18: case.build (SHAREDLIB_BUILD) success
 ---------------------------------------------------
2020-10-07 07:05:19: case.build (MODEL_BUILD) starting
 ---------------------------------------------------
2020-10-07 07:08:18: case.build (MODEL_BUILD) success
 ---------------------------------------------------
2020-10-07 07:08:18: case.submit starting
 ---------------------------------------------------
2020-10-07 07:08:41: case.submit success case.test:34862486
 ---------------------------------------------------
2020-10-07 07:11:06: case.run starting
 ---------------------------------------------------
2020-10-07 07:11:12: model execution starting
 ---------------------------------------------------
2020-10-07 07:11:47: model execution success
 ---------------------------------------------------
2020-10-07 07:11:47: case.run success
 ---------------------------------------------------
```
Does this mean the case `SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu` passed the test?

mattdturner:
If you go to the case directory, there should be a `TestStatus` file.  If you look in that file, it will say if everything PASSed or if there are some FAILs.

hyungyukang:
> If you go to the case directory, there should be a `TestStatus` file. If you look in that file, it will say if everything PASSed or if there are some FAILs.

@mattdturner , I just checked it, and it says:
```
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu CREATE_NEWCASE
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu XML
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu SETUP
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu SHAREDLIB_BUILD time=107
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu MODEL_BUILD time=179
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu SUBMIT
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu RUN time=43
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu MEMLEAK
PASS SMS.T62_oQU240.GMPAS-IAF.cori-haswell_gnu SHORT_TERM_ARCHIVER
```
So I think the case passed the test. Could you please make the changes I suggested and run the cases again? Thanks in advance!

mattdturner:
@hyungyukang Thanks for catching the namelist oversight on my part.  I thought I had made all of the necessary modifications, but I guess I missed that one.

I just pushed the change to the E3SM branch, and have started my test suite that I run prior to merging code.  It will run overnight, and I'll report back once the simulations finish.

mattdturner:
All of the `PET_*` and `PEM_*` tests that I ran failed.  The `PEM_*` failures are expected, since they check bit-for-bit between different number of MPI ranks.  I have an idea of why the `PET_*` tests might be failing.

The following 3 tests passed:
```
```

All `gnu` tests failed because I forgot to link in LAPACK.  I'm re-running the tests now, and will hopefully have something more to report soon.

mattdturner:
PASS:
```
```
FAIL:
```
FAIL SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu RUN 
FAIL SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel RUN 
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu RUN 
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel RUN 
FAIL SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu RUN 
FAIL SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel RUN 
```

The `SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu` runs fail with the error message stating that the Iteration number exceeds max iters (18465)

hyungyukang:
@mattdturner , Thanks for reporting. It seems only higher resolution simulations are failed. I'll look into it!

hyungyukang:
> @hyungyukang see if you can access this file on cori:
> /global/cscratch1/sd/mpeterse/runs/201014_QU240wISC/test_QU240wISC.tar
> you can run that, first with split explicit, then semi-implicit. That is a first easy check.
> ...

@mark-petersen , I copied that test case file to my Mac and succeeded to reproduce the same error as in Cori.
I’ll look into this problem more and report progresses to you and @mattdturner when I have some results.

hyungyukang:
@mattdturner , I made some modifications on the semi-implicit (SI) code and pushed it. Now the case `SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_gnu` is running with the SI code, but it's not finished within an hour (--walltime 01:00:00). This is the same for the split-explicit (SE) method. When is the end date of this simulation? For the SI method, the case was terminated after 2 days and 18 hours of simulation (2 days for the SE method). Anyway, can you please test the updated code again? Thanks in advance.

hyungyukang:
@mark-petersen , Thanks for pushing #6246e1e. I removed `ncpus` mistakenly during small cleanings just before pushed the code.

mark-petersen:
No problem. I just tested the head on LANL grizzly, with 
```
export LAPACK=/usr/projects/climate/mpeterse/repos/lapack/build/lapack_gnu
make gfortran CORE=ocean OPENMP=true DEBUG=true GEN_F90=true USE_LAPACK=true
```
and tested with the nightly regression suite with time stepping changed to semi-implicit. I get this:
```
 ** Running case Baroclinic Channel 10km - Default Test
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - Partition Test
   ** FAIL (See case_outputs/Global_Ocean_240km_-_Partition_Test for more information)
      PASS
      PASS
 ** Running case Global Ocean 240km - BGC Ecosys Test
      PASS
      PASS
      PASS
      PASS
   ** FAIL (See case_outputs/Baroclinic_Channel_10km_-_Decomp_Test for more information)
      PASS
      PASS
   ** FAIL (See case_outputs/Periodic_Planar_20km_-_LIGHT_particle_region_reset_test for more information)
   ** FAIL (See case_outputs/Periodic_Planar_20km_-_LIGHT_particle_time_reset_test for more information)
```
We are expecting a failure on the partition test (same as decomp test), right? I'm not sure about the LIGHT (Lagrangian particle) failure. Let's not worry about that right now.

@hyungyukang could you please write a few paragraphs on what changed with a99d0b5, specifically the change in equations that you are solving, from SSH to barotropic velocity?

@mattdturner could you repeat all your previous tests in E3SM? The commit a99d0b5 is a substantial change, so we needed to retest everything. I'm hoping it passes the tests with ice shelves now.

mattdturner:
@mark-petersen Thanks for re-running the MPAS-O test suite.

I'm in the process of running a host of tests within E3SM (the same ones I always run), and will report back w/ the results.

hyungyukang:
@mark-petersen , sure, will do. But one strange thing is that the LIGHT tests were passed last time (https://github.com/MPAS-Dev/MPAS-Model/pull/422#issuecomment-652580809). Maybe there's some problems on the initialization stage where I made some changes. I'll look into it soon.

mark-petersen:
I just rebased to the head of `ocean/develop`. This compiles and has the same pass/fail behavior as above. It's best to keep up so the merge is easier. @hyungyukang, the nightly regression suite with the updated flags is here on cori:
```
/global/cscratch1/sd/mpeterse/for_others/hyungyukang/201028_semi_implicit_rebase_with_init_cc9d8fdbd.tar
```

mattdturner:
I re-ran a host of tests.  Here are the results:

**PASS**:
```
SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu
```

**FAIL**:
```
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu RUN 
FAIL PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-haswell_gnu COMPARE_base_modpes
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel COMPARE_base_single_thread
FAIL PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-haswell_intel COMPARE_base_modpes
```

The `PEM_` test failures are expected.  The `PET_` test that failed in RUN was an issue with the queue, and I am re-running it.  I'll need to look into the `PET_` failure in COMPARE, although I suspect an OMP directive just needs a small tweak.

But the updates appear to allow the previously-failing tests to pass.  Thanks @hyungyukang

mark-petersen:
@mattdturner I compiled and tested this with gnu 5 on grizzly with the test case QU240wISC, and ran with semi-implicit time step. Unfortunately it passed the thread test, with both debug and optimized, so that does not help us solve the failed PET test above. The test is here on IC:
```
/lustre/scratch3/turquoise/mpeterse/runs/201107_QU240wISC/ocean/global_ocean/QU240wISC/thread_test
```

mattdturner:
@mark-petersen thanks for running the test.

I re-ran `PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel` after commenting out all OpenMP directives in `components/mpas-source/src`, and the test still fails.  Looking in `<test directory>/bld/cmake-bld`, there is no OpenMP code that should be executed.  

I also ran the same test, but comparing 1 thread to 1 thread (i.e, use `xmlchange` to force both runs to use 1 thread) just to confirm that nothing weird was going on.  That test PASSed. 

I'm still digging to see what other OpenMP might be causing the non-bfb results.

hyungyukang:
@mattdturner and @mark-petersen , thanks for testing the codes. And @mark-petersen , I ran the `Periodic Planar 20km - LIGHT particle region reset test` and `Periodic Planar 20km - LIGHT particle time reset test` on Cori. Have you ever run this test case using the `split_explicit`? it failed too. I think there are some problems on the `graph.info` file. When I ran `region reset test` using single core in the `forward` directory, both `split_explicit` and `semi_implicit` passed this test case. This test case originally use six cores, but both scheme did not proceed at the first time step when using six cores. For `time reset test`, I need to unlink `graph.info.part.6` first then remake it using `gpmetis graph.info 6` in `forward` directory. After that, both schemes passed this test. Can you please check this test case?
And @mattdturner and @mark-petersen , how do you think having a low-performance mode of the semi-implicit method as an additional option in the semi-implicit code? This mode will be BFB across partitions while losing computational performance. But we can use this mode for some test and comparison purposes.

mark-petersen:
@hyungyukang thanks for testing the LIGHT test case. I'll make a note of that, but those test cases are not really relevant to the time step, as it is meant to test the LIGHT Lagrangian particles.

@hyungyukang how much work would it be to add a low-performance mode? If it is a day's work I think it is worth it, as then we can set that flag when we conduct partition tests, and know that it is not hiding some other error. For example, we can try to decrease the number of halo updates, and still have confidence that we did it correctly.

mattdturner:
I agree that it would be a good option to have a bit-for-bit partition test that we can switch on for the `PEM_*` tests within E3SM.  I believe there is a bfb MPI collective routine in E3SM, but I have never used it.

hyungyukang:
@mark-petersen , I'm not sure that it'd take only a day, but it won't be that long. I'll let you know when it's ready!

philipwjones:
While we don't want to hold up this PR, I will be having Bob implement new reductions that preserve b4b but are still performant. That will at least help the reduction piece of this, though the preconditioner dependence will still need to be handled in a non-performant way for testing.

hyungyukang:
> While we don't want to hold up this PR, I will be having Bob implement new reductions that preserve b4b but are still performant. That will at least help the reduction piece of this, though the preconditioner dependence will still need to be handled in a non-performant way for testing.

@philipwjones , Thanks. The Jacobi preconditioner should be used to achieve BFB match across partitions. Absolutely, we lose computational performance as you mentioned. I just implemented a very simple BFB allreduce, but it would be great if we have BFB allreduce using good algorithms for MPAS.
@mattdturner and @mark-petersen , I just pushed a thread-match mode for the semi-implicit solver. In this mode, the Jacobi preconditioner is used. And I implemented a very simple BFB allreduce in which MPI allreduce performs using single precision (not actually `kind=4`, but I discarded double precision floating points.). Next version, this algorithm should be changed using better algorithms. Currently, at least, we can identify that this semi-implicit algorithm can achieve BFB match across MPI tasks by using the domain-independent preconditioner and the BFB allreduce algorithm. I added this mode as an option in the `semi_implicit_ts` namelist section. To turn on this mode: `config_btr_si_thread_match_mode=.true.` 
```
&semi_implicit_ts
    config_btr_si_preconditioner = 'ras'
    config_btr_si_tolerance = 1.0e-9
    config_n_btr_si_outer_iter = 2
    config_btr_si_thread_match_mode = .true.
/
```

hyungyukang:
@mattdturner and @mark-petersen , the thread match mode passed `PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu` and `PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-haswell_gnu` tests. But I had to change namelist building in `E3SM/components/mpas-ocean/bld` a lot to use the semi-implicit solver. @mattdturner , if you need the changed `build-namelist` for the semi-implicit solver, please refer to this directory in Cori: `/global/project/projectdirs/e3sm/hgkang/data_for_other/bld_for_semi_implicit_201110`.

mark-petersen:
@hyungyukang just to be clear, we have a thread-matching version now, but not the low-performance bfb on different partitions, as discussed [above](https://github.com/MPAS-Dev/MPAS-Model/pull/422#issuecomment-723394291), correct? Does the bfb partition still seem like a reasonable addition?

mattdturner:
I just confirmed that setting `config_btr_si_thread_match_mode` allows the `PEM_*` tests to pass.  I'm re-running my full suite of tests to see where we stand.

mattdturner:
Ok, I ran a suite of tests for both `config_btr_si_thread_match_mode` equals `.false.` and `.true.`.  The results are

**config_btr_si_thread_match_mode = .false.**
```
PASS:
SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu

--
FAIL:
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu COMPARE_base_single_thread
FAIL PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-haswell_intel COMPARE_base_modpes
FAIL PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-haswell_gnu COMPARE_base_modpes
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel COMPARE_base_single_thread
FAIL PET_Ln9_P8192.T62_oRRS18to6v3.GMPAS-IAF.cori-knl_gnu COMPARE_base_single_thread
FAIL PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-haswell_gnu COMPARE_base_modpes
FAIL PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel COMPARE_base_single_thread
FAIL PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-haswell_intel COMPARE_base_modpes
FAIL PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu COMPARE_base_single_thread
```

**config_btr_si_thread_match_mode = .true.**
```
PASS:
PET_Ln9_P8192.T62_oRRS18to6v3.GMPAS-IAF.cori-knl_gnu
SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu

--
FAIL:
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu COMPARE_base_single_thread
FAIL PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-haswell_gnu COMPARE_base_modpes
FAIL PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu COMPARE_base_single_thread
FAIL PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel COMPARE_base_single_thread
FAIL PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-haswell_intel COMPARE_base_modpes
```

Some of the tests (e.g., `PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel`) fail for both `intel` and `gnu`.  Other tests (e.g., `PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu` appear to only fail for `gnu`...

Some of the `PEM_*` tests PASS with thread mode enabled, but others still FAIL.

I am going to look into the `PET_*` failures today to see if I can determine which `!$omp` directives are causing a problem.

hyungyukang:
@mattdturner , thank you so much for running those test cases. I wonder why some PEM_* tests were failed, but I'll also look into those test cases. Thanks, again. And can you please tell me what `PET_*` test case is about and what is `COMPARE_base_modpes`?

mattdturner:
@hyungyukang The `PET_*` tests are similar to the `PEM_*` tests, but instead of comparing the results between simulations with 2 different MPI Rank counts it compared results from simulations with 2 different OpenMP thread counts.

So, a `PEM_*` test might compare results from a run on 32 ranks to a run on 64 ranks.  On the other hand, a `PET_*` test might compare results from a run on 64 ranks w/ 2 threads per rank to a run on 64 ranks with only 1 thread per rank.

hyungyukang:
> @hyungyukang The `PET_*` tests are similar to the `PEM_*` tests, but instead of comparing the results between simulations with 2 different MPI Rank counts it compared results from simulations with 2 different OpenMP thread counts.
> 
> So, a `PEM_*` test might compare results from a run on 32 ranks to a run on 64 ranks. On the other hand, a `PET_*` test might compare results from a run on 64 ranks w/ 2 threads per rank to a run on 64 ranks with only 1 thread per rank.

Oh, I see. Thank you so much! I'll look into the failed `PEM_*` test cases soon.

mark-petersen:
Rebased. Changed flag name to `config_btr_si_partition_match_mode`. Tested with gnu on grizzly, debug and optimized. Passes nightly regression suite with split explicit, and with semi-implicit on all. MPAS-Ocean partition test (4 versus 8):
- FAILS if `config_btr_si_partition_match_mode = .false.`
- PASS if `config_btr_si_partition_match_mode = .true.`

Thanks @hyungyukang, that is beautiful!

hyungyukang:
> Rebased. Changed flag name to `config_btr_si_partition_match_mode`. Tested with gnu on grizzly, debug and optimized. Passes nightly regression suite with split explicit, and with semi-implicit on all. MPAS-Ocean partition test (4 versus 8):
> 
> * FAILS if `config_btr_si_partition_match_mode = .false.`
> * PASS if `config_btr_si_partition_match_mode = .true.`
> 
> Thanks @hyungyukang, that is beautiful!

@mark-petersen , Thanks for testing! I'm glad that the SI solver passed all tests you've run. I'm currently looking into the filed `PEM_*` tests. Hope this problem will be solved soon. Thanks again!

mattdturner:
I've updated my local working copy of this branch to use Cray-LibSci on Cori instead of my custom-built LAPACK libraries.  The test suite I just ran had the following:
```
PASS:
PET_Ln9_P8192.T62_oRRS18to6v3.GMPAS-IAF.cori-knl_gnu
SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel
SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu
PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-haswell_intel
PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-haswell_gnu

--
FAIL:
FAIL PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu COMPARE_base_single_thread
FAIL PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel COMPARE_base_single_thread
FAIL PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu COMPARE_base_single_thread
```
From these tests, it looks like the `PEM_*` issues are no longer showing up.  I'm not entirely sure why changing the library routines would make `PEM_*` PASS or FAIL.  There is still a threading issue that I am looking into (`PET_*`)

mattdturner:
The only changes necessary to use Cray-LibSci on Cori is
```
diff --git a/cime_config/machines/config_machines.xml b/cime_config/machines/config_machines.xml
index eda5c95a9..9e7f590a9 100644
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -179,7 +179,7 @@
         <command name="rm">git</command>
         <command name="load">git</command>
         <command name="rm">cmake</command>
-        <command name="load">cmake/3.14.4</command>
+        <command name="load">cmake/3.18.2</command>
       </modules>
     </module_system>
 
@@ -331,7 +331,7 @@
         <command name="rm">git</command>
         <command name="load">git</command>
         <command name="rm">cmake</command>
-        <command name="load">cmake/3.14.4</command>
+        <command name="load">cmake/3.18.2</command>
       </modules>
 
       <!--command name="list">&gt;&amp; ml.txt</command-->
```

As of CMake 3.16.0, CMake will identify Cray-LibSci when looking for LAPACK routines.

hyungyukang:
@mattdturner , oh, that's great! I've also looked into the failed `PEM_*` tests. Anyway thank you. And for the `PET_*` tests, it seems it failed when comparing with the results of single core. I have an idea. I'll let you know after I try to fix it. Thanks again.

hyungyukang:
@mattdturner , after I changed `config_machines.xml` to use Cray-libsci, I can't run `PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel` as well as other passed tests. I've got this error message logged in `$CASE_DIR/run/e3sm.log.*`:

```
 0:  (t_initf)       profile_ovhd_measurement= F
 0:  (t_initf)       profile_add_detail=       F
 0:  (t_initf)       profile_papi_enable=      F
 0: forrtl: severe (174): SIGSEGV, segmentation fault occurred
 0: Image              PC                Routine            Line        Source
 0: e3sm.exe           0000000002602669  Unknown               Unknown  Unknown
 0: e3sm.exe           000000000232C9A0  Unknown               Unknown  Unknown
 0: e3sm.exe           000000000260224C  Unknown               Unknown  Unknown
 0: e3sm.exe           000000000232C9A0  Unknown               Unknown  Unknown
srun: error: nid00646: task 0: Exited with exit code 174
srun: Terminating job step 37203913.0
 0: slurmstepd: error: *** STEP 37203913.0 ON nid00646 CANCELLED AT 2020-12-13T05:53:19 ***
63: forrtl: error (78): process killed (SIGTERM)
```

It seems there are some failed array allocations. I removed a whole E3SM and installed again, but there was no change. And this problem is the same for explicit/implicit solvers and intel/gnu compilers.
Haven't you ever had this problem? Please let me know if you have some thoughts on this!

mattdturner:
@hyungyukang I just did a clean copy of the code, and was unable to reproduce the crash that you got.  The only local changes that I made were:
```
diff --git a/cime_config/machines/config_machines.xml b/cime_config/machines/config_machines.xml
index 70e1170cd..e535d87c6 100644
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -179,7 +179,7 @@
         <command name="rm">git</command>
         <command name="load">git</command>
         <command name="rm">cmake</command>
-        <command name="load">cmake/3.14.4</command>
+        <command name="load">cmake/3.18.2</command>
       </modules>
     </module_system>
 
@@ -331,7 +331,7 @@
         <command name="rm">git</command>
         <command name="load">git</command>
         <command name="rm">cmake</command>
-        <command name="load">cmake/3.14.4</command>
+        <command name="load">cmake/3.18.2</command>
       </modules>
 
       <!--command name="list">&gt;&amp; ml.txt</command-->
```
```
diff --git a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml b/components/mpas-ocean/
bld/namelist_files/namelist_defaults_mpaso.xml
index 2669b370a..538e8b2f8 100644
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -48,7 +48,7 @@
 <config_dt ocn_grid="oARRM60to6">'00:05:00'</config_dt>
 <config_dt ocn_grid="EC30to60E2r2">'00:30:00'</config_dt>
 <config_dt ocn_grid="WC14to60E2r3">'00:10:00'</config_dt>
-<config_time_integrator>'split_explicit'</config_time_integrator>
+<config_time_integrator>'semi_implicit'</config_time_integrator>
 
 <!-- hmix -->
 <config_hmix_scaleWithMesh ocn_grid="oQU480">.false.</config_hmix_scaleWithMesh>
@@ -383,7 +383,7 @@
 <config_btr_si_preconditioner>'ras'</config_btr_si_preconditioner>
 <config_btr_si_tolerance>1.0e-9</config_btr_si_tolerance>
 <config_n_btr_si_outer_iter>2</config_n_btr_si_outer_iter>
-<config_btr_si_partition_match_mode>.false.</config_btr_si_partition_match_mode>
+<config_btr_si_partition_match_mode>.true.</config_btr_si_partition_match_mode>
 
 <!-- ALE_vertical_grid -->
```

It looks like the error you are getting (at least for GNU) is an error within the MPI implementation.  I am not sure why changing the version of cmake would cause an MPI allocation error.

Are there any other changes that show up when you run a `git status` in your local copy of the E3SM repo?

mark-petersen:
FYI, I just added commits. Most are white space clean-up. Only substantial changes are:
- Revert Makefile, those changes need to go into a PR into develop. It still works if you set ```export LAPACK=/usr/projects/climate/mpeterse/repos/lapack/build/lapack_gnu```
etc. I'll make a PR into develop so we have that better error message, but it does not hold up this PR.
- Add package semiImplicitTime. That just makes the new variables un-allocated when not using semi-implicit.

Testing now. Will also update the E3SM PR.

hyungyukang:
@mattdturner , I also cleaned my copy and did the same things you did. Anyway, Cori is now in maintenance, so I'll try again when it's back to service. Thanks.

hyungyukang:
@mark-petersen , changes are looking good. Thanks. I'll run some tests, too.

mark-petersen (approved):
@hyungyukang Thanks for your substantial effort here.

mark-petersen:
Current HEAD passes MPAS-O nightly regression suite with gnu 6 debug on grizzly. Fails partition by default, but passes with config_btr_si_partition_match_mode = .true..

mattdturner (approved):

hyungyukang:
@mark-petersen and @mattdturner , thank you so much for testing and approving this PR!

philipwjones (approved):
Approve based on visual inspection, many discussions throughout development and the extensive testing by others. Thanks everyone!

hyungyukang:
> Approve based on visual inspection, many discussions throughout development and the extensive testing by others. Thanks everyone!

@philipwjones , thanks a lot for reviewing and approving this PR!

sarats:
Nice work @hyungyukang.

hyungyukang:
> Nice work @hyungyukang.

@sarats , thanks again!

mark-petersen:
rebased. Testing...

mark-petersen:
Passes mpas-o nightly regression suite with gnu debug and both split explicit and semi-implicit. Note that for semi-implicit, compile on grizzly with 
```
export LAPACK=/usr/projects/climate/mpeterse/repos/lapack/build/lapack_gnu
make gfortran CORE=ocean USE_LAPACK=true
```

PR #425, opened 2020-01-16
add the enthalpy-benchmark test case in COMPASS

tongzhangice:
The new enthalpy-benchmark test case for testing two Kleiner enthalpy benchmark experiments (A and B). Note that for testing benchmark A, we don't have to change the code, but for testing benchmark B, we need to do several revisions in the code (as explained in the README file).


matthewhoffman:
@tongzhangice , it looks like you are pushing new commits related to the damage/calving code to this branch.  I think that is because you are making your commits on a branch called `landice/develop` instead of a feature branch.  If you need helping getting that organized correctly, let me know.

tongzhangice:
> @tongzhangice , it looks like you are pushing new commits related to the damage/calving code to this branch. I think that is because you are making your commits on a branch called `landice/develop` instead of a feature branch. If you need helping getting that organized correctly, let me know.

Yes, I pushed it directly to my landice/develop. How can I fix it? I just PR a feature branch called damage-calving. Thanks！

matthewhoffman:
@tongzhangice , I've pushed three new commits making some minor modifications.  Look them over and make sure they are ok with you before I merge the PR.  
Also, I was originally thinking it would be good to add some version of these tests to our regression suite, but they take a little long to run (5-20 minutes), and we already have one enthalpy test in the regression suite.  I was trying to think if there was a shorter but still meaningful way to run either of these tests, but I don't think there really is.  What do you think?

tongzhangice:
@matthewhoffman I am ok with those changes. We can save the stage 1 result of Exp A somewhere and only run stage 2, and only plot the basal melt rate (since we only have the exact solution for that). But I am afraid people might get more confused about it. Another possible way is that we reduce the vertical mesh resolution, which I believe will also save some time.

tongzhangice:
@matthewhoffman We can change the vertical layer number[ here](https://github.com/tongzhangice/MPAS-Model/blob/landice/develop/testing_and_setup/compass/landice/enthalpy-benchmark/1m/A/config_setup_mesh_step.xml#L28). We may also change the size of mesh in [here](https://github.com/tongzhangice/MPAS-Model/blob/landice/develop/testing_and_setup/compass/landice/enthalpy-benchmark/1m/A/config_setup_mesh_step.xml#L11-L12). Currently it is 2x2 mesh, maybe we can change it a 1x2 mesh or something? I haven't tested it yet.

matthewhoffman:
@tongzhangice , I tried changing the number of vertical levels and it surprisingly had little effect.  (I think the code must be vectorized over the vertical or something like that.)  But I did find that reducing the output interval had a big impact on the run time, and I could also increase the time step a bit as well.  With those changes, Exp A can run in a little over a minute without having to make other modifications, which is short enough to add it to the regression suite.  I will work on doing that, and then get this merged.

tongzhangice:
@matthewhoffman Thanks Matt. It is also a surprise to me. Thanks for testing that. Glad changing timestep works!

testing_and_setup/compass/landice/enthalpy-benchmark/plot_enthalpy_benchmark_A.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Create image plot
+plotname = 'enthalpy_A_results.png'
+plt.savefig(plotname, dpi=150)
+print('Saved plot as ' + plotname)
+
+displayImage = False
+if displayImage:
+   # Note: To get interactive plot, need to comment line at beginning of script "matplotlib.use('Agg')"
+   plt.show()
```
xylar: Merged already but for future reference, here's another option that might have been closer to what you wanted in this case.

You could put `displayImage = False` earlier (before your first `plt` call) and then change to the `Agg` backend on the fly:
```python
displayImage = False
if not displayImage:
    plt.switch_backend('Agg')

...
if displayImage:
   plt.show()
```

`switch_backend` has some caveats you should look into, which is why the way you have it now is more reliable, but it would work in this particular case.

matthewhoffman:
Thanks, @xylar , that is very useful to know.  I wondered if something like that was possible, but assumed you couldn't change the backend after the figure was created.  In this particular case, it's not that important that things are particularly automated, so the solution that got merged is fine.

xylar:
> but assumed you couldn't change the backend after the figure was created

I think that's true.  You have to select the backend before you create the figure, so you would need to move `displayImage = True/False` to before you create the figure.  And I don't think you can switch back to displaying to the screen once you've switched to `Agg`, but presumably that would have been fine here.

PR #426, opened 2020-01-20
Add a stack to the init_atmosphere core

MiCurry:
This commit adds a new module to the MPAS init_atmosphere core, mpas_stack, which is a simple and extensible stack implementation. It uses Fortran 2003's polymorphic features to allow different 'payloads' to be associated with a stack; thus it can be used for a number of applications.

At current, this module is not being used by any part of the init_atmosphere core, but a future PR could make use of its features.

A copy of `mpas_init_atm_core.F`, that uses and contains tests for mpas_stack.F, has been uploaded as a gist, https://gist.github.com/MiCurry/22ad2085f0ba10e325d96449a97a8434, which can be used to test mpas_stack.F.

### DIFF
--- a/src/core_init_atmosphere/mpas_stack.F
+++ b/src/core_init_atmosphere/mpas_stack.F
@@ -1,0 +1,280 @@
+module mpas_stack
+
+   implicit none
+
+   private
+
+   ! Public Subroutines and Structures
+   public :: mpas_stack_is_empty
+   public :: mpas_stack_push
+   public :: mpas_stack_pop
+   public :: mpas_stack_free
+
+   public :: mpas_stack_type, mpas_stack_payload_type
+
+   type mpas_stack_payload_type
+   end type mpas_stack_payload_type
+
+   type mpas_stack_type
+      type (mpas_stack_type), pointer :: next => null()
+      class (mpas_stack_payload_type), pointer :: payload => null()
+   end type mpas_stack_type
+
+   !***********************************************************************
+   !
+   !  module mpas_stack
+   !
+   !> \brief   MPAS Stack module
+   !> \author  Miles A. Curry
+   !> \date    04/04/19
+   !> \details
+   !>
+   !> Introduction
+   !> ==============
+   !> The MPAS stack is a simple, extensible data stack data structure for use
+   !> within the MPAS atmospheric model. It functions as a wrapper around a
+   !> polymorphic data structure to provide usage in different areas.
+   !>
+   !>
+   !> Creating a Stack
+   !> ==================
+   !> The stack data structure (`type (mpas_stack_type)`) is defined by a single 
+   !> `next` pointer > and a pointer to a `type (mpas_stack_payload_type)`, which 
+   !> is defined as a empty derived type.
+   !>
+   !> To use the stack, create a derived type that extends the `mpas_stack_payload_type` 
+   !> type. Define your extended derived type with members that meets your application.
+   !>
+   !> For instance:
+   !> ```
+   !> type, extends(mpas_stack_payload_type) :: my_payload_name
+   !>    ! Define the members of your type as you wish
+   !> end type my_payload_name
+   !>
+   !> class (my_payload_name), pointer :: item1 => null(), item2 => null()
+   !> ```
+   !>
+   !> The extended mpas_stack_payload_type will enable a user defined type to be 
+   !> associated with a stack item. The stack stores references of a payload, thus 
+   !> a single payload can be used in multiple push operations.
+   !>
+   !> You will then need to create a stack (or multiple stacks if you desire) as
+   !> the following:
+   !>
+   !> ```
+   !> type (mpas_stack_type), pointer :: stack1 => null(), stack2 => null()
+   !> ```
+   !>
+   !>  Pushing onto a Stack
+   !>  ====================
+   !>  You can push your items onto a stack as:
+   !>
+   !> ```
+   !> allocate(item1)
+   !> stack1 => mpas_stack_push(stack1, item1)
+   !> allocate(item2)
+   !> stack1 => mpas_stack_push(stack1, item2)
+   !> ```
+   !>
+   !> Popping an item off of the stack
+   !> ================================
+   !> Popping an item off of the stack will require a bit more work than pushing.
+   !> Because the payload is a polymorphic class , we will need to use the select 
+   !> case to get our type (or multiple types) back into a usable object:
+   !> ```
+   !> ! The item to pop items into
+   !> class (mpas_stack_payload_type), pointer :: top
+   !> type (my_payload_name), pointer :: my_item
+   !>
+   !> top => mpas_stack_pop(stack1)
+   !> select type(top)
+   !>    type is(my_payload_name)
+   !>       my_item => top
+   !> end  select
+   !> ```
+   !>
+   !> Note: It is recommended to create your own `pop` function so you can reduce 
+   !> the amount of coded needed. An example is provided at the bottom of
+   !> this module as the function `user_pop(..)`
+   !
+   !-----------------------------------------------------------------------
+
+   contains
+
+   !***********************************************************************
+   !
+   !  routine mpas_stack_is_empty
+   !
+   !> \brief   Returns .true. if the stack is empty, otherwise .false.
+   !> \author  Miles A. Curry
+   !> \date    01/28/20
+   !> Returns .true. If the stack is empty and/or if the stack is unassociated.
+   !
+   !-----------------------------------------------------------------------
+   function mpas_stack_is_empty(stack) result(is_empty)
+
+      implicit none
+      type (mpas_stack_type), intent(in), pointer :: stack
+      logical :: is_empty
+
+      is_empty = .true.
+      if (associated(stack)) then
+         is_empty = .false.
+         return
+      endif
+
+   end function mpas_stack_is_empty
+
+   !***********************************************************************
+   !
+   !  routine mpas_stack_push
+   !
+   !> \brief   Push an item onto stack
+   !> \author  Miles A. Curry
+   !> \date    01/28/20
+   !> \details
+   !>
+   !> Push a mpas_stack_payload_type type, onto `stack` and return the new stack. If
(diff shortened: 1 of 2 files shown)


MiCurry:
A companion repository has also been setup, https://github.com/MiCurry/mpas_stack to help test for memory leaks.

MiCurry:
@mgduda I think this is good to go, at least for inspection. I've compiled with Intel, GNU and PGI, and tested with Intel and GNU. I'm working on testing with PGI.

There is a single pitfall when using polymorphic types and the Intel compiler (which I think we've ran into and discussed before). With GNU, you can define your item as a `type` (i.e. `type(cell_t) :: cell`) and successfully push and deallocate this item with `mpas_stack_free`, but with Intel you cannot. You're able to allocate and push the item just fine,  but when you try to deallocate the item you get the runtime error:

```
forrtl: severe (173): A pointer passed to DEALLOCATE points to an object that cannot be deallocated
```
[Which appears to be a bug](https://software.intel.com/en-us/forums/intel-fortran-compiler/topic/390944). The workaround (as Doctor Fortran posted), is to use `class` instead of `type` (i.e `class(cell_t) :: cell)`. For our usage, I don't think it will cause any problems.

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,273 @@
...
+
+   public :: node, payload_t
+
+   type payload_t
+   end type payload_t
+
+   type node
+      type (node), pointer :: next => null()
+      class (payload_t), pointer :: payload => null()
+   end type node
```
mgduda: Since this type is made public by the module, we could consider naming it with an `mpas_` prefix, and similarly for `payload_t`. Also, if the `_t` suffix is used for the payload type, perhaps we should use it for the node type as well, and consider using `_type`, which is used elsewhere in MPAS infrastructure? Also, one can imagine all sorts of situations involving "nodes" in MPAS infrastructure, so perhaps qualifying that this is a *stack* node would be good. In all, then, how about `mpas_stack_node_type` and `mpas_stack_payload_type`?

If `mpas_stack_node_type` is too long, and if the module supports pushing entire stacks onto each other, then `mpas_stack_type` might be appropriate. By allocating a variable of type `mpas_stack_type`, the user is effectively creating a stack with one item.

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,273 @@
...
+
+   !***********************************************************************
+   !
+   !  module mpas_stack
+   !
+   !> \brief   MPAS KD-Tree module
```
mgduda: Should this be `stack` rather than `KD-Tree`?

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,273 @@
...
+      type (node), intent(in), pointer :: stack
+      logical :: is_empty
+
+      is_empty = .TRUE.
+      if (associated(stack)) then
+         is_empty = .FALSE.
+         return
+      endif
+
+      return
```
mgduda: The two `return` statements in this function could be omitted for simplicity.
Stylistically, the rest of MPAS tends to not use upper-case letters for `.TRUE.` and `.FALSE.`.

src/core_init_atmosphere/mpas_stack.F
mgduda: Having looked more at how the stack works, I think it seems reasonable to name the stack derived type as `mpas_stack_type`. The user doesn't really need to know that the stack is implemented as a linked list -- only that they have a pointer to an `mpas_stack_type`, onto which they can push or pop payloads.

mgduda:
@MiCurry Functionally, I think the module looks good. Besides a bit of clean-up, I'd just like to push for a renaming of the derived types.

src/core_init_atmosphere/mpas_stack.F
MiCurry: That sounds good to me, and I agree, best to try to get unique as names as possible. I didn't even think about naming the stack and payload in such a way, so thanks for the recommendation.

So, just to confirm, I'll rename `node` to be: `mpas_stack_type` and I'll rename `payload_t` to be `mpas_stack_payload_type`?

src/core_init_atmosphere/mpas_stack.F
MiCurry: Yes!

src/core_init_atmosphere/mpas_stack.F
MiCurry: Sounds great I'll remove the `return` and change the booleans to be lower-case.

src/core_init_atmosphere/mpas_stack.F
mgduda: Yes, `mpas_stack_type` and `mpas_stack_payload_type` sound good to me.

MiCurry:
Okay, @mgduda, I've made those changes that you requested in terms of clean-up and renamed the derived types as we discussed above! So I think this PR is good to go!

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !> within the MPAS atmospheric model. It functions as a wrapper around a
+   !> polymorphic data structure to provide usage in different areas.
+   !>
+   !>
+   !> Creating a Stack
+   !> ==================
+   !> The stack data structure (`type (mpas_stack_type)`) is defined by a single `next` pointer
+   !> and a pointer to a `type (mpas_stack_payload_type)`, which is defined as a empty derived type.
+   !>
+   !> To use the stack, create a derived that extends the `mpas_stack_payload_type` type.  Define your
```
mgduda: Should "derived that" be "derived type that"?

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !> For instance:
+   !> ```
+   !> type, extends(mpas_stack_payload_type) :: my_payload_name
+   !>    ! Define the members of your type as you wish
+   !> end type my_payload_name
+   !>
+   !> class (my_payload_name) :: item1, item2
+   !> ```
+   !>
+   !> The extended mpas_stack_payload_type define type will enable it to ride along with the stack.
```
mgduda: The word "define" could be omitted. Also, we could use more technical terminology than "ride along with"; perhaps something like "enable it to be pushed onto and popped from the stack."

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !> It will also enable you to push the same payload twice (if need-be).
```
mgduda: It's not clear how the extended type inherently allows itself to be pushed twice on the stack. Maybe here we could just mention that the stack stores *references* to payloads, and a payload can therefore be used in multiple push operations?

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !>
+   !> You will then need to create a stack (or multiple stacks if you desire) as
+   !> the following:
+   !>
+   !> ```
+   !> type (mpas_stack_type) :: stack1, stack2
```
mgduda: Do these need to be pointers (and therefore also allocated before being pushed)?

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !> To use the stack, create a derived that extends the `mpas_stack_payload_type` type.  Define your
+   !> extended derived type with members that meets your application.
+   !>
+   !> For instance:
+   !> ```
+   !> type, extends(mpas_stack_payload_type) :: my_payload_name
+   !>    ! Define the members of your type as you wish
+   !> end type my_payload_name
+   !>
+   !> class (my_payload_name) :: item1, item2
```
mgduda: Do these need to be pointers (and therefore allocated before being pushed)?

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !> ```
+   !> type (mpas_stack_type) :: stack1, stack2
+   !> ```
+   !>
+   !>  Pushing onto a Stack
+   !>  ====================
+   !>  You can push your items onto a stack as:
+   !>
+   !> ```
+   !> stack1 => mpas_satck_push(stack1, item1)
```
mgduda: "satck" should be "stack".

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !> ```
+   !>
+   !> Popping an item off of the stack
+   !> ================================
+   !> Popping an item off of the stack will require a bit more work then pushing.
```
mgduda: "then" should be "than".

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !> type (my_payload_name), pointer :: my_item
+   !>
+   !> top => mpas_stack_pop(stack1)
+   !> select type(top)
+   !>    type is(my_payload_name)
+   !>       my_item => top
+   !> end  select
+   !> ```
+   !>
+   !> Note: It is recommended to create your own `pop` function so you can limit
```
mgduda: I think the word "limit" can be removed.

src/core_init_atmosphere/mpas_stack.F
```diff
@@ -0,0 +1,274 @@
...
+   !-----------------------------------------------------------------------
+   function mpas_stack_pop(stack) result(top)
+
+      implicit none
+
+      type (mpas_stack_type), intent(inout), pointer :: stack
+      type (mpas_stack_type), pointer :: next => null()
+      class(mpas_stack_payload_type), pointer :: top
+
+      if ( .NOT. associated(stack)) then
```
mgduda: Just for consistency, I'd propose we change `.NOT.` to lower-case `.not.`.

MiCurry:
@mgduda Okay, those changes have been made!

PR #435, opened 2020-01-28
Ocean/compute slope taper on edge

vanroekel:
In the current ocean/develop the slope tapering is computed at cell centers and then is used directly at edges without interpolation.  At the bottom, near sloping bathymetry, there are cases where values of slopeTaper are used past maxLevelCell and this value was set to one in the code.  So the tapering was going to zero then artificially jumped to one causing issues with global min.  This PR computes slopeTaper on edges.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -404,7 +404,7 @@ contains
 
       !compute the slope taper on edges
       !$omp do schedule(runtime)
-      do iEdge = 1, nEdge
+      do iEdge = 1, nEdges
   ! Add bottom taper slope for this column.
          ! Top taper slope is computed before the cell loop, because it is the same
          ! for all cells.
@@ -529,7 +529,6 @@ contains
       deallocate(dTdzTop)
       deallocate(dSdzTop)
       deallocate(k33Norm)
-      deallocate(slopeTaper)
       deallocate(slopeTaperSurface)
 
       ! allow disabling of K33 for testing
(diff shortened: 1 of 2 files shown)


vanroekel:
This was tested in mpas-o stand alone.  Here is a plot of a three month simulation (temperature min).  Black is ocean/develop, red is this PR.  All settings for GM/Redi are the same.
Screen Shot 2020-01-28 at 7 34 15 AM

vanroekel:
@jonbob and @milenaveneziani here is the fix we discussed yesterday.  It seems to work really well.  From my tests it is a bit more costly performance wise.  Could you add this to your E3SM test and run a G-case again?  You just need to update mpas_ocn_gm.F and Registry.xml

milenaveneziani:
looks great @vanroekel. Should I just manually copy these files to my branch and do the test?

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -516,7 +529,6 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
       deallocate(dTdzTop)
       deallocate(dSdzTop)
       deallocate(k33Norm)
-      deallocate(slopeTaper)
```
milenaveneziani: no `deallocate(slopeTaperEdge)`?

vanroekel: I put slopeTaperEdge in registry now in case we want to visualize it at some point.  This is why there is no deallocate here.

vanroekel:
@milenaveneziani, yes, I think the simplest is to just copy those two files into your E3SM source directory and build a new case.

mark-petersen:
@vanroekel don't wait for me, as I'm busy all day. Go ahead and merge when you are ready.

milenaveneziani:
@mark-petersen: I am running some tests with E3SM using the changes introduced here. I would wait for the results of these tests before merging this.

vanroekel:
Based on some Redi mixing papers and further testing.  I'm going to abandon this tapering and pursue a different kind.  I'm closing this PR.  I will issue a new PR with improved tapering soon.

PR #438, opened 2020-01-29
Add a KD-Tree to the init_atmosphere core

MiCurry:
This commit adds a new module to the MPAS init_atmosphere core, mpas_kd_tree,
which contains implementation of a K-Dimensional Tree. At present, the module
is not being used by any part of the init_atmosphere core; it is only compiled.

A copy of `mpas_init_atm_core.F`, that uses and contains tests for `mpas_kd_tree.F`, has been uploaded as a gist, https://gist.github.com/MiCurry/b5744ef36eec3b3ba40a8de806201ed1, which can be used to test the `.mpas_kd_tree` module.

### DIFF
--- a/src/core_init_atmosphere/mpas_kd_tree.F
+++ b/src/core_init_atmosphere/mpas_kd_tree.F
@@ -1,0 +1,409 @@
+module mpas_kd_tree
+
+   !***********************************************************************
+   !
+   !  module mpas_kd_tree
+   !
+   !> \brief   MPAS KD-Tree module
+   !> \author  Miles A. Curry
+   !> \date    01/28/20
+   !> A KD-Tree implementation to create and search perfectly balanced
+   !> KD-Trees.
+   !>
+   !> Use `mpas_kd_type` dervied type to construct points for mpas_kd_construct:
+   !>
+   !> real (kind=RKIND), dimension(:,:), allocatable :: array
+   !> type (mpas_kd_type), pointer :: tree => null()
+   !> type (mpas_kd_type), dimension(:), pointer :: points => null()
+   !>
+   !> allocate(array(k,n)) ! K dims and n points
+   !> allocate(points(n))
+   !> array(:,:) = (/.../)  ! Fill array with values
+   !>
+   !> do i = 1, n
+   !>    allocate(points(i) % point(k))    ! Allocate point with k dimensions
+   !>    points(i) % point(:) = array(:,i)
+   !>    points(i) % id = i                ! Or a value of your choice
+   !> enddo
+   !>
+   !> tree => mpas_kd_construct(points, k)
+   !>
+   !> call mpas_kd_free(tree)
+   !> deallocate(points)
+   !> deallocate(array)
+   !>
+   !
+   !-----------------------------------------------------------------------
+   use mpas_kind_types, only : RKIND
+
+   implicit none
+
+   private
+
+   public :: mpas_kd_type
+
+   ! Public Subroutines
+   public :: mpas_kd_construct
+   public :: mpas_kd_search
+   public :: mpas_kd_free
+
+   type mpas_kd_type
+      type (mpas_kd_type), pointer :: left => null()
+      type (mpas_kd_type), pointer :: right => null()
+
+      integer :: split_dim
+      real (kind=RKIND), dimension(:), pointer :: point => null()
+
+      integer :: id
+   end type mpas_kd_type
+
+   contains
+
+   !***********************************************************************
+   !
+   !  recursive routine mpas_kd_construct_internal
+   !
+   !> \brief   Create a KD-Tree from a set of k-Dimensional points
+   !> \author  Miles A. Curry
+   !> \date    01/28/20
+   !> \details
+   !> Private, recursive function to construct a KD-Tree from an array
+   !> of mpas_kd_type, points, and return the root of the tree.
+   !>
+   !> ndims should be the dimensioned of each individual point found
+   !> in points and npoints should be the number of points. dim represents
+   !> the current split dimensioned and is used internally. Upon calling
+   !> this function, dim should always be set to 0.
+   !
+   !-----------------------------------------------------------------------
+   recursive function mpas_kd_construct_internal(points, ndims, npoints, dim) result(tree)
+
+      implicit none
+
+      ! Input Variables
+      type (mpas_kd_type), dimension(:), target :: points
+      integer, intent(in) :: ndims
+      integer, value :: npoints
+      integer, value :: dim
+
+      ! Return Value
+      type (mpas_kd_type), pointer :: tree
+
+      ! Local Variables
+      integer :: median
+
+      if (npoints < 1) then
+         tree => null()
+         return
+      endif
+
+      ! Sort the points at the split dimension
+      dim = mod(dim, ndims) + 1
+      call quickSort(points, dim, 1, npoints, ndims)
+
+      median = (1 + npoints) / 2
+
+      points(median) % split_dim = dim
+      tree => points(median)
+
+      ! Build the right and left sub-trees but do not include the median
+      ! point (the root of the current tree)
+      if (npoints /= 1) then
+          points(median) % left => mpas_kd_construct_internal(points(1:median-1), ndims, median - 1, points(median) % split_dim)
+          points(median) % right => mpas_kd_construct_internal(points(median+1:npoints), ndims, npoints - median, &
+                                                                                             points(median) % split_dim)
+      endif
+
+   end function mpas_kd_construct_internal
+
+
+   !***********************************************************************
+   !
+   !  routine mpas_kd_construct
+   !
+   !> \brief   Construct a balanced KD-Tree
+   !> \author  Miles A. Curry
+   !> \date    01/28/20
+   !> \details
+   !> Create and return a perfectly balanced KD-Tree from an array of
+   !> mpas_kd_type, points. The point member of every element of the points
+   !> array should be allocated and set to the points desired to be in the
+   !> KD-Tree and ndims should be the dimensions of the points.
+   !>
+   !> Upon error, the returned tree will be unassociated.
+   !
+   !-----------------------------------------------------------------------
+   function mpas_kd_construct(points, ndims) result(tree)
+
+      implicit none
+
+      ! Input Varaibles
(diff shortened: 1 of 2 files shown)


MiCurry:
A companion repository has been setup, https://github.com/MiCurry/mpas_kd_tree, which can be used to help tests for memory leaks.

MiCurry:
@mgduda, what do you think of constructing the KD-Tree with an array of `mpas_kd_tree`? Do you think it might be better to create the KD-Tree with an array of floats? I think the first implementation of this module did just that, but was changed to the present configuration so nodes could contain the cell ID.

We could add a new argument to `mpas_kd_construct`, `integer, dimension(:), intent(in) :: ids`, which would represent the id's of each individual point. `mpas_kd_construct` could then allocate assign the corresponding ID with the associated point.

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,394 @@
...
+   !> \author  Miles A. Curry
+   !> \date    01/28/20
+   !> A KD-Tree implementation to create and search perfectly balanced
+   !> KD-Trees.
+   !>
+   !> Use `mpas_kd_type` dervied type to construct points for mpas_kd_construct:
+   !>
+   !> real, dimension(:,:), allocatable :: array
+   !> type (mpas_kd_tree), pointer :: tree => null()
+   !> type (mpas_kd_tree), dimension(:), pointer :: points => null()
```
mgduda: Should these variables be of type `mpas_kd_type` rather than `mpas_kd_tree`?

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,394 @@
...
+   !>
+   !> allocate(array(k,n)) ! K dims and n points
+   !> allocate(points(n))
+   !> array(:,:) = (/.../)  ! Fill array with values
+   !>
+   !> do i = 0, n
```
mgduda: Should this be `1, n` or `0, n-1`?

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,394 @@
...
+   !
+   !> \brief   MPAS KD-Tree module
+   !> \author  Miles A. Curry
+   !> \date    01/28/20
+   !> A KD-Tree implementation to create and search perfectly balanced
+   !> KD-Trees.
+   !>
+   !> Use `mpas_kd_type` dervied type to construct points for mpas_kd_construct:
+   !>
+   !> real, dimension(:,:), allocatable :: array
```
mgduda: For consistency, we might consider using `real (kind=RKIND)` in this example rather than simply `real`. Also, in the example below, after the elements of `array` are copied to the `point` members of the `points` array, we can `deallocate(array)` to make clear to readers that the memory used by `array` should be freed to prevent a memory leak.

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,394 @@
...
+   !> do i = 0, n
+   !>    allocate(points(i) % point(k))    ! Allocate point with k dimensions
+   !>    points(i) % point(:) = array(:,i)
+   !>    points(i) % id = i                ! Or a value of your choice
+   !> enddo
+   !>
+   !> tree => mpas_kd_construct(points, k)
+   !
+   !-----------------------------------------------------------------------
+   use mpas_derived_types
```
mgduda: Rather than using the entire derived_types module, we can just
```fortran
   use mpas_kind_types, only : RKIND
```

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,394 @@
...
+      else if (query(kdtree % split_dim) < kdtree % point(kdtree % split_dim)) then
+         if (associated(kdtree % left)) then ! Search left
+            call mpas_kd_search_internal(kdtree % left, query, res, distance)
+         endif
+         if ((kdtree % point(kdtree % split_dim) - query(kdtree % split_dim))**2 <= distance .AND. associated(kdtree % right)) then
+            call mpas_kd_search_internal(kdtree % right, query, res, distance) ! Check the other subtree
+         endif
+      else ! Nearest point could be in either left or right subtree, so search both
+         if(associated(kdtree % right)) call mpas_kd_search_internal(kdtree % right, query, res, distance)
+         if(associated(kdtree % left)) call mpas_kd_search_internal(kdtree % left, query, res, distance)
```
mgduda: For consistency, I'd suggest not using a one-line if-test here, and instead using
```fortran
if (associated(kdtree % right)) then
   call mpas_kd_search_internal(kdtree % right, query, res, distance)
end if
```

mgduda:
@MiCurry As a general consistency issue, could you change all occurrences of `.AND.`, `.TRUE.`, etc. to lower-case, and also make consistent the use of whitespace in control constructs? In various places, we have
```fortran
if(something) then
```
```fortran
if (something) then
```
```fortran
if ( something ) then
```
for example.

Otherwise, I think the new kd-tree module looks good.

mgduda:
@MiCurry Regarding the current construction of kd-trees with arrays of `mpas_kd_type`, I think that the module is good as-is. If we constructed trees from arrays of `real` and `integer` values, would that imply a need to either duplicate the values in those arrays (in kd-tree nodes) or to require the user to never deallocate the `real` and `integer` arrays for the lifetime of the kd-tree?

MiCurry:
Thanks for the review @mgduda. I'll get started on making the reviews that you requested.

> @MiCurry Regarding the current construction of kd-trees with arrays of mpas_kd_type, I think that the module is good as-is. If we constructed trees from arrays of real and integer values, would that imply a need to either duplicate the values in those arrays (in kd-tree nodes) or to require the user to never deallocate the real and integer arrays for the lifetime of the kd-tree?

I suppose we would have the two options above if we constructed trees from arrays of reals and integers. As you suggested, one would allocate a new node for each value and duplicate the values into the new nodes; while the other could allocate new nodes and then point to the original values.

I guess my thought with passing in real and integer arrays would be to avoid having the user allocate and deallocate an array of nodes, but I see that this would require us to duplicate data, which is probably not preferable. The other method (pointing to the original values) would be similar to the current method, so perhaps it is best if we use the current method.

MiCurry:
Okay, @mgduda I have just pushed an update to this branch with the changes you requested. Please let me know if there is anything else!

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,402 @@
...
+   !> enddo
+   !>
+   !> tree => mpas_kd_construct(points, k)
+   !>
+   !> call mpas_kd_free(tree)
+   !> deallocate(points)
+   !>
+   !
+   !-----------------------------------------------------------------------
+   use mpas_derived_types, only : RKIND
```
mgduda: The `RKIND` constant is ultimately defined in the `mpas_kind_types` module, so it may be cleaner to obtain `RKIND` from there rather than from `mpas_derived_types`.

src/core_init_atmosphere/mpas_kd_tree.F
mgduda: Sorry, I should have emphasized the change from `mpas_derived_types` to `mpas_kind_types` in addition to the use of an `only` clause.

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,402 @@
...
+   !> do i = 1, n
+   !>    allocate(points(i) % point(k))    ! Allocate point with k dimensions
+   !>    points(i) % point(:) = array(:,i)
+   !>    points(i) % id = i                ! Or a value of your choice
+   !> enddo
+   !>
+   !> tree => mpas_kd_construct(points, k)
+   !>
+   !> call mpas_kd_free(tree)
+   !> deallocate(points)
```
mgduda: Should we also `deallocate(array)` in this example code? This might help users to appreciate that the `mpas_kd_free` routine doesn't free up the memory that was allocated for the `array` array earlier,
and that this memory has to be explicitly freed to prevent a memory leak.

mgduda:
Other than the points about `RKIND` and `deallocate(array)`, I think this PR is ready to be merged.

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: Ah right, I always have a hard time finding the right module to pull RKIND from in the framework directory, so good to know!

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: Yes, that's a good idea! Sorry if I missed this addition earlier!

MiCurry:
Okay, the last push contains fixes to the `RKIND` and `deallocate(array)`, so this PR is ready to be merged!

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -0,0 +1,403 @@
...
+   subroutine mpas_kd_search(kdtree, query, res, distance)
+
+      implicit none
+      type (mpas_kd_type), pointer, intent(in) :: kdtree
+      real (kind=RKIND), dimension(:), intent(in) :: query
+      type (mpas_kd_type), pointer, intent(inout) :: res
+      real (kind=RKIND), intent(out), optional :: distance
+
+      real (kind=RKIND) :: dis
+
```
mgduda: If the kd-tree was constructed with zero points, then the tree pointer returned by `mpas_kd_construct` will be an unassociated pointer. If this is then passed to `mpas_kd_search`, a segfault will likely result. We could add a check like
```fortran
     if (.not. associated(kdtree)) then
        res => null()
        return
     end if
```
here in the `mpas_kd_search` routine to protect against this. We could then add a note in the description for `mpas_kd_search` that if no nearest neighbor was found, an unassociated `res` pointer is returned (with the only likely circumstance under which this would occur being an empty tree).

MiCurry: That sounds like a great idea, I'll go ahead and implement that.

MiCurry:
Okay @mgduda, I have added into the `mpas_kd_search` protection for searching on an unassociated kdtree!

PR #433, opened 2020-01-28
Change global_ocean ISC test cases to use BedMachine

xylar:
This merge adds support for a new topography for runs with ice-shelf cavities (wISC).  The new topography combines BedMachineAntarctica and the GEBCO global topography.  The combined data set is on a 1/20 degree grid (2x higher res than the previous topography based on Bedmap2 and ETOPO2).

PR #440, opened 2020-02-03
Ocean init mode: add subroutine for smoothing topography

### LINKED ISSUE #436: Topography smoothing
I'm concerned that we are not performing an adequate level of topography smoothing in MPAS-Ocean `global_ocean` configurations.  As an example, see bathymetry plot in #433.

I think it would be appropriate to do at least a few iterations of nearest-neighbor smoothing (with weights TBD) as part of initialization.  The step currently called smoothing:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/mode_init/Registry_global_ocean.xml#L178-L181
is actually just filling in holes in topography:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F#L976-L1007
which, while important, is not what I would think of when setting  the flag `config_global_ocean_smooth_topography`.

Given that I'm introducing a new bathymetry and ice draft for E3SM Cryosphere simulations, now might be the right time to talk about this issue.

xylar:
This feature performs an iterative weighted average of a field with a given smoothing weight and for a given number of iterations.  Currently, it is only used for Cryosphere test cases (with ice-shelf cavities, `wISC`).

Also renames config option `...fill_bathymetry_holes` from `...smooth_topography` because this is a more accurate description of what is done when this flag is `.true.`

closes #436

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -412,7 +412,7 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_Redi_BLD_taper_layers" type="integer" default_value="3" units="non-dimensional"
-					description="Number of layers below boundary layer depth, over which Redi coefficient is tapered." 
+					description="Number of layers below boundary layer depth, over which Redi coefficient is tapered."
 					possible_values="Integer between 1 and nVertLevels. Use 1 for no tapering."
 		/>
 	</nml_record>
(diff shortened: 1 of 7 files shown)


xylar:
Currently, I'm having trouble with scratch fields (at least on my laptop).  I'll try on LANL IC tomorrow and ask @mark-petersen for help if I'm still not able to figure it out.

xylar:
## Results EC60to30wISC

Below are the bathymetry, ice draft and fraction of ice cover, without smoothing followed by with smoothing with a weight (of the central cell relative to the average of its neighbors) of 0.9 and 6 iterations.

bot_orig
bot_smooth
draft_orig
draft_smooth
frac_orig
frac_smooth

xylar:
weight of 0.9 and 6 iterations might be too much for this data set, particularly because it was already smoothed at 0.1 degrees on the original grid, but it does demonstrate the concept.  I think this amount of smoothing might be more appropriate at higher res and in particular for the BedMachine bathymetry (unsmoothed) that we will hopefully move to for Cryosphere configurations.

xylar:
## Results SO60to10wISC

Same as above but at higher resolution.

bot_orig
bot_smooth
draft_orig
draft_smooth
frac_orig
frac_smooth

In my testing, the SSH adjustment phase failed, so there's clearly some work still to do to make that more robust but I will do that in #433 with the new bathymetry.

xylar:
@mark-petersen, I assume you can handle the registry conflict but let me know.

PR #280, opened 2019-06-13
Convert GM-Redi to triad formulation

mark-petersen:
Currently, the Redi along-isopycnal mixing formulation used simple spatial averaging to compute the isopycnal slope. This PR introduces triad averaging, which uses a very specific form of an expansion of the slope using three cells, the thermal expansion and saline contraction coefficients.

PR #445, opened 2020-02-06
Bug fixes in Redi and new tapering options

vanroekel:
This PR fixes a few issues with the Redi mixing implementation and also adds tapering based on Danabasoglu and McWilliams 95 and Large et al 1997.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -2761,14 +2761,6 @@
 			 description="Magnitude of slope of isopycnal surface, using triad through this cell and edge, angled up. Uses expansion of equation of state."
 			 packages="forwardMode;analysisMode"
     />
-    <var name="slopeTaperEdgeUp" type="real" dimensions="nVertLevels nEdges Time" units="non-dimensional"
-       description="tapering of Redi mixing at cell edges"
-       packages="forwardMode;analysisMode"
-    />
-   <var name="slopeTaperEdgeDown" type="real" dimensions="nVertLevels nEdges Time" units="non-dimensional"
-       description="tapering of Redi mixing at cell edges"
-       packages="forwardMode;analysisMode"
-    />
 		<var name="k33" type="real" dimensions="nVertLevelsP1 nCells Time" units="m^2 s^{-1}"
 			 description="The (3,3) entry of the Redi diffusion tensor. Added to the model vertical diffusion. Defined at the top of cell k"
 			 packages="forwardMode;analysisMode"
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -395,33 +395,17 @@
 					description="The Redi diffusion coefficient"
 					possible_values="Positive real numbers"
     />
-    <nml_option name="config_REDI_half_slope" type="real" default_value="0.004" units="non-dimensional"
+    <nml_option name="config_Redi_half_slope" type="real" default_value="0.004" units="non-dimensional"
         description="value of slope where set to half max"
         possible_values="positive real numbers less than slope max"
     />
-    <nml_option name="config_REDI_half_width" type="real" default_value="0.001" units="non-dimensional"
+    <nml_option name="config_Redi_half_width" type="real" default_value="0.001" units="non-dimensional"
         description="value of half width of tanh function limiter"
         possible_values="very small positive real numbers"
     />
 		<nml_option name="config_Redi_closure" type="character" default_value="constant" units="unitless"
 					description="Control what type of function is used for Redi $\kappa$. Only 'constant' is currently supported."
 					possible_values="'constant', 'data', 'match_GM'"
-		/>
-		<nml_option name="config_Redi_surface_taper_layers" type="integer" default_value="3" units="non-dimensional"
-					description="Number of layers below surface over which Redi coefficient is tapered."
-					possible_values="Integer between 1 and nVertLevels. Use 1 for no tapering, and Redi will be 0 at k=1 and full at k=2."
-		/>
-		<nml_option name="config_Redi_bottom_taper_layers" type="integer" default_value="3" units="non-dimensional"
-					description="Number of layers above bottom over which Redi coefficient is tapered."
-					possible_values="Integer between 1 and nVertLevels. Use 1 for no tapering, and Redi will be 0 at bottom and full one layer up."
-		/>
-		<nml_option name="config_Redi_zero_in_boundary_layer" type="logical" default_value=".true." units="unitless"
-					description="If true, Redi kappa is set to zero above the boundary layer depth"
-					possible_values=".true. or .false."
-		/>
-		<nml_option name="config_Redi_BLD_taper_layers" type="integer" default_value="3" units="non-dimensional"
-					description="Number of layers below boundary layer depth, over which Redi coefficient is tapered." 
-					possible_values="Integer between 1 and nVertLevels. Use 1 for no tapering."
 		/>
 	</nml_record>
 	<nml_record name="GM_eddy_parameterization" mode="forward">
@@ -2752,7 +2736,11 @@
 		<var name="surfaceBuoyancyForcing" type="real" dimensions="nCells Time" units="m^2 s^{-3}"
 			 description="CVMix/KPP: diagnosed surface buoyancy flux due to heat, salt and freshwater fluxes. Positive flux increases buoyancy."
 			 packages="forwardMode;analysisMode"
-		/>
+    />
+    <var name="rediRossbyRadius" type="real" dimensions="nEdges Time" units="m"
+       description="Rossby radius of deformation used to compute Redi limiting near surface"
+       packages="forwardMode;analysisMode"
+    />
 		<var name="slopeTriadUp" type="real" dimensions="nVertLevels TWO nEdges Time" units="non-dimensional"
 			 description="Magnitude of slope of isopycnal surface, using triad through this cell and edge, angled up. Uses expansion of equation of state."
 			 packages="forwardMode;analysisMode"
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -394,7 +394,15 @@
 		<nml_option name="config_Redi_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="The Redi diffusion coefficient"
 					possible_values="Positive real numbers"
-		/>
+    />
+    <nml_option name="config_REDI_half_slope" type="real" default_value="0.004" units="non-dimensional"
+        description="value of slope where set to half max"
+        possible_values="positive real numbers less than slope max"
+    />
+    <nml_option name="config_REDI_half_width" type="real" default_value="0.001" units="non-dimensional"
+        description="value of half width of tanh function limiter"
+        possible_values="very small positive real numbers"
+    />
 		<nml_option name="config_Redi_closure" type="character" default_value="constant" units="unitless"
 					description="Control what type of function is used for Redi $\kappa$. Only 'constant' is currently supported."
 					possible_values="'constant', 'data', 'match_GM'"
@@ -2753,7 +2761,11 @@
 			 description="Magnitude of slope of isopycnal surface, using triad through this cell and edge, angled up. Uses expansion of equation of state."
 			 packages="forwardMode;analysisMode"
     />
-    <var name="slopeTaperEdge" type="real" dimensions="nVertLevels nEdges Time" units="non-dimensional"
+    <var name="slopeTaperEdgeUp" type="real" dimensions="nVertLevels nEdges Time" units="non-dimensional"
+       description="tapering of Redi mixing at cell edges"
+       packages="forwardMode;analysisMode"
+    />
+   <var name="slopeTaperEdgeDown" type="real" dimensions="nVertLevels nEdges Time" units="non-dimensional"
        description="tapering of Redi mixing at cell edges"
        packages="forwardMode;analysisMode"
     />
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -2752,7 +2752,11 @@
 		<var name="slopeTriadDown" type="real" dimensions="nVertLevels TWO nEdges Time" units="non-dimensional"
 			 description="Magnitude of slope of isopycnal surface, using triad through this cell and edge, angled up. Uses expansion of equation of state."
 			 packages="forwardMode;analysisMode"
-		/>
+    />
+    <var name="slopeTaperEdge" type="real" dimensions="nVertLevels nEdges Time" units="non-dimensional"
+       description="tapering of Redi mixing at cell edges"
+       packages="forwardMode;analysisMode"
+    />
 		<var name="k33" type="real" dimensions="nVertLevelsP1 nCells Time" units="m^2 s^{-1}"
 			 description="The (3,3) entry of the Redi diffusion tensor. Added to the model vertical diffusion. Defined at the top of cell k"
 			 packages="forwardMode;analysisMode"
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -103,6 +103,7 @@ module ocn_time_integration_rk4
       type (mpas_pool_type), pointer :: forcingPool
       type (mpas_pool_type), pointer :: scratchPool
       type (mpas_pool_type), pointer :: swForcingPool
+      type (mpas_pool_type), pointer :: rediPool
 
       integer :: rk_step
 
@@ -938,7 +939,7 @@ module ocn_time_integration_rk4
 
       type (mpas_pool_type), pointer :: meshPool, verticalMeshPool
       type (mpas_pool_type), pointer :: statePool, diagnosticsPool, forcingPool
-      type (mpas_pool_type), pointer :: scratchPool, tendPool, provisStatePool
+      type (mpas_pool_type), pointer :: scratchPool, rediPool, tendPool, provisStatePool
       type (mpas_pool_type), pointer :: swForcingPool, tracersPool
 
       real (kind=RKIND), dimension(:), pointer :: sshCur
@@ -960,6 +961,7 @@ module ocn_time_integration_rk4
       call mpas_pool_get_subpool(block % structs, 'forcing', forcingPool)
       call mpas_pool_get_subpool(block % structs, 'scratch', scratchPool)
       call mpas_pool_get_subpool(block % structs, 'tend', tendPool)
+      call mpas_pool_get_subpool(block % structs, 'redi', rediPool)
       call mpas_pool_get_subpool(block % structs, 'provis_state', provisStatePool)
       call mpas_pool_get_subpool(block % structs, 'shortwave', swForcingPool)
 
@@ -995,7 +997,7 @@ module ocn_time_integration_rk4
       endif
       call mpas_threading_barrier()
 
-      call ocn_tend_tracer(tendPool, provisStatePool, forcingPool, diagnosticsPool, meshPool, swForcingPool, &
+      call ocn_tend_tracer(tendPool, rediPool, provisStatePool, forcingPool, diagnosticsPool, meshPool, swForcingPool, &
                            scratchPool, dt, activeTracersOnlyIn=.false., timeLevelIn=1)
       call mpas_threading_barrier()
 
@@ -1009,7 +1011,7 @@ module ocn_time_integration_rk4
 
       type (mpas_pool_type), pointer :: meshPool, verticalMeshPool
       type (mpas_pool_type), pointer :: statePool, diagnosticsPool, forcingPool
-      type (mpas_pool_type), pointer :: scratchPool, tendPool, provisStatePool
+      type (mpas_pool_type), pointer :: rediPool, scratchPool, tendPool, provisStatePool
       type (mpas_pool_type), pointer :: swForcingPool, tracersPool
 
       real (kind=RKIND), dimension(:), pointer :: sshCur
@@ -1031,6 +1033,7 @@ module ocn_time_integration_rk4
       call mpas_pool_get_subpool(block % structs, 'forcing', forcingPool)
       call mpas_pool_get_subpool(block % structs, 'scratch', scratchPool)
       call mpas_pool_get_subpool(block % structs, 'tend', tendPool)
+      call mpas_pool_get_subpool(block % structs, 'redi', rediPool)
       call mpas_pool_get_subpool(block % structs, 'provis_state', provisStatePool)
       call mpas_pool_get_subpool(block % structs, 'shortwave', swForcingPool)
 
@@ -1084,7 +1087,7 @@ module ocn_time_integration_rk4
       endif
       call mpas_threading_barrier()
 
-      call ocn_tend_tracer(tendPool, provisStatePool, forcingPool, diagnosticsPool, meshPool, swForcingPool, &
+      call ocn_tend_tracer(tendPool, rediPool, provisStatePool, forcingPool, diagnosticsPool, meshPool, swForcingPool, &
                            scratchPool, dt, activeTracersOnlyIn=.false., timeLevelIn=1)
       call mpas_threading_barrier()
 
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
@@ -105,6 +105,7 @@ module ocn_time_integration_split
       type (mpas_pool_type), pointer :: forcingPool
       type (mpas_pool_type), pointer :: scratchPool
       type (mpas_pool_type), pointer :: swForcingPool
+      type (mpas_pool_type), pointer :: rediPool
 
       type (dm_info) :: dminfo
       integer :: iCell, i,k,j, iEdge, cell1, cell2, split_explicit_step, split, &
@@ -1501,7 +1502,8 @@ module ocn_time_integration_split
             call mpas_pool_get_subpool(block % structs, 'mesh', meshPool)
             call mpas_pool_get_subpool(block % structs, 'scratch', scratchPool)
             call mpas_pool_get_subpool(block % structs, 'shortwave', swForcingPool)
-            call ocn_tend_tracer(tendPool, statePool, forcingPool, diagnosticsPool, meshPool, swForcingPool, scratchPool, &
+            call mpas_pool_get_subpool(block % structs, 'redi', rediPool)
+            call ocn_tend_tracer(tendPool, rediPool, statePool, forcingPool, diagnosticsPool, meshPool, swForcingPool, scratchPool, &
                     dt, activeTracersOnly, 2)
 
             block => block % next
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -404,7 +404,7 @@ contains
 
       !compute the slope taper on edges
       !$omp do schedule(runtime)
-      do iEdge = 1, nEdge
+      do iEdge = 1, nEdges
   ! Add bottom taper slope for this column.
          ! Top taper slope is computed before the cell loop, because it is the same
          ! for all cells.
@@ -529,7 +529,6 @@ contains
       deallocate(dTdzTop)
       deallocate(dSdzTop)
       deallocate(k33Norm)
-      deallocate(slopeTaper)
       deallocate(slopeTaperSurface)
 
       ! allow disabling of K33 for testing
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -106,16 +106,18 @@ contains
          layerThickness, &
          areaCellSum, inSituThermalExpansionCoeff, inSituSalineContractionCoeff, kappaGM3D
 
-      real(kind=RKIND), dimension(:), pointer   :: gmBolusKappa,  cGMphaseSpeed, bottomDepth, indexBoundaryLayerDepth
+      real(kind=RKIND), dimension(:), pointer   :: gmBolusKappa,  cGMphaseSpeed
       real(kind=RKIND), dimension(:,:,:), pointer :: slopeTriadUp, slopeTriadDown
+      real(kind=RKIND), dimension(:), pointer :: rediRossbyRadius
       real(kind=RKIND), dimension(:), pointer   :: areaCell, dcEdge, dvEdge, tridiagA, tridiagB, tridiagC, rightHandSide
       integer, dimension(:), pointer   :: maxLevelEdgeTop, maxLevelCell, nEdgesOnCell
       integer, dimension(:,:), pointer :: cellsOnEdge, edgesOnCell, edgeSignOnCell
       integer                          :: i, k, iEdge, cell1, cell2, iCell, N, iter, iCellSelf
       real(kind=RKIND)                 :: h1, h2, areaEdge, c, BruntVaisalaFreqTopEdge, rtmp, stmp, maxSlopeK33
-      real(kind=RKIND)                 :: bottomAv, sumN2, countN2, maxN, kappaSum, ltSum
+      real(kind=RKIND)                 :: sumN2, countN2, maxN, kappaSum, ltSum
       real(kind=RKIND) :: dcEdgeInv, drhoDx, drhoDT, drhoDS, dTdx, dSdx
-      real(kind=RKIND) :: slopeTaperUp, slopeTaperDown, surfaceTaperCoef, bottomTaperCoef, bldEdge
+      real(kind=RKIND) :: slopeTaperUp, slopeTaperDown, sfcTaperUp, sfcTaperDown
+      real(kind=RKIND) :: verticalDisplacement, rValue, zEdge
       real(kind=RKIND), dimension(:), allocatable :: dzTop, dTdzTop, dSdzTop, k33Norm
       
       ! Dimensions
@@ -164,9 +166,8 @@ contains
       call mpas_pool_get_array(diagnosticsPool, 'gmStreamFuncTopOfEdge', gmStreamFuncTopOfEdge)
       call mpas_pool_get_array(diagnosticsPool, 'gmStreamFuncTopOfCell', gmStreamFuncTopOfCell)
       call mpas_pool_get_array(diagnosticsPool, 'gmBolusKappa', gmBolusKappa)
-      call mpas_pool_get_array(diagnosticsPool, 'indexBoundaryLayerDepth', indexBoundaryLayerDepth)
+      call mpas_pool_get_array(diagnosticsPool, 'rediRossbyRadius', rediRossbyRadius)
 
-      call mpas_pool_get_array(meshPool, 'bottomDepth', bottomDepth)
       call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop',  maxLevelEdgeTop)
       call mpas_pool_get_array(meshPool, 'maxLevelCell',  maxLevelCell)
       call mpas_pool_get_array(meshPool, 'cellsOnEdge',  cellsOnEdge)
@@ -453,16 +454,29 @@ contains
 
                ! set taper of slope ('F' function from Danabasoglu and McWilliams 95)
                slopeTaperUp = 0.5_RKIND*(1.0 - tanh((abs(slopeTriadUp(k,iCellSelf,iEdge)) - &
-                  config_REDI_half_slope)/config_REDI_half_width))
+                  config_Redi_half_slope)/config_Redi_half_width))
                slopeTaperDown = 0.5_RKIND*(1.0 - tanh((abs(slopeTriadDown(k,iCellSelf,iEdge)) - &
-                  config_REDI_half_slope)/config_REDI_half_width))
+                  config_Redi_half_slope)/config_Redi_half_width))
+
+               zEdge = 0.5_RKIND*(zMid(k,cell1) + zMid(k,cell2))
+               verticalDisplacement = rediRossbyRadius(iEdge)*abs(slopeTriadUp(k,iCellSelf,iEdge))
+               rValue = zEdge / verticalDisplacement
+
+               sfcTaperUp = 0.5_RKIND*(1.0_RKIND + sin(3.141592_RKIND*(rValue - 0.5_RKIND)))
+               sfcTaperUp = max(0.0_RKIND, min(sfcTaperUp, 1.0_RKIND))
+
+               verticalDisplacement = rediRossbyRadius(iEdge)*abs(slopeTriadDown(k,iCellSelf,iEdge))
+               rValue = zEdge / verticalDisplacement
+
+               sfcTaperDown = 0.5_RKIND*(1.0_RKIND + sin(3.141592_RKIND*(rValue - 0.5_RKIND)))
+               sfcTaperDown = max(0.0_RKIND, min(sfcTaperDown, 1.0_RKIND))
 
                ! Slope can be unbounded in regions of neutral stability.
                ! Reset slope to maximum slope limit.
                slopeTriadUp(k,iCellSelf,iEdge) = &
-                  slopeTaperUp * slopeTriadUp(k,iCellSelf,iEdge)
+                  slopeTaperUp * sfcTaperUp * slopeTriadUp(k,iCellSelf,iEdge)
                slopeTriadDown(k,iCellSelf,iEdge) = &
-                  slopeTaperDown * slopeTriadDown(k,iCellSelf,iEdge)
+                  slopeTaperDown * sfcTaperDown * slopeTriadDown(k,iCellSelf,iEdge)
 
                ! Griffies 1998 eqn 34
                k33(k  ,iCell) = k33(k  ,iCell) + areaEdge*dzTop(k  )*slopeTriadUp(k,iCellSelf,iEdge)**2
@@ -769,7 +783,7 @@ contains
       integer :: iEdge
       integer, pointer :: nEdges
       integer, pointer :: nVertLevels
-      real(kind=RKIND), dimension(:), pointer   :: gmBolusKappa
+      real(kind=RKIND), dimension(:), pointer   :: fEdge, gmBolusKappa, rediRossbyRadius
       integer, dimension(:,:), pointer :: cellsOnEdge, edgesOncell
 
       err = 0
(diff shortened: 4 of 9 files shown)


vanroekel:
Here is a plot of minimum temperature from the tests (black is the latest from this PR).  This is an E3SM G-case.    We are evaluating after 30 years of simulation.

Screen Shot 2020-02-04 at 10 00 35 PM

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -450,14 +401,34 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
+                          / dzTop(k)
+            end do
+            ! mrp performance: could move above cell loop later
+            dzTop(1) = -1e-15_RKIND
+            dTdzTop(1) = -1e-15_RKIND
+            dSdzTop(1) = -1e-15_RKIND
+            dzTop(maxLevelCell(sVal)+1) = -1e-15_RKIND
+            dTdzTop(maxLevelCell(sVal)+1) = -1e-15_RKIND
+            dSdzTop(maxLevelCell(sVal)+1) = -1e-15_RKIND
+
```
vanroekel: I think we need to compute dtdz and dsdz each time.  There are times when the stratification is computed for iCell but this is not consistent with iCellSelf.  This makes most sense to me comparing to the picture in Griffies et al 1998

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -650,6 +655,15 @@ subroutine ocn_tend_tracer(tendPool, statePool, forcingPool, diagnosticsPool, me
...
 
+               modifiedGroupName = trim(groupItr % memberName) // "_redi_term1"
+               call mpas_pool_get_array(redi1Pool, trim(modifiedGroupName), redi_term1)
+
+               modifiedGroupName = trim(groupItr % memberName) // "_redi_term2"
+               call mpas_pool_get_array(redi2Pool, trim(modifiedGroupName), redi_term2)
+
+               modifiedGroupName = trim(groupItr % memberName) // "_redi_term3"
+               call mpas_pool_get_array(redi3Pool, trim(modifiedGroupName), redi_term3)
+
```
vanroekel: these output the individual redi terms (except vertical mixing). we can revert these when we want to merge.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -216,12 +218,13 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThicknessEdge,
 
                ! For top layer, only use triads pointing down:
                flux_term2 = coef * RediKappa(iCell) * layerThicknessEdge(k, iEdge) * &
-                  0.5_RKIND * &
+                  0.25_RKIND * &
```
vanroekel: I think we need to treat the top and bottom as an average of 4 things even though there are just two.  I think of it as an average of 4 triads still but 2 are zero.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -317,7 +323,8 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThicknessEdge,
                flux_term3 = RediKappa(iCell) * 2.0_RKIND * &
                    (fluxRediZTop(iTr,k)/areaSum(k) - fluxRediZTop(iTr,k+1)/areaSum(k+1))
 
-               tend(iTr,k,iCell) = tend(iTr,k,iCell) + flux_term3
+               redi_term3(iTr,k,iCell) = redi_term3(iTr,k,iCell) - flux_term3
+               tend(iTr,k,iCell) = tend(iTr,k,iCell) - flux_term3
```
vanroekel: This seems to be the most critical change.  It was not clear to me why flux_term3 had a positive on it and was the dominant term of 1-3 in my testing that seemed to cause the large negative temperatures.  This seems to fit best with the equation for the Redi term you had on confluence.

mark-petersen: Term 3 is `d/dz ( h S dot grad phi) = ( S dot grad phi)` where `grad phi` is computed on the edge and middle vertically, and then multiplied by the slope triad that goes through that edge. I don't expect a negative sign here. But just in case, I made this change in the current head of `ocean/develop` and reran this idealized SO test case.
output
Compare to identical repo with positive sign on term 3:
https://github.com/MPAS-Dev/MPAS-Model/pull/280#issuecomment-576533942
image
Also, the analytic polynomial test cases match with a positive sign but not a negative sign.

vanroekel: Thanks @mark-petersen, I agree the sign is correct as positive, I still don't fully understand why though.

vanroekel:
@mark-petersen I've been thinking more about redi term3 and had a question about the fluxRediZtop code (here https://github.com/MPAS-Dev/MPAS-Model/blob/600fa0a6733ebbdcc4e55023c3f81ddab2ce46f2/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F#L303-L313).  When I look at this it seems there are only two triads being considered for each edge.  The previous code (redi term2) has an explicit loop over iCellSelf (including 1 and 2) but here only seems to have one value per edge.  Shouldn't this be an average of the 4?  I'm sure I'm missing something in the implementation.

vanroekel:
@mark-petersen I think this PR is ready now.  Here is the latest plot of temperature min from @jonbob's E3SM run.  There is something weird at year 20, but overall this looks good.
Screen Shot 2020-02-18 at 10 11 23 PM

mark-petersen:
That plot is with the single-cell pits, right? When the single-cell pits were removed, did the min remain at -1.8, even if it was just MPAS-Ocean stand alone?

milenaveneziani:
Aren't you concerned about all the values with T<-2?

vanroekel:
The min temperature remained above -2 for most of the run.  Something weird is indeed happening at year 20.  It does recover to be above -2 again quickly.  I will take a look at that time period.  Other than those few years things look good.

This is the case with no single cell pits.  So I'm not sure why we are getting that period.  Jon saved extra output so I should be able to see what is happening.

jonbob:
@vanroekel - let me know if I should rerun that time. And I'll check the output to make sure nothing suspicious has happened....

jonbob:
I looked more closely at the output and the log files -- there's an almost linear decrease in the minimum temperature starting very close to the beginning of year 20. Nothing in the logs jumps out to indicate something going off the rails

vanroekel:
@mark-petersen I've looked through the 30 years of simulation from @jonbob.  The negatives are most often right near the bottom, but there are a few further up in the column (layer 14 when max level cell = 21).  So it looks like bottom tapering could help, but may not completely fix the problem.

mark-petersen:
OK. It sounds like bottom tapering is still worth trying.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -481,22 +452,46 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
+               rValue = zEdge / verticalDisplacement
+
+               sfcTaperDown = 0.5_RKIND*(1.0_RKIND + sin(3.141592_RKIND*(rValue - 0.5_RKIND)))
+               sfcTaperDown = max(0.0_RKIND, min(sfcTaperDown, 1.0_RKIND))
+
+               ! LPV move vertical lines up to not limit the vertical mixing term
+               ! Griffies 1998 eqn 34
+               k33(k  ,iCell) = k33(k  ,iCell) + slopeTaperUp * sfcTaperUp * &
+                              areaEdge*dzTop(k)*slopeTriadUp(k,iCellSelf,iEdge)**2
+               k33(k+1,iCell) = k33(k+1,iCell) + slopeTaperDown * sfcTaperDown * &
```
mark-petersen: @vanroekel, these k33 lines were moved up in the last commit, but they still have the taper coefficients, same as slopeTriadUp and slopeTraidDown. The plan was to *not* taper k33, right? So shouldn't the slope taper variables not appear in the above lines?

vanroekel:
update, after the new commit, here are the results of min temperature after 1 year.  Blue is before the latest commit and orange is the latest
Screen Shot 2020-02-25 at 1 57 49 PM

milenaveneziani:
Looks great!

vanroekel:
Okay, I've gone through and made a number of new changes, including tapering based on N2 (like Danabasoglu and Marshall 1997), I've also added a convective trigger.  I've tested briefly in stand alone, things are looking good so far!

Screen Shot 2020-02-28 at 11 52 44 AM

vanroekel:
I should mention the orange line is the commit with the r value fix, and blue is the latest commit.

milenaveneziani:
Great. I can run a G-case whit E3SM when we are ready.

vanroekel:
Thanks @milenaveneziani for offering to do the run.  I had asked @jonbob to pull the code together for the E3SM run and he already submitted the test.  4 months in, things look good!

Also, through 6 months of stand alone results look good as well.

vanroekel:
@mark-petersen this branch now has options to disable all tapering.  To remove tapering on everything do

```
config_Redi_diabatic_limit = 1e6
config_Redi_use_slope_taper = .false.
config_Redi_use_surface_taper = .false.
config_Redi_N2_limit_term1 = .false.
config_GM_closure = 'constant'
```

Jon is testing the new limiting of term1 in a g-case now.

vanroekel:
@mark-petersen I just pushed my latest code with some changes to tapering and disabling terms 2 and 3 in the bottom cell.  Could you test your convergence on this commit and let me know how it goes.

mark-petersen:
@vanroekel I find these descriptions confusing:
```
 398     <nml_option name="config_Redi_half_slope" type="real" default_value="0.004" units="non-dimensional"
 399         description="value of slope where set to half max"
...
 402     <nml_option name="config_Redi_half_width" type="real" default_value="0.001" units="non-dimensional"
 403         description="value of half width of tanh function limiter"
```
Could you replace the description with something like
```
nml_option name="config_Redi_half_slope"
"The isopycnal slope used in the Redi calculation is limited by the 
function $equation here$ (see Authors et al eqn. XX). 
config_Redi_half_slope is $\variable$ and is the slope value where the limiting is 1/2 
(i.e. the center of the tanh function)"
```
and similar for config_Redi_half_width. Thanks!

vanroekel:
@mark-petersen, I agree. I have to look up the definition every time when I look at this.  I'll make the changes when I push some other changes I'm working on.

mark-petersen:
Here are the automated test results with today's repo head. Convergence tests with exact solutions 
convergence
I think it's right, but I need to look into it a bit further.  Here are the qualitative tests:
output
That all looks correct, and about the same as in January. I am satisfied that the operators are working correctly when no tapering is applied.

vanroekel:
@mark-petersen I've pushed up what I hope is the final code for this PR.  I'll run mpas-analysis in the morning and if that looks reasonable and the code passes your convergence tests I think we might be able to be done.

vanroekel:
@mark-petersen could you also check global conservation of T/S with the latest commit in stand alone?

vanroekel:
@mark-petersen I've pushed my latest changes.  But my rediLimitCounter always is zero and I'm not sure why.  In addition to your testing, could you look at my code related to rediLimitCounter and let me know if you see any issues?

mark-petersen:
@vanroekel Here is the rebased version, on today's head. Let's work from here.

vanroekel:
@mark-petersen regarding the error, the test case now requires this change to namelist
```
config_AM_mixedLayerDepths_enable = .true.
```

I've made a number of other changes and will rebase onto what you have here and then push back.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -196,323 +198,265 @@ subroutine ocn_gm_compute_Bolus_velocity(diagnosticsPool, meshPool, scratchPool)
...
+               ! cell, even though they are never used. This avoids an if
+               ! statement or separate computation for top and bottom.
+               slopeTriadUp(k,iCellSelf,iEdge) = &
+                  - drhoDx / &
+                  (  drhoDT * dTdzTop(k) &
+                   + drhoDS * dSdzTop(k) )
+               slopeTriadDown(k,iCellSelf,iEdge) = &
+                  - drhoDx / &
+                  (  drhoDT * dTdzTop(k+1) &
+                   + drhoDS * dSdzTop(k+1) )
```
mark-petersen, line 358: The QU240 gets a `floating invalid` on this line with a standard run (performance_test) after I turned on the MLD AM. I'm looking into it. @vanroekel can you run a standard test with debug on and see if it trips it? Here are my flags, in case there is some combination that might cause this:

<details>
<summary> flags </summary> <p>

```
in /lustre/scratch4/turquoise/mpeterse/runs/nightly/200402_rebase_rediFixesAndNewTapering_Luke_PR/ocean/global_ocean/QU240/performance_test/forward
 70 &Redi_isopycnal_mixing
 71     config_use_Redi = .true.
 72     config_Redi_kappa = 600.0
 73     config_Redi_maximum_slope = 0.3
 74     config_Redi_closure = 'constant'
 75     config_Redi_diabatic_limit = 5e-7
 76     config_Redi_use_slope_taper = .true.
 77     config_Redi_use_surface_taper = .true.
 78     config_Redi_N2_limit_term1 = .true.
 79 /
 80 &GM_eddy_parameterization
 81     config_use_GM = .true.
 82     config_GM_kappa = 600.0
 83     config_GM_closure = 'N2_dependent'
 84     config_GM_min_stratification_ratio = 0.1
 85     config_GM_constant_gravWaveSpeed = 0.3
 86 /
 87 &mesoscale_eddy_parameterizations
 88     config_eddying_resolution_taper = 'ramp'
 89     config_eddying_resolution_ramp_min = 20e3
 90     config_eddying_resolution_ramp_max = 30e3
```
</details>

vanroekel:
@mark-petersen I just ran this in debug and saw the same issues you did in QU240.  I think I've got it all fixed now.  I was able to run the QU240 "test" case and it ran successfully with the latest commit in debug.  Let me know if it doesn't work.

mark-petersen:
Rebased on today's head. Previous problem with `floating invalid` is solved. On nightly regression suite, these fail:
```
   - 'Global Ocean 240km - Restart Test': -o ocean -c global_ocean -r QU240 -t restart_test
   - 'Global Ocean 240km - SE Blocks Test': -o ocean -c global_ocean -r QU240 -t se_blocks_test
   - 'Global Ocean 240km - RK4 Blocks Test': -o ocean -c global_ocean -r QU240 -t rk4_blocks_test
```
because of an actual mismatch for all three. All the other tests pass with debug and optimized. I can look into it tomorrow.

vanroekel:
@mark-petersen, what do you mean by "actual mismatch"

I assume the last two imply a halo type issue?  I think the first one requires an addition to the restart stream, I'm using a mixed layer depth based index field.  I'm guessing this code is being called before the mixed layer AM resulting in nonBFB across restarts.  I can try fix this one for sure.

I can also look for a halo issue too if that's what is implied by the last two tests.

mark-petersen:
When a regression test says `FAIL`, it can be because it did not complete the simulation, or for restart and blocks because the files did not match. By "actual mismatch", I just meant the later. Yes, it is a halo update issue.

vanroekel:
Thanks for the explanation.  I see the problem is related to indMLD, but it is not clear to me what the fix is.  I tried adding this to the restart stream and it did not help.  I also tried halo exchanges before the gm routine, but it also didn't work.  I'd appreciate any ideas you have.

vanroekel:
@mark-petersen yes to both.  I can see it there, but it doesn't help.  But if I hard code indMLD(:) = 1 then the restart problem is fixed.

Also, the block issue appears to be in redi_term2, but I'm stumped as to what the problem is

vanroekel:
@mark-petersen I think I understand the SE block failure.  I'll work on a fix shortly.  If you have any thoughts on the Restarts test, I'd appreciate it.

mark-petersen:
I'm trying now. Are you setting mixedLayerDepthsOutput AM to compute every time step? i.e.,
```
<stream name="mixedLayerDepthsOutput"
...
  output_interval="00-00-00_00:00:01" (something short)
in namelist:
config_AM_mixedLayerDepths_compute_interval = 'output_interval'
```
If the compute time is not every time step, that could mess up bfb restarts. I'm testing now by throwing every variable possible into the restart file. 

Also, it's possible that the halo problem was causing the restart problem.

vanroekel:
I just made the change and am testing.  I've been able to get SE blocks to pass, but RK4 blocks still fails.  Any ideas why they might be different?

vanroekel:
Another update, with MLD every time step restart passes.

vanroekel:
@mark-petersen I can get the full nightly to pass now, but it requires some changes to the test case (which MLD values are calculated).  I need to disable the gradient based calculations of MLD in the newest branch.  However, given these are never used as far as I can tell, I would like to simply remove those options.  What do you think?  pinging @maltrud and @milenaveneziani as well.

vanroekel:
Okay, I've just push changes that pass all tests for me.  NOTE that to get passing I need to compute MLD every time step. and set the following flags
```
config_AM_mixedLayerDepths_Tgradient = .false.
config_AM_mixedLayerDepths_Dgradient = .false.
```

vanroekel:
I've re pushed all the changes without the cvmix deletion.  it should work now.  I've also fixed the MLD issue in the prior commit.  You can now leave Tgradient and Dgradient on.  I would still flip to off as they are rarely used.

milenaveneziani:
I am not worried at all about the disabling of the MLD gradient calculation.

vanroekel:
@mark-petersen the latest commit passes the entire nightly regression for me, all that needs to change is outputting MLD every time step.

mark-petersen:
Added bfb block test and threading test for Redi:
```
  43: -o ocean -c Redi_verification -r SouthernOceanSlice40 -t se_blocks_test
  44: -o ocean -c Redi_verification -r SouthernOceanSlice40 -t thread_test
```
It currently passes the block test but fails the thread test. Looking into it now...

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -196,323 +195,204 @@ subroutine ocn_gm_compute_Bolus_velocity(diagnosticsPool, meshPool, scratchPool)
...
-      ! slope can be unbounded in regions of neutral stability, reset to the large, but bounded, value
-      ! values is hardwrite to 1.0, this is equivalent to a slope of 45 degrees
-      !$omp do schedule(runtime) private(k)
-      do iEdge = 1, nEdges
-         do k = 1, nVertLevels
-            relativeSlopeTopOfEdge(k, iEdge) = max( min( relativeSlopeTopOfEdge(k, iEdge), 1.0_RKIND), -1.0_RKIND)
+      nCells = nCellsArray( 3 )
+      !$omp do schedule(runtime) private(invAreaCell, iEdge, cell1, cell2, k, iCellSelf) &
+      !$omp& private(dcEdgeInv, areaEdge, drhoDT, drhoDS, dTdx, dSdx, drhoDx) &
+      !$omp& private(sfcTaper, sfcTaperUp, sfcTaperDown, sfcTaperFactor)
```
mark-petersen, line 275: @philipwjones, @mattdturner or @amametjanov: I am failing a test for bfb match between 1 and 2 threads. Could you please check the above three lines? I added all variables that need to be thread-private. I have four allocated column arrays: 
```dzTop, dTdzTop, dSdzTop, k33Norm```
Do I put those in `omp private`? I tried, but got a compile error.

If you could check for any omp directive errors in this file, that would be helpful.

File: src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F [hunk 8]
```diff
@@ -165,285 +166,345 @@ subroutine ocn_tracer_hmix_redi_tend(meshPool, scratchPool, layerThicknessEdge,
...
+      redi_term3_topOfCell(:,:,:) = 0.0_RKIND
+
+      allocate(fluxRediZTop(nTracers,nVertLevels+1))
+      redi_term3(:,:,:) = 0.0_RKIND
+      redi_term1(:,:,:) = 0.0_RKIND
+      ! Term 1: this is the "standard" horizontal del2 term, but with RediKappa coefficient.
+      ! \kappa_2 \nabla \phi on edge
+      ! mrp performance: merge term 1 into main cell loop later
+      !$omp do schedule(runtime) private(invAreaCell, i, iEdge, cell1, cell2, iCellSelf) &
+      !$omp& private(r_tmp, coef, kappaRediEdgeVal, tracer_turb_flux, flux, k, iTr, flux_term2)
```
mark-petersen, line 192: Same thing here and in this file, if you could check my choice of private variables.

src/core_ocean/shared/mpas_ocn_gm.F
amametjanov: It looks like `iCellSelf` is being initialized inside the loop but is not declared `private`.

amametjanov: Please try with this update that has additional private variables:
```
azamatm@thetalogin2:~/repos/E3SM-clone/components/mpas-source> git diff
diff --git a/src/core_ocean/shared/mpas_ocn_gm.F b/src/core_ocean/shared/mpas_ocn_gm.F
index 2490cf07..2d152190 100644
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -270,9 +270,9 @@ contains
       endif
 
       nCells = nCellsArray( 3 )
-      !$omp do schedule(runtime) private(invAreaCell, iEdge, cell1, cell2, k, iCellSelf) &
-      !$omp& private(dcEdgeInv, areaEdge, drhoDT, drhoDS, dTdx, dSdx, drhoDx) &
-      !$omp& private(sfcTaper, sfcTaperUp, sfcTaperDown, sfcTaperFactor)
+      !$omp do schedule(runtime) private(invAreaCell, iEdge, cell1, cell2, k, iCellSelf, &
+      !$omp& dcEdgeInv, areaEdge, drhoDT, drhoDS, dTdx, dSdx, drhoDx, &
+      !$omp& sfcTaper, sfcTaperUp, sfcTaperDown, sfcTaperFactor, i, slopeTaperDown, slopeTaperUp)
       do iCell = 1, nCells
          invAreaCell = 1.0_RKIND/areaCell(iCell)
          k33(1:maxLevelCell(iCell)+1,iCell) = 0.0_RKIND
@@ -376,8 +376,8 @@ contains
                slopeTriadDown(k,iCellSelf,iEdge) = &
                   slopeTaperDown * sfcTaperDown * slopeTriadDown(k,iCellSelf,iEdge)
 
-            end do
-        end do
+            end do ! maxLevelEdgeTop(iEdge)
+        end do ! nEdgesOnCell(iCell)
 
          ! Normalize k33
         do k = 2, maxLevelCell(iCell)
```

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
amametjanov: An update for this file:
```
azamatm@thetalogin2:~/repos/E3SM-clone/components/mpas-source> git diff src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
diff --git a/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F b/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
index 6d8141ff..06ec35bc 100644
--- a/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
@@ -188,8 +188,8 @@ contains
       ! Term 1: this is the "standard" horizontal del2 term, but with RediKappa coefficient.
       ! \kappa_2 \nabla \phi on edge
       ! mrp performance: merge term 1 into main cell loop later
-      !$omp do schedule(runtime) private(invAreaCell, i, iEdge, cell1, cell2, iCellSelf) &
-      !$omp& private(r_tmp, coef, kappaRediEdgeVal, tracer_turb_flux, flux, k, iTr, flux_term2) 
+      !$omp do schedule(runtime) private(invAreaCell, i, iEdge, cell1, cell2, iCellSelf, &
+      !$omp& r_tmp, coef, kappaRediEdgeVal, tracer_turb_flux, flux, k, iTr, flux_term2, dTracerDx, flux_term3)
       do iCell = 1, nCells
          invAreaCell = 1.0_RKIND / areaCell(iCell)
          fluxRediZTop(:,:) = 0.0_RKIND
@@ -369,7 +369,7 @@ contains
      !$omp end do
 
      !now go back and reapply all tendencies
-     !$omp do schedule(runtime) private(invAreaCell, k, iTr, i)
+     !$omp do schedule(runtime) private(invAreaCell, k, iTr, i, iEdge)
      do iCell=1,nCells
        invAreaCell = 1.0_RKIND / areaCell(iCell)
         do k = 1, maxLevelCell(iCell)
```

mark-petersen:
Thanks, @amametjanov that was helpful. It still does not have a thread match, so I need a better debugging method.  Do you agree that I can use `!$omp single`  and `!$omp end single` around parts of the code, to isolate the problem? I am trying this:
```
diff --git a/src/core_ocean/shared/mpas_ocn_gm.F b/src/core_ocean/shared/mpas_ocn_gm.F
index 3f8a39b01..efe7a5a86 100644
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -195,46 +195,47 @@ contains
       nCells = nCellsArray( size(nCellsArray) )
       nEdges = nEdgesArray( size(nEdgesArray) )

-      !$omp do schedule(runtime) private(k)
+      !  !$omp do schedule(runtime) private(k)
+      !$omp single
       do iEdge = 1, nEdges
          do k = 1, nVertLevels
             gradDensityEdge(k, iEdge) = 0.0_RKIND
             normalGMBolusVelocity(k, iEdge) = 0.0_RKIND
          end do
       end do
-      !$omp end do
+      !  !$omp end do
+      !$omp end single
```
and commenting out the previous omp statements. Does that force that look to use a single thread?

mattdturner:
@mark-petersen does just commenting out the `!$omp` directives not solve your problem?

mark-petersen:
No, but my 1 versus 2 thread mismatch may be caused in some other part of the code. If I just comment out the `!$omp` directives, is that an appropriate way to debug? I did try that first, thinking the same thing. But I was concerned that there would be multiple threads running the full loop then, and they might step on the same memory, so I thought `!$omp single` is safer.

mattdturner:
I would think just commenting out the OpenMP directives would be an appropriate way to debug.  I'm not sure what loops you need it for (aside from the trivial loop you included in the comment), but if each thread has a private copy of the array and you use `!$omp single` then you will run into an issue with some threads having incorrect data at the end of the loop because their private copy wasn't updated.

Perhaps @philipwjones or @amametjanov might want to weight in though, since this threading approach of having the entire Ocean model run in the `!$omp parallel` region (instead of parallel regions around specific parts of the code) is still somewhat new to me.  I'm also not the most familiar with MPAS-O yet.

amametjanov:
```
!$omp master
...
!$omp end master
!$omp barier !if needbe
```
is preferred over `!$omp single` here because `!$omp single` does not guarantee that master thread will execute the region.

philipwjones:
Just to respond to a comment above, commenting out the omp directives within a parallel region does indeed result in all threads executing the full loop, so unless you do a manual form of static scheduling by making the loop limit depend on thread count, an undecorated loop is probably not a good idea.  So using master is probably better, but note also that using master is fine for shared arrays, but if the loop is updating any threadprivate variables that are used later, only master will have the right data.  Looks like you're getting help from Az, but let me know if you need additional eyes.

mark-petersen:
Thanks @amametjanov, @philipwjones and @mattdturner for the comments. I've changed the sections in question to use `!$omp master` for debugging.

mark-petersen:
@vanroekel, I set up a new test that compares one and two thread runs. Note that compile requires `OPENMP=true`
```
  81: -o ocean -c global_ocean -r QU240 -t thread_test
```
It fails on the current head, 2a15092, with everything in these files:
```
 src/core_ocean/shared/mpas_ocn_gm.F
 src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```
set to only run on the master thread. It passes with this set:
```
 70 &Redi_isopycnal_mixing
 71     config_use_Redi = .false.
 72     config_Redi_kappa = 600.0
 73     config_Redi_maximum_slope = 0.3
 74     config_Redi_closure = 'constant'
 75     config_Redi_use_slope_taper = .false.
 76     config_Redi_use_surface_taper = .false.
 77     config_Redi_N2_limit_term1 = .false.
 78 /
 79 &GM_eddy_parameterization
 80     config_use_GM = .true.
 81     config_GM_kappa = 600.0
 82     config_GM_closure = 'N2_dependent'
 83     config_GM_min_stratification_ratio = 0.1
 84     config_GM_constant_gravWaveSpeed = 0.3
 85 /
```
but fails by changing the one flag:
```
 71     config_use_Redi = .true.
```

mark-petersen:
OK, I think I found it. But two specific questions for @philipwjones, @mattdturner or @amametjanov:

1. There are some array initializations like this:
```
redi_term2(:,:,:) = 0.0_RKIND
```
I would guess that we should always loop over the outer index explicitly for good practice, but especially for threading:
```
!$omp do schedule(runtime)
do iCell = 1, nCells
    redi_term2(:,:,iCell) = 0.0_RKIND
end do
!$omp end do
```
Agreed? That was my bug - I think there was thread contention on those arrays.

2. We have two types of temporary arrays: some with an iCell index, and some without. Where should these allocate/deallocate statements be relative to `$omp` statements? My best guess is:
```
allocate(columnTemp(nVertLevels))

!$omp master
allocate(blockTemp(nVertLevels,nCells))
!$omp end master
!$omp barrier

do iCell = 1, nCells
   do k = 1, nVertLevels
      ! calculations with columnTemp and blockTemp
   end do
end do
!$omp end do

deallocate(columnTemp)

!$omp master
deallocate(blockTemp)
!$omp end master
!$omp barrier
```
That way, every thread has its own copy of `columnTemp` but there is one shared version of `blockTemp`. If we did not put `master` lines around `blockTemp`, it would not cause an error, but it would allocate excessive memory. Does that all sound correct?

philipwjones:
Yes on 1. Almost on 2. The column arrays are fine and should be allocated by all threads. However, your blockTemp array will only be allocated on master so other threads won't have a version since all arrays are thread private by default in MPAS-O. But we also don't want every thread to allocate a full size beast. One way to force it to be shared is to make it a module variable rather than inside the subroutine, then it will be shared and your master allocate will do as you want.

philipwjones:
BTW, this is an argument for merging Matt's localized threading PR soon. So we can finally get out of this always threaded paradigm and the surprises it always causes.

mark-petersen:
@philipwjones Agreed. That is the next one after Redi. For today on this one, should I do this?
```
module ocn_tracer_hmix_Redi
...
! Private module variables
real (kind=RKIND), dimension(:,:), allocatable :: blockTemp

subroutine foo(...)
real (kind=RKIND), dimension(:), allocatable :: columnTemp

allocate(columnTemp(nVertLevels))
allocate(blockTemp(nVertLevels,nCells))

do iCell = 1, nCells
   do k = 1, nVertLevels
      ! calculations with columnTemp and blockTemp
   end do
end do
!$omp end do

deallocate(columnTemp)
deallocate(blockTemp)
```
I didn't realize that a module-level allocatable variable is shared.

philipwjones:
You still want the blockTemp one inside a master directive so it's the only one allocating the shared array.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -165,285 +166,351 @@ subroutine ocn_tracer_hmix_redi_tend(meshPool, scratchPool, layerThicknessEdge,
...
+              if(tempTracer < minimumVal(iTr)) then
+
+                  do i = 1, nEdgesOnCell(iCell)
+                     iEdge = edgesOnCell(i, iCell)
+                     redi_term2_edge(iTr,k,iEdge) = 0.0_RKIND
+                  enddo
+
+                  redi_term3_topOfCell(iTr,k,iCell) = 0.0_RKIND
+                  redi_term3_topOfCell(iTr,k+1,iCell) = 0.0_RKIND
+                  rediLimiterCount(k,iCell) = 1.0_RKIND
```
mark-petersen, line 365: @vanroekel this line sets the counter to 1 for any tracer group, but you set rediLimiterCount to 0 above only for active tracers. You just want it for active tracers, right? We can also increment in case T&S are both reset:
`rediLimiterCount(k,iCell) = rediLimiterCount(k,iCell) + 1`

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -165,285 +166,351 @@ subroutine ocn_tracer_hmix_redi_tend(meshPool, scratchPool, layerThicknessEdge,
...
-      !$omp do schedule(runtime)
+      allocate(minimumVal(nTracers))
+      allocate(redi_term2_edge(nTracers,nVertLevels,nEdges))
+      allocate(redi_term3_topOfCell(nTracers,nVertLevels+1,nCells))
+
+      nCells = nCellsArray(1)
+      nCellsP1 = nCellsArray( size(nCellsArray) ) + 1
+      redi_term2(:,:,:) = 0.0_RKIND
+      redi_term2_edge(:,:,:) = 0.0_RKIND
+      redi_term3_topOfCell(:,:,:) = 0.0_RKIND
```
mark-petersen, line 183: @vanroekel I think `redi_term2_edge` and `redi_term3_topOfCell` do not need to be zeroed out, because they are not incremented upon. Do you agree? I will keep these zero lines for `redi_term1`, `redi_term2`, `redi_term3` because those are incremented, but make them thread safe.

vanroekel: @mark-petersen -- yes I agree.  Thanks for catching that

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
vanroekel: oh, yes, definitely just active tracers, I am not sure if incrementing would work well.  There could be times when salinity is reset and not temperature and vice versa and so rediLimiterCount=1 wouldn't tell us which, but = 2 would say both.  I would say we only do it for temperature, or I could go back and make that part of tracer groups.

mark-petersen:
Finally passes all tests with multiple threads and debug on, and all Redi and GM flags on.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -793,86 +676,109 @@ subroutine ocn_gm_init(domain, err)!{{{
...
          call mpas_pool_get_array(meshPool, 'cellsOnEdge', cellsOnEdge)
          call mpas_pool_get_array(diagnosticsPool, 'gmBolusKappa', gmBolusKappa)
          call mpas_pool_get_array(meshPool, 'areaCell', areaCell)
+         call mpas_pool_get_array(diagnosticsPool, 'rediRossbyRadius', rediRossbyRadius)
+         call mpas_pool_get_array(meshPool, 'fEdge', fEdge)
+
+         !Impose tapering of Large et al. 1997, wrap in if statement later
+         !$omp do schedule(runtime)
+         do iEdge=1,nEdges
+           rediRossbyRadius(iEdge) = max(15000.0_RKIND,min(100000.0_RKIND,2.0_RKIND / abs(fEdge(iEdge))))
```
mark-petersen, line 706: @vanroekel it looks like `rediRossbyRadius` is computed but never used. May I remove it?

vanroekel: @mark-petersen correct, I had meant to remove that, please do so if it is easy.  I'm happy to do it as well.

mark-petersen:
Commit f751f53 produces a bfb match with @vanroekel's last commit  2f74e37 on nightly regression suite when both are compiled with `OPENMP=false`, so I can say that the changes I've made since last Thursday are only for threading. In addition, I show a bfb match on the current head f751f53 between threaded (`OPENMP=true`) and not (`OPENMP=false`).

mark-petersen:
@vanroekel The head is still bfb with your last commit 2f74e37 on the MPAS-Ocean nightly regression suite. Please check in your standard E3SM set-up. Should be bfb there as well, as changes were all omp and formatting, and removing unused variables.

This is almost ready to merge. I'm rechecking convergence and idealized Redi test now.

vanroekel:
@mark-petersen I'm building a case with your latest changes now.  I should have something for you by tomorrow morning at the latest.

vanroekel:
@mark-petersen I can confirm that the latest commit is BFB with my previous e3sm tests.

mark-petersen:
Automated tests all look good. 
convergence
output

vanroekel:
Great!  Thanks for all the help getting this across the finish line.  I agree this is ready to merge.

src/core_ocean/Registry.xml
```diff
@@ -1647,7 +1626,8 @@
 				mode="forward;analysis">
 
 			<stream name="mesh"/>
-			<var_struct name="tracers"/>
+      <var_struct name="tracers"/>
+      <var name="indMLD"/>
```
mark-petersen, line 1630: @vanroekel I'm taking a last careful look through. We added `indMLD` to the restart file here when we were struggling with the restart bug, but I don't think it is needed. The restart test passes without it, with Redi and GM on. Do you agree it can be removed?

src/core_ocean/Registry.xml
```diff
@@ -1657,6 +1637,7 @@
 			<var name="xtime"/>
 			<var name="simulationStartTime"/>
 			<var name="boundaryLayerDepth"/>
+			<var name="indexBoundaryLayerDepth"/>
```
mark-petersen, line 1640: Same thing here. My restart test passes using QU240 and cvmix on, using kpp, with both `boundaryLayerDepth` and `indexBoundaryLayerDepth` removed. Are there conditions where those variables are used from the previous timestep, rather than recomputed?

vanroekel: agreed.  I think this was from your first tapering implementation when it was based on BLD.  I'm sure it can be removed.

src/core_ocean/Registry.xml
vanroekel: I think so.  I believe rediGM is now called before the timestep but after the analysis compute, but this might require that MLD be computed on start up.  Is that flag set in your testing?  If we don't compute on start up we may want to keep indMLD in the restart file.

mark-petersen: I guess it depends on the timing of the `mixedLayerDepths` AM. We can only safely remove the variable if the AM is computed just before the restart is written, to get exact restarts. For example, if we use either of these settings:
```
config_AM_mixedLayerDepths_compute_interval = 'dt'
config_AM_mixedLayerDepths_compute_interval = '01_00:00:00'
```
for a global run, we can remove `indMLD` from the restart file. The safest option is to leave it in, at the cost of a slightly larger restart file.

src/core_ocean/Registry.xml
mark-petersen: What about `boundaryLayerDepth` in the restart file? That was added a long time ago. Probably before the cvmix:
```
cb8195861b (Douglas Jacobsen   2015-02-02 13:45:52 -0700 1659)          <var name="boundaryLayerDepth"/>
```

vanroekel: I can't think of any reason to include boundaryLayerDepth in the restart file.  It is always computed before mixing and is only used within the mixing routine.  So I don't think that should affect restarts.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -1823,9 +1884,10 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
         call mpas_pool_get_subpool(block % structs, 'diagnostics', diagnosticsPool)
         call mpas_pool_get_subpool(block % structs, 'scratch', scratchPool)
         ! Compute normalGMBolusVelocity; it will be added to the baroclinic modes in Stage 2 above.
-        if (config_use_standardGM) then
-           call ocn_gm_compute_Bolus_velocity(diagnosticsPool, meshPool, scratchPool)
-        end if
+ !       if (config_use_GM.or.config_use_Redi) then
+ !          call ocn_gm_compute_Bolus_velocity(statePool, diagnosticsPool, &
```
mark-petersen, line 1888: @vanroekel I saw that this line is commented out in `ocean/develop`, and looks like we were in a rush and left some confusing code. This was commented out in this PR in:
   66a13d973 fixes for restart test
We changed E3SM to call this routine from the driver:
 ```
components/mpas-ocean/driver/ocn_comp_mct.F:968:
              call ocn_gm_compute_Bolus_velocity(statePool, diagnosticsPool, meshPool, scratchPool, timeLevelIn=1)
```
in https://github.com/E3SM-Project/E3SM/commit/b29cfe5b370 (Luke Van Roekel   2020-05-05 16:01:43 -0500  968)
but it appears that it is not called from the driver in stand-alone. I can make an issue to clean this up, but wanted to ping you on these lines.

File: src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F [hunk 15]
```diff
@@ -1920,10 +1982,11 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
          ! Update the effective desnity in land ice if we're coupling to land ice
          call ocn_effective_density_in_land_ice_update(meshPool, forcingPool, statePool, scratchPool, err)
 
-         ! Compute normalGMBolusVelocity; it will be added to normalVelocity in Stage 2 of the next cycle.
-         if (config_use_standardGM) then
-            call ocn_gm_compute_Bolus_velocity(diagnosticsPool, meshPool, scratchPool)
-         end if
+!         ! Compute normalGMBolusVelocity; it will be added to normalVelocity in Stage 2 of the next cycle.
+!         if (config_use_GM.or.config_use_Redi) then
+!            call ocn_gm_compute_Bolus_velocity(statePool, diagnosticsPool, &
```
mark-petersen, line 1987: This line was also commented out, and needs to be removed or cleaned up.

src/core_ocean/mode_forward/mpas_ocn_forward_mode.F
```diff
@@ -597,9 +595,13 @@ function ocn_forward_mode_run(domain) result(ierr)!{{{
            call mpas_timer_stop("land_ice_build_arrays")
 
            call ocn_frazil_forcing_build_arrays(domain, meshPool, forcingPool, diagnosticsPool, statePool, err)
-           
            call ocn_tidal_forcing_build_array(domain, meshPool, forcingPool, diagnosticsPool, statePool, err)
 
+           ! Compute normalGMBolusVelocity, relativeSlope and RediDiffVertCoef if respective flags are turned on
+           if (config_use_Redi.or.config_use_GM) then
+               call ocn_gm_compute_Bolus_velocity(statePool, diagnosticsPool, meshPool, scratchPool, timeLevelIn=1)
```
mark-petersen, line 602: Looks like the stand-alone call is here, so maybe this is just a matter of removing the commented out lines for clarity.

src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
```diff
@@ -726,12 +727,13 @@ subroutine ocn_time_integrator_rk4(domain, dt)!{{{
          !$omp end do
 
          ! Compute normalGMBolusVelocity and the tracer transport velocity
-         if (config_use_standardGM) then
-             call ocn_gm_compute_Bolus_velocity(diagnosticsPool, meshPool, scratchPool)
+         if (config_use_GM) then
+             call ocn_gm_compute_Bolus_velocity(statePool, diagnosticsPool, &
```
mark-petersen, line 731: In rk4, the corresponding lines here should also be removed, since it is called from the driver.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
vanroekel: yes @mark-petersen I removed those calls and moved them to the driver, I just never removed them.  I can clean this up if you want.  It should be in stand alone as you note below

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
vanroekel: yes those can be removed

PR #447, opened 2020-02-09
compute landIceMask using geometric_features

xylar:
This merge changes `landIceMask` in the `global_ocean` test cases with ice-shelf cavities (`wISC`) to being computed from the `AntarcticIceCoverange` feature from Bedmap2 (and soon BedMachine, see #433).  Previously, `landIceFraction` was interpolated to the MPAS mesh and then used to compute `landIceMask`, but this meant that the land-ice mask wasn't necessarily consistent with the culled mesh used to produce mapping files for E3SM.

This merge also computes the culled mesh without ice-shelf cavities within the `culled_mesh` step, rather than as part of a second, largely redundant step.  The (currently broken) `e3sm_coupling` step is also updated to point to this new location for the culled mesh without ice-shelf cavities.

In the various test cases, including `global_ocean` and also in the `diagnostics` module, `landIceMask` is no longer computed based on `landIceFraction`.  Instead, `landIceFraction` is set to zero where `landIceMask == 0`.  Eventually, when both these fields become dynamic in time, the coupling (or forcing) code will need to update both these fields at each coupling (or forcing) interval.  Since `landIceMask` cannot currently change in time, this is not a present concern.

Bug fix: A missing link to the driver script is added to the `SO60to10wISC` test case.


xylar:
## Testing

I successfully ran the following test cases on LANL IC:
* `global_ocean/EC60to30wISC/init`
* `global_ocean/QU240wISC/init`
* `isomip/10km/expt1.01`
* `isomip_plus/2km/Ocean0`
* `sub_ice_shelf_2D/5km/Haney_number_iterative_init`

I still plan to run:
* `global_ocean/EC60to30wISC/spin_up`
* `global_ocean/SO60to10wISC/init`
and will update once those are done.

xylar:
I don't think it's worth taking too careful a look at the changes in `landIceMask` because of these changes, because I will be changing both coastlines and bathymetry in #433.

xylar:
@jonbob, as a review of this, all I'm asking from you is a verification that this produces more useful results for you than our previous attempt.  I'll send an email tomorrow or Monday with more details on where things stand and what I presume will need to be redone.

PR #448, opened 2020-02-10
Alter the e3sm_coupling README to add warnings

xylar:
This merge updates the `readme` file added to the `e3sm_coupling` step in COMPASS to warn users that it is still in development and that files it produces should not be uploaded to the LCRC server.


xylar:
@mark-petersen, there are no code changes here.  Could you include this in your next set of changes to `ocean/develop`?

PR #450, opened 2020-02-13
Remove single-cell pits when initializing global bathymetry

mark-petersen:
Before this, single-cell pits were allowed to be one layer deeper than all surrounding cells. These cells can collect tracer extrema and are not able to use horizontal diffusion or advection to clear them out. This PR reduces pits to make them level with the next deepest neighbor cell.

### DIFF
--- a/src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
+++ b/src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
@@ -214,7 +214,7 @@ contains
 
       call mpas_log_write( 'Initializing vertical coordinate with ssh = 0.')
       ! compute the vertical grid (layerThickness, restingThickness, maxLevelCell, zMid)
-      ! based on bottomDepth and refBottomDepth and apply PBCS if requested
+      ! based on bottomDepth and refBottomDepth and apply PBCs if requested
       call ocn_init_ssh_and_landIcePressure_vertical_grid(domain, iErr)
 
       if(iErr .ne. 0) then
@@ -842,19 +842,19 @@ contains
 
        type (block_type), pointer :: block_ptr
 
-       type (mpas_pool_type), pointer :: meshPool, scratchPool, verticalMeshPool, criticalPassagesPool
+       type (mpas_pool_type), pointer :: meshPool, verticalMeshPool, criticalPassagesPool
 
        real (kind=RKIND), dimension(:), pointer :: latCell, lonCell, bottomDepth, bottomDepthObserved, &
             refBottomDepth, refLayerThickness, refZMid, oceanFracObserved
 
        integer, pointer :: nCells, nCellsSolve, nVertLevels
 
-       type (field1DInteger), pointer :: maxLevelCellField, smoothedLevelsField
+       type (field1DInteger), pointer :: maxLevelCellField
        type (field1DReal), pointer :: bottomDepthObservedField
        integer, dimension(:), pointer :: maxLevelCell, nEdgesOnCell
        integer, dimension(:, :), pointer :: cellsOnCell
 
-       integer :: iCell, coc, j, k, maxLevel
+       integer :: iCell, coc, j, k, maxLevelNeighbors
        integer :: minimum_levels
 
        character (len=StrKIND), pointer :: config_global_ocean_topography_method
@@ -880,7 +880,6 @@ contains
        call mpas_pool_get_config(domain % configs, 'config_global_ocean_topography_has_ocean_frac', &
                                  config_global_ocean_topography_has_ocean_frac)
 
-       call mpas_pool_get_subpool(domain % blocklist % structs, 'scratch', scratchPool)
 
        block_ptr => domain % blocklist
        do while(associated(block_ptr))
@@ -895,13 +894,10 @@ contains
           call mpas_pool_get_array(meshPool, 'oceanFracObserved', oceanFracObserved)
           call mpas_pool_get_array(meshPool, 'refBottomDepth', refBottomDepth)
 
+          ! Alter bathymetry for minimum depth
           do k = 1, nVertLevels
              if (refBottomDepth(k).gt.config_global_ocean_minimum_depth) then
                 minimum_levels = k
-                print '(a,f8.2,2a,i5,a,f8.2,a)', 'config_global_ocean_minimum_depth=', &
-                   config_global_ocean_minimum_depth,' m.  ', 'Setting minimum layer index to ', &
-                   minimum_levels, ' with a bottom depth of ', refBottomDepth(k), ' m.'
-                exit
              end if
           end do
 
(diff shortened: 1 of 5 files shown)


mark-petersen:
Testing now...

xylar:
I was wondering what the logic was in having these values previously.  Could you rename the config option following my work in #436 as well?
https://github.com/MPAS-Dev/MPAS-Model/pull/440/commits/c4993f3785b93624d25767fe8e99d0267191ad4e
This PR seems like a more logical place to rename this config option.  And I must insist that filling holes isn't smoothing in the sense that a user would expect.

src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
```diff
@@ -996,8 +998,8 @@ subroutine ocn_init_setup_global_ocean_create_model_topo(domain, iErr)!{{{
                 end do
 
                 if (maxLevel < maxLevelCell(iCell) ) then
-                   smoothedLevelsField % array(iCell) = maxLevel + 1
-                   bottomDepth(iCell) = refBottomDepth(maxLevel + 1)
+                   smoothedLevelsField % array(iCell) = maxLevel
```
vanroekel: @mark-petersen I tested this code and there are still ~700 locations that need adjustment (pits).  In my offline code I had to iterate twice to remove all issues.  Can you add a loop around this adjustment?

mark-petersen:
@xylar this change is the same place as #440. It makes more sense to close this PR and make this minor change for single-cell pits in #440, then review it all together. Do you agree? If so, I'll close this PR and add a commit to #440.

xylar:
No I think these need to be separate PRs.  I don't want any additional confusion about what we mean by "smoothing" and if we start modifying filling holes in a PR about smoothing, that would do exactly that in my view.

src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
xylar: @mark-petersen, the loop needs to be until there are no more pits to fill (with a maximum of `maxEdges` because I believe that's the max).

mark-petersen:
rebased. testing now.

src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
```diff
@@ -998,37 +994,35 @@ subroutine ocn_init_setup_global_ocean_create_model_topo(domain, iErr)!{{{
...
-          ! Fill holes. Enforce that no cell has maxLevelCell that is more than
-          ! 1 vertical level higher than its neighbors.
-          if (config_global_ocean_fill_bathymetry_holes) then
-             call mpas_pool_get_subpool(block_ptr % structs, 'scratch', scratchPool)
-
-             call mpas_pool_get_field(scratchPool, 'smoothedLevels', smoothedLevelsField)
+          ! Alter cells for partial bottom cells before filling in bathymetry
+          ! holes.
+          do iCell = 1, nCells
+             call ocn_alter_bottomDepth_for_pbcs(bottomDepth(iCell), refBottomDepth, maxLevelCell(iCell), iErr)
```
mark-petersen: @xylar this is now complete. The critical change was the line above. The bathymetry holes would be filled, and then in a later call the cells were altered. Specifically, if the bottom bathymetry is within 5% of the layer thickness to the top of the cell, we round it up and change maxLevelCell to one less. 

My solution here is to call the PBC routine before filling the bathymetry. I left the original call to
`ocn_alter_bottomDepth_for_pbcs`
because it is later in the sequence, and it is within your iterations for SSH for land ice. I thought there was no harm done leaving the later one in place.

mark-petersen:
I tested this with the EC60to30 in COMPASS, and this little script:
https://gist.github.com/mark-petersen/52af011b559786a2d7206e4c225ea992
which showed single-cell holes before, and now shows there are none.

The other changes in this PR are just a little clean-up, and moving init flags to the init_mode Registry.xml file.

mark-petersen:
@xylar and @vanroekel if either of you want to make comments or test, please do.

src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
```diff
@@ -998,37 +994,35 @@ subroutine ocn_init_setup_global_ocean_create_model_topo(domain, iErr)!{{{
...
+          ! Fill bathymetry holes, i.e. single cells that are deeper than all neighbors.
+          ! These cells can collect tracer extrema and are not able to use
+          ! horizontal diffusion or advection to clear them out. Reduce pits to
+          ! make them level with the next deepest neighbor cell.
+          if (config_global_ocean_fill_bathymetry_holes) then
              call mpas_pool_get_array(meshPool, 'nEdgesOnCell', nEdgesOnCell)
              call mpas_pool_get_array(meshPool, 'cellsOnCell', cellsOnCell)
 
-             call mpas_allocate_scratch_field(smoothedLevelsField, .true.)
-
```
xylar: I think I agree that the temporary field isn't needed but want to talk it through.  If a cell is deeper than its neighbors, it will get filled in.  By definition, it is now at the same maxLevelCell as its deepest neighbor.  Therefore, no neighbor can be deeper than this cell if it was altered and no neighbors will be filled.  So it should be safe to update maxLevelCell as we go.

src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
xylar: I agree, I think this is safe to do. I believe I had `ocn_alter_bottomDepth_for_pbcs` in `ocn_init_vertical_grid` because it currently only gets called if you are *not* constraining the vertical grid base don the Haney number, which handles partial bottom cells after making substantial alternations to the grid.  This earlier call might slightly modify the outcome of the later stage but not in a way that I expect to be important (just probably not bit-for-bit).

xylar:
I'd like to do some testing tomorrow.

src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F
mark-petersen: Yeah, I went through the same logic, and came to the same conclusion.

xylar:
@mark-petersen and @vanroekel, I tested including my 2 commits with `QU240` and `QU240wISC` on my laptop (with gnu) and everything went fine. 

I tested `EC60to30` and `EC60to30wISC` on grizzly with intel, and was able to run `init` successfully.  I had CFL problems in both setups during `spin_up` but I suspect this is unrelated to this PR.  This PR is also not the right place to address those problems.

 I used the following script to check that `maxLevelCell` had the desired property in `init/initial_state.nc` for each of the 4 setups listed above:
```python
#!/usr/bin/env python

import xarray
import argparse
import numpy


parser = argparse.ArgumentParser()
parser.add_argument("-i", "--infile", dest='infile', required=True,
                    help="MPAS file with maxLevelCell to check")
args = parser.parse_args()

ds = xarray.open_dataset(args.infile)

cellsOnCell = ds.cellsOnCell.values - 1

maxLevelCell = ds.maxLevelCell.values - 1

maxLevelNeighbor = numpy.zeros(maxLevelCell.shape, int)
for index in range(ds.sizes['maxEdges']):
    coc = cellsOnCell[:, index]
    mask = coc >= 0
    maxLevelNeighbor[mask] = numpy.maximum(maxLevelNeighbor[mask],
                                           maxLevelCell[coc[mask]])

mask = maxLevelCell >= 0

assert(numpy.all(maxLevelCell[mask] <= maxLevelNeighbor[mask]))
```

xylar:
I suspect that the changes related to reverting Redi https://github.com/MPAS-Dev/MPAS-Model/commit/d3667cc77d66d3b587225709b26db44e2812c65f might have also reverted some changes to the EC60to30 setups that made it possible to spin them up (and aren't related directly to Redi).

mark-petersen:
@xylar I ran the these, with the following results:
1. EC60to30 on this PR: init and spin-up complete.
2. EC60to30wISC on this PR: init and SSH adjustment works. Spin-up dies, as you noted.
3. EC60to30wISC on ocean/develop: init works. SSH adjustment dies on first iteration (go figure).

I think the EC60to30wISC spin-up can work by adjusting Rayleigh damping and time step. I would rather make those adjustments when we have everything in place (64 levels, new mesh, etc), rather than get it to work for this one PR. If you agree, I'll merge it - perhaps after #495

xylar:
@mark-petersen,
> EC60to30 on this PR: init and spin-up complete.

That's strange, it didn't work for me.  Did you change anything?

> I think the EC60to30wISC spin-up can work by adjusting Rayleigh damping and time step. I would rather make those adjustments when we have everything in place (64 levels, new mesh, etc), rather than get it to work for this one PR. If you agree, I'll merge it - perhaps after #495

Our agreement was that we were *not* going to change to 64 vertical levels for EC60to30wISC for now.  So we need to set this up as is, and get spin-up working.  This is quite a high priority, since we have to start burning through our time on NERSC before June 30th.

xylar:
I will keep working on spin-up today, so we can hopefully get this IC in place before #495

xylar:
I think this is not working (anymore).  `maxLevelCell = 60` everywhere and `bottomDepth = 5500` everywhere in my `EC60to30wISC` testing.

xylar:
It's probably the changes I made to try to fill pits in the Haney-number constrained  initialization.  I'm going to try again with the commit before that one and see if things look normal.

mark-petersen:
This is definitely something I introduced. It occurred without ISC as well. I'm debugging now, but have calls interspersed this morning.

mark-petersen:
@xylar I'm sorry I wasted your time on this. Bug fix is in. I had deleted a print statement and deleted the exit on the next line by mistake. You can proceed with your ISC testing.

xylar:
Okay, great!  Thanks for finding that.  Then we all need to re-test because all our "tests" passed trivially with a flat bottom...  @vanroekel, can you run a quick test when you have time?

xylar:
I'm going to start again with my EC60to30wISC spin-up starting with this branch and modifying the stages of spin-up if needed.

vanroekel:
sure, I should be able to do that today.  Would it be enough to run init again?  Or would you like me to run the spin up as well?

mark-petersen:
I'm running EC60to30 init and spinup now. I'll post some plots.

xylar:
> Would it be enough to run init again? Or would you like me to run the spin up as well?

Just running the test you did before should be enough, given the spin-ups @mark-petersen and I are doing.

vanroekel:
okay, happy to do that.  I'll let you know what I find.

xylar:
@mark-petersen, something is not working right with Haney-number constrained initialization at the moment but it's a problem with `ocean/develop` as well, so not related to this PR.  I'll look into it tomorrow but I don't have the energy for it today.  Anyway, that is to say that I won't be able to test this PR on EC60to30wISC until I figure out why the Haney Number is comping out as 10572.8534636064!!!

vanroekel:
I just retested and the result is good, no single cell pits.

xylar:
@mark-petersen, I somehow missed linking in (and using) the land-ice mask from cell culling in the EC60to30wISC initial_state step.  This is now fixed in https://github.com/MPAS-Dev/MPAS-Model/pull/499.  After cherry-picking #499 onto this branch, I was able to create an EC60to30wISC mesh that doesn't have any pits and looks sensible in paraview.  I'm working on spinning it up now.

xylar:
The EC60to30wISC spin-up was also successful, so we should be good to go (for the maint-1.2 run) as soon as this, #499 and #495 get merged.  I don't want to create the initial condition from my current branch because I would like to be able to point to a specific commit on `ocean/develop` that it was generated from.

mark-petersen:
rebased. retesting now.

mark-petersen:
Passed nightly regression with debug, and also QU240wISC.

mark-petersen:
@xylar ocean/develop is ready to go. I am running the EC60to30wISC right now here:
```
/lustre/scratch4/turquoise/mpeterse/runs/200401_EC60to30wISC/ocean/global_ocean/EC60to30wISC/init
```
remember to use the directory `initial_state_60_layer`. I renamed it to `initial_state` so all the links work, i.e.
```
mv initial_state initial_state_renamed_64
mv initial_state_60_layer initial_state
```

PR #453, opened 2020-02-21
Modified logic to determine when FEM needs to be recomputed

matthewhoffman:
Fix logic to determine whether the finite element mesh needs to be recomputed. Note: in order to compare the cell mask with the previous time steps, we would need the cell mask to have 2 time levels. However, if we set the cell mask to have 2 time levels we get different results. This needs to be fixed. At the moment the cell mask is set to have 1 time level, so it might happen that the mesh is not recomputed when it should.


mgduda:

PR #455, opened 2020-02-25
Adds Ocean Heat Content AM

vanroekel:
Adds the ability to calculate ocean heat content during a forward model run

It computes heat content over standard layers
0-700m
700-2000m
2000m - bot
full Depth

### DIFF
--- a/src/core_ocean/analysis_members/Makefile
+++ b/src/core_ocean/analysis_members/Makefile
@@ -25,7 +25,8 @@ MEMBERS = mpas_ocn_global_stats.o \
           mpas_ocn_rpn_calculator.o \
           mpas_ocn_transect_transport.o\
           mpas_ocn_eddy_product_variables.o\
-          mpas_ocn_moc_streamfunction.o
+          mpas_ocn_moc_streamfunction.o\
+          mpas_ocn_ocean_heat_content.o
 
 all: $(OBJS)
 
--- a/src/core_ocean/analysis_members/Registry_analysis_members.xml
+++ b/src/core_ocean/analysis_members/Registry_analysis_members.xml
@@ -29,3 +29,4 @@
 #include "Registry_transect_transport.xml"
 #include "Registry_eddy_product_variables.xml"
 #include "Registry_moc_streamfunction.xml"
+#include "Registry_ocean_heat_content.xml"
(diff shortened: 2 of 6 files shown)


vanroekel:
I've tested this in a brief 60to30 stand alone test case.  Results look reasonable. Here is the sfctobottom anomaly (units of 10^22 J)

Screen Shot 2020-02-25 at 3 23 15 PM

vanroekel:
I've added four arrays (nCells dimension)

<var name="oceanHeatContentSFCtoBot"/>
<var name="oceanHeatContentSfcto700m"/>
<var name="oceanHeatContent700mto2000m"/>
<var name="oceanHeatContent2000mtoBot"/>

milenaveneziani:
@vanroekel: we normally compute the OHC from the layerWeigthedAvg AM (and then define the depth ranges over which to do the accumulated quantity through the MPAS-Analysis config file). So I am not sure why we need this AM? Does the layerWeigthedAvg AM have a problem?

vanroekel:
my read of that AM is it can do the globally averaged OHC, this returns a horizontal map.  This PR is needed to return the map but more importantly for work on a LDRD-DR effort where we want to compute the 2-D maps at high frequency while the model runs and do statistics on the data.

milenaveneziani:
Indeed the layerWeightedAvg AM computes quantities over specified regions (including the global region). I was thinking that we could potentially modify it, to output the T and S content globally over specified depth ranges, but perhaps better to keep things separate.

So I assume this AM does what your postprocessing script does to get the global maps?

mark-petersen:
rebased. Added this AM and output variables to analysis test in nightly regression suite.

On above comments, it sounds like this AM is sufficiently different from others. If it is useful, it's fine to have some redundancy between AMs and MPAS-Analysis. We can also write these out at higher frequency.

mark-petersen:
This can can be merged in next week. It requires E3SM flag additions, so will be put in an E3SM PR for that.

mark-petersen:
Rebased. Passes nightly regression with gnu and intel/debug, including this AM in the analysis test.

mark-petersen:
@xylar and @milenaveneziani I would like to merge this. Do you have any comments?

xylar:
Sorry, I didn't realize my review had been requested. I don't have any input so will take myself off as a reviewer.

PR #459, opened 2020-02-28
Add static data set index file parser

MiCurry:
This commit creates a new module, mpas_parse_geoindex, which can parse index
files that describe static data sets. The index files and data sets are also used by WRF and the WPS.

For more information on valid fields for an index file please see section 3-53 "Description of Index Options" of the WRF Users Guide.

Currently, mpas_parse_geoindex is not used anywhere in the model; it is only compiled.

### DIFF
--- a/src/core_init_atmosphere/mpas_parse_geoindex.F
+++ b/src/core_init_atmosphere/mpas_parse_geoindex.F
@@ -1,0 +1,263 @@
+module mpas_parse_geoindex
+
+   use mpas_log, only : mpas_log_write
+   use mpas_derived_types, only : MPAS_LOG_ERR, MPAS_LOG_WARN
+   use mpas_pool_routines
+
+   implicit none
+
+   private
+
+   public :: mpas_parse_index
+
+   contains
+
+   !***********************************************************************
+   !
+   !  subroutine mpas_parse_index
+   !
+   !> \brief   Parse a geogrid's index file and put the results into an MPAS pool
+   !> \author  Miles A. Curry
+   !> \date    02/20/2020
+   !> \details
+   !> Parse an index file of a static data set into an MPAS pool, allocating
+   !> each keyword=value pair into the pool with the pool member key being
+   !> keyword, and the value being value.
+   !>
+   !> This function can parse index files with one keyword=value pair
+   !> per line; a "#" at the start of a line, which will cause the line to be
+   !> ignored; or an empty line containing only a newline/return character, which
+   !> will also be ignored. Spaces or tabs before, between or after the
+   !> keyword=value tokens are > ignored.
+   !>
+   !> If a line contains anything but the above valid syntaxes, a syntax
+   !> error will raised and -1 will be returned.
+   !>
+   !> Case is ignored.
+   !>
+   !> The definitions of a keyword, which can found in section 3-53
+   !> of the WRF-AWR User's Guide, will determine the corresponding type
+   !> of that keyword. A keyword that has been assigned the wrong type
+   !> will raise a type error and -1 will be returned.
+   !>
+   !> Keywords that are not handled explicitly by this function will produce
+   !> a warning that the keyword is unrecognized.
+   !
+   !-----------------------------------------------------------------------
+   function mpas_parse_index(path, geo_pool) result(ierr)
+
+      use mpas_io_units
+
+      implicit none
+      ! Input Variables
+      character (len=*), intent(in) :: path
+      type (mpas_pool_type), intent(inout) :: geo_pool
+      integer :: ierr
+
+      ! Local Variables
+      character (len=StrKIND) :: line, lhs, rhs
+      character (len=StrKIND) :: read_err_msg, open_msg
+      integer :: geo_unit
+      integer :: open_stat, read_stat, line_read_stat
+      integer :: i, k
+      logical :: res
+
+      character (len=StrKIND), pointer :: char_t
+      integer :: iceiling, ifloor
+      integer, pointer :: int_t
+      real(kind=RKIND), pointer :: real_t
+
+      ierr = 0
+
+      inquire(file=trim(path), exist=res)
+      if ( .not. res) then
+         call mpas_log_write("Could not find or open the file at: "//trim(path), messageType=MPAS_LOG_ERR)
+         ierr = -1
+         return
+      endif
+
+      call mpas_new_unit(geo_unit)
+      open_stat = 0
+      open(geo_unit, FILE=trim(path), action='READ', iostat=open_stat, iomsg=open_msg)
+      if (open_stat /= 0) then
+         call mpas_release_unit(geo_unit)
+         call mpas_log_write("Could not open 'index' file at:'"//trim(path)//"'", messageType=MPAS_LOG_ERR)
+         call mpas_log_write(trim(open_msg), messageType=MPAS_LOG_ERR)
+         ierr = -1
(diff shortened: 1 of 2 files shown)


MiCurry:
Like the mpas_kd_tree and mpas_stack, this gist, https://gist.github.com/MiCurry/d32539c1bfaff58b34410d0a4a1d4cba, contains an alternative `mpas_init_atm_core.F` which contains tests for `mpas_parse_geoindex`. It also contains a test index file.

To use the index file and the alternative `mpas_init_atm_core.F`, create a directory at the same level as the `init_atmosphere` executable named `geog_test` and place the index file above inside it.

src/core_init_atmosphere/mpas_parse_geoindex.F
```diff
@@ -0,0 +1,238 @@
...
+      index_path = trim(path)//"index"
+      inquire(file=index_path, exist=res)
+
+      if ( .not. res) then
+         call mpas_log_write("No 'index' file found in the geographic data set: "//index_path, messageType=MPAS_LOG_CRIT)
+      endif
+
+      open_stat = 0
+
+      open(10, FILE=index_path, action='READ', iostat=open_stat, iomsg=open_msg)
```
mgduda: Rather than using a hard-wired unit number of 10 here, we should instead use the `mpas_new_unit` and `mpas_release_unit` routines from `mpas_io_units`.

src/core_init_atmosphere/mpas_parse_geoindex.F
```diff
@@ -0,0 +1,238 @@
...
+      if (open_stat /= 0) then
+         call mpas_log_write("Could not open 'index' file at:'"//trim(index_path)//"'", messageType=MPAS_LOG_ERR)
+         call mpas_log_write(trim(open_msg), messageType=MPAS_LOG_CRIT)
+      endif
+
+      line_read_stat = 0
+      read_stat = 0
+      k = 0
+      do while ( .true. )
```
mgduda: While there are good reasons for using a `while (.true.)` loop in some situations, I think it's harder for the reader to immediately appreciate the conditions under which the loop will terminate. I think it's worth considering a loop structure like:
```fortran
      k = 1
      read(10,'(A)', iostat=line_read_stat) line
      do while (line_read_stat == 0)

         ... work ...

         k = k + 1 ! Keep track of line numbers for error reporting
         read(10,'(A)', iostat=line_read_stat) line
      end do
```

src/core_init_atmosphere/mpas_parse_geoindex.F
```diff
@@ -0,0 +1,238 @@
...
+      character (len=StrKIND) :: index_path, line, lhs, rhs
+      character (len=StrKIND) :: read_err_msg, open_msg
+      integer :: open_stat, read_stat, line_read_stat
+      integer :: i, k
+      logical :: res
+
+      character (len=StrKIND), pointer :: char_t
+      integer :: iceiling, ifloor
+      integer, pointer :: int_t
+      real, pointer :: real_t
```
mgduda: Currently, if MPAS is built with double-precision reals, we define RKIND=8 and add compiler flags to promote default reals to 64-bits; but, in future, we may remove the compiler flags. So it would be best if `real` variables were always assigned an explicit kind, e.g., `real(kind=RKIND)`.

src/core_init_atmosphere/mpas_parse_geoindex.F
MiCurry: Sounds great, thanks for pointing out the mpas_io_units module!

src/core_init_atmosphere/mpas_parse_geoindex.F
MiCurry: I thought about doing this method, but I wasn't sure which method would be most readable. I think the version you suggested is most readable; however, I'm not sure that it will work with the current implementation.

On lines 94 and 95, we check to see if the line is comment (Starts with '#') or an empty line. If it is, `cycle` is called, the loops starts over, execution returns to line 94 and 95 with the same value for line, and again calls cycle. Thus the program enters an infinite loop.

https://github.com/MPAS-Dev/MPAS-Model/blob/b262a9cf4a720130823a1dee6338da6dd437b72d/src/core_init_atmosphere/mpas_parse_geoindex.F#L94-L95

If we move the cycle lines to below the read at the end of the `do while` loop, then we potentially could try to parse a comment or an empty line that was read by the first read.

To fix it, we could wrap the `do` loop on line 97 in an `if` statement that checks for either a comment or an empty line and avoid using the `cycle` statement all together.

Do you have any thoughts on this @mgduda?

mgduda: Would it work to replace the two lines (94 and 95) that handle comment and blank lines with something like the following?
```fortran
         !
         ! If a blank or comment line is encountered, just move on to the next line
         !
         if (line(1:1) == '#' .or. len_trim(line) == 0) then
            k = k + 1
            read(10,'(A)', iostat=line_read_stat) line
            cycle
         end if
```

MiCurry:
Okay, @mgduda I have just pushed all of those changes that you requested and I think this PR is ready to go.

src/core_init_atmosphere/mpas_parse_geoindex.F
```diff
@@ -0,0 +1,265 @@
...
+   !
+   !-----------------------------------------------------------------------
+   function mpas_parse_index(path, geo_pool) result(ierr)
+
+      use mpas_io_units
+
+      implicit none
+      ! Input Variables
+      character (len=StrKIND), intent(in) :: path
+      type (mpas_pool_type), intent(inout), pointer :: geo_pool
```
mgduda: Ultimately it probably doesn't matter, but I don't think we actually need to require that `geo_pool` be a pointer. This would in principle allow calling code to pass an actual argument that was not a pointer; and, it avoids the need verify that the `geo_pool` pointer is associated in this routine.

src/core_init_atmosphere/mpas_parse_geoindex.F
```diff
@@ -0,0 +1,265 @@
...
+      logical :: res
+
+      character (len=StrKIND), pointer :: char_t
+      integer :: iceiling, ifloor
+      integer, pointer :: int_t
+      real(kind=RKIND), pointer :: real_t
+
+      ierr = 0
+
+      index_path = trim(path)//"index"
```
mgduda: As a design consideration, we could require that the caller provide the full path (including filename) that this routine will parse. It adds a little extra work on the caller's side, but it makes this code more general, in that we could parse any file we like rather than just files named "index".

src/core_init_atmosphere/mpas_parse_geoindex.F
MiCurry: I think that's a good idea. There isn't any reason for it to be a pointer. I'll remove the `pointer` attribute from `geo_pool`.

src/core_init_atmosphere/mpas_parse_geoindex.F
MiCurry: I like that idea as well. Best to be as general as we possible. I'll update that as well.

MiCurry:
Okay, @mgduda. I just pushed an updated that addresses the last changes you requested.

MiCurry:
@mgduda I have just pushed an update which replaces `path` usages with `trim(path)`.

PR #461, opened 2020-03-02
add the Bassis & Ma damage calculation and calving scheme

tongzhangice:
Rebase onto commit e1eef63
In the calculare_damage subroutine, the damage is initialized with the Nye zero stress and varies in [0, 1] on the ice shelf (0 for all grounded ice) according to the theory of Bassis & Ma (2015). The ice cell adjacent to the ocean will be calved away if the damage value exceeds a certain threshold value (can be prescribed in the namelist), if we turn on the entry, 'damagecalving', in the namelist. The damage is able to keep its value from healing (restore_damage). We can also connect the damage to rheology by updating stiffness = 1 - damage.


matthewhoffman:
Replaced by #747 .

PR #463, opened 2020-03-03
Fix race condition for ezxml library in parallel builds

mgduda:
This merge fixes a race condition affecting the ezxml library when using
a parallel build to compile MPAS.

Under the right conditions, a race condition in the compilation of the ezxml
library could lead to a failed build. The race condition occurs when the MPAS
build tools (input_gen, parse) try to build the ezxml library for use on
login nodes, while the MPAS framework simultaneously tries to build the ezxml
library for use on batch nodes.

The fix implemented in this merge resolves this issue by modifying the Makefile
for the ezxml library to compile an object file with a specified name. The MPAS
tools specify an object file name of ezxml_tools.o, while the framework
specifies an object file name of ezxml.o.

### DIFF
--- a/src/external/Makefile
+++ b/src/external/Makefile
@@ -6,7 +6,7 @@ esmf_time:
 	( cd esmf_time_f90; $(MAKE) FC="$(FC) $(FFLAGS)" CPP="$(CPP)" CPPFLAGS="$(CPPFLAGS) -DHIDE_MPI" GEN_F90=$(GEN_F90) )
 
 ezxml-lib:
-	( cd ezxml; $(MAKE) )
+	( cd ezxml; $(MAKE) OBJFILE="ezxml.o" )
 
 clean:
 	( cd esmf_time_f90; $(MAKE) clean )
--- a/src/external/ezxml/Makefile
+++ b/src/external/ezxml/Makefile
@@ -1,14 +1,5 @@
-.SUFFIXES: .c .o
-
-OBJS = ezxml.o
-
-all: clean
-	$(MAKE) -j 1 library
-
-library: $(OBJS)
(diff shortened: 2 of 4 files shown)


mgduda:
@mark-petersen and @matthewhoffman After this PR, I have just one other small software fix for the `init_atmosphere` core, after which I'm hoping we can release MPAS v7.1.

MiCurry:
This fixes the race condition I was getting quite frequently on a number of MPAS projects, which I had assumed were caused by me messing up dependencies, but I've been able to reliably recreate the problem (and confirm this fix works).

The problem occurs after a re-making a core without cleaning, with a fast enough compiler or a short compile. For instance, the init_atmosphere core, or dummy_core (https://github.com/mgduda/MPAS-Model/tree/dummy_core) compiled, and recompiled with gfortran will run into a race condition. This also seems to occur with any number of make threads greater than 2.

ifort seems to not run into this problem, perhaps because the timing of its compiles of specific files line up differently.

Mergin the reconstruct_u core, https://github.com/MiCurry/MPAS-Model/tree/reconstruct_u, fixes the failed re-compiles I was getting! Thanks for finding this @mgduda.

mark-petersen:
@matthewhoffman one small note, any Makefile change might imply a corresponding CMake change is required. However, we have never had a failure with the ezxml build with Make or CMake, so in this case I think "if it ain't broke, don't fix it", since a build fail would be obvious when it occurs.

matthewhoffman:
@mark-petersen , that makes sense to me.  And thanks for testing.

mgduda:
Thanks, @MiCurry , @mark-petersen , and @matthewhoffman !

PR #464, opened 2020-03-04
Add Coast of Maine refined to COMPASS

mark-petersen:
Add Gulf of Maine variable resolution mesh, which we made as an example for Wolfram's proposal. We did not run simulations, but I wanted to record it in COMPASS.


mark-petersen:
CUSP mesh, with coast of Maine embedded resolution
Coast_of_Maine_cell_width_1
Coast_of_Maine_cell_width_3

mark-petersen:
@pwolfram if you could just upload your mesh picture, and approve this based on the mesh I created, that would be great. No need to review the files.

pwolfram:
GoMmesh

PR #465, opened 2020-03-04
Remove geom data

xylar:
This data can now be included in the `geometric_features` conda package, and the `compass_0.1.0` environment includes the version with this data.

### DIFF
--- a/testing_and_setup/compass/README_ocean.md
+++ b/testing_and_setup/compass/README_ocean.md
@@ -34,31 +34,6 @@ In six places, replace `FULL_PATH_TO_MPAS_MODEL_REPO` with the path where you
 have checked out (and built) the branch of MPAS-Model you are planning to use.
 Five other paths are required, as explained below.
 
-### geometric\_data
-
-```
-geometric_data = FULL_PATH_TO_PATH_OF_CACHED_GEOMETRIC_FEATURES_DATA
-```
-In general, this can be any directory where geojson data files will be
-automatically downloaded and cached as test cases require them.  However, if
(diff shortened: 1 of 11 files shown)


xylar:
I'm having trouble updating compass 0.1.0 on cori at the moment because of some file system issue.  We don't have a compass environment there at all at the moment.  Everywhere else (LANL IC, compy and blues) we have environments that include geometric_data in them, so we no longer need to point to a local cache of the data.

testing_and_setup/compass/README_ocean.md
```diff
@@ -6,21 +6,25 @@ To set up and run ocean test cases from COMPASS, you will need a conda
 environment.  First, install Miniconda3 (if miniconda is not already
 installed), then create a new conda environment as follows:
 ``` bash
-conda create -n compass_py3.7 -c conda-forge -c xylar python=3.7 \
-    geometric_features=0.1.4 mpas_tools=0.0.6 jigsaw=0.9.11 jigsawpy=0.0.2 \
-    metis pyflann scikit-image basemap pyamg ffmpeg
+conda create -n compass_0.1.0 -c conda-forge -c e3sm python=3.7 compass=0.1.0
```
mark-petersen: This is now out of date. I'll just use the second commit, as the first commit updated the conda version.

mark-petersen:
Kept second commit, removing geometric_data links.

testing_and_setup/compass/README_ocean.md
xylar: Correct.  This is now taken care of in the `compass` metapackage.

PR #466, opened 2020-03-05
Add QU480 global_ocean test case

xylar:
This will be useful for validating analysis and other infrastructure tools

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/QU480/init/config_base_mesh.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/QU480/init/config_base_mesh.xml
@@ -1,0 +1,1 @@
+../../config_files//config_base_mesh.xml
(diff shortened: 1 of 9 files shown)


xylar:
@mark-petersen and @jonbob, this is the new mesh I was talking about in the email I just sent.  This would be useful in E3SM, together with ne4, in an ultra-low-res configuration that could be used both for smoke-testing E3SM itself and for smoke-testing analysis tools such as MPAS-Analysis, E3SM-Diags and processflow.

xylar:
A question is what the mesh should be called in E3SM, since oQU480 is already taken and would indicate a v1 mesh in any case.

xylar:
## Testing

I build and ran the test case on my laptop (Ubuntu with gfortran).  Everything ran successfully.  (I had to reduce the number of cores to 8 from 16 for my laptop.)

xylar:
I was able to get an ultra-low-resolution E3SM run going with the existing oQU480 mesh, so I don't think this PR is needed anymore.

PR #417, opened 2019-12-13
Create E3SM coupling files within COMPASS

mark-petersen:
Adds a script that produces coupling files for E3SM. This PR adds it to the QU240 as a test, but it is set up in a general way, so that any global case can add it to the init process.

The goals for this system are:
- speed: An automated method
- simplicity: so others can use and alter it
- process documentation: ability to record a standard sequence of steps
- expandability: so special cases, like ice shelves and data icebergs, are easy to handle.

I think this method of a single python script within COMPASS satisfies these.

To add this to any case, add a link 
```
cd testing_and_setup/compass/ocean/global_ocean/RESOLUTION/init
ln -isf ../../config_files/config_e3sm_coupling.xml  .
```
Then set up an init case for that resolution, and there is a new directory e3sm_coupling with a run.py script.

Documentation of all the steps for coupling files are at
- [Confluence: Simplify E3SM Ocean File Generation](https://acme-climate.atlassian.net/wiki/spaces/OCNICE/pages/763854989/Simplify+E3SM+Ocean+File+Generation)
- [Confluence: Making mapping, runoff, domain files and adding grids to ACME](https://acme-climate.atlassian.net/wiki/spaces/OCNICE/pages/22052884/Making+mapping+runoff+domain+files+and+adding+grids+to+ACME)

PR #467, opened 2020-03-06
Add support for copy_file tag

xylar:
This is needed for copying config files that may then be altered by the user within a test case.

Since there is a significant amount of redundant code between this and and support for the `add_link` tag, a helper function has been added to encapsulate the redundancy.

Some error messages have been modified to raise exceptions.

### DIFF
--- a/testing_and_setup/compass/setup_testcase.py
+++ b/testing_and_setup/compass/setup_testcase.py
@@ -25,6 +25,7 @@ import subprocess
 from six.moves import configparser
 import textwrap
 import netCDF4
+import shutil
 
 try:
     from collections import defaultdict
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen, let me know if `develop` is not where this belongs.  I think this is what we've done in the past with regard to changing the shared COMPASS infrastructure.

xylar:
@matthewhoffman, https://github.com/MPAS-Dev/MPAS-Model/pull/417 merged in @mark-petersen's work in progress on making initial conditions, mapping files, etc. that are ready to go into E3SM.  One of the things we realized we couldn't do in COMPASS was copy (as opposed to symlink) a file.  In this particular case, we want to copy a config file that a user would then alter as needed for a particular config (e.g for debugging).

xylar:
## Testing

I built the QU240 `init` and `test` stages, and ran all steps of the `init` stage.  All the symlinks are good and the `file_copy` step worked as expected.

It would be good to setup and run as many test cases as practical before merging this, since the `add_link` code has a lot of `if` statements that might not be tested in a typical setup.  For now, I'm setting up all of the `global_ocean` test cases to make sure they don't have any broken symlinks (other than those pointing to other steps and stages of the test cases themselves).

mark-petersen:
@mgduda this goes into `develop`, but I'm not asking for your review because it is only in COMPASS.

@matthewhoffman, this is adding a new COMPASS function `copy_file` as only `add_link` was there before. This is really useful for config files where it is useful to edit locally, but you don't want to alter a file linked back to the repo.

xylar:
@matthewhoffman, I know this is already merged, but could you run a test case or 2 from `landice` to make sure `add_link` still works as before for you?  I did quite a bit of testing on the `ocean` side so I'm not worried but it would be better to know sooner rather than later if there's a problem.

mark-petersen:
@xylar, I added this one commit to `ocean/develop` so we have it now. I did a local test merge of `ocean/develop` into `develop` and there are no conflicts, as the `setup_testcase.py` file is identical on both branches.

matthewhoffman:
@xylar , I'll try to double check this today when I'm doing some COMPASS stuff.  This is a valuable feature that I've wanted in the past.  Thanks for putting it together.

xylar:
>  I added this one commit to ocean/develop so we have it now. I did a local test merge of ocean/develop into develop and there are no conflicts, as the setup_testcase.py file is identical on both branches.

In this case, I'm okay with this.  In general, we don't want to put in redundant commits because it makes debugging a lot more challenging because the same change is effectively introduced in two places.

mark-petersen:
I agree. The alternative today was to merge develop back to all three cores at once and make a separate E3SM PR for just that, which is way too much overhead for this small change on a lone file in COMPASS.

matthewhoffman:
@xylar , I included this in a COMPASS run today, and everything worked. Nice work.

PR #468, opened 2020-03-07
Create mask and mapping files for online MOC and MPAS-Analysis

xylar:
This merge adds functionality for creating MOC, Antarctic-region, ocean-basin, ice-shelf and transport-transect masks using `mpas_tools` and `geometric_features`.  It does not require any pre-cached files other than those already part of `geometric_features` (which now includes all the geometric data when it is installed as part of the `compass` conda environment).

Mapping files between the mesh and the default comparison grids in MPAS-Analysis are now also created and staged for uploading.

This merge also performs some clean-up:
* Fixes a typo in `ocean_initial_state`
* Removes trailing white space and performs some other PEP8 clean-up (mostly 2 new lines between functions)
* Points to the initial condition from `ssh_adjustment` rather than `initial_state` for test cases with ice-shelf cavities. (This should eventually point to the last restart file from spin-up for most runs, so this is only an intermediate fix.)

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/config_files_ISC/config_e3sm_coupling.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/config_files_ISC/config_e3sm_coupling.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <config case="e3sm_coupling">
 
-	<add_link source="../initial_state/initial_state.nc" dest="init.nc"/>
+	<add_link source="../ssh_adjustment/init.nc" dest="init.nc"/>
 	<add_link source="../initial_state/graph.info" dest="graph.info"/>
 	<add_link source="../culled_mesh/no_ISC_culled_mesh.nc" dest="no_ISC_culled_mesh.nc"/>
     <add_link source_path="script_configuration_dir" source="scripts/create_E3SM_coupling_files.py" dest="create_E3SM_coupling_files.py"/>
--- a/testing_and_setup/compass/ocean/global_ocean/scripts/config_E3SM_coupling_files.ini
+++ b/testing_and_setup/compass/ocean/global_ocean/scripts/config_E3SM_coupling_files.ini
@@ -20,7 +20,6 @@ enable = true
 
 [transects_and_regions]
 enable = true
-transect_region_geojson = /usr/projects/climate/mpeterse/analysis_input_files/geojson_files/SingleRegionAtlanticWTransportTransects.geojson
 
 [mapping_CORE_Gcase]
 enable = true
@@ -34,7 +33,7 @@ atm_scrip_tag = JRA025
 [mapping_ne30]
 enable = false
 # need to add complete name here:
-atm_scrip_tag = ne30 
+atm_scrip_tag = ne30
 
 [domain_CORE_Gcase]
 enable = true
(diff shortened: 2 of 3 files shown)


xylar:
I have cherry-picked the commit from #467 so I added the "don't merge" flag.  Once #467 is merged and `develop` is merged into `ocean/develop`, I'll rebase and remove the redundant commit.

xylar:
## Testing

On my laptop, I have run the `init` stage including the `e3sm_coupling` step for:
* `QU240`
* `QU240wISC`
* `EC60to30wIsC` (was not able to complete the `ssh_adjustment` on my laptop)

The masks are created as expected.  For example:
```
$ ls assembled_files_for_upload/inputdata/ocn/mpas-o/oQU240wISC/
...
oQU240wISC_moc_masks_and_transects.nc

$ ls assembled_files_for_upload/diagnostics/mpas_analysis/region_masks/
oQU240wISC_antarcticRegions.nc      oQU240wISC_oceanBasins.nc
oQU240wISC_iceShelfMasks.nc         oQU240wISC_transportTransects.nc
oQU240wISC_moc_masks_and_transects.nc
```

Further testing is needed to make sure the resulting files actually work as expected in MPAS-Analysis.

The online MOC analysis member will also need to be tested, and may require changes (in particular with the new file name that includes 4 regional MOC basins, not just the Atlantic).

xylar:
@milenaveneziani and @darincomeau, don't worry about reviewing this just yet.  At some point, I might ask you to perform an MPAS-Analysis run with the masks and mapping files created by `e3sm_coupling`.  But for now we don't have very many runs to try it out on.

@mark-petersen, your feedback would be appreciated anytime, since you're familiar with the code I'm altering.

milenaveneziani:
@xylar: would it make sense to add a section  for making Arctic regions? I can add it, since I know the different region groups I created.
And I could also test it on an ARRM60to10 run.

xylar:
@milenaveneziani , sure, feel free.

xylar:
@milenaveneziani, if you just want to give me a list of regions, I'll be happy to add it, too.

milenaveneziani:
Hi @xylar, thanks for offering. I seem to understand that you need the full list of regions, not just the group or tag name, right?

xylar:
@milenaveneziani, a tag would be okay.  `geometric_features` doesn't have a concept of a "group", so that isn't sufficient.  If you want to send me a python script that creates the feature collection, I can use that, too.  Many of the others come from the "examples" directory.

milenaveneziani:
A tag is fine. I have added a 'Arctic' tag to regions of the Arctic_Basin and Atlantic_Basin (two other tags) that are relevant to the Arctic and sub-Arctic (such as the Canada Basin and the Labrador Sea). So the 'Arctic' regions would be great to add a this point.

There is also another parallel category, that @proteanplanet (Andrew) requested for comparisons with sea-ice observations, and that is called 'Arctic_NSIDC'. We can add this later if necessary.

milenaveneziani:
Here is the list of regions associated with the 'Arctic' tag:
Baffin_Bay
Baltic_Sea
Barents_Sea
Canada_Basin
Canadian_Archipelago
Central_Arctic
Chukchi_Sea
East_Siberian_Sea
Greenland_Sea
Hudson_Bay
Irminger_Sea
Kara_Sea
Labrador_Sea
Laptev_Sea
North_Sea
Norwegian_Sea

milenaveneziani:
ooh, that was nice and quick. thanks. I will start a test soon.

xylar:
I tested the Arctic regions on the QU240 grid.  I get masks for these regions:
```
 regionNames =
  "Baffin Bay",
  "Baltic Sea",
  "Barents Sea",
  "Canada Basin",
  "Canadian Archipelago",
  "Central Arctic",
  "Chukchi Sea",
  "East Siberian Sea",
  "Greenland Sea",
  "Hudson Bay",
  "Irminger Sea",
  "Kara Sea",
  "Labrador Sea",
  "Laptev Sea",
  "North Sea",
  "Norwegian Sea" ;
```
Looks the same as your list.

milenaveneziani:
oops, wait, now I understand why you say we should wait about testing.. This is something that comes during the init step.

xylar:
Right, this is for generating new meshes, so that's the stage at which it's worth testing.  Presumably once @mark-petersen has got enough of these features on a branch for testing, we'll run one of the test cases through the paces all the way to analysis.

PR #469, opened 2020-03-09
Add COMPASS step for remapping prescribed ISMF

xylar:
The data is from the Rignot et al. (2013) data set (not publicly available).

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/config_files_ISC/config_e3sm_coupling.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/config_files_ISC/config_e3sm_coupling.xml
@@ -8,6 +8,7 @@
     <add_link source_path="script_configuration_dir" source="scripts/copy_cell_indices_ISC.py" dest="copy_cell_indices_ISC.py"/>
     <copy_file source_path="script_configuration_dir" source="scripts/config_E3SM_coupling_files.ini" dest="config_E3SM_coupling_files.ini"/>
     <add_link source_path="script_configuration_dir" source="scripts/readme_E3SM_coupling_files" dest="readme"/>
+	<add_link source_path="initial_condition_database" source="MeltRatesRignot2013.nc" dest="MeltRatesRignot2013.nc"/>
 
 	<run_script name="run.py">
 		<step executable="python">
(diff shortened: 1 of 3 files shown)


xylar:
This work is based off of #468 and should ideally be rebased after that is merged.

mark-petersen:
Please rebase later today or tomorrow.

xylar:
I think we decided we weren't bothering with this one just yet, right?

mark-petersen:
Sure. We can put it lower on the list.

xylar:
I went ahead and rebased onto #433 and there were no conflicts.

mark-petersen:
@xylar could you rebase this again? There were some conflicts on some python import lines that I was unsure of.

xylar:
I rebased and there was no conflict.

mark-petersen:
@xylar, can you describe the case, computer, and python package you tested with, and what the directory `init/e3sm_coupling/prescribed_ismf` contained when you were done?
I tried on grizzly on a compute node using `QU240wISC` and package `compass_0.1.2`. I first run the init step. Then I get this:
```
cd e3sm_coupling/
./run.py
****** prescribed_ismf ******
running: mpirun -np 36 /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/base/envs/compass_0.1.2/bin/ESMF_RegridWeightGen --source /tmp/tmpar9c01qz/src_mesh.nc --destination /tmp/tmpar9c01qz/dst_mesh.nc --weight ./map_5600.0x5600.0km_1.0km_Antarctic_stereo_to_oQU240wISC.nc --method conserve --netcdf4 --no_log --src_regional --dst_regional --ignore_unmapped
Operating system error: Cannot allocate memory
Allocation would exceed memory limit
```
But this is a brand new functionality. So if it worked for you, I'm happy to merge it and work out the kinks for this grizzly set-up later.

pwolfram:
@mark-petersen, sometimes a user takes up all the memory on the compute node.  I've had this issue before, sure is frustrating.

mark-petersen:
@pwolfram Are you thinking of a login node? I should be the only one on that compute node. There must be some problem with the size of the files it is loading.

pwolfram:
Sorry @mark-petersen , I was thinking head node -- I agree that the file might be too big.

xylar:
@mark-petersen, I believe you are running with the serial version of `compass`.  If that's the case, you will do horrible things if you try to run `ESMF_RegridWeightGen` with `mpirun`.  It might work but it will just be running 36 independent processes trying to write the same file.  You need to set `nprocs` to 1 until we can get the parallel environment working again (https://github.com/MPAS-Dev/MPAS-Model/pull/480).  This problem should be with any mapping files you're creating, not just this case, by the way.

PR #473, opened 2020-03-14
Add support for computing signed distance in COMPASS

xylar:
This is accomplished by first computing a mask of inside vs. outside a region, then computing the distance to the boundary of that region, and finally combining these to compute the signed distance.


xylar:
## Testing

Here is a test ran in the `testing_and_setup/compass/ocean` directory.

```python
#!/usr/bin/env python

import numpy
import json
from jigsaw_to_MPAS.coastal_tools import signed_distance_from_geojson
import matplotlib.pyplot as plt


with open('high_res_region.geojson') as fp:
    geojson_data = json.load(fp)
lon = numpy.arange(-180., 180., 0.25)
lat = numpy.arange(-90., 90., 0.25)


signed_distance = signed_distance_from_geojson(geojson_data, lon, lat,
                                               max_length=0.25)

plt.pcolormesh(signed_distance)
plt.colorbar()
plt.savefig('test.png', dpi=200)
```

Here is the geojson file:
https://gist.github.com/xylar/d41fee67db6609b87ee97913b53fbc0e

The resulting signed distance:
test

xylar:
## More Testing

I used the existing `compute_cell_width` function to convert the signed distance into a `tanh` transition.

```python
#!/usr/bin/env python

import numpy
import json
from jigsaw_to_MPAS.coastal_tools import signed_distance_from_geojson, \
    compute_cell_width
import matplotlib.pyplot as plt


with open('high_res_region.geojson') as fp:
    geojson_data = json.load(fp)
lon = numpy.arange(-180., 180.01, 0.25)
lat = numpy.arange(-90., 90.01, 0.25)

nlon = len(lon)
nlat = len(lat)

signed_distance = signed_distance_from_geojson(geojson_data, lon, lat,
                                               max_length=0.25)

width = 30.*numpy.ones((nlat, nlon))

trans_width = 5000e3
# this cancels out the offset included in compute_cell_width so the transition is at distance zero
trans_start = -0.5*trans_width
dx_min = 10.

width = compute_cell_width(signed_distance, width, lon, lat, dx_min,
                           trans_start, trans_width,
                           restrict_box={'include': [], 'exclude': []})

Lon, Lat = numpy.meshgrid(lon, lat)

ax = plt.subplot(111)
plt.pcolormesh(Lon, Lat, width)
plt.colorbar()
ax.set_aspect('equal')
ax.autoscale(tight=True)
plt.tight_layout()
plt.savefig('test.png', dpi=200)
```

test

xylar:
Using the new capability, my last commit is a draft of the `SO60to10wISC` with this capability.  Here's what the resolution looks like:
test

xylar:
@maltrud, could you post or send me the PowerPoint you had with the rough contour you want the resolution transition to follow?

@kristin-hoch, I'd benefit from a phone/skype/whatever call to discuss how to make the new version of the mesh as similar to the old as possible -- where to make the transition relative to the region boundary and how broad (in km) to make the transition width.

xylar:
Here are plots of `sqrt(areaCell)` (as a proxy for cell size) for `ocean/develop` and this branch:

`ocean/develop`:
```
	nCells = 728014 ;
	nEdges = 2198922 ;
	nVertices = 1470485 ;
```
ocean_develop

This branch (~20% more cells):
```
	nCells = 869633 ;
	nEdges = 2624024 ;
	nVertices = 1753967 ;
```
signed_distance

xylar:
Assuming we have a noon meeting on Wednesday, I probably won't make further revisions until we've discussed these results at that meeting.

maltrud:
@xylar can you please point me to the grid file for this?  it looks pretty good!

xylar:
On anvil:
```
/lcrc/group/acme/xylar/mpas/test_signed_distance/ocean/global_ocean/SO60to10wISC/init/culled_mesh
```
I believe you should be able to read it.

xylar:
Oops, there's a seam along the +/- 180 degree line:
`ocean/develop`:
ocean_develop_ross

This branch:
signed_distance_ross

Should be easy enough to fix.  Stay tuned...

xylar:
Latest iteration of the transition for this branch (with higher resolution Drake Passage):
high_res_drake
```
	nCells = 739904 ;
	nEdges = 2234478 ;
	nVertices = 1494156 ;
```
Only about 1.6% more cells than the `ocean/develop` version of the mesh.

And the view of the Ross Sea/Pacific sector:
high_res_drake_ross

xylar:
@mark-petersen, please let me know if the last commit doesn't look like what you expected.

xylar:
Moving target... I altered the wrong SO test case the first time but now it should be right.

PR #479, opened 2020-03-18
Add surfaceAirTemperature, basalHeatFlux to ascii mesh converter

matthewhoffman:
This adds the surfaceAirTemperature, basalHeatFlux fields to the ascii mesh converter
used to convert an MPAS input file to the ascii format that Albany can ingest.

Also pass maps of floating/marine boundaries by reference, but then comment out their usage for observational fields  - for now it seems the extension for floating/marine boundaries may not be needed.

Other minor fixes useful for the ascii mesh converter.


matthewhoffman:
Tested by confirming that two new ascii files get generated and they have appropriate values in them.  @mperego , please try it out and also go ahead and push any commits you want to this branch to fix other issues you know about (e.g. smb units).

mperego:
@matthewhoffman  Thanks! will do.

PR #482, opened 2020-03-23
Add a script for loading the compass conda env

xylar:
Many supported machines are included in this script, and more can be added as needed.

### DIFF
--- a/testing_and_setup/compass/load_compass_env.bash
+++ b/testing_and_setup/compass/load_compass_env.bash
@@ -13,7 +13,7 @@ elif [[ $HOSTNAME = "cooley"* ]]; then
   base_path="/lus/theta-fs0/projects/ccsm/acme/tools/e3sm-unified/base"
 elif [[ $HOSTNAME = "compy"* ]]; then
   base_path="/share/apps/E3SM/conda_envs/base"
-elif [[ $HOSTNAME = "gr-fe"* ]] || [[ $HOSTNAME = "wf-fe"* ]]; then
+elif [[ $HOSTNAME = "gr-fe"* ]] || [[ $HOSTNAME = "ba-fe"* ]]; then
   base_path="/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/base"
 elif [[ $HOSTNAME = "burnham" ]]; then
   base_path="/home/xylar/miniconda3/"
--- a/testing_and_setup/compass/load_compass_env.bash
+++ b/testing_and_setup/compass/load_compass_env.bash
@@ -1,7 +1,5 @@
 version=0.1.2
 
-default_python=3.7
-
 # The rest of the script should not need to be modified
 if [[ $HOSTNAME = "cori"* ]] || [[ $HOSTNAME = "dtn"* ]]; then
   base_path="/global/cfs/cdirs/acme/software/anaconda_envs/base"
--- a/testing_and_setup/compass/load_compass_env.bash
+++ b/testing_and_setup/compass/load_compass_env.bash
@@ -1,0 +1,36 @@
+version=0.1.2
+
+default_python=3.7
+
+# The rest of the script should not need to be modified
+if [[ $HOSTNAME = "cori"* ]] || [[ $HOSTNAME = "dtn"* ]]; then
+  base_path="/global/cfs/cdirs/acme/software/anaconda_envs/base"
+elif [[ $HOSTNAME = "acme1"* ]] || [[ $HOSTNAME = "aims4"* ]]; then
+  base_path="/usr/local/e3sm_unified/envs/base"
+elif [[ $HOSTNAME = "blueslogin"* ]]; then
+  base_path="/lcrc/soft/climate/e3sm-unified/base"
+elif [[ $HOSTNAME = "rhea"* ]]; then
+  base_path="/ccs/proj/cli900/sw/rhea/e3sm-unified/base"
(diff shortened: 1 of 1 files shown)


xylar:
We would need to update the version of the `compass` env that gets loaded here each time the requirements for COMPASS change.  While we should keep `README_ocean.md` in sync, presumably this new script would serve as the "official" indication of which `compass` env is compatible with the current code.

testing_and_setup/compass/load_compass_env.bash
```diff
@@ -0,0 +1,36 @@
...
+  base_path="/lcrc/soft/climate/e3sm-unified/base"
+elif [[ $HOSTNAME = "rhea"* ]]; then
+  base_path="/ccs/proj/cli900/sw/rhea/e3sm-unified/base"
+elif [[ $HOSTNAME = "cooley"* ]]; then
+  base_path="/lus/theta-fs0/projects/ccsm/acme/tools/e3sm-unified/base"
+elif [[ $HOSTNAME = "compy"* ]]; then
+  base_path="/share/apps/E3SM/conda_envs/base"
+elif [[ $HOSTNAME = "gr-fe"* ]] || [[ $HOSTNAME = "wf-fe"* ]]; then
+  base_path="/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/base"
+elif [[ $HOSTNAME = "burnham" ]]; then
```
pwolfram: Wouldn't it make more sense here to have this customized for individuals via sourcing a dot file?

xylar: This was just me testing the script on my personal laptop.  I'm happy to take it out but I figured it did no harm...

xylar: And, yes, no one on their personal laptop should bother using this script.  They should just activate their env in the usual way.

testing_and_setup/compass/load_compass_env.bash
```diff
@@ -0,0 +1,36 @@
+version=0.1.2
+
+default_python=3.7
```
sbrus89: This doesn't look like it's used in the script right now. This must be needed as an environment variable?

testing_and_setup/compass/load_compass_env.bash
pwolfram: Sounds good

testing_and_setup/compass/load_compass_env.bash
```diff
@@ -0,0 +1,36 @@
...
+  base_path="/usr/local/e3sm_unified/envs/base"
+elif [[ $HOSTNAME = "blueslogin"* ]]; then
+  base_path="/lcrc/soft/climate/e3sm-unified/base"
+elif [[ $HOSTNAME = "rhea"* ]]; then
+  base_path="/ccs/proj/cli900/sw/rhea/e3sm-unified/base"
+elif [[ $HOSTNAME = "cooley"* ]]; then
+  base_path="/lus/theta-fs0/projects/ccsm/acme/tools/e3sm-unified/base"
+elif [[ $HOSTNAME = "compy"* ]]; then
+  base_path="/share/apps/E3SM/conda_envs/base"
+elif [[ $HOSTNAME = "gr-fe"* ]] || [[ $HOSTNAME = "wf-fe"* ]]; then
```
sbrus89: Wolf can be left out since it's decommissioned, as far as I know. Should Badger be included?

testing_and_setup/compass/load_compass_env.bash
xylar: Good call, that's an artifact from the script I adapted this from.

testing_and_setup/compass/load_compass_env.bash
xylar: Yep, absolutely (another artifact from a previous script).

xylar:
> We could consider automatically linking this script to the work_dir or in the case (or test) directory so that the conda environment can be easily loaded before running a case.

I'm trying to decide how I feel about that.  COMPASS should ideally work fine on unsupported machines where `load_compass_env.sh` won't work.  In such cases, it would be potentially distracting to have it.  Also, should it be in every step of every test case?

Anyway, happy to keep discussing it but maybe premature to add that in this PR until we have a sense of how we would use it.

xylar:
@sbrus89, let me know if you notice any other issues.

sbrus89:
@xylar, I don't see any additional issues. You took care of everything with the last set of commits.

sbrus89:
@xylar, I see your point about non-supported machines. I agree that something to address that point doesn't belong in this PR.

sbrus89:
I just brought it up because I find myself reverse searching my history a lot to find the right `source` command to load the conda environment (which isn't a huge deal). Making the file available somewhere in the `work_dir` would make it easier to load on supported machines.

Maybe the best answer is for me to make an alias.

xylar:
> Maybe the best answer is for me to make an alias.

That wouldn't work well for me at least because the environment to load might be different for different `MPAS-Model` branches (therefore different directories for the script to source).  So I see the value of an "idiot-proof" (no offense to anyone, I would benefit from this, too) version where your test case tells you which env to load.

I generally find myself wanting to put a job script in the case directory (one up from the individual steps) so this would be an appropriate place to put a link to the `load_compass_env.sh` script.  I guess if someone tries to source it and it doesn't work, they will get a message telling them why...  Easy enough to add in a future PR.

PR #483, opened 2020-03-23
Fix bug in pointwiseStats locations for Sandy cases

sbrus89:
Add missing +1 offset to cell IDs used in the points.nc file generated in the init step of the hurricane Sandy cases.


pwolfram:
Thanks @sbrus89 !

sbrus89:
@pwolfram, thanks for the quick merge!

PR #487, opened 2020-03-24
Update maint1.2: 3D varying GM

mark-petersen:
This update adds the following PRs from ocean/develop:
- Adds options for 3d varying GM bolus and 2d varying phase speed #288
- Add GM bolus eddy stats #339
- Fix threading issue in MPAS-O GM routine #376
- compute landIceMask using geometric_features #447
- bug fixes for the nonlocal source term in KPP #305

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -375,6 +375,22 @@
 					description="If true, the standard GM for the tracer advection and mixing is turned on. This includes the redi portion which is captured in hmix."
 					possible_values=".true. or .false."
 		/>
+		<nml_option name="config_gm_lat_variable_c2" type="logical" default_value=".false." units="NA"
+					description="If true, spatially variable phase speed is used Following Ferrari et al (2010)"
+					possible_values=".true. or .false."
+		/>
+		<nml_option name="config_gm_kappa_lat_depth_variable" type="logical" default_value=".false." units="NA"
+					description="If true, spatitally and deph varying Bolus Kappa is used, following Danabasoglu et al (2007)"
+					possible_values=".true. or .false."
+		/>
+		<nml_option name="config_gm_min_stratification_ratio" type="real" default_value="0.1" units="NA"
+					description="minimum value for N2/Nmax2 ratio"
+					possible_values="small positive reals"
+		/>
+		<nml_option name="config_gm_min_phase_speed" type="real" default_value="0.1" units="m s^{-1}"
+					description="minimum allowed phase speed for spatially variable computation"
+					possible_values="small positive reals"
+		/>
 		<nml_option name="config_use_Redi_surface_layer_tapering" type="logical" default_value=".false." units="NA"
 					description="If true, the Redi K33 vertical mixing is limited in and just below the ocean boundary layer."
 					possible_values=".true. or .false."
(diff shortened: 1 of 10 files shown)


mark-petersen:
I used cherry-pick on these commits, and there were no conflicts. Compiles with gnu and intel.

Are there any other PRs needed from `ocean/develop`?

src/core_ocean/Registry.xml
```diff
@@ -375,6 +375,22 @@
 					description="If true, the standard GM for the tracer advection and mixing is turned on. This includes the redi portion which is captured in hmix."
 					possible_values=".true. or .false."
 		/>
+		<nml_option name="config_gm_lat_variable_c2" type="logical" default_value=".false." units="NA"
```
mark-petersen: @jonbob note there are 6 new GM flags added to maint-1.2 with this update.

darincomeau:
@mark-petersen I think we need https://github.com/MPAS-Dev/MPAS-Model/pull/392 to prevent build errors when using python 3.

mark-petersen:
@darincomeau thanks, I just added that.

@xylar can you think of anything else we might need for maint-1.2?

xylar:
@mark-petersen, yes, I believe we need: https://github.com/MPAS-Dev/MPAS-Model/pull/447

This disables computing the land-ice mask in forward mode, which I think is necessary with the new initial condition we're about to set up.

xylar:
Actually, given the complexity of COMPASS changes in #447, you might want to just cherry-pick https://github.com/xylar/MPAS-Model/commit/da1f7c959dee9b1c43046d0ebebe793c09b81575.

jonbob:
@mark-petersen , @darincomeau - what about changing ocean output to single precision in this? It would help with run storage...

mark-petersen:
Done, thanks. I'll merge this in and set up an E3SM PR.

mark-petersen:
> @mark-petersen , @darincomeau - what about changing ocean output to single precision in this? It would help with run storage...

@jonbob good idea, I was thinking the same thing. I can add that on the E3SM PR in a minute.

jonbob:
@mark-petersen - I was just looking at PR's from the last year or so. Do any of these make sense to include:
[2948](https://github.com/E3SM-Project/E3SM/pull/2948)
[3056](https://github.com/E3SM-Project/E3SM/pull/3056)
I can't remember what updates to master we actually also merged into maint-1.2, but it doesn't seem like many...

mark-petersen:
@xylar or @matthewhoffman, it looks like we should include
https://github.com/E3SM-Project/E3SM/pull/2948
on the E3SM update for the cryo branch (maint-1.2). Do you agree?

darincomeau:
> @matthewhoffman will have a better sense than me. It seems like including this would do no harm. It only matters if we set `config_land_ice_flux_mode = 'standalone'`. Is that how we run typically in E3SM? I guess maybe so. (I should know this!)

Yes, this is the namelist option used in ISMF compsets.

mark-petersen:
> @mark-petersen - I was just looking at PR's from the last year or so. Do any of these make sense to include:
> https://github.com/E3SM-Project/E3SM/pull/3056

@jonbob thanks for the suggestion. I added the KPP nonlocal fix, which is that 3056 (MPAS PR https://github.com/MPAS-Dev/MPAS-Model/pull/305)

jonbob:
@mark-petersen - just let me know when you think you're done and we can start the corresponding E3SM PR

src/core_ocean/Registry.xml
jonbob: @mark-petersen - I'm only seeing 4? Aren't the others just new arrays?

mark-petersen: Yes, just 4. My mistake.

PR #489, opened 2020-03-24
Smooth topography before interp to MPAS

xylar:
This merge adds a python script for iteratively smoothing topograpthy (from for now, the dataset combining BedMachine and GEBCO) proportional to the resolution of the MPAS mesh, as determined in the base_mesh step.

Local weights for the iterative smoothing are determined based on the resolution of the lon/lat grid, the MPAS mesh resolution, the number of iterations of smoothing and a constant multiplicative factor `smoothing` that can be increased or decreased to provide more or less smoothing overall.  The default value, `smoothing = 1.0`, smooths the topography by one MPAS cell width.

Increasing the number of iterations will typically not increase the amount of smoothing, since less smoothing is performed at each iteration.  The exception is that the maximum amount of smoothing is an unweighted nearest-neighbor averaging at each iteration.  If a region (commonly the poles) is so stretched on the lon/lat grid that an MPAS cell covers more grid cells than the number of smoothing iterations, it is not possible to smooth over the full width of an MPAS grid cell.  In practice, this is not likely to be an important issue.

The cost of this approach is significant, taking several minutes even for coarse MPAS meshes (it is performed on the topographic grid, so is largely agnostic to the MPAS resolution).  We may
need to make the QU240wISC test case a special case where we either skip this process or simplify it in some way to reduce runtime of regression testing.

Currently, the approach is only being used for test cases with ice-shelf cavities (the same ones using BedMachine/GEBCO bathymetry).  It could be expanded to other test cases if it proves useful.

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/config_files_ISC/config_initial_state.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/config_files_ISC/config_initial_state.xml
@@ -29,12 +29,14 @@
 	<add_link source="../culled_mesh/culled_graph.info" dest="graph.info"/>
 	<add_link source="../culled_mesh/critical_passages_mask_final.nc" dest="critical_passages.nc"/>
 	<add_link source="../culled_mesh/land_ice_mask.nc" dest="land_ice_mask.nc"/>
+	<add_link source="../base_mesh/cellWidthVsLatLon.nc" dest="cellWidthVsLatLon.nc"/>
 	<add_link source_path="initial_condition_database" source="layer_depth.80Layer.180619.nc" dest="layer_depth.nc"/>
 	<add_link source_path="initial_condition_database" source="PTemp.Jan_p3.filled.mpas100levs.160127.nc" dest="temperature.nc"/>
 	<add_link source_path="initial_condition_database" source="Salt.Jan_p3.noBlackCaspian.filled.mpas100levs.160127.nc" dest="salinity.nc"/>
 	<add_link source_path="initial_condition_database" source="windStress.ncep_1958-2000avg.interp3600x2431.151106.nc" dest="wind_stress.nc"/>
 	<add_link source_path="bathymetry_database" source="BedMachineAntarctica_and_GEBCO_2019_0.05_degree.200128.nc" dest="topography.nc"/>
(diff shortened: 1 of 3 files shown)


xylar:
@mark-petersen, this is the work I mentioned today where I would smooth bathymetry *before* resampling to the MPAS mesh rather than after.  It seems to be working well so far.

I'm putting this here as a placeholder for now, not because it's ready to review or merge.

@jonbob, I might get your help doing another round of E3SM testing with a new EC60to30wISC mesh once we're a little further along (e.g. we have 64 layers as an option).

xylar:
To determine the relationship between nearest-neighbor weights, number of iterations and the width (in cells) of the resulting smoothing, I used this script:
https://gist.github.com/xylar/ffe3317daa4203b7fe58397381969770
The finding was that interpolation weights can be determined from the width over which to smooth and the number of iterations by
``` python
A = 2.65
p_iter = 0.5
p_width = 2.25
weight = (width/(A*iterations**p_iter))**p_width
```
This formula is used in the script `smooth_topo.py` in COMPASS.

testing_and_setup/compass/ocean/global_ocean/config_files_ISC/config_initial_state.xml
```diff
@@ -53,6 +56,12 @@
 	</streams>
 
 	<run_script name="run.py">
+		<step executable="./smooth_topo.py">
+			<argument flag="--iter">100</argument>
+			<argument flag="--smoothing">1.0</argument>
+			<argument flag="--mpi_tasks">4</argument>
```
xylar: It's annoying that the number of MPI tasks is repeatedly hard coded and not part of the config file or some kind of machine config file.  We need to think if there's a better way to do this.  This is just one example of where the problem arises.

xylar:
@prajvala00123, take a look at this when you get a chance.  It's probably not the easiest entry point for learning about setting up a new test case for MPAS-Ocean in E3SM, but it will eventually be helpful if you are interested in playing with bathymetric smoothing.

xylar:
Needs to be rebased once the next set of COMPASS merges have happened.

xylar:
This branch has been moved here: https://github.com/xylar/compass/tree/add_better_topo_smoothing

A PR will be made to https://github.com/MPAS-Dev/compass when the time is right.

PR #480, opened 2020-03-20
Fixes to COMPASS to support conda MPI

xylar:
This merge converts several script calls to function calls, which seems to work more reliably with conda MPI:
* The paraview extractor can now be called as a function rather than a script, and this is done during base-mesh generation and culling
* SCRIP files are now created with a function call

With these changes, calls to python scripts that use NetCDF in the parallel conda enviornment will now work as long as they are called with `mpirun -np 1`

This merge also adds support for a `conda_mpi` attribute to `step` tags in COMPASS XML files. 
If this is set to `true` or `false`, the step will have `mpirun -np 1` prepended to the executable in conda environments with MPI support.  If no `conda_mpi` attribute is specified, `mpirun -np 1` is prepended only to python scripts (calls starting with `python` or ending with `.py`).  

This is needed to support compass conda environments with MPI. Python scripts and modules that use the `netcdf4` package with mpich support will work property if they are called with `mpirun`.

Changes are made to `setup_testcase.py` in b1a2b3b, so this commit should be merged to `develop` in a separate PR.

PR #492, opened 2020-03-26
Optionally add links to load_compass_env.sh in test cases

### LINKED ISSUE #490: link in test cases to load_compass_env.bash
@sbrus89 requested a way to link to [load_compass_env.bash](https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/load_compass_env.bash) within a test case so this file can be sourced more easily to run the test case with the right environment.

xylar:
This is specified either in the config file or at the command line.

Like #480, this involves changes to common COMPASS infrastructure and we should consider making a separate PR to `develop` instead of merging those changes to `ocean/develop`.

closes #490


xylar:
@sbrus89, thanks for testing!  Yes, you're right, this will only work with `compass_0.1.3`, which is waiting for https://github.com/MPAS-Dev/MPAS-Model/pull/480.

Also, the other file edits you mention are from https://github.com/MPAS-Dev/MPAS-Model/pull/480, which this PR is based off of.  Sorry for not mentioning that before.

xylar:
@mark-petersen, this has been rebased onto #480 and should also now be ready for testing and merge.

PR #493, opened 2020-03-26
Modified mesh rescale code

akturner:
Added option to rescale and rescaling radius to configs
Removed check for inconsistent radius of input


mgduda:

PR #494, opened 2020-03-26
fixed incorrect path for ocean cvmix after submodule change

### LINKED ISSUE #488: PGI build of ocean on summit fails with bad dependency
Build of latest ocean/develop is failing on Summit with PGI19.9, claiming a weird dependency:
PGF90-F-0004-Unable to open MODULE file cvmix_background.mod (mpas_subdriver.F: 14) 
where no such dependency should exist. While this might appear to be a compiler issue, a build of ocean/develop from ~week ago builds fine with this same compiler. 

Trying to narrow down further...

philipwjones:
When changing to submodules for the ocean cvmix package, the path to cvmix source had changed, but one instance of this path was not changed, leading to build failures with PGI on Summit. This PR fixes that bad path.

Fixes #488

### DIFF
--- a/src/core_ocean/build_options.mk
+++ b/src/core_ocean/build_options.mk
@@ -6,7 +6,7 @@ NAMELIST_SUFFIX=ocean
 FCINCLUDES += -I$(ROOT_DIR)/core_ocean/driver 
 FCINCLUDES += -I$(ROOT_DIR)/core_ocean/mode_forward -I$(ROOT_DIR)/core_ocean/mode_analysis -I$(ROOT_DIR)/core_ocean/mode_init
 FCINCLUDES += -I$(ROOT_DIR)/core_ocean/shared -I$(ROOT_DIR)/core_ocean/analysis_members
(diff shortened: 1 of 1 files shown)


mark-petersen:
@philipwjones I ran a test compile in a fresh directory, and this compiles on grizzly with both gnu and intel, both with make (not cmake). But I don't understand how gnu and intel worked before with this error. Don't they use that line too?

Also, is `build_options.mk` used by both the make and cmake systems? Which one were you testing with?

I'm merging in a bunch of COMPASS changes right now. I'll put this one on right after that. It's best to give in a single E3SM update, since we are changing a make file.

philipwjones:
@mark-petersen the CMake system has its own definitions and the correct paths are in ocean.cmake and in the actual Makefile in core_ocean. Not sure why Intel, gnu don't hit this, but I suspect, since the error occurs in the forward mode driver, that PGI treats nested "use" statements differently and tries to find this nested dependency while the others might wait and resolve it at the link stage?

PR #496, opened 2020-03-30
New ocean mesh structure with GPU replication

philipwjones:
This change adds a mesh structure with all current mesh pool data and replicates that data on a GPU (or other accelerator) for the duration of the simulation to reduce the need for data transfers of static mesh data. It is a critical piece of infrastructure needed for future GPU modifications being staged. In a later change, this structure was replaced by public module variables instead since a large user-defined type did not perform well on GPUs. Mesh variables can now be accessed just by "use"-ing this module.

This module also has a revised index naming in which nCells, nCellsSolve and nCellsArray are replaced by nCellsAll, nCellsOwned, and nCellsHalo(n) where the latter differs from nCellsArray by only including the index of each halo depth and no longer includes nCellsOwned as the first entry. Similar changes were introduced for Edges, Vertices. 

NOTE: this change requires only one block per MPI task and will abort if more than one block per MPI task is attempted. Any multi-block tests will fail.

This has been tested on summit in a QU240 test case and is bit-for-bit since the structure itself is not yet being used anywhere. However, I have verified the data is correctly transferred to the GPU accelerator on Summit.

[b4b]

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_mesh.F
+++ b/src/core_ocean/shared/mpas_ocn_mesh.F
@@ -13,13 +13,13 @@
 !  ocn_mesh
 !
 !>  \brief MPAS ocean mesh structure with GPU support
-!!  \author Rob Aulwes and Phil Jones
-!!  \date   14 Jan 2020
-!!  \details
-!!  This module creates and maintains a primary ocean mesh structure
-!!  and ensures all mesh variables are copied to an accelerator device
-!!  if needed. Currently it consists of pointers to the existing MPAS mesh pool
-!!  variables, but is intended to eventually replace the mesh pool later.
+!> \author Rob Aulwes and Phil Jones
+!> \date   14 Jan 2020
+!> \details
+!> This module creates and maintains a primary ocean mesh structure
+!> and ensures all mesh variables are copied to an accelerator device
+!> if needed. Currently it consists of pointers to the existing MPAS mesh pool
+!> variables, but is intended to eventually replace the mesh pool later.
 !
 !-------------------------------------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -0,0 +1,842 @@
...
+      block => domain % blocklist
+      do while ( associated(block) )
+
+         !*** we now only support one block per task
+         blockCount = blockCount + 1
+         if (blockCount > 1) then
+            call mpas_log_write(&
+               'ocn_meshCreate: more than one block no longer supported', &
+                MPAS_LOG_CRIT)
+         endif
```
mattdturner: If this only supports 1 block per task, why is there the `do while` loop?  Is that the only way to get the number of blocks to confirm that there is only 1 block?

src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -0,0 +1,842 @@
...
+!!  and ensures all mesh variables are copied to an accelerator device
+!!  if needed.
+!
+!-------------------------------------------------------------------------------
+
+   subroutine ocn_meshCreate(domain)
+
+      ! Input arguments
+
+      type(domain_type) :: &
```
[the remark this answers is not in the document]
philipwjones: I had intent there originally but the compiler complained, maybe because intent attributes for pointers is a F03 feature and may not be fully supported? Not sure, but had to remove it for a successful build.

src/core_ocean/shared/mpas_ocn_mesh.F
philipwjones: You guessed it. The only way to get a block count (that I know of anyway).

src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -0,0 +1,842 @@
...
+         do k=1,ocnMesh%nVertLevels
+            ocnMesh%vertexMask(k,n)       = real(vertexMaskTmp(k,n),     RKIND)
+            ocnMesh%edgeSignOnVertex(k,n) = &
+                                          real(edgeSignOnVertexTmp(k,n), RKIND)
+            ocnMesh%boundaryVertex(k,n)   = real(boundaryVertexTmp(k,n), RKIND)
+         end do
+         end do
+
+         call mpas_log_write('In meshCreate copy to device')
+#ifdef MPAS_OPENACC
```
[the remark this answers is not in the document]
philipwjones: Normally, I put these data transfers inside an ifdef because we also have an openmp master to make sure we only transfer from one thread in a mixed openmp/openacc setup. But there's an openmp master around most of this routine, so you are correct this can be removed.

src/core_ocean/shared/mpas_ocn_mesh.F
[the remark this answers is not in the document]
mattdturner: Ah, gotcha.

src/core_ocean/shared/mpas_ocn_mesh.F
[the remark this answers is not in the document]
philipwjones: Just pushed a new version that removed this redundant ifdef

src/core_ocean/shared/mpas_ocn_mesh.F
mattdturner: Well that's a bummer.  Although if we are getting rid of blocks soon then it won't be an issue for much longer.

mattdturner:
Approved per visual inspection and @philipwjones testing.

philipwjones:
While resolving the redundant OpenACC ifdef, found some debug output that hadn't been removed and realized I hadn't added this file to the Cmake build, so just committed/pushed those mods while also removing the redundant ifdef

src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -0,0 +1,829 @@
...
+      !***
+      !*** end of preamble, begin code
+      !***
+
+      err = 0 
+
+      !$omp barrier
+      !$omp master
+
+#ifdef MPAS_OPENACC
```
mattdturner, line 603: Another `#ifdef MPAS_OPENACC`.

philipwjones: Thanks - removed in latest version

philipwjones:
Pushed a bug fix for bad dimensions on two edgeSign arrays

mattdturner:
Should there be a `nCellsSolve` variable added to the mesh type?  Some routines pull the `nCellsSolve` variable from `meshPool`.  (e.g., https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/core_ocean/shared/mpas_ocn_tracer_advection_std.F#L95)

`      call mpas_pool_get_dimension(meshPool, 'nCellsSolve', nCellsSolve)`

philipwjones:
Thanks @mattdturner  I also caught another one (nVerticesArray) and there are probably some more. I caught everything that I found documented in Registry and the struct include file, but there are clearly a number of fields in the dimension subpool that I didn't pick up. This stuff is so opaque, I can't find a place with a definitive list so might have to resort to a massive grep...

mattdturner:
> This stuff is so opaque, I can't find a place with a definitive list so might have to resort to a massive grep...

As we work through updating the files for the GPU-related work, any variable that was missed will make itself known.

philipwjones:
Added the relevant missing variables discussed above. However, I have renamed/refactored them a bit. So instead of nCells, nCellsSolve and nCellsArray, the new structure has nCellsAll, nCellsOwned, and nCellsHalo(n) where nCellsHalo refers only to the halo levels, i.e. nCellsHalo(1) has the number of owned+halo cells in the first halo level. The old nCellsArray had the first entry as nCellsOwned/Solve so this is a conceptual change. Similar constructs for Edges and Vertices have been introduced. The old names/structures still exist in the old mesh pool.

src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -851,30 +851,413 @@ end subroutine ocn_meshDestroy
...
+      ! convert previous index limits into new arrangement
+      ocnMesh%nCellsAll       = nCellsTmp
+      ocnMesh%nEdgesAll       = nEdgesTmp
+      ocnMesh%nVerticesAll    = nVerticesTmp
+
+      n = size(nEdgesArrayTmp)
+      ocnMesh%nCellsOwned = nCellsArrayTmp(1)
+      do i=2,n
+         ocnMesh%nCellsHalo(i-1) = nCellsArrayTmp(i) 
+      end do
```
mattdturner, line 964: After the most recent commit, when running the executable that was compiled with `DEBUG=true`, I get a segfault with the following traceback:
```
forrtl: severe (408): fort: (2): Subscript #1 of the array NCELLSARRAYTMP has value 5 which is greater than the upper bound of 4

ocean_model        0000000003332286  Unknown               Unknown  Unknown
ocean_model        0000000001EEB7BF  ocn_mesh_mp_ocn_m         963  mpas_ocn_mesh.F
ocean_model        0000000001B64939  ocn_forward_mode_         362  mpas_ocn_forward_mode.F
ocean_model        0000000001B5FB8B  ocn_core_mp_ocn_c          76  mpas_ocn_core.F
ocean_model        000000000041DE9F  mpas_subdriver_mp         331  mpas_subdriver.F
ocean_model        00000000004192E2  MAIN__                     14  mpas.F
ocean_model        0000000000419252  Unknown               Unknown  Unknown
libc-2.26.so       00002AAAADD4234A  __libc_start_main     Unknown  Unknown
ocean_model        000000000041916A  Unknown               Unknown  Unknown
```

I think line 960 should be `n = size(nCellsArrayTmp)` instead of `n = size(nEdgesArraytmp)`

mattdturner: Making this change allowed me to run to completion with `DEBUG=true`

philipwjones: Thanks. Yeah a cut-and-paste error. Fixed and pushed the change.

philipwjones:
Just committed a new version that eliminates the structure in favor of public mesh variables that are accessed with a module use. The prior structure did not perform well on GPU due to the need to traverse such a large structure to find variables. I also squashed some of the prior commits for bug fixes so that this is a little cleaner. I will update the PR text above to better reflect these latest changes.

src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -0,0 +1,1187 @@
...
+!-------------------------------------------------------------------------------
+
+   subroutine ocn_meshDestroy(err)
+
+      ! Input variables
+
+      ! Since the ocnMesh is currently a public module variable, no inputs 
+      ! here, but eventually may want to treat ocnMesh as a specific 
+      ! instantiation instead and pass via args everywhere. If so, need an
+      ! input mesh here
```
mattdturner, line 625: This still references the ocnMesh derived type, which has been removed.

philipwjones:
These reference arrays weren't used that often since they only refer to the reference values and not the time-variable depths. But happy to include them if needed.

mattdturner:
I'm not sure if its totally necessary.  I just noticed that the `verticalMeshPool` is passed as an argument to `ocn_ALE_thickness` and wasn't sure if those variables not being included in `ocn_mesh` was an oversight.

mark-petersen:
Rebased, passes nightly regression suite with gnu and intel, debug and optimized. bfb with previous, except block tests with 2 blocks/core fail, as expected.

src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -0,0 +1,1187 @@
...
+   private
+
+   !----------------------------------------------------------------------------
+   !
+   ! Public parameters
+   !
+   !----------------------------------------------------------------------------
+
+   integer, public :: &! mesh, array sizes
+      nCellsAll,      &! total number of local (owned+halo) cells in primary
```
mark-petersen, line 43: I really like the variables `nCellsAll, nCellsOwned, nCellsHalo(n)`. @philipwjones thanks for thinking of it. I myself can never remember which direction `nCellsArray` proceeds, and `nCells` sometimes has different definitions within the same module. As you use this `ocn_mesh` module in other modules, it will be good to switch from `nCells` to these exact specifications.

philipwjones: Thanks and yeah, @mattdturner and I have been gradually replacing these as we start use-ing the module with the optimizations.

Issue #497, opened 2020-03-30
Tidal forcing does not work with `ssh_gradient` for RK4


pwolfram:
Tidal forcing currently modifies `zMid` for the RK4 implementation.  However, there is no direct feedback on pressure forcing when `ssh_gradient` is used because the tendency does not include `zMid` but instead `ssh`.

pwolfram:
cc @sbrus89

pwolfram:
Reference code from `src/core_ocean/shared/mpas_ocn_vel_pressure_grad.F`
```
160       if (config_pressure_gradient_type.eq.'ssh_gradient') then                                                             
161                                                                                                                             
162          ! pressure for sea surface height                                                                                  
163          ! - g grad ssh                                                                                                     
164                                                                                                                             
165          !$omp do schedule(runtime) private(cell1, cell2, invdcEdge, k)                                                     
166          do iEdge=1,nEdges                                                                                                  
167             cell1 = cellsOnEdge(1,iEdge)                                                                                    
168             cell2 = cellsOnEdge(2,iEdge)                                                                                    
169             invdcEdge = 1.0_RKIND / dcEdge(iEdge)                                                                           
170                                                                                                                             
171             do k=1,maxLevelEdgeTop(iEdge)                                                                                   
172                tend(k,iEdge) = tend(k,iEdge) - gravity * edgeMask(k,iEdge) * invdcEdge * ( ssh(cell2) - ssh(cell1) )        
173             end do                                                                                                          
174          end do                                                                                                             
175          !$omp end do                                                                                                       
176                                                                                                                             
177       elseif (config_pressure_gradient_type.eq.'pressure_and_zmid') then                                                    
178                                                                                                                             
179          ! pressure for generalized coordinates                                                                             
180          ! -1/density_0 (grad p_k + density g grad z_k^{mid})                                                               
181                                                                                                                             
182          !$omp do schedule(runtime) private(cell1, cell2, invdcEdge, k)                                                     
183          do iEdge=1,nEdges                                                                                                  
184             cell1 = cellsOnEdge(1,iEdge)                                                                                    
185             cell2 = cellsOnEdge(2,iEdge)                                                                                    
186             invdcEdge = 1.0_RKIND / dcEdge(iEdge)                                                                           
187                                                                                                                             
188             do k=1,maxLevelEdgeTop(iEdge)                                                                                   
[...]

Issue #458, opened 2020-02-27
Verify SAL for RK4 (wetting / drying cases)


pwolfram:
Tidal implementation of SAL for RK4 may be incorrect in #371 at https://github.com/MPAS-Dev/MPAS-Model/pull/371/files#diff-d2a9414bd0f5332537025544c6ed7ea9R357

Currently, SAL is computed as a perturbation off the layer potential:
`- config_self_attraction_and_loading_beta * zMid(k,iCell)` and this may not be correct.

Given the approximation really is only deep water as the reference below implies, maybe we should do something like this:
`- config_self_attraction_and_loading_beta * (bottomDepth > 100)*ssh(iCell)`

We wouldn't do any SAL for shallow waters with less than 100m depth.  A cleaner way to do this is to use a tanh function that goes from 0 for `bottomDepth < 100m` and to 1 by `bottomDepth ~ 1000m`.

Reference:
![image](https://user-images.githubusercontent.com/4295853/75493491-3990e980-5977-11ea-8232-bb7f7c6ce172.png)

pwolfram:
Needed for #366 final case with full SAL, which is only a minor correction to current approach.

pwolfram:
Closed by #498, thanks @sbrus89!

pwolfram:
Reopening-- unclear if SAL is working correctly based on preliminary testing and it is currently turned off in #520

PR #498, opened 2020-03-30
Fix tidal potential for RK4

sbrus89:
Adds tidal potential as a separate forcing tendency.

This generalizes tidal forcing for different `pressure_gradient` options.

Addresses #497


pwolfram:
Thanks @sbrus89!  I'll test this out for the case I was using this for earlier today and report back on what I find.

sbrus89:
@pwolfram, just so you know, I haven't even complied this yet. I just wanted get something started.

pwolfram:
Thanks @sbrus89, looking forward to testing it out now!

src/core_ocean/shared/mpas_ocn_vel_tidal_potential.F
```diff
@@ -137,15 +138,17 @@ subroutine ocn_vel_tidal_potential_tend(meshPool, forcingPool, ssh, tend, err)!{
...
          cell2 = cellsOnEdge(2,iEdge)
          invdcEdge = 1.0_RKIND / dcEdge(iEdge)
 
+         potentialGrad = gravity * invdcEdge * ( tidalPotentialEta(cell2) - tidalPotentialEta(cell1) &
+                                 + config_self_attraction_and_loading_beta*(ssh(cell2) - ssh(cell1)))
+
          do k=1,maxLevelEdgeTop(iEdge)
-            tend(k,iEdge) = tend(k,iEdge) + gravity * edgeMask(k,iEdge) * invdcEdge * ( tidalPotentialEta(cell2) - tidalPotentialEta(cell1) &
-                                                                        + config_self_attraction_and_loading_beta*(ssh(cell2) - ssh(cell1)))
+            tend(k,iEdge) = tend(k,iEdge) -  edgeMask(k,iEdge) * potentialGrad
```
pwolfram: Good  catch

src/core_ocean/shared/mpas_ocn_vel_tidal_potential.F
```diff
@@ -0,0 +1,210 @@
...
+      ! call individual init routines for each parameterization
+      !
+      !-----------------------------------------------------------------
+      logical, pointer :: config_use_tidal_potential_forcing
+
+      err = 0
+
+      call mpas_pool_get_config(ocnConfigs, 'config_use_tidal_potential_forcing', config_use_tidal_potential_forcing)
+
+      tidalPotentialOn= .true.
```
pwolfram: Should this be initialized to false?  Obviously having it set to true helps for testing but in the final version it seems like it should be false.

src/core_ocean/shared/mpas_ocn_vel_tidal_potential.F
```diff
@@ -0,0 +1,210 @@
...
+      if (.not. tidalPotentialOn) return
+
+      call mpas_pool_get_dimension(meshPool, 'nEdgesArray', nEdgesArray)
+      call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
+      call mpas_pool_get_array(meshPool, 'cellsOnEdge', cellsOnEdge)
+      call mpas_pool_get_array(meshPool, 'dcEdge', dcEdge)
+      call mpas_pool_get_array(meshPool, 'edgeMask', edgeMask)
+      call mpas_pool_get_array(forcingPool, 'tidalPotentialEta', tidalPotentialEta)
+
+      call mpas_pool_get_config(ocnConfigs, 'config_use_tidal_potential_forcing', config_use_tidal_potential_forcing)
```
pwolfram: Don't need this line

src/core_ocean/shared/mpas_ocn_vel_tidal_potential.F
sbrus89: Yes, absolutely.

src/core_ocean/shared/mpas_ocn_vel_tidal_potential.F
sbrus89: Agreed.

src/core_ocean/shared/mpas_ocn_vel_tidal_potential.F
pwolfram: Have a minus sign error here.

pwolfram: Actually, for consistency with `ssh_gradient`, make the minus at line 147.

pwolfram: Look at `src/core_ocean/shared/mpas_ocn_vel_pressure_grad.F` line 172 for contrast.

sbrus89: Thanks for catching this. It's taken care of now.

src/core_ocean/shared/Makefile
```diff
@@ -68,13 +68,14 @@ OBJS = mpas_ocn_init_routines.o \
...
+mpas_ocn_tendency.o: mpas_ocn_high_freq_thickness_hmix_del2.o mpas_ocn_tracer_surface_restoring.o mpas_ocn_thick_surface_flux.o mpas_ocn_tracer_short_wave_absorption.o mpas_ocn_tracer_advection.o mpas_ocn_tracer_hmix.o mpas_ocn_tracer_nonlocalflux.o mpas_ocn_surface_bulk_forcing.o mpas_ocn_surface_land_ice_fluxes.o mpas_ocn_tracer_surface_flux_to_tend.o mpas_ocn_tracer_interior_restoring.o mpas_ocn_tracer_exponential_decay.o mpas_ocn_tracer_ideal_age.o mpas_ocn_tracer_TTD.o mpas_ocn_vmix.o mpas_ocn_constants.o mpas_ocn_config.o mpas_ocn_frazil_forcing.o mpas_ocn_tidal_forcing.o mpas_ocn_tracer_ecosys.o mpas_ocn_tracer_DMS.o mpas_ocn_tracer_MacroMolecules.o mpas_ocn_diagnostics.o mpas_ocn_wetting_drying.o mpas_ocn_tidal_potential_forcing.o mpas_ocn_vel_tidal_potential.o
```
pwolfram: `mpas_ocn_config.o` causes this to not build on intel for me, but removing this string fixes the problem.

sbrus89: You may need to rebase on `ocean/coastal` it looks like @mark-petersen added `mpas_ocn_config.F` 2019-12-06

pwolfram:
Previously, tidal signal was not being picked up when `ssh_gradient` was used.  This fixes that issue and tidal forcing is now activated when `ssh_gradient` is used.

sbrus89:
As discussed with @pwolfram, it may make sense to merge `mpas_ocn_tidal_potential_forcing.F` into `mpas_ocn_vel_tidal_potential.F`.

sbrus89:
Also, this address #458.

pwolfram:
# Testing with new USDEQU240at120cr20rr5 case 
 * no wetting /drying
 * `config_pressure_gradient_type = 'ssh_gradient'  `
 * `config_dt = '00:01:00'`
 * `config_time_integrator = 'RK4'`
 * `config_mom_del4 = 4.0e8 `
 * `config_self_attraction_and_loading_beta = 0.09`


image

Results are reasonable with the exception of the weird resonance near 10-23

pwolfram:
@xylar and @mark-petersen, this is ready for your review, especially related to @sbrus89 's question about 

Merge `mpas_ocn_tidal_potential_forcing.F` into `mpas_ocn_vel_tidal_potential.F.`?

My take is it is probably best to do this but would be good to get additional feedback on this issue.

sbrus89:
I think it makes sense to do this, but definitely want to see what @xylar and @mark-petersen think. Just for context, right now:

 `mpas_ocn_tidal_potential_forcing.F` contains all the routines to initialize the astronomical factors and compute the equilibrium potential forcing.

`mpas_ocn_vel_tidal_potential.F` computes the gradient of the equilibrium potential and does the tendency update.

Both have separate `init` routines, which is redundant.

pwolfram:
Same as above but with `config_pressure_gradient_type = 'pressure_and_zmid'`

image

sbrus89:
@pwolfram, are these with or without wetting/drying?

xylar:
I don't see any reason you can't just combine those two files and avoid some redundancy. I have done something similar with land ice fluxes because momentum and tracer fluxes are interrelated.

pwolfram:
@sbrus89, can you also make `ssh_gradient` the default pressure for the hurricane cases?  Please let me know when you are ready for me to review.  I think it safe to merge the two files before @mark-petersen, @xylar and myself do the formal review prior to merge.

pwolfram:
@mark-petersen and @xylar, @sbrus89 is running some final tests but this is now a good time for you both to do a code review prior to our merge.

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -336,31 +336,6 @@ subroutine ocn_tend_vel(tendPool, statePool, forcingPool, diagnosticsPool, meshP
...
-        do iCell = 1, nCells
-          do k = 1, maxLevelCell(iCell)
-            tidalPotentialZMid(k,iCell) = zMid(k,iCell) - tidalPotentialEta(iCell) &
-                                        - config_self_attraction_and_loading_beta * zMid(k,iCell)
-          end do 
-        end do
-
-        call mpas_pool_get_array(forcingPool, 'tidalPotentialZMid', zMid)
-      end if
-
```
pwolfram, line 363: Glad to have this removed!

testing_and_setup/compass/ocean/hurricane/forward_template.xml
```diff
@@ -30,6 +30,8 @@
                 <option name="config_tidal_potential_reference_time">'2012-10-10_00:00:00'</option>
                 <option name="config_tidal_potential_ramp">10.0</option>
```
pwolfram, line 32: I don't think we need this white space here in this file.  Can you please remove?

sbrus89: I was trying to keep different namelist groups separated because I think it's more readable. However, if you disagree, I'll change it.

pwolfram: You are right, best to leave as-is.

sbrus89:
Comparison between split explict and RK4 with `ssh_gradient`

sbrus89:
Comparison between previous split explicit (with `pressure_and_zmid`), current split explicit (with `pressure_and_zmid`), and current split explicit (with `ssh_gradient`)

sbrus89:
@pwolfram, I think everything checks out here. The SE and RK4 results with `ssh_gradient` are nearly identical, and the SE with `pressure_and_zmid` matches the previous results. 

Interestingly for this case, it's a mixed bag on whether `pressure_and_zmid` or `ssh_gradient` gives better results. Some stations are improved, others are worse.

pwolfram:
Thank you all for the hard work!

PR #495, opened 2020-03-30
Add script to compute vertical coordinate for initial condition

mark-petersen:
Previously, the vertical coordinate (layer depths) was read from a file when generating initial conditions in COMPASS. This adds a script with input parameters and adds it to the COMPASS init sequence. This will be used to create a 64-layer domain for the EC60to30 and others right now.

PR #502, opened 2020-04-02
Fix 60-layer initial state for EC60to30wISC

xylar:
This step inadvertently got replaced with the non-ice-shelf-cavity version in #495.


xylar:
@mark-petersen, you removed the local version of `config_initial_state.xml` in `EC60to30wISC` and linked to the 60-layer version that's only appropriate for runs without ice-shelf cavities.  That's not going to work, as I found out while doing a spin-up.  I'll test with this branch but need you to merge this to `ocean/develop` so we can be sure I'm not spinning up from a branch that we will throw away or a commit that will get lost.

mark-petersen:
I woke up and saw the same thing on my test. Running this one quick...

xylar:
@mark-petersen, you were a little too fast.  It doesn't work and I'm going to have to do another PR.

PR #506, opened 2020-04-03
Clean up e3sm_coupling script and config

xylar:
Now that we have a little breathing room after getting EC60to30 initial conditions in place, I'd like to clean up the E3SM coupling step and a few other small things:
* move `cull_cells.py` into `ocean/global_ocean/scripts`, where it seems to belong
* fix some incorrect tabs in an XML file
* switch `init.nc` to `mesh.nc` to distinguish the "mesh" from the "initial condition" during the E3SM coupling step.


testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -307,195 +334,158 @@ def mapping_analysis(config, mesh_name, date_string, ice_shelf_cavities):
...
                     '--source', atm_scrip_file,
                     '--destination', ocn_scrip_file,
-                    '--weight', 'map_' + atm_scrip_tag + '_TO_' + mesh_name + nomaskStr + '_'
-                       + short + '.' + date_string + '.nc',
+                    '--weight', mapping_file,
                     '--ignore_unmapped']
             run_command(args)
-
-    except OSError:
-        print('mapping must be run on one compute node')
```
xylar, line 391: This `try` / `except` around a do loop is something I would advise against.  I appreciate that there was a meaningful exception name `OSError` here.  But it would be fine to just let the `OSError` happen I think.  I could imagine various sources of `OSErrors` here, only some of which would relate to not being on a compute node.  It also is hard to figure out what went wrong if you catch the error and don't print the error message that was produced.  Which call failed?

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -307,195 +334,158 @@ def mapping_analysis(config, mesh_name, date_string, ice_shelf_cavities):
...
-        nomaskStr=''
-
-    # make links
-    make_link(domain_exe, 'domain_exe')
-    mapping_file = 'map_' + mesh_name + nomaskStr + '_TO_' + \
-        atm_scrip_tag + '_aave.' + date_string + '.nc'
-    make_link('../mapping_CORE_Gcase/' + mapping_file, mapping_file)
+def domain_CORE_Gcase(config):  # {{{
+    make_domain_files(config, 'CORE_Gcase')
+# }}}
```
xylar, line 455: There are 3 essentially identical functions with a single string that differs between them.  It makes a lot more sense to have one function than 3 copies.

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -307,195 +334,158 @@ def mapping_analysis(config, mesh_name, date_string, ice_shelf_cavities):
...
-    args = ['./domain_exe', '-m', mapping_file, '-o', mesh_name, '-l', 'T62']
-    run_command(args)
-
-    # make links in output directories
-    files = glob.glob('domain*.nc')
-    os.chdir('../assembled_files_for_upload/inputdata/share/domains')
-    for file in files:
-        make_link('../../../../domain/' + file, './' + file)
-# }}}
+def make_domain_files(config, mapping_suffix):  # {{{
```
xylar, line 468: This function replaces the 3 `domain` functions above.  We could add more arguments in the future if needed (e.g. different atmosphere grids.

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -180,97 +189,113 @@ def graph_partition_ocean(config, mesh_name, date_string, ice_shelf_cavities):
...
-            'seaice.' + mesh_name + '.nc'
-            ]
+    args = ['ncks', '-v',
+            'areaCell,cellsOnCell,edgesOnCell,fCell,indexToCellID,latCell,'
+            'lonCell,meshDensity,nEdgesOnCell,verticesOnCell,xCell,yCell,zCell,'
+            'angleEdge,cellsOnEdge,dcEdge,dvEdge,edgesOnEdge,fEdge,'
+            'indexToEdgeID,latEdge,lonEdge,nEdgesOnCell,nEdgesOnEdge,'
+            'verticesOnEdge,weightsOnEdge,xEdge,yEdge,zEdge,areaTriangle,'
+            'cellsOnVertex,edgesOnVertex,fVertex,indexToVertexID,'
+            'kiteAreasOnVertex,latVertex,lonVertex,xVertex,yVertex,zVertex',
```
xylar, line 225: I think this is the complete list of fields we want to keep for the sea-ice initial condition but this should *definitely* have another pair of eyes on it.

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -180,97 +189,113 @@ def graph_partition_ocean(config, mesh_name, date_string, ice_shelf_cavities):
...
-            mesh_name + '.nc',
-            'seaice.' + mesh_name + '.nc'
-            ]
+            'bottomDepth, refBottomDepth,restingThickness,temperature,salinity,'
+            'temperatureSurfaceValue,salinitySurfaceValue,surfaceVelocityZonal,'
+            'surfaceVelocityMeridional,SSHGradientZonal,SSHGradientMeridional,'
+            'vertNonLocalFluxTemp,normalVelocity,layerThickness,'
+            'normalBarotropicVelocity,vertCoordMovementWeights,'
+            'boundaryLayerDepth,seaIcePressure,atmosphericPressure,'
+            'filteredSSHGradientZonal,filteredSSHGradientMeridional',
```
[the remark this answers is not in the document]
xylar: I went ahead and made this switch, as you'll see below.

xylar:
## Testing

I tested a merge of this PR, #507 , #508, #510 and #511 together through spin-up for the following test cases:
* `EC60to30`
* `EC60to30wISC`
* `SO60to10wISC`

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -27,8 +27,7 @@
 # }}}
 
 
-def main():
-# {{{
+def main():  # {{{
```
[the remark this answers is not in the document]
mark-petersen: Thanks. I'll keep that in mind. Folds are really great in vim, but I can see why they are excessive in other editors.

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -180,97 +189,113 @@ def graph_partition_ocean(config, mesh_name, date_string, ice_shelf_cavities):
...
+    mesh_name = config.get('main', 'mesh_name')
+    date_string = config.get('main', 'date_string')
+    ice_shelf_cavities = config.getboolean('main', 'ice_shelf_cavities')
 
     if ice_shelf_cavities:
-        nomaskStr='.nomask'
+        nomaskStr = '.nomask'
     else:
-        nomaskStr=''
+        nomaskStr = ''
```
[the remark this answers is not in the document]
mark-petersen: Strange, I always try to use `autopep8` before I check in python code. I must have added these lines after that.

PR #507, opened 2020-04-03
Add metadata to MPAS Meshes

xylar:
This metadata is formatted according to the specs of @proteanplanet, see examples below.

To do this, each case with an `e3sm_coupling` step has been given its own config file that contains either the mesh metadata or ways to construct it from other keys of `autodetect` it from mesh fields.

Config files:
- [x] QU240
- [x] QU240wISC
- [x] EC60to30
- [x] EC60to30wISC
- [x] CUSP12
- [x] CUSP8
- [x] SO60to10wISC


xylar:
Example from the QU240wISC test case:
```
$ ncdump -h mesh.nc
...
		:MPAS\ Mesh\ Short\ Name = "QU.wISC.E3SMv2.240km.64L.rev001" ;
		:MPAS\ Mesh\ Long\ Name = "Quasi-uniform mesh Ice Shelf Cavities for E3SM Version 2, 240km resolution, 64 vertical levels" ;
		:MPAS\ Mesh\ Description = "MPAS quasi-uniform mesh at 240-km resolution includes cavities under the ice shelves in the Antarctic" ;
		:MPAS\ Mesh\ E3SM\ Version = 2 ;
		:MPAS\ Mesh\ QU.wISC\ Version = "001" ;
		:MPAS\ Mesh\ QU.wISC\ Version\ Author = "Xylar Asay-Davis" ;
		:MPAS\ Mesh\ QU.wISC\ Version\ Author\ Institution = "Los Alamos National Laboratory" ;
		:MPAS\ Mesh\ QU.wISC\ Version\ Creation\ Date = "04/03/2020 15:57:06" ;
		:MPAS\ Mesh\ QU.wISC\ Minimum\ Resolution\ \(km\) = 240. ;
		:MPAS\ Mesh\ QU.wISC\ Maximum\ Resolution\ \(km\) = 240. ;
		:MPAS\ Mesh\ QU.wISC\ Maximum\ Depth\ \(m\) = 6000. ;
		:MPAS\ Mesh\ QU.wISC\ Number\ of\ Levels = 64. ;
		:MPAS\ Mesh\ Ice\ Shelf\ Cavities = "OFF" ;
		:MPAS\ Mesh\ Runoff\ Description = "<<<Spreading function described here>>>" ;
		:MPAS\ Mesh\ COMPASS\ Version = "0.1.2" ;
		:MPAS\ Mesh\ JIGSAW\ Version = "0.9.12" ;
		:MPAS\ Mesh\ JIGSAW-Python\ Version = "0.2.1" ;
		:MPAS\ Mesh\ MPAS-Tools\ Version = "0.0.9" ;
		:MPAS\ Mesh\ NCO\ Version = "4.9.2" ;
		:MPAS\ Mesh\ ESMF\ Version = "8.0.0" ;
		:MPAS\ Mesh\ Geometric\ Features\ Version = "0.1.6" ;
		:MPAS\ Mesh\ Metis\ Version = "5.1.0" ;
		:MPAS\ Mesh\ pyremap\ Version = "0.0.5" ;
```

xylar:
@proteanplanet, the spaces in the attribute names are kind of non-standard and don't format well in `ncdump`.  Do you want to use underscores or something else?  Or are you okay with the escape characters?  (The same is true of parentheses, like in the units.)

testing_and_setup/compass/ocean/global_ocean/EC60to30/init/config_E3SM_coupling_files.ini
```diff
@@ -0,0 +1,73 @@
...
+creation_date = autodetect
+min_res = 30
+max_res = 60
+max_depth = autodetect
+levels = autodetect
+runoff_description = <<<Spreading function described here>>>
+
+# The following options need to be added by the author
+author =
+institution =
```
xylar, line 28: @mark-petersen (and others), I haven't come up with a good way to fill these in in an automated way.  It might be helpful, e.g. for regression testing, to have a command-line flag that supplies these.  Needs a little thought.

proteanplanet:
Hi @xylar, Underscores are fine and I agree that formatting with ncdump is important.  With regard to Author and Institution, the only way that's ever worked in a semi-automatic way for me is with environment variables.

proteanplanet:
@xylar :  Another thought that occurred to me on the Author and Institution problem.  Instead of using "Institution" as a field, we could instead use email address, which usually is reflective of the institution anyway.  Both name and email address can be grabbed from .gitconfig.   Anyhow, just an idea.

xylar:
@proteanplanet, that's an idea.  I was thinking of the `.gitconfig` as well.  I use my personal email for GitHub so I'd need to override that for myself but that's not a big deal.

I went ahead and added command-line options for filling in author and institution but that's not a complete solution yet.

proteanplanet:
I am pasting Jon's comment here from confluence: "We may need to revisit the naming convention – E3SM tests use periods to separate fields in a test name, meaning

 create_test SMS.T62_oEC60to30v3.GMPAS-IAF.anvil_intel

has the test type (SMS) separated from the grid resolution (T62_oEC60to30v3) and compset (GMPAS-IAF), etc by periods. So the naming convention as it stands would cause problems. Or we at least need a “short” naming convention that would work with these limitations."

Do we need to consider this in the current PR?

vanroekel:
Just so I understand, on mesh creation this metadata gets created on its own?  This is a ton of metadata to expect someone to have to fill in on creation by hand.

@xylar if using the gitconfig to fill email address, can you share how you would swap your personal email for lanl?  I would need to do the same.

Also, thinking more about the name for meshes, I'd like to propose a change.  I think the E3SMv2 is in an odd place.  I'd prefer something like
```
CUSP.12kmto60km.L64.E3SMv2.rev001
```

xylar:
@proteanplanet 
> Do we need to consider this in the current PR?

Yes, absolutely!  That's why this is a discussion.  What would the proposed change be?  Underscores instead of periods?

@vanroekel 
> Just so I understand, on mesh creation this metadata gets created on its own? This is a ton of metadata to expect someone to have to fill in on creation by hand.

First, this doesn't happen at mesh creation but rather at the `e3sm_coupling` stage.  We could switch this to happening at an earlier stage (`culled_mesh` or `initial_state`) but it seems like we don't need this metadata for a typical standalone configuration, and the `e3sm_coupling` stage (essentially the last step after spin-up of an initial condition) is good enough.

This also gives us the option of having slightly different metadata for the sea-ice initial condition if desired.

Most of the metadata comes from a config file that that the user would not need to alter (it is specific to the test case).  Other parts come from the mesh itself, or the conda environment, and are automatic.  Only the author and institution are not automatic at the moment but the `git` idea is a good one for solving that.

> @xylar if using the gitconfig to fill email address, can you share how you would swap your personal email for lanl? I would need to do the same.

My plan would be to have a config option for email address with a default value of `autodetect`.  You could fill in your email before running the `e3sm_coupling` step in `config_E3SM_coupling_files.ini`.  In other words, you would need to remember to change this config file each time you are creating E3SM coupling files.

> Also, thinking more about the name for meshes, I'd like to propose a change. I think the E3SMv2 is in an odd place. I'd prefer something like CUSP.12kmto60km.L64.E3SMv2.rev001

Yes, I think that makes more sense to me, too.  It's more similar to what we've done so far and it puts the more relevant info earlier.

Did you want `km` twice in `12kmto60km`?  That wasn't @proteanplanet's suggestion and I prefer to avoid the redundancy if possible.

xylar:
Here's the updated version of the metadata:
```
		:MPAS_Mesh_Short_Name = "QU_wISC_240km_64L_E3SMv2_rev001" ;
		:MPAS_Mesh_Long_Name = "Quasi-uniform mesh Ice Shelf Cavities for E3SM Version 2, 240km resolution, 64 vertical levels" ;
		:MPAS_Mesh_Description = "MPAS quasi-uniform mesh at 240-km resolution includes cavities under the ice shelves in the Antarctic" ;
		:MPAS_Mesh_E3SM_Version = 2 ;
		:MPAS_Mesh_QU_wISC_Version = "001" ;
		:MPAS_Mesh_QU_wISC_Version_Author = "Xylar Asay-Davis" ;
		:MPAS_Mesh_QU_wISC_Version_Author_E-mail = "xylarstorm@gmail.com" ;
		:MPAS_Mesh_QU_wISC_Version_Creation_Date = "04/04/2020 10:53:30" ;
		:MPAS_Mesh_QU_wISC_Minimum_Resolution_km = 240. ;
		:MPAS_Mesh_QU_wISC_Maximum_Resolution_km = 240. ;
		:MPAS_Mesh_QU_wISC_Maximum_Depth_m = 6000. ;
		:MPAS_Mesh_QU_wISC_Number_of_Levels = 64. ;
		:MPAS_Mesh_Ice_Shelf_Cavities = "ON" ;
		:MPAS_Mesh_Runoff_Description = "<<<Spreading function described here>>>" ;
		:MPAS_Mesh_COMPASS_Version = "0.1.2" ;
		:MPAS_Mesh_JIGSAW_Version = "0.9.12" ;
		:MPAS_Mesh_JIGSAW-Python_Version = "0.2.1" ;
		:MPAS_Mesh_MPAS-Tools_Version = "0.0.9" ;
		:MPAS_Mesh_NCO_Version = "4.9.2" ;
		:MPAS_Mesh_ESMF_Version = "8.0.0" ;
		:MPAS_Mesh_geometric_features_Version = "0.1.6" ;
		:MPAS_Mesh_Metis_Version = "5.1.0" ;
		:MPAS_Mesh_pyremap_Version = "0.0.5" ;
```
This is without me replacing my personal email address (bots, knock yourselves out!).  It uses my work email address if I put into the config file.

testing_and_setup/compass/ocean/global_ocean/EC60to30wISC/init/config_E3SM_coupling_files.ini
```diff
@@ -0,0 +1,78 @@
...
+creation_date = autodetect
+min_res = 30
+max_res = 60
+max_depth = autodetect
+levels = autodetect
+runoff_description = <<<Spreading function described here>>>
+
+# The following options are detected from .gitconfig if not explicitly entered
+author = autodetect
+email = autodetect
```
xylar, line 30: @vanroekel, a copy of this file will end up in the `e3sm_coupling` directory.  You could edit the `email` option here and put in your LANL email address.

testing_and_setup/compass/ocean/global_ocean/EC60to30wISC/init/config_E3SM_coupling_files.ini
```diff
@@ -0,0 +1,78 @@
...
+              (30 km) and poles (35 km) with 60-km resolution at mid latitudes
+              and includes cavities under the ice shelves in the Antarctic.
+e3sm_version = 2
+mesh_version = 001
+creation_date = autodetect
+min_res = 30
+max_res = 60
+max_depth = autodetect
+levels = autodetect
+runoff_description = <<<Spreading function described here>>>
```
xylar, line 26: @vanroekel, This is where a lot of the metadata comes from.  the various `${blah}` entries come from other config options.  (This uses the so-called [ExtendedInterpolation](https://docs.python.org/3.3/library/configparser.html#configparser.ExtendedInterpolation) for config files.)  Anything with `autodetect` will get looked up somehow or other.

vanroekel:
@xylar, thanks for the explanations.  and regarding the tag, that was a mistake, I don't think we should have km twice in there, 12to60km is much better.

xylar:
## Testing

I tested a merge of this PR, #506, #508, #510 and #511 together through spin-up for the following test cases:
* `EC60to30`
* `EC60to30wISC`
* `SO60to10wISC`

jonbob:
@xylar - I think we need a *really* short name as well. As useful as it is to have this much information embedded in even the short name, the reality may be that we can't afford the short name that gets used by E3SM to be that long. The resolution for any coupled case has three grid specifications now that we have tri-grids, and I don't believe it will be workable if each of those grids requires 30+ characters to define it (like QU_wISC_240km_64L_E3SMv2_rev001). So we also need a shorthand notation -- my first recommendation is dropping the underscores completely, since they are typically used to separate the meshes for the different components in an E3SM resolution definition (i.e. T62_oEC60to30v3). Can we use the QU240 name as an example, and see exactly what information is necessary for an E3SM name?
1. QU_wISC_240km_64L_E3SMv2_rev001
2. QUwISC240km64LE3SMv2rev001 -- dropping underscores
3. QUwISC240km64LE3SMv2r01 -- replacing rev001 with r01 (are we really likely to have more than 99 revisions of any grid?)
4. QUwISC240km64LE2r01 -- replacing E3SMv2 with E2
5. QUwISC24064LE2r01 -- getting rid of km (do we ever use anything else?)
6. QU240wISC64LE2r01 -- reordered to keep some of the numbers separated

This is not great, but has gone from 31 characters to 17 and still has all the information? Anyone else?

xylar:
The verbose names were at @maltrud's and @proteanplanet's request, so they'll need to weigh in.  I don't see a value in a short name distinct from the "really short name".  When would we use the short name?  @maltrud and @proteanplanet, does this essentially leave us back where we started with mesh names, with the exception that we now include the number of vertical layers and the E3SM version?

The main problem with 5 and 6 above are the numbers running together.  For a case without ice shelves, we would have `QU24064LE2r01`.  However, the simple solution is to move the `L` in front of the number of levels: `QU240L64E2r01`, or with ice shelves, `QUwISC240L64E2r01`.  I don't have a strong preference but it seems odd to specify "with ice-shelf cavities" (`wISC`) in the middle of specifying the resolution (`240L64`) so I'd keep it toward the beginning.

@jonbob and @vanroekel, does anyone outside of the COSIM team need to weight in on these names?  Anyone else from infrastructure, water cycle, etc.?

jonbob:
@xylar - I agree that L64 is more consistent with E2 and r01, that is, the identifier first and the value second. And it has the benefit of keeping the numbers distinct.

I think we are squeezed between keeping the names short and wanting as much information as possible. I think that's where the metadata you've added is especially critical. We'll let others weigh in about how much has to be in the actual name...

proteanplanet:
I am wondering if the solution to this is to have a short_name that addresses the point of @jonbob (e.g. EC30to50L64), then having a long_name following the suggestion of @maltrud  (e.g.QU.wISC.E3SMv2.240km.64L.rev001), and then using the metadata description to account for the rest.   I think this could accommodate everyone's needs.

xylar:
@proteanplanet, that sounds fine.  @maltrud, where would the `long_name` get used in your vision?

jonbob:
@proteanplanet - do you still think we would need the r01 identifier in the short name as well? I can still envision having different revisions of a horizontal grid, and each would require separate initial and mapping files

proteanplanet:
@jonbob - Good question.  I guess we do, otherwise things could get pretty confused.  I do wonder whether we should really just leave out the E3SM bit, and keep the rest as our short name, e.g. ECwISC30to60L64rev002.  Otherwise, we really aren't addressing the issue that @xylar and @mark-petersen are trying to address, and we are back to where we started.   In an effort to shorten the above short name a bit, maybe instead of adding special identifiers for wISC, for example, we should give the entire grid that includes the cavities a completely difference name, e.g. ECwISC could just be call EI (purely abitrary for discussion here).  Also, maybe we can replace the rev with just an underscore, hence the name above becoming EI30to60L64_002.  What do you think @xylar, @maltrud, @jonbob?

maltrud:
All, how about the following?
QUwISC240kmL64E3SMv2r01 
ECwISC60to30kmL64E3SMv2r01 
very close to @jonbob 's suggestion, but with L64 and wISC before the grid resolution?  also i agree that 2 digit revision should be sufficient.

maltrud:
actually just noticed my suggestion is essentially identical to @proteanplanet

xylar:
I don't think we're converging on a solution yet.  Issues are:
* Do we include the E3SM version or not?  The desired to include it was a large part of the motivation to change the naming scheme.  Switching from `v3` to `r01` doesn't really alleviate the confusion that @mark-petersen and I were hoping to address
* Do we include `km` or not?  I think it adds clarity, not because you know what the units are but because you know that the number refers to something to do with units at all -- it suggests that it can only be horizontal resolution.  But I don't feel strongly.
* Are we okay with `wISC` or do we want that to be shorter?  I proposed `wISC` as preferable to `ISC` for various reasons (the `w` separates a bunch of capital letters and `wISC` sounds more pronouncable to me than `ISC`).  Anything shorter is less clear(`wC` would potentially be mis-identified as "water cycle", `wI`, `wIS`, `wS`, etc. are not accurate, since we're not talking about the ice-shelves themselves and `IS` in particular would more likely to refer to "ice sheet" than "ice shelf".

For the short names, I would suggest:
* `QU240L64E2r01` (Likely irrelevant since it won't get put in E3SM)
* `QUwISC240L64E2r01`  (Again, likely irrelevant since it won't get put in E3SM)
* `EC30to60L64E2r01`
* `ECwISC30to60L64E2r01`
* `CUSP12to60L64E2r01`
* `CUSP8to60L64E2r01`
* `SO10to60L64E2r01`

(Not sure which of the `CUSP` meshes is considered likely v2 mesh.)

I am fine with converting the short name that @proteanplanet first proposed to the `long_name`, and making sure the current long name and description are combined in some way.

vanroekel:
My two cents.
1) We should definitely include some way to tag e3sm version.  I agree with @xylar that if we don't the r tag is again meaningless.
2) I don't feel super strongly about km, but as ICoM gets farther along we will definitely need it as they think about sub km meshes.  But on the other hand the EC does imply km scale resolution.

I like the names you suggest @xylar

One other possible suggestion, what if we tag these to campaigns?  In this case we could have either WC or CR or BGC as the start, and the E3SM version would give the same info as the front specifier.  However, I'm sure there are lots of use cases I'm missing that wouldn't work with this idea.

@jonbob -- is there a target length we need to shoot for?

maltrud:
my opinions on @xylar 's last post:
1) yes, i feel pretty strongly that we have E3SMvX in the name.
2) completely agree with the logic for including km and feel pretty strongly about it, too.
3) i think wISC is the best choice.

so ECwISC60to30kmL64E3SMv2r01 would be in all of the file names.

xylar:
@maltrud, so @jonbob's 3rd suggestion from above?  I'm good with that, too.

vanroekel:
I'm good with that one too.

maltrud:
> @maltrud, so @jonbob's 3rd suggestion from above? I'm good with that, too.

yup, except using L64 instead of 64L

xylar:
> yup, except using L64 instead of 64L

Ah, right!

xylar:
My sense is that, assuming @jonbob agrees to this longer form, there would be no need for a "long name" that is only slightly longer.

jonbob:
I'm not sure what to say -- if all three grids necessary to define an E3SM resolution were this long, it could be intractable. But if all the information is necessary to distinguish one grid from another, then there isn't much choice. The only change I think we make from @maltrud's last suggestion is for the short name, let "E" replace "E3SMv". If it's a naming convention, the meaning will become clear if it isn't already.

maltrud:
just to be clear--i think this 'long' name (eg, ECwISC60to30kmL64E3SMv2r01 ) should appear on the mpas-ocean/seaice files (grid, IC, forcing, etc).  i agree with @jonbob that this needs to be significantly shortened when defining B, C, G, etc cases within e3sm.  say, ECwI60Ev2r01.  maybe L64 though i'm guessing we will only have 1 vertical grid for any official revision.

xylar:
@maltrud, thanks for clarifying, it was not clear to me that you were on board with the "really short name".  I think we should then call the thing needed in E3SM case names `short_name` and the name used in ocean and sea-ice file names the `long_name`.  @maltrud had:
```
ECwI60Ev2r1
```
I would suggest some changes.  My preferred option would be:
```
ECwI30to60E2r1
```
I could be okay with:
```
ECwI30E2r1
```
I would prefer to continue to include both min and max resolution in the name, since this might be the only distinguishing feature of some mesh short names.  If we're going to quote one resolution and not both, it should be the min, not the max. 

If we're shortening `E3SMv`, it might as well be `E` and not `Ev`, since I don't think the latter makes it any clearer to the uninitiated what the `E` is for.  

As I said above, i'm not big in `wI` as I said above but I can live with it.

xylar:
@jonbob, is it going to be possible to have a short name for E3SM cases and tests, and a different long name for ocean and sea-ice file names, as @maltrud would like?  Is the infrastructure set up for this?

jonbob:
@xylar - yes, we should be able to do that if necessary. I would prefer to use the short name for all mapping files, if only because I think it's clearer and keeps filenames at a reasonable length. Most of those already have more information included in their names, like
```map_ne30np4_to_oEC60to30wLI_nomask_aave.160915.nc```
where we need the atm mesh name, plus the type of mapping and the date. But ocn/ice filenames themselves are defined mostly in our streams files, which offers plenty of latitude in naming. For example:
``` 
   elif ocn_grid == 'oEC60to30v3wLI':
        ic_date += '171031'
        ic_prefix += 'oEC60to30v3wLI60lev'
        decomp_date += '170328'
        decomp_prefix += 'mpas-o.graph.info.'
        restoring_file += 'sss.monthlyClimatology.PHC2_salx_040803.oEC60to30v3wLI.nc'
        analysis_mask_file += 'masks_SingleRegionAtlanticWTransportTransects_EC60to30v3wLI_171116.nc'```

jonbob:
@xylar - I also agree with your latest suggestion for short names. I think
```
ECwI30to60E2r1
```
is very clear and concise, but I also agree it may make more sense to include "wISC" instead of "wI".
```
ECwISC30to60E2r1
```
is not really any longer than what we have been using, like
```
oEC60to30v3wLI
```
but much clearer in that we get rid of the "v3" which has been used for too many things and replace it with "E2r1" which adds more information.

milenaveneziani:
I think `wI` is actually better than `wISC`.

xylar:
Latest update (still a work in progress, of course):
```
		:MPAS_Mesh_Short_Name = "EC30to60E2r01" ;
		:MPAS_Mesh_Long_Name = "EC30to60kmL64E3SMv2r01" ;
		:MPAS_Mesh_Description = "MPAS Eddy Closure mesh for E3SM version 2 with enhanced resolution around the equator (30 km) and poles (35 km), 60-km resolution at mid latitudes, and 64 vertical levels" ;
		:MPAS_Mesh_E3SM_Version = 2 ;
		:MPAS_Mesh_EC_Version = "01" ;
		:MPAS_Mesh_EC_Version_Author = "Xylar Asay-Davis" ;
		:MPAS_Mesh_EC_Version_Author_E-mail = "xylarstorm@gmail.com" ;
		:MPAS_Mesh_EC_Version_Creation_Date = "04/07/2020 17:02:05" ;
		:MPAS_Mesh_EC_Minimum_Resolution_km = 30. ;
		:MPAS_Mesh_EC_Maximum_Resolution_km = 60. ;
		:MPAS_Mesh_EC_Maximum_Depth_m = 6000. ;
		:MPAS_Mesh_EC_Number_of_Levels = 64. ;
		:MPAS_Mesh_Ice_Shelf_Cavities = "OFF" ;
		:MPAS_Mesh_Runoff_Description = "<<<Spreading function described here>>>" ;
		:MPAS_Mesh_COMPASS_Version = "0.1.2" ;
		:MPAS_Mesh_JIGSAW_Version = "0.9.12" ;
		:MPAS_Mesh_JIGSAW-Python_Version = "0.2.1" ;
		:MPAS_Mesh_MPAS-Tools_Version = "0.0.9" ;
		:MPAS_Mesh_NCO_Version = "4.9.2" ;
		:MPAS_Mesh_ESMF_Version = "8.0.0" ;
		:MPAS_Mesh_geometric_features_Version = "0.1.6" ;
		:MPAS_Mesh_Metis_Version = "5.1.0" ;
		:MPAS_Mesh_pyremap_Version = "0.0.5" ;
```

xylar:
@jonbob 
> I would prefer to use the short name for all mapping files, if only because I think it's clearer and keeps filenames at a reasonable length.

I think that seems fair, and it's an easy enough thing for me to change in `e3sm_coupling` (and I would guess for you in your workflow

@milenaveneziani 
> I think wI is actually better than wISC.

It's concise for sure.  But aren't you worried that people will think this is "with ice"?  Isn't that about as ambiguous as possible (do we mean sea ice, ice sheets, ice shelves)?

milenaveneziani:
> But aren't you worried that people will think this is "with ice"?

it's a mesh name though, not a compset, so I'd think it can only be 'ice-shelves'?

jonbob:
@xylar  I think the metadata and naming convention both look great -- thanks for all your work on this. I myself prefer wISC but not picky. The extra characters shouldn't be a big deal since they aren't in all grids, and it seems more extensible whenever we have more descriptors that may become necessary

xylar:
@jonbob 
> is not really any longer than what we have been using, like `oEC60to30v3wLI`

Speaking of which, do we need the `o` prefix on these grids?

jonbob:
@xylar - I think the "o" is from early times and it was used to describe it as an ocean grid. I don't believe we need that, especially since the grids are also used for seaice.

xylar:
Great, the `o` always bothered me precisely because it's an ocean/sea ice grid so I think it'll be nice to drop that.

xylar:
Okay, I think there's sufficient consensus on this.  The only issue of remaining contention seems to be `wI` vs. `wISC` but I'm going to use the latter because @jonbob is the one pushing for shorter names and he still thinks it's okay.  @milenaveneziani, I agree that we would know that `wI` means "with ice-shelf cavities" but I think it takes us back toward the rather deceptive `wLI`, which did get misunderstood by some as meaning that it was coupled to the land-ice model.

xylar:
@vanroekel, @proteanplanet, @maltrud and @milenaveneziani, if you want to review this formally, please do.  If not, please take yourselves off as reviewers.

@mark-petersen, I'd appreciate you running this through the paces on a test case or 2 to make sure it works for you (and makes sense to you).

mark-petersen:
Tested `e3sm_coupling` with 1 and 36 cores.  Appears to work with QU240, QU240wISC, EC60to30. Thanks @xylar!

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
```diff
@@ -1142,6 +1147,162 @@ def remap_rignot(inFileName, meshFileName, meshName, outFileName,
...
+        if 'global_ocean' in path:
+            index = path.index('global_ocean') + 1
+            mesh_name = 'o{}'.format(path[index])
+            print("- mesh name autodetected from path: {}".format(mesh_name))
+            config.set('mesh', 'long_name', mesh_name)
+        else:
+            raise ValueError("mesh name not found in path. Please specify "
+                             "the mesh_name in config_E3SM_coupling_files.ini.")
+    else:
+        print("- mesh name specified in config file: ", mesh_name)
```
mark-petersen, line 1237: would be nice to see both long and short name here.

mark-petersen:
It looks like the long name is used everywhere in E3SM file names, not the short name.  Was that intentional? I like the short name better.

xylar:
> It looks like the long name is used everywhere in E3SM file names, not the short name. Was that intentional? I like the short name better.

That was the consensus as I understood it.  Long names where possible but short names in mapping files (and eventually in test and case names) where the long name is too long to be practical.

https://github.com/MPAS-Dev/MPAS-Model/pull/507#issuecomment-610045535

@mark-petersen, I'm reluctant to revisit this after a 50+ message conversation but if you feel strongly, we need to have you and @maltrud come to an agreement on this.

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
xylar: Okay, I can add this.

mark-petersen:
OK, thanks. I see now. Thanks for pointing that out. I'll go with the decision above.

testing_and_setup/compass/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
xylar: Fixed.

PR #508, opened 2020-04-03
Copy define_base_mesh.py in 3 test cases

xylar:
This behavior got changed only in the default, not in test cases that have their own version of `config_base_mesh.xml`

PR #510, opened 2020-04-04
Configuration updates to SO_wISC_10to60km_64L

xylar:
Default to spinning up on 36 nodes (1296 cores) on IC.  This is what I'm spinning up with right now.  Spin-up on 4 nodes is extraordinarily slow!

Also, switch to 6 iterations of topography smoothing.  0 and 3 iterations were producing a worrying spot of high kinetic energy near the Ross Ice Shelf, see details below.

Remove an unused (and potentially confusing) template file.


xylar:
## Testing

I tested a merge of this PR, #506, #507, #508 and #511 together through spin-up for the following test cases:
* `EC60to30`
* `EC60to30wISC`
* `SO60to10wISC`

xylar:
In my testing of `SO_wISC_10to60km_64L_E3SMv2_rev001` (i.e. `SO60to10wISC`), the KE is surprisingly high under a little ice shelf to the west to of the Ross.  I re-ran with 3 iterations and 6 iterations of smoothing, and the KE is somewhat reduced in the first case and and region doesn't stand out at all in the second.  As a result, I think we should test with 6 iterations of smoothing for now.  I think I will add that change to this PR, just to keep the number of PRs somewhat under control.

xylar:
## Kinetic Energy (log scale)

### No Smoothing

ke_smooth0

### 3 Iterations of Smoothing

ke_smooth3

### 6 Iterations of Smoothing

ke_smooth6

xylar:
If we decide we don't want topographic smoothing globally, we would restrict it to the region near ice shelves in much the same way that the vertical coordinate is only altered from z-star in the region near ice shelves.

PR #512, opened 2020-04-07
Fix masking for initial state plotting

xylar:
Cells below `maxLevelCell` are no longer included.

### DIFF
--- a/testing_and_setup/compass/ocean/scripts/plots/plot_initial_state.py
+++ b/testing_and_setup/compass/ocean/scripts/plots/plot_initial_state.py
@@ -3,13 +3,14 @@
 This script creates plots of the initial condition.
 """
 # import modules
-import matplotlib.pyplot as plt
 from netCDF4 import Dataset
 import numpy as np
 import argparse
 import datetime
 import matplotlib
 matplotlib.use('Agg')
+import matplotlib.pyplot as plt
+from matplotlib.font_manager import FontProperties
 
 
 def main():
@@ -51,6 +52,7 @@ def main():
     plt.subplot(3, 3, 2)
     varName = 'maxLevelCell'
     var = ncfile.variables[varName]
+    maxLevelCell = var[:]
     plt.hist(var, bins=nVertLevels - 4)
     plt.ylabel('frequency')
     plt.xlabel(varName)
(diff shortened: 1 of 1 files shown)


xylar:
I also switched the text with max/min in the plot to a monospace font so the columns align better.

xylar:
initial_state

testing_and_setup/compass/ocean/scripts/plots/plot_initial_state.py
```diff
@@ -3,13 +3,14 @@
...
 # import modules
-import matplotlib.pyplot as plt
 from netCDF4 import Dataset
 import numpy as np
 import argparse
 import datetime
 import matplotlib
 matplotlib.use('Agg')
+import matplotlib.pyplot as plt
+from matplotlib.font_manager import FontProperties
```
xylar, line 13: Usually, you need to switch to the `Agg` backend before importing `plt`.  I guess that isn't required in the most recent `matplotlib` but it's still safer to do it in this order.  PEP8 will complain about imports not being at the top but it is necessary in this case.

testing_and_setup/compass/ocean/scripts/plots/plot_initial_state.py
```diff
@@ -65,45 +67,60 @@ def main():
...
+        cell1 = cellsOnEdge[:, 1]-1
+        edgeMask[:, k] = np.logical_and(np.logical_and(cellMask[cell0, k],
+                                                       cellMask[cell1, k]),
+                                        np.logical_and(cell0 >= 0,
+                                                       cell1 >= 0))
+
     plt.subplot(3, 3, 4)
     varName = 'temperature'
-    var = ncfile.variables[varName]
-    plt.hist(np.ndarray.flatten(var[:]), bins=100, log=True)
```
xylar, line 71: The mask takes care of of flattening the array in this case but a more efficient way of flattening a typical numpy array is to call its `ravel` method: `var[:].ravel()`

testing_and_setup/compass/ocean/scripts/plots/plot_initial_state.py
```diff
@@ -65,45 +67,60 @@ def main():
     txt = txt + ' {:9.2e}'.format(np.min(var)) + \
         ' {:9.2e}'.format(np.max(var)) + ' ' + varName + '\n'
```
xylar, line 68: I didn't have the energy to change all the formatting, but I would not use `str` addition like this.  For example, here, I would do:
```suggestion
    txt = '{}{:9.2e} {:9.2e} {}\n'.format(txt, np.amin(var), np.amax(var), varName)
```
I think `amin` and `amax` are considered to be the better than the aliases `min` and `max` because it clarifies that you want the min/max over the array vs. `minimum` and `maximum`, which give the element-wise min/max between two arrays.

testing_and_setup/compass/ocean/scripts/plots/plot_initial_state.py
```diff
@@ -65,45 +67,60 @@ def main():
...
 
+    font = FontProperties()
+    font.set_family('monospace')
+    font.set_size(12)
     print(txt)
     plt.subplot(3, 3, 1)
-    plt.text(0, 1, txt, fontsize=12, verticalalignment='top')
+    plt.text(0, 1, txt, verticalalignment='top', fontproperties=font)
```
xylar, line 123: Apparently, this is how you change fonts.  Never had to do this before.

mark-petersen:
@xylar you are too good! Thanks for the changes. I appreciate the python improvements as well!

PR #513, opened 2020-04-07
Localized OpenMP Threading

mattdturner:
This PR replaces the OpenMP threading implementation with the parallel region being initiated before the timestep loop with a new implementation that created parallel regions around each OpenMP loop (i.e., more localized threading regions instead of a global region).

### DIFF
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -252,6 +252,7 @@ module ocn_time_integration_rk4
          call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
          call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
 
+         !$omp parallel
          !$omp do schedule(runtime) private(k)
          do iEdge = 1, nEdges
             do k = 1, maxLevelEdgeTop(iEdge)
@@ -267,6 +268,7 @@ module ocn_time_integration_rk4
             end do
          end do
          !$omp end do
+         !$omp end parallel
 
          call mpas_pool_begin_iteration(tracersPool)
          do while ( mpas_pool_get_next_member(tracersPool, groupItr) )
@@ -277,6 +279,7 @@ module ocn_time_integration_rk4
                call mpas_pool_get_array(tracersPool, trim(groupItr % memberName), tracersNew, 2)
 
                if ( associated(tracersCur) .and. associated(tracersNew) ) then
+                  !$omp parallel
                   !$omp do schedule(runtime) private(k)
                   do iCell = 1, nCells  ! couple tracers to thickness
                      do k = 1, maxLevelCell(iCell)
@@ -284,24 +287,29 @@ module ocn_time_integration_rk4
                      end do
                   end do
                   !$omp end do
+                  !$omp end parallel
                end if
             end if
          end do
 
          if (associated(highFreqThicknessCur)) then
+              !$omp parallel
               !$omp do schedule(runtime)
               do iCell = 1, nCells
                  highFreqThicknessNew(:, iCell) = highFreqThicknessCur(:, iCell)
               end do
               !$omp end do
+              !$omp end parallel
          end if
 
          if (associated(lowFreqDivergenceCur)) then
-              !$omp do schedule(runtime)
+              !$omp parallel
+              !$omp do schedule(runtime) private(iCell)
               do iCell = 1, nCells
                  lowFreqDivergenceNew(:, iCell) = lowFreqDivergenceCur(:, iCell)
               end do
               !$omp end do
+              !$omp end parallel
          end if
 
          block => block % next
@@ -453,12 +461,14 @@ module ocn_time_integration_rk4
 
               call mpas_pool_get_dimension(block % dimensions, 'nCells', nCells)
 
-              !$omp do schedule(runtime)
+              !$omp parallel
+              !$omp do schedule(runtime) private(iCell)
               do iCell = 1, nCells
                  highFreqThicknessProvis(:, iCell) = highFreqThicknessCur(:, iCell) + rk_substep_weights(rk_step) &
                     * highFreqThicknessTend(:, iCell)
               end do
               !$omp end do
+              !$omp end parallel
               block => block % next
            end do
 
@@ -675,19 +685,23 @@ module ocn_time_integration_rk4
 
 
          if (config_prescribe_velocity) then
-            !$omp do schedule(runtime)
+            !$omp parallel
+            !$omp do schedule(runtime) private(iEdge)
             do iEdge = 1, nEdges
                normalVelocityNew(:, iEdge) = normalVelocityCur(:, iEdge)
             end do
             !$omp end do
+            !$omp end parallel
          end if
 
          if (config_prescribe_thickness) then
-            !$omp do schedule(runtime)
+            !$omp parallel
+            !$omp do schedule(runtime) private(iCell)
             do iCell = 1, nCells
                layerThicknessNew(:, iCell) = layerThicknessCur(:, iCell)
             end do
             !$omp end do
+            !$omp end parallel
          end if
 
          ! direct application of tidal boundary condition
@@ -719,11 +733,13 @@ module ocn_time_integration_rk4
          ! ------------------------------------------------------------------
          ! Accumulating various parameterizations of the transport velocity
          ! ------------------------------------------------------------------
-         !$omp do schedule(runtime)
+         !$omp parallel
+         !$omp do schedule(runtime) private(iEdge)
          do iEdge = 1, nEdges
             normalTransportVelocity(:, iEdge) = normalVelocityNew(:, iEdge)
          end do
          !$omp end do
+         !$omp end parallel
 
          ! Compute normalGMBolusVelocity and the tracer transport velocity
          if (config_use_GM) then
@@ -733,11 +749,13 @@ module ocn_time_integration_rk4
          call mpas_threading_barrier()
 
          if (config_use_GM) then
-            !$omp do schedule(runtime)
+            !$omp parallel
+            !$omp do schedule(runtime) private(iEdge)
             do iEdge = 1, nEdges
                normalTransportVelocity(:, iEdge) = normalTransportVelocity(:, iEdge) + normalGMBolusVelocity(:, iEdge)
             end do
             !$omp end do
+            !$omp end parallel
          end if
          ! ------------------------------------------------------------------
          ! End: Accumulating various parameterizations of the transport velocity
@@ -754,7 +772,8 @@ module ocn_time_integration_rk4
                           includeHalos = .true.)
          call mpas_threading_barrier()
 
-         !$omp do schedule(runtime)
+         !$omp parallel
+         !$omp do schedule(runtime) private(iCell)
          do iCell = 1, nCells
             surfaceVelocity(indexSurfaceVelocityZonal, iCell) = velocityZonal(1, iCell)
             surfaceVelocity(indexSurfaceVelocityMeridional, iCell) = velocityMeridional(1, iCell)
@@ -763,6 +782,7 @@ module ocn_time_integration_rk4
             SSHGradient(indexSSHGradientMeridional, iCell) = gradSSHMeridional(iCell)
          end do
          !$omp end do
+         !$omp end parallel
 
          call ocn_time_average_coupled_accumulate(diagnosticsPool, statePool, forcingPool, 2)
 
@@ -1163,7 +1183,8 @@ module ocn_time_integration_rk4
       call mpas_pool_get_array(diagnosticsPool, 'wettingVelocity', wettingVelocity)
       call mpas_threading_barrier()
 
-      !$omp do schedule(runtime) private(k)
+      !$omp parallel
+      !$omp do schedule(runtime) private(k, iEdge)
       do iEdge = 1, nEdges
          do k = 1, maxLevelEdgeTop(iEdge)
             normalVelocityProvis(k, iEdge) = normalVelocityCur(k, iEdge) + rkSubstepWeight &
@@ -1173,8 +1194,7 @@ module ocn_time_integration_rk4
       end do
       !$omp end do
 
-
-      !$omp do schedule(runtime) private(k)
+      !$omp do schedule(runtime) private(k, iCell)
       do iCell = 1, nCells
          do k = 1, maxLevelCell(iCell)
             layerThicknessProvis(k, iCell) = layerThicknessCur(k, iCell) + rkSubstepWeight &
@@ -1182,6 +1202,7 @@ module ocn_time_integration_rk4
          end do
       end do
       !$omp end do
+      !$omp end parallel
 
       call mpas_pool_begin_iteration(tracersPool)
       do while ( mpas_pool_get_next_member(tracersPool, groupItr) )
@@ -1196,7 +1217,8 @@ module ocn_time_integration_rk4
                modifiedGroupName = trim(groupItr % memberName) // 'Tend'
                call mpas_pool_get_array(tracersTendPool, modifiedGroupName, tracersGroupTend)
                if ( associated(tracersGroupProvis) .and. associated(tracersGroupCur) .and. associated(tracersGroupTend) ) then
-                  !$omp do schedule(runtime) private(k)
+                  !$omp parallel
+                  !$omp do schedule(runtime) private(k, iCell)
                   do iCell = 1, nCells
                      do k = 1, maxLevelCell(iCell)
                         tracersGroupProvis(:, k, iCell) = ( layerThicknessCur(k, iCell) * tracersGroupCur(:, k, iCell)  &
@@ -1206,34 +1228,41 @@ module ocn_time_integration_rk4
 
                   end do
                   !$omp end do
+                  !$omp end parallel
                end if
             end if
          end if
       end do
 
       if (associated(lowFreqDivergenceCur)) then
-         !$omp do schedule(runtime)
+         !$omp parallel
+         !$omp do schedule(runtime) private(iCell)
          do iCell = 1, nCells
             lowFreqDivergenceProvis(:, iCell) = lowFreqDivergenceCur(:, iCell) + rkSubstepWeight &
                                               * lowFreqDivergenceTend(:, iCell)
          end do
          !$omp end do
+         !$omp end parallel
       end if
 
       if (config_prescribe_velocity) then
-         !$omp do schedule(runtime)
+         !$omp parallel
+         !$omp do schedule(runtime) private(iEdge)
          do iEdge = 1, nEdges
             normalVelocityProvis(:, iEdge) = normalVelocityCur(:, iEdge)
          end do
          !$omp end do
+         !$omp end parallel
       end if
 
       if (config_prescribe_thickness) then
-         !$omp do schedule(runtime)
+         !$omp parallel
+         !$omp do schedule(runtime) private(iCell)
          do iCell = 1, nCells
             layerThicknessProvis(:, iCell) = layerThicknessCur(:, iCell)
          end do
          !$omp end do
+         !$omp end parallel
       end if
       call mpas_threading_barrier()
 
@@ -1243,11 +1272,13 @@ module ocn_time_integration_rk4
       ! ------------------------------------------------------------------
       ! Accumulating various parametrizations of the transport velocity
       ! ------------------------------------------------------------------
-      !$omp do schedule(runtime)
+      !$omp parallel
+      !$omp do schedule(runtime) private(iEdge)
       do iEdge = 1, nEdges
          normalTransportVelocity(:, iEdge) = normalVelocityProvis(:, iEdge)
       end do
       !$omp end do
+      !$omp end parallel
 
       ! Compute normalGMBolusVelocity, relativeSlope and RediDiffVertCoef if respective flags are turned on
       if (config_use_GM) then
@@ -1257,11 +1288,13 @@ module ocn_time_integration_rk4
       call mpas_threading_barrier()
 
       if (config_use_GM) then
-         !$omp do schedule(runtime)
+         !$omp parallel
+         !$omp do schedule(runtime) private(iEdge)
          do iEdge = 1, nEdges
             normalTransportVelocity(:, iEdge) = normalTransportVelocity(:, iEdge) + normalGMBolusVelocity(:,iEdge)
          end do
          !$omp end do
+         !$omp end parallel
       end if
       ! ------------------------------------------------------------------
       ! End: Accumulating various parametrizations of the transport velocity
@@ -1327,7 +1360,8 @@ module ocn_time_integration_rk4
 
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
 
-      !$omp do schedule(runtime) private(k)
+      !$omp parallel
+      !$omp do schedule(runtime) private(k, iCell)
       do iCell = 1, nCells
          do k = 1, maxLevelCell(iCell)
             layerThicknessNew(k, iCell) = layerThicknessNew(k, iCell) + rkWeight * layerThicknessTend(k, iCell)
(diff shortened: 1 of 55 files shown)


mattdturner:
Testing was done with Intel v19 on Cori for a QU240 `performance_test` case.  The results are bit-for-bit when compiling with `DEBUG`.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -599,11 +624,15 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
+         !$omp end parallel
       endif
 
       nEdges = nEdgesArray( 3 )
 
-      !$omp do schedule(runtime)
+      !!$omp parallel
+      !!$omp do schedule(runtime) &
+      !!$omp private(iEdge, cell1, cell2, k, BruntVaisalaFreqTopEdge, tridiagB, tridiagC, &
+      !!$omp         rightHandSide, tridiagA, N)
```
mattdturner, line 635: This loop has the OpenMP directives commented out since there is a race condition that is yet to be resolved.  This will be resolved in a future PR (the work in this PR need to be merged for future work to build upon).

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -353,7 +357,16 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, diagnost
...
+      ! TODO, mdt: determine where the openmp data race is coming from in this loop
+      !!$omp parallel
+      !!$omp do schedule(runtime) &
+      !!$omp private(iCell, invAreaCell, k, RiSmoothed, RiTemp, BVFSmoothed, nsmooth, &
+      !!$omp         langmuirNumber, langmuirEnhancementFactor, bulkRichardsonNumberStop, &
+      !!$omp         bulkRichardsonFlag, topIndex, kIndexOBL, deltaVelocitySquared, sigma, &
+      !!$omp         OBLDepths, interfaceForcings, surfaceAverageIndex, sfc_layer_depth, kav, &
+      !!$omp         i, iEdge, factor, normalVelocitySum, NormalVelocityAv, &
+      !!$omp         delU2, potentialDensitySum, blTemp) &
+      !!$omp firstprivate(cvmix_variables, Nsqr_iface, turbulentScalarVelocityScale)
```
mattdturner, line 369: This loop has the OpenMP directives commented out since there is a race condition that is yet to be resolved.  This will be resolved in a future PR (the work in this PR need to be merged for future work to build upon).  I'm also putting off resolving this race condition until the other work on CVMIX is merged.

mark-petersen:
Rebased. @mattdturner, these two files had so many previous changes, I thought it was dangerous to try to resolve the conflicts:
```
src/core_ocean/shared/mpas_ocn_gm.F
src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```
Would you mind redoing the process on those two?

Currently, intel with debug and OpenMP=true produces an error in an ALE routine. With OpenMP=false everything passes. That error could be from my rebasing, and I can look tomorrow.

mattdturner:
Yeah, I'll redo the process on those files.  I'll also look into the ALE routines while I'm at it to see what might be causing the problem

mattdturner:
@mark-petersen I just pushed 2 commits to re-do the localized threading in `mpas_ocn_tracer_hmix_redi.F` and `mpas_ocn_gm.F`.  The additional changes were bfb on Intel with debug flags enabled for the QU240 performance test.  Note that they are only bfb if I use `-O0` as a compile option.

What problems were you seeing with the ALE routines?  I didn't encounter any crashes because of them, but I'm not sure if my test case actually ran those routines.

mark-petersen:
@mattdturner you said this PR is BFB in debug. Do you expect this PR to be bfb in optimized as well?

mattdturner:
In my QU240 testing, this wasn't even BFB with `DEBUG=true`.  I had to modify the Makefile to add `-O0` to the debug flags in order to get BFB.  So yes, it would probably be non-BFB for the E3SM regression suite.

src/core_ocean/shared/mpas_ocn_thick_ale.F
```diff
@@ -151,7 +151,8 @@ subroutine ocn_ALE_thickness(meshPool, verticalMeshPool, SSH, ALE_thickness, err
       !
       if (configALEthicknessProportionality==1) then ! restingThickness_times_weights
 
-         !$omp do schedule(runtime)
+         !$omp parallel
+         !$omp do schedule(runtime) private(kMax, thicknessSum, k, SSH_ALE_Thickness)
```
mark-petersen, line 155: This section is producing an error on grizzly using 
```
module load intel/17.0.1 openmpi/1.10.5 netcdf/4.4.1 parallel-netcdf/1.5.0 pio/1.7.2
make ifort CORE=ocean DEBUG=true OPENMP=true GEN_F90=true
```
It looks like iCell is missing in the private list. I added it, and still get this same error:
```
forrtl: severe (174): SIGSEGV, segmentation fault occurred
...
ocean_model        0000000001B89B2B  ocn_thick_ale_mp_         172  mpas_ocn_thick_ale.f90
```
which is this line, just below:
```
160                SSH_ALE_Thickness(k) = SSH(iCell) * vertCoordMovementWeights(k) * restingThickness(k, iCell)
```
I have not found the root problem yet. @mattdturner if your tests on cori don't reproduce it, it might be an intel version thing. Could you double check that the allocatable array `SSH_ALE_Thickness(k)` is declared correctly? It does not need to be module level with these changes?

I can confirm that this error does not occur with optimization on, or with `OPENMP=false`.

mark-petersen:
@mattdturner could you do this:
```
cd src/core_ocean
git grep -in 'private(' |grep omp | grep -v iCell|grep -v iEdge|grep -v iVertex
```
It produces 86 lines where the partitioned index is not in the private pragma. Could you check those? You can ignore the modules removed in https://github.com/MPAS-Dev/MPAS-Model/pull/569. I was not sure if the others were oversights.

src/core_ocean/shared/mpas_ocn_thick_ale.F
mattdturner: As the code currently stands, I should be seeing the error that you are seeing.  I'm not sure why it didn't show up in my test on Cori, but the `SSH_ALE_Thickness` array is uninitialized within the OpenMP region.  I'll look into this now.

mattdturner: I just pushed a commit that changes `SSH_ALE_Thickness` to `firstprivate`, and I was able to successfully run a test on Grizzly.

mattdturner:
@mark-petersen I'm not sure what you want me to look into?  Is the issue that the index over which a loop is iterating isn't declared `private`?  In an `!$omp do` directive, the index of the initial `do` loop is inherently private.  E.g.,
```
!$omp do
do i = 1, 10
```
In this example OpenMP automatically assigns `i` to be private.  If that is what you were asking, I can explicitly include them as `private` for completeness-sake if you would prefer.

src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
```diff
@@ -277,31 +279,37 @@ subroutine ocn_time_integrator_rk4(domain, dt)!{{{
...
                  highFreqThicknessNew(:, iCell) = highFreqThicknessCur(:, iCell)
               end do
               !$omp end do
+              !$omp end parallel
          end if
 
          if (associated(lowFreqDivergenceCur)) then
-              !$omp do schedule(runtime)
+              !$omp parallel
+              !$omp do schedule(runtime) private(iCell)
```
mark-petersen, line 307: @mattdturner when I looked through this PR, it appeared that you were intentionally adding the main index loop to the private list, so I thought that went with the addition of the parallel line before it. Are you saying that we don't need to put iCell on the private list here, or in all the other places in this PR?

mark-petersen:
@mattdturner thanks for the update. I can confirm that this PR now passes the nightly regression suite using intel and debug, with OPENMP=true.

src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
mattdturner: Correct, the `iCell` does not need to be in the private list here.  I thought that I had removed all of the main loop indices from private lists, but it looks like there are quite a few that I missed.

mark-petersen: Sorry for the extra work, but could you remove all those? Otherwise it leads to confusion. I think you could use sed within a shell script to automate it on all the files.

mattdturner: Sure.  I'll take care of that today and push the changes once I'm done.

mattdturner: I just pushed changes that remove the main loop index from the `!$omp private` directives.  It PASSED the thread test when compiled with DEBUG using Intel v19 on Cori

mark-petersen:
@mattdturner thanks for the update. This fails the 1 vs 2 threads bfb comparison on grizzly with both gnu/optimized and intel/debug. Passes all other tests. I'll see if I can find it.

mark-petersen:
@mattdturner I believe I see two problems - one was on this branch before, and one introduced by 268275878. I added an RK4 1 vs 2 thread test to provide more information.  Here is what I see:
1. ocean/develop (80d2f30d3) passes bfb thread test with all: gnu optimized, gnu debug, intel debug
2. this branch, previous commit (70f782913): **fails gnu debug, gnu opt**, but passes intel debug
3. this branch, last commit (268275878): fails gnu debug, gnu opt. Passes intel debug with RK4 but fails for split explicit

Let's just concentrate on the bold in (2) for now. Let's see if you can reproduce my error. Please start with 480252a on grizzly, and try the following:
```
# standard gcc modules:
module unload python; 
make gfortran CORE=ocean OPENMP=true GEN_F90=true
cd testing_and_setup/compass/
./manage_regression_suite.py -m runtime_definitions/mpirun.xml -t ocean/regression_suites/nightly.xml -f config.ocean_turq -s --work_dir $WORKDIR
```
with your own `config.ocean_turq` file. I'm sure you know all that, but wanted to be clear how I got the error. Then on a compute node reload the modules and run, and you should see this:
```
cd $WORKDIR
./nightly_ocean_test_suite.py
      PASS
      PASS
      PASS
   ** FAIL (See case_outputs/Global_Ocean_240km_-_RK4_Thread_Test for more information)
```
where the first uses split_explicit (the default) and the second is the newly added RK4 test. We can look at the individual error. I like to use one core for simplicity now:
```
cd ocean/global_ocean/QU240/thread_test/2thread_run/
export OMP_NUM_THREADS=2
mpirun -n 1 ./ocean_model
```

<details>
<summary> and we get the error: </summary> <p>

```
mpirun -n 1 ./ocean_model

Backtrace for this error:

Backtrace for this error:
--------------------------------------------------------------------------
Sorry!  You were supposed to get help about:
    mpi_init:warn-fork
from the file:
    help-mpi-runtime.txt
But I couldn't find that topic in the file.  Sorry!
--------------------------------------------------------------------------

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x2AADD1FF2C57
#0  0x2AADD1FF2C57
#1  0x2AADD1FF1E50
#1  0x2AADD1FF1E50
#2  0x2AADD2EC33EF
#2  0x2AADD2C8062F
#3  0x2AADD4F41B4B
#4  0x2AADD4F3F7B3
#5  0x2AADD4C34533
#6  0x2AADD1BABD51
#7  0x2AADD2F528CB
#8  0x2AADD1FF2B42
#9  0x2AADD1FF1E50
#10  0x2AADD2EC33EF
#3  0xFC49C0 in __ocn_time_integration_split_MOD_ocn_time_integrator_split._omp_fn.10 at mpas_ocn_time_integration_split.F:966
#4  0x2AADD260808D
#5  0x2AADD2C78EA4
#6  0x2AADD2F8B8CC
#7  0xFFFFFFFFFFFFFFFF
#11  0xFC4B90 in __ocn_time_integration_split_MOD_ocn_time_integrator_split._omp_fn.10 at mpas_ocn_time_integration_split.F:968 (discriminator 20)
#12  0x2AADD26049DE
#13  0xFB9B6D in __ocn_time_integration_split_MOD_ocn_time_integrator_split at mpas_ocn_time_integration_split.F:940 (discriminator 4)
#14  0xFB3818 in __ocn_time_integration_MOD_ocn_timestep at mpas_ocn_time_integration.F:110
#15  0xE28CDF in __ocn_forward_mode_MOD_ocn_forward_mode_run at mpas_ocn_forward_mode.F:621
#16  0xF65AF9 in __ocn_core_MOD_ocn_core_run at mpas_ocn_core.F:111
#17  0x409AC1 in __mpas_subdriver_MOD_mpas_run at mpas_subdriver.F:347
#18  0x408893 in MAIN__ at mpas.F:16

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 16467 RUNNING AT gr0162
=   EXIT CODE: 11
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Segmentation fault (signal 11)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```
</details>

so we get an `invalid memory reference` at `mpas_ocn_time_integration_split.F:966`.  I looked:
```
 940                 !$omp parallel
 941                 !$omp do schedule(runtime) &
 942                 !$omp private(i, iEdge, cell1, cell2, sshEdge, thicknessSum, flux)
...
 964 flux = ((1.0-config_btr_gam1_velWt1) * normalBarotropicVelocitySubcycleCur(iEdge) &
 965 + config_btr_gam1_velWt1 * normalBarotropicVelocitySubcycleNew(iEdge)) &
 966 * thicknessSum
```
and the error is not obvious. But I find gnu debug messages uninformative. Maybe it's a race condition, or related to an allocated array like the last few problems.

@mattdturner could you look at this?

mark-petersen:
@mattdturner There are some loops without OpenMP directives:
```
src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
2205                do iEdge = 1, nEdges
2223                do iEdge = 1, nEdges
```
These are in the init subroutine, so performance doesn't matter. After this PR, what happens to these? Are they run by one thread or all? Do they potentially cause an error with multiple threads, or just not use all threads effectively?

The answers also apply to the init mode files in here (initi mode is only used to generate the initial conditions):
```
src/core_ocean/mode_init/
```
where we don't care about performance. Can we just go without OpenMP directives in init mode? If we do, and someone runs init mode with multiple threads, will it cause any problems?

philipwjones:
As long as there are no parallel directives, there are no threads so this is just running unthreaded. While it might not be important to execute in parallel, if these are initializing arrays that are being used in threaded regions later, it would be interesting to see if we get any first-touch affinity if we thread during init. Not something to worry about now, but a future experiment...

amametjanov:
Just curious about impact on throughput. Could you please post SYPD (or seconds/day) for 2-thread, Intel-optimized, oEC60to30v3 runs for comparison of numbers `before` and `after` the update?

mattdturner:
> Can we just go without OpenMP directives in init mode? If we do, and someone runs init mode with multiple threads, will it cause any problems?

@mark-petersen For this PR, I only made modifications to the OpenMP directives that already existed.  I didn't add OpenMP directives to any new loops, or remove them from any loops.  There will have to be a more extensive threading analysis at a later date (as @philipwjones mentioned).

But if we remove all OpenMP directives in init and somebody still runs with 2 threads, the only problem will be that the person would not be utilizing the threads.  The simulation would run without error, but they would have requested more resources in the queue than they are utilizing (wasting a bit of allocation).

I'll start looking at the error you reported in https://github.com/MPAS-Dev/MPAS-Model/pull/513#issuecomment-636090075

> Just curious about impact on throughput. Could you please post SYPD (or seconds/day) for 2-thread, Intel-optimized, oEC60to30v3 runs for comparison of numbers before and after the update?

@amametjanov My initial tests after making these changes showed a 3% performance slowdown as a result of these changes.  That was a while ago and when I had some OpenMP directives commented out, so the performance degredation should be smaller now.  Once I am able to get all the bugs resolved in this PR, I'll run a couple new tests to look at the performance impact.  I will note, though, that a small performance degradation is expected.  But the modifications allow for other optimizations in the future that should result in overall better performance.

mattdturner:
@mark-petersen I just pushed a new commit that commented out the `!$omp` directives for a single loop in `mpas_ocn_gm.F`.  After removing those directives, the `thread_test` and `rk4_thread_test` both passed on Grizzly using gnu w/ debug.  I'll have to come back to that loop (and 1 other in vmix, I believe) later to see if I can get those loops threaded properly...

mark-petersen:
@mattdturner Thanks, I think omitting omp directives on a few loops is fine for now. I rebased, and confirmed that it passes gnu with debug. Running additional tests now...

One question (perhaps for @philipwjones): This PR does not include analysis members. Should that be done as well in a future PR, or should we just let those be single-threaded?

mark-petersen:
This passes the nightly regression suite for MPAS-Ocean with gnu-debug, intel-debug, and gnu-optimized. This includes a bfb comparison between 1 and 2 threads. It is bit-for-bit on all tests against `ocean/develop`.

I merged it into `e3sm/testing`, which is available as a submodule in E3SM. It fails to complete on E3SM. See description at https://github.com/E3SM-Project/E3SM/pull/3596.

The advantage of `e3sm/testing` is that we can continue to revise this MPAS PR while testing in E3SM. I'm moving the E3SM discussion to https://github.com/E3SM-Project/E3SM/pull/3596.

philipwjones:
Anything that can be threaded, should be eventually or we have the usual Amdahl's law bottlenecks.  But I think we might want to explore some scheduling and affinity strategies for optimization first. 
For the analysis members, it's not always clear what to prioritize because we don't know which are used regularly (other than the ones currently turned on in the production sims).

mark-petersen:
This is now passes nightly regression suite with intel 19 on grizzly (thanks @pwolfram, your modules were already useful!). There was a variable `RediKappa` with `package='gm'` that was passed into a subroutine without being initialized. Only intel 19 with debug catches this as an error, not gnu or intel 17. Now passes:
```
```
in E3SM. Still fails
```
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_intel
```

mark-petersen:
```
PASS:
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-IAF.cori-knl_gnu
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_gnu
FAIL:
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-IAF.cori-knl_intel
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_intel
```
Dies with `State validation failed` which means T,S or u are out of bounds.

mattdturner:
@mark-petersen I just corrected a handful of bugs in the OpenMP directives.  It passed all `global_ocean` tests for Intel and GNU on Cori.  I still need to test these changes in E3SM to see if the FAILs are resolved.

mattdturner:
With the updates from yesterday, merging this branch into E3SM allows the following tests to all PASS for me:
- PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel
- PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu
- PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_intel

The nightly regression suite for MPAS-O also passes for intel and gnu

mattdturner:
Rebased.  

After rebasing, the MPAS-O nightly regression suite PASSes all tests if compiling with `DEBUG=true` and adding the `-O0` compilation flag.  If compiling with just `DEBUG=true`, but not including `-O0` then there are DIFFs of around O(-15).

I'll start running E3SM tests w/ these changes this afternoon...

mattdturner:
The following E3SM tests PASS:
```
```

The following tests FAIL:
```
PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_gnu
PEM_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel
```

The `PET_Ln9` run fails before it even enters the Ocean code.  It fails w/ a `Program received signal SIGILL: Illegal instruction.` error.  I'm not entirely sure what is going on there.

The `PEM_Ln9` run fails the `COMPARE_base_modpes` stage.  

I'll start trying to debug these 2 `P1024.ne30` tests.

---

Note that I tested this by doing the following in E3SM (partially here for my reference):
```
git checkout master
cd components/mpas-source
git fetch mturner     # I previously setup my MPAS fork as a remote
git merge mturner/ocean/local_openmp_threading
```

I had to also remove the `!$omp parallel` region from `components/mpas-ocean/driver/ocn_comp_mct.F`

I haven't done any comparison to baseline datasets, but I do expect some DIFFs in the E3SM tests based on what I saw in the MPAS-O standalone tests.

mattdturner:
Oddly enough, while `PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_gnu` FAILs, `PET_Ln9_D_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel` passes.

philipwjones:
@mattdturner I sometimes get that illegal instruction error when the compiler has built for the wrong target processor. Looks like this was the only combination of gnu/threading/knl.  Might try rebuilding just to see if it's a bad build or try building an older version to see if this particular combination config is messed up?

mattdturner:
@philipwjones I reran the `PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_gnu` test, and it failed with the same illegal instruction error.  I'll have to checkout `master` and re-run to confirm that the fail isn't a result of my changes.

mattdturner:
I ran `PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_gnu` and `PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel` w/ E3SM `master`.

`PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_gnu` FAILs with the same `illegal instruction` error.  

I'm assuming that I can ignore this failure for this PR, since `master` has an identical failure and it is not related to (or solvable by) threading changes.

`PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel` also FAILs the comparison test.  

@philipwjones and/or @mark-petersen , do either of you have any insights on which files I should focus on to try and determine the source of the comparison failure?  The lower resolution simulations (`QU120`, `QU240`, `EC60to30`) are all bit-for-bit between thread counts, but the `ne30_oECv3_ICG` resolution is not.  Even though `master` has the same failure, since it is likely a threading error I think it should be fixed in this PR.

mattdturner:
Quick update on the `PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel` failure.  I commented out all OpenMP directives in the ocean code, and re-ran the test.  It still fails the comparison step.  So I am assuming that the error is not in the ocean code.  

Although, the same test ([PET_Ln9_PS.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel.allactive-mach-pet](https://my.cdash.org/test/7226789)) appears to PASS on sandiatoss3 so I am not sure why the test is failing for me on Cori...

mattdturner:
Since the 2 above FAILures also fail in `master`, even with all openmp disabled in the ocean model, I've decided that the errors should not be investigated / resolved in this PR.

I ran `PET_Ln9_P4096.T62_oRRS18to6v3.GMPAS-IAF.cori-knl_intel` w/ both `master` and `master` updated with the changes in this branch.  The test FAILs on `master`, but PASSes with this branch.

So, based on all of the testing for both lower and higher resolutions, I think this branch is ready to be merged.

PR #419, opened 2020-01-06
Add Arctic Regional Mesh to COMPASS

mark-petersen:
We had created the Arctic Regional Mesh (ARM) in 2018 with a matlab version. This adds the ability to auto-generate these meshes in COMPASS using our standard python tools, and calling jigsaw from the python package. In particular, a new function `AtlanticPacificGrid` was added to allow different mesh distributions in the two basins.

PR #515, opened 2020-04-08
Switch ARM creation to signed distance function

mark-petersen:
As a warm-up to revising the EC60to30 and CUSP mesh, I switched ARM from the hard-wired function `AtlanticPacificGrid` to signed distance based on a geojson file. The new steps are:
- Create cell width vs latitude for Atlantic and Pacific basins
- Expand from 1D to 2D
- Signed distance of Atlantic region
- Merge Atlantic and Pacific distrubutions smoothly
- Merge Atlantic and Pacific distrubutions with step function
- Create a land mask that is 1 over land
- Merge: step transition over land, smooth transition over water


mark-petersen:
The region (used http://geojson.io/) and signed distance function:
image

testing_and_setup/compass/ocean/global_ocean/ARM60to10/init/define_base_mesh.py
```diff
@@ -1,41 +1,123 @@
...
+    #                      name='signed_distance')
+    #cw_filename = 'signed_distance.nc'
+    # da.to_netcdf(cw_filename)
+
+# replace this:
+    #cellWidth = mdt.AtlanticPacificGrid(lat, lon, AtlGrid, PacGrid)
 
     import matplotlib.pyplot as plt
     import matplotlib
     matplotlib.use('Agg')
```
xylar, line 86: Not directly related to the PR, but this should get moved to the top and should be reordered:
```
    import matplotlib
    matplotlib.use('Agg')
    import matplotlib.pyplot as plt
```
The switch to the `Agg` backend should be before importing `plt`.

testing_and_setup/compass/ocean/global_ocean/ARM60to10/init/define_base_mesh.py
```diff
@@ -1,41 +1,123 @@
...
 import jigsaw_to_MPAS.mesh_definition_tools as mdt
+from jigsaw_to_MPAS.coastal_tools import signed_distance_from_geojson, \
+    mask_from_geojson
+from geometric_features import read_feature_collection
+import xarray
+
+# Uncomment to plot the cell size distribution.
+# import matplotlib
+# matplotlib.use('Agg')
+# import matplotlib.pyplot as plt
```
xylar, line 11: You're doing this later so should just comment this in.

mark-petersen:
See previous results here: https://github.com/MPAS-Dev/MPAS-Model/pull/419#issuecomment-571161757
@xylar, I'm not quite ready for a detailed review. I just wanted to show you that your signed distance function is great for these meshes. I'll update this code later today.

mark-petersen:
This shows the use of land masks to combine a step transition over land, smooth transition over water. Here I just use 1 and 2 to clearly show the transitions.
cellWidthGlobal

testing_and_setup/compass/ocean/global_ocean/ARM60to10/init/define_base_mesh.py
```diff
@@ -1,41 +1,123 @@
...
+    ax.grid(True)
     plt.xlabel('latitude')
     plt.title('Grid cell size, km')
     plt.legend()
-    plt.savefig('cellWidthVsLat.png')
+
+    var = signed_distance
+    ax = axs[1, 0]
+    ax = plt.axes(projection=ccrs.PlateCarree())
+    ax.set_global()
```
mark-petersen, line 105: subplot with cartopy

mark-petersen:
Here is the final plot, with masks and cell widths so the user can understand the steps. I was able to use subplot and cartopy together.
mesh_construction

mark-petersen:
@xylar this is done. I'm happy to hear your comments, as this method will be used in the other upcoming meshes.

testing_and_setup/compass/ocean/global_ocean/ARM60to10/init/define_base_mesh.py
```diff
@@ -1,41 +1,131 @@
...
+    Americas_land_mask = mask_from_geojson(fc, lon, lat)
+    fc = read_feature_collection('Europe_Africa_land_mask.geojson')
+    Europe_Africa_land_mask = mask_from_geojson(fc, lon, lat)
+    landMask = np.fmax(Americas_land_mask, Europe_Africa_land_mask)
+
+    # Merge: step transition over land, smooth transition over water
+    cellWidth = cellWidthStep * landMask + cellWidthSmooth * (1 - landMask)
+
+    # save signed distance to a file
+    # da = xarray.DataArray(signedDistance,
```
mark-petersen, line 97: @xylar this is why we were confused on the phone. In the SO60to10 you had these lines to save signed distance, not cell width. Is signed distance useful in a netcdf file, since I'm making a plot of it below? I think I can delete these lines.

xylar: No, that was just for debugging.  But I definitely do want the cell widths saved to a NetCDF file but I think the code that calls `define_base_mesh.py` actually does the saving so you don't need to do it.

xylar:
It's late here but I'll give this a look tomorrow.

testing_and_setup/compass/ocean/global_ocean/ARM60to10/init/define_base_mesh.py
```diff
@@ -1,41 +1,131 @@
...
+        'signedDistance',
+        'maskSmooth',
+        'cellWidthSmooth',
+        'maskSharp',
+        'cellWidthSharp',
+        'landMask',
+        'cellWidth']
+    j = 2
+    for varName in varNames:
+        plot_cartopy(j, varName, vars()[varName])
```
xylar, line 126: I've never used `vars()` to get the dictionary of local variables like this but it seems like a clever approach to me.

xylar:
It would be interesting to see a plot of cell area on the culled mesh if you happen to have that easily accessible in ParaView.  if not, no worries.

PR #518, opened 2020-04-12
SOwISC12to60E2r02 mesh: coarser Southern Ocean, high res Arctic and North Atlantic

xylar:
The "background" mesh before regional refinement now has 45-km resolution in southern mid latitudes and 25-km resolution near the south pole, before the 12-km refinement is applied around Antarctica.  Resolution is also enhanced in the Arctic, with 15-km resolution around Greenland, tapering off to 30-km resolution at the Arctic margins and in the northern part of the Gulf Stream in the North Atlantic.


xylar:
This has been done at @maltrud's request (will he ever be satisfied with the SO grid?!).

@maltrud's cartoon:
image

The requested resolution (passed to JIGSAW) at these longitudes:
res_vs_lat

The map of resolution passed to JIGSAW:
cellWidthGlobal

xylar:
@mark-petersen has also mentioned wanting to add a region of enhanced resolution (30 km?) in the North Atlantic, based on discussions with FESOM developers at AWI.  This might be a good place to add that.

xylar:
Current | Proposed
-- | --
SOwISC_current  | SOwISC_coarser
SOwISC_Ross_current | SOwISC_Ross_coarser  
nCells = 739904 ; nEdges = 2234478 ; nVertices = 1494156 ; | nCells = 729522 ;nEdges = 2203046 ;nVertices = 1473097 ;

xylar:
## Current
image

```
nCells = 739904 ;
nEdges = 2234478 ;
nVertices = 1494156 ;
```
## Coarser

image

```
nCells = 729522 ;
nEdges = 2203046 ;
nVertices = 1473097 ;  
```

It seems like the reduction in cells because of the coarsening was compensated (mostly) by the high-res region creeping north.  This is likely because I made the base polar mesh have 25 km resolution instead of 30 km.  I'm going to change that and see what happens.

xylar:
@maltrud and anyone else, the data for this is available on LANL IC at:
/lustre/scratch4/turquoise/xylar/test_SOwISC_coarser_SH/ocean/global_ocean/SO60to10wISC/init/

xylar:
## Tighter 30-km resolution around the equator

To further reduce the number of cells, @mark-petersen has suggested reducing the high-res region around the equator by a factor of 2 (from 30 degrees latitude wide to 15 degrees wide).  Below are results from that mesh.

res_vs_lat

cellWidthGlobal

drake_res_vs_lat


The resolution of the proposed mesh:
SOwISC_tight_eq

For comparison, here is the current SOwISC (before this PR) at the same angle:
SOwISC_current_eq

xylar:
## Another variant: 12 km finest resolution, 1600 km transition 

Previous was 10 km finest, 1100 km transition

@maltrud's request:
unnamed

What I came up with:
res_vs_lat

cellWidthGlobal

drake_res_vs_lat

SOwISC_12km_weddell

Current version from `ocean/develop` for comparison:

SOwISC_current_weddell

```
nCells = 389042 ;
nEdges = 1178412 ;
nVertices = 789030 ;
```

xylar:
## Higher resolution in the North Atlantic

As requested at the last E3SM Cryosphere meeting, @maltrud came up with the following request for enhanced resolution:

image

"Maybe something like a linear increase in grid scale from 15 km to ~30 km away from the coast of Greenland, but enclosed by the circle drawn in black."

Here's what I came up with as a first iteration:
SOwISC_12km_high_na_arctic

```
nCells = 420112 ;
nEdges = 1274050 ;
nVertices = 853561 ;
```
Number of cells is only about 30k more than the previous version (a nearly negligible change)!

For comparison, the "current" mesh (`ocean/develop`) with the same colorbar:
SOwISC_current_arctic

Here are the usual plots of approximate resolution vs. lat (and sometimes lon):

res_vs_lat
cellWidthGlobal
drake_res_vs_lat

Here are the shapes that define the contour of ~35 km resolution:
https://github.com/xylar/MPAS-Model/blob/coarser_SOwISC/testing_and_setup/compass/ocean/global_ocean/SO60to10wISC/init/north_mid_res_region.geojson

and of 15 km resolution (Greenland):
https://github.com/xylar/MPAS-Model/blob/coarser_SOwISC/testing_and_setup/compass/ocean/global_ocean/SO60to10wISC/init/greenland.geojson

These are very easy to edit (e.g. on http://geojson.io) so feel free to suggest changes.

xylar:
I notice that the new mesh has the Black Sea and the "current" one doesn't.  I wonder if we should put in a critical passage to make sure we do have the Black Sea or a critical land bridge to make sure we don't.

milenaveneziani:
Thanks @xylar!
I think the transition from 20 to 30 km (current transition zone for variable GM) is covering about half of the subpolar gyre at the moment.  My suggestion is to move the center of the circle more  toward the center of the subpolar gyre, and also make  the transition at lower latitudes (toward the GS extension) sharper, so that we don't transition GM right in the middle of the GS extension either.
I personally wouldn't worry about the Black Sea.

xylar:
> I think the transition from 20 to 30 km (current transition zone for variable GM) is covering about half of the subpolar gyre at the moment. My suggestion is to move the center of the circle more toward the center of the subpolar gyre, and also make the transition at lower latitudes (toward the GS extension) sharper, so that we don't transition GM right in the middle of the GS extension either.

I don't speak Arctic or North Atlantic.  What and where is the GS?

If you want to modify the shape(s) as you see fit, I'm happy to rerun.

> I personally wouldn't worry about the Black Sea.

Let it randomly appear in different meshes?  Or take it out?

mark-petersen:
I think Black and Caspian Seas are in the same boat :). We should decide if they should both be in or both be out, and then force them in or out. If they don't influence the global ocean, it seems like they should both be out.

xylar:
> I think Black and Caspian Seas are in the same boat :). We should decide if they should both be in or both be out, and then force them in or out. If they don't influence the global ocean, it seems like they should both be out.

There's probably no danger of getting the Caspian Sea even at high res because we don't resolve the connection and it gets culled.  I can add a critical land bridge so the same happens with the Black Sea if there's general agreement.

milenaveneziani:
GS=Gulf Stream.
Let me see if I can render in graphics what I am thinking about.

xylar:
> GS=Gulf Stream.

Ah, of course!

> Let me see if I can render in graphics what I am thinking about.

That'd be great!

milenaveneziani:
Here is one cartoon, but imagine that the 20-30 km transition wraps around  the high-res circle.
Please also wait for @maltrud to weigh in before changing anything.
SOmeshNArefinement

jonbob:
As for the Black and Caspian Seas, we do need to be consistent because it impacts other e3sm components. We had some trouble awhile back when MOSART assumed they were ocean and tried to add runoff to those cells -- and all the other components considered them inland seas.

xylar:
Good to know!

maltrud:
i agree with @milenaveneziani --if it's possible to move the center of the circle southward to move the 30km transition as she diagrammed.  @xylar this is really excellent.
as for Black/Caspian:  we need a land bridge as @jonbob noted.  for example, the 18to6 has them filled in.

xylar:
Okay, adding a Black Sea land bridge is a totally separate issue to be dealt with elsewhere but definitely noted.

I'm working on a mesh that (roughly at least) follows @milenaveneziani's suggestions.  I'll refine as I have time and repost tomorrow.  Thanks everyone for the input!  Keep it coming!

xylar:
Here's the place where I'm dealing with the Black Sea issue: https://github.com/MPAS-Dev/geometric_features/pull/136

xylar:
@milenaveneziani, what do you think of this -- the colorbar is 15 to 30 km:

SOwISC_12km_high_na2_15to30_arctic

And with the original color bar:
SOwISC_12km_high_na2_arctic

maltrud:
@xylar i think this looks really good.  if others are on board, i'd like to test it.

xylar:
Size of the latest mesh:
```
nCells = 425809 ;
nEdges = 1291198 ;
nVertices = 865006 ;
```
Very similar to the last iteration.

milenaveneziani:
I  agree, this looks great. Thanks @xylar!

xylar:
I'm initializing this here on LANL IC:
/users/xylar/scratch4/test_SOwISC_coarser_SH_high_res_NA/ocean/global_ocean/SO60to10wISC/init/
If all goes well, I'll run spin-up tomorrow morning.

xylar:
I also noticed that the Black Sea is absent again in this iteration on my laptop (unrelated to MPAS-Dev/geometric_features#136, which I am not using yet) but that certainly might not be true for my LANL IC run, since JIGSAW behaves differently on different machines.

xylar:
I rebased this branch onto `ocean/develop` after realizing the mesh metadata wasn't there.  I'm going to re-test the initialization and spin-up, still in:
```
/lustre/scratch4/turquoise/xylar/test_SOwISC_coarser_SH_high_res_NA/ocean/global_ocean/SO60to10wISC/
```
I'll let everyone know once it's done.

I think we should consider this revision 2 of the mesh (and what @darincomeau is currently using revision 1) even if neither of these meshes live very long.  Because of this, I think it is important to make sure this PR gets merged to `ocean/develop` and a new spin-up gets run so that the official revision 2 mesh can more easily be recreated in the future if need be.  Because of that, I'll ask everyone to approve this PR based on my plots or whatever else we can discern from the spin-up rather than waiting for an E3SM run.  There will most certainly be a revision 3...

mark-petersen:
@darincomeau @maltrud and @proteanplanet, we would like to finalize this "r02" revision based on the discussion above. Approving this PR just means approving the mesh version to be tested in E3SM, though it can be revised to a new "r03" based on that experience.

Please approve or make further comments today, if possible.

xylar:
> minor point - we didn't use 'L64' in the r01 version of this mesh; should it be included here? I just want to make sure I'm consistent with the agreed upon naming convention.

Good point, the "short name" is without that.  I'll edit.

>  Also, where'd we end up with the horizontal cell count?

I believe https://github.com/MPAS-Dev/MPAS-Model/pull/518#issuecomment-614834164 is still accurate or close enough, so
 ```
nCells = 425809
```

maltrud:
@mark-petersen this is just for mpas, right?  this won't be propagated to e3sm, correct?

xylar:
Right, the COMPASS changes might propagate to E3SM but that isn't the same as the mesh being available for use in E3SM.  That part won't happen just based on this PR.

maltrud:
thanks for the clarification, @xylar

mark-petersen:
@xylar since it's the end of the day here, I'll let you rebase and make any adjustments. You can merge with the button if you like.

xylar:
## Latest

Here's the latest resolution plot.  I'm not going to re-run spin-up until this gets merged to `ocean/develop` but will run through `initital_state` to make sure no problems emerge.

cellWidthGlobal

mark-petersen:
PS: A mesh generation timing test took 21:58 on Chrysalis on one node.

PR #457, opened 2020-02-27
Ocean: Remove scratch allocates and calls to get_config

mattdturner:
This PR replaces almost all calls to `mpas_allocate_scratch_field` in the ocean model with straight Fortran allocates.  This is done as part of the MPAS framework rewrite, per @philipwjones .  

This PR also replaces almost all calls to `mpas_pool_get_config` in subroutines with definitions in `ocn_config` via `use ocn_config`.  This addresses #394 , and allows #406 to be closed without being merged.

There are still a few calls to `mpas_allocate_scratch_field` that cannot be replaced until more of the MPAS framework rewrite has been completed (namely removal of `blocks`).

This PR only changes the files in the shared subdirectory, not the mode_forward subdirectory.

PR #519, opened 2020-04-14
Ocean/tracer advection optimization

mattdturner:
Optimizations to tracer transport that include
- Removing pointer retrievals
- Removing unused variables and calculations
- Trimming argument lists to remove Pool variables (no longer used in these routines)
- Add the use of `ocnMesh` (relies on PR #496 being merged)

This PR requires the following PRs be merged first:
- #496 
- #457 
- #513 

Some of the changes in those PRs were either `git cherry-pick`'ed or manually added.  As a result, once those PRs are merged there might need to be some commits squashed or changes reverted in this PR.

### DIFF
--- a/src/core_ocean/mode_init/Makefile
+++ b/src/core_ocean/mode_init/Makefile
@@ -83,6 +83,8 @@ mpas_ocn_init_hurricane.o: $(UTILS)
 
 mpas_ocn_init_tidal_boundary.o: $(UTILS)
 
+mpas_ocn_init_dam_break.o: $(UTILS)
+
 #mpas_ocn_init_TEMPLATE.o: $(UTILS)
 
 clean:
--- a/src/core_ocean/shared/Makefile
+++ b/src/core_ocean/shared/Makefile
@@ -82,7 +82,7 @@ mpas_ocn_mesh.o:
 
 mpas_ocn_thick_ale.o: mpas_ocn_constants.o mpas_ocn_config.o
 
-mpas_ocn_time_average_coupled.o: mpas_ocn_constants.o mpas_ocn_config.o mpas_ocn_tracer_ecosys.o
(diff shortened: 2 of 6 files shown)


mattdturner:
This was tested using a QU240 test case, and a EC60to30 test case (both generated via Compass).  The results are bit-for-bit for debug cases (and should be for optimized cases as well) with Intel compiler on Cori.

As a result of the optimizations in this PR (specifically, removing the unused calculations for `computeBudgets`), there is a reduction of the time spent in `tracer adv` by 16.4% (average of 3 runs for both baseline and optimized)

mattdturner:
I updated this PR to include the recent changes to PR #496 and the necessary updates to tracer advection resulting from the changes in #496

mattdturner:
I ran into some issues resolving merge conflicts during rebase, so instead I started over and re-applied the relevant changes.  

I had to leave the `computeBudgets` code in since it calculates values used elsewhere.  I either missed that previously, or there were recent changes that now use the arrays.  As a result, the performance improvement drops from 16.4% in `tracer adv` to 1.6%.  This PR is still necessary, though, since it updates the tracer advection routines to use the `ocn_mesh` module.

This is BFB when compiled w/ intel (both DEBUG and OPT).

I need to test this in E3SM before its deemed ready.

mattdturner:
PASSes
```
```

So I think this is ready to go into E3SM.

mark-petersen:
Rebased to be safe. Then tested nightly regression suite. Passes all with gnu 6 debug on badger, intel 19 debug on grizzly, and is bfb in optimized with gnu 6 and intel 19 against same compilers with branch `ocean/develop`.

I also looked through the code. This all looks correct. Thanks @mattdturner for your work. I love the simplification of obtaining mesh variables from the module.

mark-petersen (approved):

mark-petersen:
@jonbob this is ready and will need it's own E3SM PR. We can merge this PR when we have a free slot.

PR #522, opened 2020-04-17
New Mesh: EC15to60kmL64E3SMv2r01: Draft of new EC mesh with refinement around Greenland

xylar:
Mesh long name: `EC15to60kmL64E3SMv2r01`

Adds a new EC15to60 test case with refinement around Greenland

This is just do document a mesh @maltrud and I are playing around with that might eventually inform a future EC, low res mesh for v2.


xylar:
## Plots of resolution

cellWidthGlobal

arctic

south_atlantic

testing_and_setup/compass/ocean/global_ocean/EC15to60/spin_up/config_prep_spin_up1.xml
```diff
@@ -0,0 +1,44 @@
...
+
+
+	<namelist name="namelist.ocean" mode="forward">
+		<template file="template_forward.xml" path_base="script_configuration_dir"/>
+		<template file="template_forward.xml" path_base="script_resolution_dir"/>
+		<option name="config_run_duration">'00-00-10_00:00:00'</option>
+		<option name="config_dt">'00:10:00'</option>
+		<option name="config_btr_dt">'00:01:00'</option>
+		<option name="config_Rayleigh_friction">.true.</option>
+		<option name="config_Rayleigh_damping_coeff">1.0e-4</option>
```
xylar, line 15: Probably need to be adjusted

testing_and_setup/compass/ocean/global_ocean/EC15to60/template_forward.xml
```diff
@@ -0,0 +1,12 @@
+<template>
+	<namelist>
+		<option name="config_dt">'00:10:00'</option>
+		<option name="config_btr_dt">'00:00:20'</option>
+		<option name="config_pio_num_iotasks">4</option>
+		<option name="config_pio_stride">36</option>
+		<option name="config_run_duration">'0000_01:00:00'</option>
+		<option name="config_mom_del4">1.5e10</option>
+		<option name="config_hmix_scaleWithMesh">.true.</option>
+		<option name="config_use_GM">.true.</option>
```
xylar, line 10: Probably need to be adjusted

xylar:
The new EC grid can be found here:
```
/lustre/scratch4/turquoise/xylar/test_EC15to60/ocean/global_ocean/EC15to60/
```
I'm still spinning it up (and probably doing so very inefficiently) but this should be a good starting point for you to play around with the new mesh.

For now (for comparison with the EC60to30 mesh), we're running with 60, not 64 vertical levels:
```
nVertLevels = 60 ;
nCells = 220321 ;
nEdges = 671300 ;
nVertices = 450694 ;
```

vanroekel:
@xylar or @mark-petersen  before this develops too much further, can we socialize this mesh with water cycle as well, especially if there are discussions within cryosphere of this potentially being the default v2 low res mesh?  I don't think anyone outside cryosphere has been a part of this discussion for developing this mesh.  Looking at this, I like the cell count, but am worried the equatorial region is too coarse.  I'm concerned that the transition zone will harm tropical instability wave fidelity and potentially degrade ENSO fidelity as well.

maltrud:
@vanroekel i've been fully involved with this, even safe to say i've been driving it.  i want to emphasize that it is not intended to be the default v2 mesh, but if it looks good it could be a prototype.  my main purpose is to see if a ec^0to30-like grid with ~15km resolution in the subpolar gyre gives us a more realistic AMOC.  i didn't notice if the latest has the narrower equatorial band--if it does, that could be interesting to test, as well.

xylar:
@vanroekel, I can't find Chris Golaz's GitHub username.  Feel free to mention him or add him as a reviewer.

xylar:
@mark-petersen's argument for tighter resolution around the equator was based on:
1. a discussion with FESOM developers at AWI that suggested this and
2. The following figure suggesting that this should be sufficient to capture the major equatorial current:
image

However, I'm sure discussion is welcome, and would also help us with revisions to our SO refined mesh (#518)

vanroekel:
@xylar perhaps I'm not clear from the figures about the equatorial region.  Is the 30 km region 7.5degrees or 10 or something else?  The figure you post suggests we should use 10, which I think is perfect, but one of your figures, I think on the SO mesh one seem to show a transition at 7.5.  Could you let me know what your cutoff is?  Thanks

vanroekel:
one other question.  Would you like me to suggest any parameter settings for this mesh?  I saw a few spots saying values should be changed.  I'm happy to help.  One thing I want to make sure of is that we use the config_eddy_resolution_taper = 'ramp' and not constant.

testing_and_setup/compass/ocean/global_ocean/EC15to60/init/define_base_mesh.py
```diff
@@ -0,0 +1,101 @@
...
+    dlat = dlon
+    nlon = int(360./dlon) + 1
+    nlat = int(180./dlat) + 1
+    lon = np.linspace(-180., 180., nlon)
+    lat = np.linspace(-90., 90., nlat)
+
+    cellWidthSouth = mdt.EC_CellWidthVsLat(lat, cellWidthEq=30.,
+                                           cellWidthMidLat=60.,
+                                           cellWidthPole=35.,
+                                           latPosEq=7.5, latWidthEq=3.0)
```
xylar, line 35: @vanroekel, the edge of the 30-km region around the equator is at +/- 7.5 degrees (`latPosEq`).  This could be increased to 10.0.  I would probably also increase `latWidthEq` to 4.0.

testing_and_setup/compass/ocean/global_ocean/EC15to60/init/define_base_mesh.py
```diff
@@ -0,0 +1,101 @@
...
+    print(cellWidthSouth)
```
xylar, line 36: Needs to be removed.

testing_and_setup/compass/ocean/global_ocean/EC15to60/init/define_base_mesh.py
xylar: @mark-petersen had suggested that the counter current might not need high resolution, which is why we had halved the high-res region from 30 degrees total to 15 degrees total.  It sounds like @vanroekel thinks it needs high res.

xylar:
> one other question. Would you like me to suggest any parameter settings for this mesh? I saw a few spots saying values should be changed. I'm happy to help. One thing I want to make sure of is that we use the config_eddy_resolution_taper = 'ramp' and not constant.

It doesn't hurt to change these parameters here for the spin-up but it's probably more important for the E3SM pull request when that happens.  I'll point out where these changes would go and you can use the GitHub suggestion feature to essentially edit the code here if you want.

testing_and_setup/compass/ocean/global_ocean/EC15to60/template_forward.xml
xylar: @vanroekel, this would also be where to add other namelist changes.  Hit the "Insert a suggestion" icon and edit the code as you would like.

xylar:
@vanroekel, here's the same but with a 20-degree Equatorial region instead of 15:

cellWidthGlobal

testing_and_setup/compass/ocean/global_ocean/EC15to60/init/define_base_mesh.py
xylar: I updated this as stated above.

lconlon:
I think this version looks pretty good, subpolar gyre-wise. What is the max GM that you're using?

xylar:
> What is the max GM that you're using?

I don't think we've got far enough to decide that yet, and would certainly welcome input.  @maltrud, @milenaveneziani, etc. what is the default?

vanroekel:
I'd imagine we would use a value of 600, which is the default for other EC6030 Gcases, but it is something we could and maybe should test.

lconlon:
> I'd imagine we would use a value of 600, which is the default for other EC6030 Gcases, but it is something we could and maybe should test.

Sounds good. Happy to help with any testing if needed.

milenaveneziani:
We used a max GM of 600 for the ARRM meshes, but in those simulations we don't have 3d GM.

xylar:
@maltrud, is testing progressing on this mesh at all?  Or are you still stuck?  No rush, since you were the one that requested this one but figured we could put in some results here if we wanted.

At some point, we probably want to merge this just to document this version of the mesh, even if it isn't our final E3SM v2 EC mesh.

vanroekel:
@maltrud could you summarize your testing on this mesh?  I seem to remember you saying we won't want this for v2, but don't remember the reason.  As @xylar mentions in a comment above we should merge this to document, but it would be great to have a summary of what we might want to change for an r02 version of this mesh.

xylar:
It sounds like this hadn't made it far enough in testing that we can use it for E3SM v2.  Nevertheless, I would like to have this rebased and merged for documentation purposes.

xylar:
Actually, I don't want to merge until #563 and #577

xylar:
Unfortunately, this branch was never merged and COMPASS has moved on to the point where it is not feasible to merge now.

I have created a branch on the new COMPASS repo to preserve this branch in case we need to recreate the mesh: https://github.com/MPAS-Dev/compass/tree/EC15to60E2r1

PR #523, opened 2020-04-18
Convert CUSP to signed distance

mark-petersen:
Re-design of CUSP mesh using a number of overlaid regions. Each region begins with a signed distance from a geojson region file. A mask is computed using a tanh function on the signed distance.


mark-petersen:
@xylar, it would be great to hear your thoughts on this. The details are preliminary, but the process is in place. It automatically generates this plot, which shows how there are five masks, and each one adds to the high resolution region. There are very few points in each geojson file (also added to this PR) so the regions are easy to alter.
mesh_construction
cellWidthGlobal

testing_and_setup/compass/ocean/jigsaw_to_MPAS/mesh_definition_tools.py
```diff
@@ -65,7 +65,16 @@ def mergeCellWidthVsLat(
...
+def EC_CellWidthVsLat(lat,
+    cellWidthEq = 30.0,
+    cellWidthMidLat = 60.0,
+    cellWidthPole = 35.0,
+    latPosEq = 15.0,
+    latPosPole = 73.0,
+    latTransition = 40,
+    latWidthEq = 6.0,
+    latWidthPole = 9.0
+    ):
```
xylar, line 77: @mark-petersen, I did this work in #518 so we're competing with one another on this.  I made the docstring have numpydoc formatting (allowing it to be automatically parsed and formatted the arguments in a more standard way -- one line per argument really isn't a standard practice.  In any case, we need to reconcile these differences before the next mega-merge.

testing_and_setup/compass/ocean/global_ocean/CUSP12/init/define_base_mesh.py
```diff
@@ -1,187 +1,375 @@
...
+    # prepare plots
+    fig = plt.figure()
+    plt.clf()
+    fig.set_size_inches(10.0, 14.0)
+    cmapBluesHalf = truncate_colormap(cmapIn='Blues', minval=0.0, maxval=0.7)
+
+    # Create cell width vs latitude for Atlantic and Pacific basins
+    QU1 = np.ones(lat.size)
+    EC60to30 = mdt.EC_CellWidthVsLat(lat)*km
+    EC60to30Narrow = mdt.EC_CellWidthVsLat(lat, latPosEq = 8.0, latWidthEq = 3.0)*km
```
xylar, line 85: @vanroekel seems to think in #522 that we want the equator region to be at least 10 degrees wide (probably with a 4.0 degree transition for consistency).  I think we want to be consistent between all the meshes, so this should probably do the same, as should #518

testing_and_setup/compass/ocean/global_ocean/CUSP12/init/define_base_mesh.py
```diff
@@ -1,187 +1,375 @@
...
+    plot_cartopy(2, 'narrow EC60to30', cellWidth, 'rainbow')
+    plotFrame = 3
+
+    # global settings for regionally refines mesh
+    highRes = 12.0*km
+
+    fileName = 'coastline_CUSP'
+    transitionOffset = 800.0*km
+    transitionWidth = 600.0*km
+    fc = read_feature_collection(fileName + '.geojson')
```
xylar, line 98:
```suggestion
    fc = read_feature_collection('{}.geojson'.format(fileName))
```

xylar: And similarly elsewhere.

xylar:
Please rebase onto #526 and #527 (and make any changes to those PRs if you're not happy with them).  Otherwise, we're competing with one another.

testing_and_setup/compass/ocean/global_ocean/CUSP12/init/define_base_mesh.py
mark-petersen: done, thanks.

mark-petersen:
Rebased on ocean/develop. This takes care of @xylar's requests above.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/mesh_definition_tools.py
mark-petersen: done.

mark-petersen:
Rebased and updated.
cellWidthGlobal
mesh_construction

mark-petersen:
@xylar, I would like to have two PRs for the CUSP mesh. This one is just implementing the signed distance function and the plots, and it is complete. Then I will open a new one with the mesh label, that can be open for discussion for a while. That way we separate the first implementation from the group discussions.

You already reviewed it once, but I'm happy to hear further comments. Otherwise, please approve and I'll merge it in.

PR #524, opened 2020-04-18
Fix missing MLD and globalStats streams

### LINKED ISSUE #521: Missing stream(s) in spin-up configs
As part of #445, changes were made to how streams are added to some forward config files.  However, it seems that not all files were modified that needed to be and, as a result, the `spin_up` stage of at least the EC60to30 and SO60to10wISC test case, and perhaps others, are broken.

Specifically, https://github.com/MPAS-Dev/MPAS-Model/commit/447df4e5d608496b0a6420ec250433fad9dbf174 missed the file `EC60to30/spin_up/config_prep_spin_up1.xml` and `SO60to10wISC/spin_up/config_prep_spin_up1.xml`.

xylar:
The template for forward global_ocean runs was updated to turn on the MLD analysis member but the associated output stream was only included in some steps, not all, causing many test cases to be broken.

In many steps, the global stats analysis was being turned on in the namelist but the associated stream was missing.

This merge also adds links to `forcing_data.nc` in `ssh_adjustment` steps because, although this data should not be needed in these steps, error log files are generated because the stream points to non-existent file.

closes #521

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/CUSP12/spin_up/config_spin_up1.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/CUSP12/spin_up/config_spin_up1.xml
@@ -32,6 +32,7 @@
 			<attribute name="output_interval">00-00-00_06:00:00</attribute>
 		</stream>
 
+		<template file="global_stats.xml" path_base="script_core_dir" path="templates/analysis_members"/>
 		<template file="template_forward.xml" path_base="script_configuration_dir"/>
 	</streams>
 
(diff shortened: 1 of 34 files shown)


xylar:
## Testing

I'm testing this on Grizzly and my laptop right now.  I'll check off the affected test cases as they pass:
- [ ] CUSP12
- [ ] CUSP8
- [ ] EC60to30
- [ ] EC60to30wISC
- [x] QU240
- [x] QU240wISC
- [ ] QU60
- [ ] SO60to10wISC

xylar:
`CUSP12/spin_up/est_final_settings` fails for other reasons (see #525) that I will address here shortly.

PR #526, opened 2020-04-19
Generalize EC cell widths function

xylar:
Many supposed parameters are actually hard-coded. Documentation has been updated to adhere to numpydoc formatting. Unnecessary looping has been removed.

This merge also improves plotting of cell widths in `define_base_mesh`.  Imports are now at the top of the file. Colormap is now perceptually uniform (not "jet"!) Plot layout is now tighter (though there's still considerable whitespace).


xylar:
@sbrus89 and @pwolfram, these changes may affect you depending on what you use as your background mesh.  Either way, these changes would be a good template for future changes to `mesh_definition_tools.py`

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
```diff
@@ -53,17 +56,14 @@ def build_mesh(
...
             fig = plt.figure()
             fig.set_size_inches(16.0, 8.0)
             plt.clf()
             ax = plt.axes(projection=ccrs.PlateCarree())
             ax.set_global()
-            im = ax.imshow(cellWidth, origin='lower', transform=ccrs.PlateCarree(
-            ), extent=[-180, 180, -90, 90], cmap='jet')
+            im = ax.imshow(cellWidth, origin='lower',
+                           transform=ccrs.PlateCarree(),
+                           extent=[-180, 180, -90, 90], cmap='viridis_r')
```
xylar, line 66: Make this `rainbow`.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
```diff
@@ -76,6 +76,8 @@ def build_mesh(
             gl.ylabels_right = False
             plt.title('Grid cell size, km')
             plt.colorbar(im, shrink=.60)
+            fig.canvas.draw()
+            plt.tight_layout()
```
xylar, line 80: This cleans up some whitespace.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
pwolfram: Please don't-- this doesn't add value and rainbow is a notoriously bad colormap.

pwolfram: (unless rainbow is the perceptually uniform one, not "jet")  In that case this is ok.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
pwolfram: :+1:

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
xylar: @pwolfram, thanks for the feedback.  I agree.  How about another solution?  I have been using a color map from paraview for resolution that is made up of 3 perceptually uniform segments ("Blue - Green - Orange").  It will be more complicated to add this to COMPASS, since it's not standard in matplotlib, but I'll see what I can do.  Would that work for you?  @mark-petersen, what do you think?

xylar: The problem with `viridis` is that it's too dark and the black outlines from `cartopy` don't show up well, but this could be fixed by drawing in the continents, not just the outlines.

xylar:
@mark-petersen, I'm still tinkering with this one.  Please don't merge yet.

xylar:
@pwolfram, @mark-petersen and anyone else who wants to weigh in, which is better?
1.
cellWidthGlobal_bgo
2.
cellWidthGlobal_viridis

I personally prefer 2 (`viridis_r`) because it's perceptually uniform. I can see how you could get more granularity from the blue-green-orange version.  But it's much harder to tell which is higher and which is lower res.

xylar:
Pending @pwolfram and @mark-petersen's approval of using `viridis` now that I'm drawing the continents and not just their outlines, I think this is ready to merge.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
```diff
@@ -53,30 +59,34 @@ def build_mesh(
...
-                linestyle='-')
+                linestyle='-', zorder=2)
             gl.xlabels_top = False
             gl.ylabels_right = False
             plt.title('Grid cell size, km')
             plt.colorbar(im, shrink=.60)
-            plt.savefig('cellWidthGlobal.png')
+            fig.canvas.draw()
+            plt.tight_layout()
+            plt.savefig('cellWidthGlobal.png', bbox_inches='tight')
```
xylar, line 88: `bbox_inches='tight'` eliminates some of the remaining white space around the image so it looks right.

mark-petersen:
@xylar I strongly prefer your first one above (de-saturated rainbow?) because I can perceive six colors, and they are in an order I expect. In `viridis_r` I only see three colors, and the top half of the colorbar looks the same to me, all dark blue. I made a similar one at the same time, but with slightly muted colors compared to above. I didn't realize we were both working on it:
cellWidthGlobal
[Here is the colorbar code.](https://gist.github.com/mark-petersen/c2e45b437fa0a17c2c9bafc91607b080)
I really like your continents filled in. Thanks for changing it.
Despite my work on this one, I would go with your de-saturated rainbow above, out of these three.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
mark-petersen: Thanks, good to know.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/3Wbgy5.xml
```diff
@@ -0,0 +1,55 @@
+<ColorMaps>
+<ColorMap space="CIELAB" indexedLookup="false" name="3Wbgy5">
+<Point x="0.000000" o="1" r="0.831373" g="0.909804" b="0.980392"/>
+<Point x="0.012500" o="1" r="0.749020" g="0.862745" b="0.960784"/>
+<Point x="0.025000" o="1" r="0.694118" g="0.827451" b="0.941176"/>
```
mark-petersen, line 5: It looks like you downloaded this directly from Paraview? That was smart.

xylar: No, copied from MPAS-Analysis, which gives proper credit and got it from the original source (LANL's own SciVisColor web page, I believe).

xylar:
@mark-petersen, so the problem is a really hard one.  More colors does give more granularity in some sense but it really leaves you with no intuition about where is high res and where is low.  As Leif pointed out, he couldn't even tell how many regions of different resolution there were in my plot 1. above -- he thought there were 9, and that the yellow bands were really distracting as if they had extra meaning.  This is the general complaint about color maps that don't transition uniformly from dark to light or visa versa, and the reason for the push for perceptual uniformity.

xylar:
I played around with a bunch more color maps and I do think `3Wbgy5`, my option 1. above, might be the best option.  As @mark-petersen points out, anything that's perceptually uniform tends to be made up of 3 distinct colors or so (the ones that aren't are gawd-awfully ugly).  So I think we sacrifice the intuitive transition from light meaning high res to dark meaning low res in order to get more granularity in what isocontours correspond to what resolutions.

mark-petersen:
Added a min/max to title because it is easier to read there:
image

mark-petersen:
@pwolfram and @sbrus89 I think this is a nice improvement. I'd like to merge today, so we can rebase the mesh PRs on it.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
pwolfram: I'm happy with perceptually uniform solutions, but don't like it when you can see "lines" like yellow, in the colormap.   But, I'm ok with non-jet solutions like proposed below-- thanks!

pwolfram:
@sbrus89, can you please review the above?

mark-petersen:
Thanks. Tested one last time with QU240, QU240wISC, EC60to30wISC, EC60to30, SO60to10wISC. Everything appears to be working and making the plots correctly.

PR #527, opened 2020-04-19
Buffer mask for signed distance by dlat

xylar:
This should make sure the mask includes the various boundaries.


xylar:
@mark-petersen, I noticed that the signed distance function had the wrong function for the row containing the North Pole for any shape that extends that far north.  This seems to be because the `rasterize` function is sensitive to round-off errors (or something along those lines).  The easiest solution I could find was to expand the mask used to determine the sign by `dlat` so we can be sure it cover everything we want.  This might mean the signed distance isn't that accurate right around zero, so I want you to let me know if you see any side effects of this.

PR #528, opened 2020-04-20
Modify freezing temperature to match MPAS-Seaice

xylar:
To match the "mushy" formulation of the freezing temperature as
a function of salinity in MPAS-Seaice (the high-temperature limit
only), a new "mushy_az1_liq" coefficient has been added, multiplying
a new term with the appropriate value and associated functional form:
```
Tf(S) = (1/mushy_az1_liq)*(S/(1 - S/1000))
```

The value `config_open_ocean_freezing_temperature_coeff_mushy_az1_liq = -18.48` is taken from:
 https://github.com/MPAS-Dev/MPAS-Model/blob/seaice/develop/src/core_seaice/column/ice_mushy_physics.F90#L34

This merge also gets rid of support for a reference pressure in the freezing temperature, which we have not made use of since initial testing.  (To my knowledge, the value has always been set to zero).  I do not see the need to keep this added complexity if we are not using it.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -961,8 +961,8 @@
 					description="Character string to choose EOS formulation"
 					possible_values="Jackett McDougall EOS = 'jm' and Linear EOS = 'linear'"
 		/>
-		<nml_option name="config_open_ocean_freezing_temperature_coeff_0" type="real" default_value="-1.8" units="C"
-					description="The freezing temperature at zero pressure in open ocean."
+		<nml_option name="config_open_ocean_freezing_temperature_coeff_0" type="real" default_value="0.0" units="C"
+					description="The freezing temperature at zero pressure and salinity in open ocean."
 					possible_values="Any real number"
 		/>
 		<nml_option name="config_open_ocean_freezing_temperature_coeff_S" type="real" default_value="0.0" units="C PSU^{-1}"
@@ -970,19 +970,19 @@
 					possible_values="Any real number"
 		/>
 		<nml_option name="config_open_ocean_freezing_temperature_coeff_p" type="real" default_value="0.0" units="C Pa^{-1}"
-					description="The coefficient for the term proportional to the (limited) pressure in the freezing temperature in the open ocean."
+					description="The coefficient for the term proportional to the pressure in the freezing temperature in the open ocean."
 					possible_values="Any real number"
 		/>
 		<nml_option name="config_open_ocean_freezing_temperature_coeff_pS" type="real" default_value="0.0" units="C PSU^{-1} Pa^{-1}"
 					description="The coefficient for the term proportional to salinity times pressure in the freezing temperature in the open ocean."
 					possible_values="Any real number"
 		/>
-		<nml_option name="config_open_ocean_freezing_temperature_reference_pressure" type="real" default_value="0.0" units="Pa"
-					description="The reference pressure above which the freezing temperature is equal to config_freezing_temperature_coeff_0 in the open ocean."
-					possible_values="Any positive real number"
+		<nml_option name="config_open_ocean_freezing_temperature_coeff_mushy_az1_liq" type="real" default_value="-18.48" units="PSU C^{-1}"
+					description="The coefficient for the mushy sea-ice physics term az1_liq in the open ocean."
+					possible_values="Any non-positive number"
 		/>
 		<nml_option name="config_land_ice_cavity_freezing_temperature_coeff_0" type="real" default_value="6.22e-2" units="C"
-					description="The freezing temperature at zero pressure in land-ice cavities."
+					description="The freezing temperature at zero pressure and salinity in land-ice cavities."
 					possible_values="Any real number"
 		/>
 		<nml_option name="config_land_ice_cavity_freezing_temperature_coeff_S" type="real" default_value="-5.63e-2" units="C PSU^{-1}"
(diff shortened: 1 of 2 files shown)


xylar:
@proteanplanet, I particularly would be interested in your feedback at this stage.  I made `config_open_ocean_freezing_temperature_coeff_mushy_az1_liq` a namelist option rather than a constant to be as consistent and as easily backwards compatible with the previous implementation of the freezing point as possible.  I could imagine writing the freezing point computation to be a lot more efficient but didn't feel like this was the time to do it.

xylar:
## Forumlations

A quick explanation of how the formulation in MPAS-Ocean was derived from the one in MPAS-Seaice.

MPAS-Seaice has:
``` Fortran
    t_high = merge(c1, c0, (Sbr <= Sb_liq))

    zTin = ((Sbr / (M1_liq + N1_liq * Sbr)) + O1_liq) * t_high + &
          ((Sbr / (M2_liq + N2_liq * Sbr)) + O2_liq) * (c1 - t_high)
```
The value of `Sb_liq = 123.66702800276086` suggests that this limit will never be reached in MPAS-Ocean simulation.  We can discuss if there is a reason to support this limit but it is difficult for me to imagine.  Supstituting `t_high = 1`,
``` Fortran
    zTin = ((Sbr / (M1_liq + N1_liq * Sbr)) + O1_liq)
```
Here are the values of the coefficients, and the coefficients used to define them:
```
c1000 = 1000.0
az1_liq = -18.48
bz1_liq = 0.0
az1p_liq = az1_liq / c1000
M1_liq = az1_liq
N1_liq = -az1p_liq
O1_liq = -bz1_liq / az1_liq
```
Substituting all this, we can simplify to:
``` Fortran
zTin = Sbr / (az1_liq * (1.0 - Sbr/1000.0))
```

xylar:
> Coefficients defined here are also available from the equation of state in the ocean model, and need not be defined separately, but that can be refined later on. 

@proteanplanet, can you explain more?  I could see this being true in the (distant) future when we switch to a non-Boussinesq formulation and TEOS-10 but for now with the JM equation of state, I don't believe the coefficients for the freezing point is available.

xylar:
> How about I get this moving on Anvil?

That sounds good.  I guess you probably know better than I do the process for updating the E3SM source in your local build to include these changes?

proteanplanet:
> @proteanplanet, can you explain more? I could see this being true in the (distant) future when we switch to a non-Boussinesq formulation and TEOS-10 but for now with the JM equation of state, I don't believe the coefficients for the freezing point is available.

Maybe I'm thinking of the wrong ocean model.  The coefficients are in POP's equation of state. In any case, it doesn't matter for now.

xylar:
@proteanplanet, this has been rebased as requested.  I believe the changes are identical to those on the final commit on your branch, https://github.com/proteanplanet/MPAS-Model/commits/Andrew/ocean/freezingtemp, but let me know if you see differences.

@vanroekel, I would ask you to take another look at these changes.  We could keep the old freezing point in E3SM by using:
```
config_open_ocean_freezing_temperature_coeff_0 = -1.8
config_open_ocean_freezing_temperature_coeff_S = 0.0
config_open_ocean_freezing_temperature_coeff_p = 0.0
config_open_ocean_freezing_temperature_coeff_pS = 0.0
config_open_ocean_freezing_temperature_coeff_mushy_az1_liq = 0.0
```
Then, we could update to the coefficients in `ocean/develop` once we're comfortable that the climate is unchanged.

@mark-petersen and @akturner, are you able to review?

mark-petersen (approved):
@xylar this all looks fine. I will trust the scientific reviews by others. 

I ran some quick tests, and it passes the nightly regression suite with intel and gnu, both debug.

For global ocean cases, this is non-bfb with `ocean/develop`. That is expected, correct?

For the corresponding E3SM PR, please specify what the default flag settings should be. That will also be a non-bfb PR, right? In that case, we need to make sure it is tested in E3SM cases. Should E3SM defauts be set the same as MPAS-O defaults?
```
253 &eos
254     config_eos_type = 'linear'
255     config_open_ocean_freezing_temperature_coeff_0 = 0.0
256     config_open_ocean_freezing_temperature_coeff_S = 0.0
257     config_open_ocean_freezing_temperature_coeff_p = 0.0
258     config_open_ocean_freezing_temperature_coeff_pS = 0.0
259     config_open_ocean_freezing_temperature_coeff_mushy_az1_liq = -18.48
260     config_land_ice_cavity_freezing_temperature_coeff_0 = 6.22e-2
261     config_land_ice_cavity_freezing_temperature_coeff_S = -5.63e-2
262     config_land_ice_cavity_freezing_temperature_coeff_p = -7.43e-8
263     config_land_ice_cavity_freezing_temperature_coeff_pS = -1.74e-10
```

xylar:
@mark-petersen, which settings we want in E3SM and whether it's bit-for-bit is up to @vanroekel and @proteanplanet to negotiate.  I think @proteanplanet is working on a run with his own cherry-pick of this commit and the results of that run will determine whether this goes in with the new freezing point or the new settings.

These settings should be bit-for-bit:
```
253 &eos
254     config_eos_type = 'linear'
255     config_open_ocean_freezing_temperature_coeff_0 = -1,8
256     config_open_ocean_freezing_temperature_coeff_S = 0.0
257     config_open_ocean_freezing_temperature_coeff_p = 0.0
258     config_open_ocean_freezing_temperature_coeff_pS = 0.0
259     config_open_ocean_freezing_temperature_coeff_mushy_az1_liq = 0.0
260     config_land_ice_cavity_freezing_temperature_coeff_0 = 6.22e-2
261     config_land_ice_cavity_freezing_temperature_coeff_S = -5.63e-2
262     config_land_ice_cavity_freezing_temperature_coeff_p = -7.43e-8
263     config_land_ice_cavity_freezing_temperature_coeff_pS = -1.74e-10
```

But if we want the new freezing point, we should copy the new settings and, yes, the results are expected to be non-bit-for-bit.

vanroekel:
@mark-petersen this is being tested in E3SM right now by @proteanplanet -- depending on throughput we should wait to make the PR.  If we can get enough simulated years we can do a nonBFB PR Monday, or if not we can go with the settings @xylar mentions.  The code changes have to be to E3SM master by 9/30.

vanroekel:
to add a bit more, the plan is to all source changes must be in, but we (Chris and I) intend to make one final PR in October with namelist settings for v2

proteanplanet:
@xylar @jonbob @vanroekel This branch does not compile within the E3SM testing branch with all other sea ice changes switched on (https://github.com/E3SM-Project/E3SM/pull/3846). However, importantly, the freezing temperature changes in #fdad3d757ce5e74b6a17c5c5eb36d4dfa441408d do compile and run if cherry picked. G-case initial conditions are currently being generated on Anvil.

xylar:
@proteanplanet, hopefully, we can figure out the compilation error as part of updating `e3sm/develop` but I'm glad you're able to make headway in any case.

proteanplanet:
Please refer to analysis on including changes in the last commit to this branch in E3SM at:

https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/1852506366/20200925.alpha4+fallback.piControlSIFreez.ne30pg2+r05+EC30to60E2r2.anvil

xylar:
Thanks for reviewing, @akturner.

mark-petersen:
Merged to `ocean/develop` locally, passes nightly regression in debug with gnu and intel.

PR #531, opened 2020-04-21
Change ARM colormap

mark-petersen:
Update ARM to use new colormap. Add xml colormap functions to `mesh_definition_tools.py`.


mark-petersen:
@xylar, thanks for your example on today's PR. I used the tight spacing, filled continents, and new color map here as well.
image

testing_and_setup/compass/ocean/jigsaw_to_MPAS/mesh_definition_tools.py
```diff
@@ -11,13 +11,41 @@
...
+from matplotlib.colors import LinearSegmentedColormap
 plt.switch_backend('agg')
 
 
 ##########################################################################
 # Functions
 ##########################################################################
 
+def read_xml_colormap(xmlFile, mapName):
+    """Read in an XML colormap"""
```
mark-petersen, line 24: I copied this function from `build_mesh.py`. I think this `mesh_definitions_tools.py` file is for lower-level subroutines. Should we remove the original, and move it here instead, or just leave it both places? Why did yours start with an underscore?

xylar: That's fine.  Yes, we should definitely remove the redundant copy.  You will need the other function for registering the color map if you didn't copy it.

The underscore is a convention to indicate that it's a private function that users aren't intended to use directly and which might change its API without warning.  In this case, I just copied that form MPAS-Analysis, where it is a helper function for a public function that registers all the color maps that MPAS-Analysis supports.

testing_and_setup/compass/ocean/jigsaw_to_MPAS/mesh_definition_tools.py
```diff
@@ -198,97 +218,31 @@ def AtlanticPacificGrid(lat, lon, cellWidthInAtlantic, cellWidthInPacific):
                     cellWidthOut[j, i] = cellWidthInAtlantic[j]
     return cellWidthOut
 
-#
-# def  circleOnGrid(lon, lat, centerLon, centerLat, radius, tanhWidth)
```
xylar, line 202: @mark-petersen, is it okay to get rid if this commented-out function?

mark-petersen: Yes. That one and `lldistkm` are the last of the matlab helper functions I wrote two years ago, the first time we interfaced with the jigsaw front end (matlab at that time). It's nice to see how far we've come since then!

mark-petersen:
@xylar thanks for making the colormap support more general. This is great. Are you done with your revisions on this one?

xylar:
Yes, I'm done here.  Merge when you're ready.

PR #535, opened 2020-04-23
Add a daily output test case to QU240

xylar:
This test case outputs the same fields as E3SM `timeSeriesStatsMonthly` but is meant to be cheaper to run (since it only requires a day, not a month, of simulation).

The test case will be used to make sure MPAS-Ocean output is CF compliant.

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/QU240/daily_output_test/config_driver.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/QU240/daily_output_test/config_driver.xml
@@ -1,0 +1,14 @@
+<driver_script name="run.py">
+	<case name="test">
+		<step executable="./run.py" quiet="true" pre_message=" * Running test" post_message=" - Complete"/>
+	</case>
+	<validation>
+		<compare_fields file1="test/analysis_members/mpaso.hist.am.timeSeriesStatsDaily.0001-01-01.nc">
+			<field name="timeDaily_avg_temperature" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="timeDaily_avg_salinity" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="timeDaily_avg_layerThickness" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="timeDaily_avg_normalVelocity" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="timeDaily_avg_ssh" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+		</compare_fields>
+	</validation>
+</driver_script>
--- a/testing_and_setup/compass/ocean/global_ocean/QU240/daily_output_test/config_test.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/QU240/daily_output_test/config_test.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0"?>
+<config case="test">
+	<add_link source="../../init/initial_state/initial_state.nc" dest="init.nc"/>
+	<add_link source="../../init/initial_state/graph.info" dest="graph.info"/>
+	<add_link source="../../init/initial_state/init_mode_forcing_data.nc" dest="forcing_data.nc"/>
+
+	<namelist name="namelist.ocean" mode="forward">
+		<template file="template_forward.xml" path_base="script_configuration_dir"/>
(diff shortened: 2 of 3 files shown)


xylar:
@wenshanw, could you see if you can add my fork, check out this branch, build and do a test run?  Here are the steps I would recommend (keeping in mind that I might have forgotten some):
```
git remote add xylar/MPAS-Model git@github.com:xylar/MPAS-Model.git
git fetch --all -p
```
Do one of the following two options:
1. If you want to, I would recommend trying out the `git worktree` functionality.  Do a google search to learn more:
```
git worktree add ../add_QU240_time_daily_mean
cd ../add_QU240_time_daily_mean
git reset --hard xylar/MPAS-Model/add_QU240_time_daily_mean
```
2. If you prefer, you can just create a new branch right where you are (I'm a little rusty on this):
```
git checkout -b add_QU240_time_daily_mean xylar/MPAS-Model/add_QU240_time_daily_mean
```
From here, it's the same with either approach:
```
make clean
make gfortran
cd testing_and_setup/compass
./list_testcases.py | grep QU240
```
Figure out the numbers for the `init` and the `daily_output_test` test case.
```
./setup_testcase.py -n 75,76 -m runtime_definitions/mpirun.xml -f config.ocean \
    --work_dir </path/to/work/directory>
```
These instructions quite likely aren't perfect, so please let me know if you run into trouble or corrections are needed.

Please run the `init` and then the `daily_output_test` test cases.  If everything works okay, as it did for me, you should be able to look at the results with:
```
ncdump -h analysis_members/mpaso.hist.am.timeSeriesStatsDaily.0001-01-01.nc | more
```
I wouldn't worry about CF-compliance just yet, I just want to make sure you can get this far.  Then, we can merge this new test case to `ocean/develop` and you can use `git merge` or `git rebase` to update your branch in #533 (or start a new branch).

xylar:
@mark-petersen, if you could run a quick test of `daily_output_test`, that would also be great!  It worked fine for me on my laptop.

mark-petersen:
@xylar I think it is good to have a test for timeSeriesStats. But is there a reason we can't evaluate CF compliance in a standard output file with other streams? Should we look at both?

xylar:
> @xylar I think it is good to have a test for timeSeriesStats. But is there a reason we can't evaluate CF compliance in a standard output file with other streams? Should we look at both?

Yes, that's the plan.  @wenshanw is starting with the `QU240/init/test` case and just adding variable to the `output` stream.  That's working fine in that the `_FillValue` attribute is there (though fields have the wrong fill value where they're not valid).  But the most significant place where we want CF-compliant output is in `timeSeriesStatsMonthlyOutput`, and we're pretty sure that adding `_FillValue` to `temperature` will not necessarily mean that `timeMonthly_avg_activeTracers_temperature` has it.  What is more, we don't want to be time averaging the fill values in MPAS data arrays because we may end up with overflows or weird results or, at the very least, cases where the time average of the fill value isn't quite the same as the fill value because of round-off.  So there's good reason to have a solid test case with `timeSeriesStats` to play around with early in the process.

wenshanw:
Thank you, @xylar!
It all worked well until I `./run.py `in `init/`
`Error: ModuleNotFoundError: No module named 'mpas_tools.viz.paraview_extractor'; 'mpas_tools.viz' is not a package`

I conda installed mpas_tools. It didn't help.

xylar:
@wenshanw, I think I know what the issue is.  You will need to install the latest `compass` metapackage.  The one you installed (0.1.2) is for a specific set of tools and will prevent you from upgrading those tools to incompatible versions.  I would suggest you keep you old environment for now and make a new one:
```
conda create -y -n compass_0.1.4 -c conda-forge -c e3sm python=3.7 compass=0.1.4
```
This should give you the latest compass environment and it should work with what you have.  You can find out what compass is needed by looking at either of these files:
```
testing_and_setup/compass/README_ocean.md
```
on your particular branch.  I this case, it will say 0.1.3 is needed but there is no harm in using 0.1.4 instead, since it is backwards compatible and the changes are very minor.  Sorry about that!  I can understand that this is confusing.

wenshanw:
@xylar Yes! It solves the problem! :bowing_woman::bowing_woman::bowing_woman:

It all worked out. 
`analysis_members/mpaso.hist.am.timeSeriesStatsDaily.0001-01-01.nc` looks normal to me.
Thank you very much!

xylar:
Great, @wenshanw!  Glad that worked.  I'll wait for @mark-petersen to test this an merge to `ocean/develop`.  Then, we can talk about how to update your branch to either merge or "rebase" onto `ocean/develop` so you can test with it.

PR #536, opened 2020-04-23
Added build support for Summit at OLCF and OpenACC for GPU acceleration

philipwjones:
This change adds a pgi-summit entry to the top level Makefile and adds an OPENACC option
similar to the OpenMP option to enable GPU acceleration for OpenACC decorated code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,36 @@ pgi:
 	"OPENMP = $(OPENMP)" \
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI" )
 
+pgi-summit:
+	( $(MAKE) all \
+	"FC_PARALLEL = mpif90" \
+	"CC_PARALLEL = mpicc" \
+	"CXX_PARALLEL = mpicxx" \
+	"FC_SERIAL = pgf90" \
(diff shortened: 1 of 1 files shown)


mark-petersen:
@mgduda @mattdturner @akturner 
This simply adds a DOE machine to the makefile configuration and an OpenACC flag, similar to the OpenMP flag.

Could you please comment or approve this week?

mark-petersen:
@philipwjones are there accompanying changes to the cmake build system required here? Are you altering this for stand-alone only right now, or do you want to run within E3SM for pgi-summit and OpenACC?

philipwjones:
Yes. The target and compiler flags are set elsewhere for the Cmake build, so I think the only change in our repo is to set the MPAS_OPENACC ifdef (like for OpenMP) but I didn't know that the equivalent of build_threaded is in the E3SM build. So maybe go ahead with a standalone mod only unless @jonbob or @jgfouca can let us know what the correct OpenACC mods are in e3sm (if there are any?)

mark-petersen:
Changed base to develop, since this is top level makefile. @philipwjones you will see this in e3sm when we merge develop into e3sm/develop, in the next week or two.

PR #537, opened 2020-04-27
Ocean/alter3 dgm

vanroekel:
Adds horizontal variability of GM bolus kappa value of [Visbeck et al 1997.](https://journals.ametsoc.org/doi/full/10.1175/1520-0485%281997%29027%3C0381%3ASOETCI%3E2.0.CO%3B2)  The Danabasoglu and Marshall 2007 modifications still function independently of this option.  The scheme contains options for min and max bolus value and more work is needed to properly pair this scheme with the ramp function for variable resolution.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -286,19 +286,23 @@
 		<nml_option name="config_Redi_closure" type="character" default_value="constant" units="unitless"
 					description="Control what type of function is used for Redi $\kappa$. Only 'constant' is currently supported."
 					possible_values="'constant', 'data', 'match_GM'"
-		/>
-		<nml_option name="config_Redi_use_slope_taper" type="logical" default_value=".true." units="unitless"
-					description="If true, Redi is tapered based on Danabasoglu and McWilliams 1995 (slope tapering)"
-					possible_values=".true. or .false."
-		/>
-		<nml_option name="config_Redi_use_surface_taper" type="logical" default_value=".true." units="unitless"
-					description="If true, Redi slope is tapered near sfc based on Large et al 1997"
-					possible_values=".true. or .false."
-		/>
-		<nml_option name="config_Redi_N2_limit_term1" type="logical" default_value=".true." units="unitless"
-					description="If true, the N2 limiting is applied to the horizontal diffusion term"
-					possible_values=".true. or .false."
-		/>
+    />
+    <nml_option name="config_Redi_use_slope_taper" type="logical" default_value=".true." units="unitless"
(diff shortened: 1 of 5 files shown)


vanroekel:
Here is a quick proof of concept test with the QU240 test case.  The values look reasonable, but we need to wait for the Redi PR into E3SM prior to testing this there.

Screen Shot 2020-04-27 at 7 53 12 AM

vanroekel:
@mark-petersen I just pushed code that fixes the taper and moves redikappa to edges.  Could you take a look, especially at fluxTerm3 for redi to make sure my interpolation is right?  Also if you could run the nightly regression on this I would appreciate it.  I've only confirmed this builds so far.  I will run the QU240 soon.

vanroekel:
The branch currently fails blocks test (passing all others).  I'm looking into this now.

vanroekel:
@mark-petersen I just pushed a branch that passes nightly for me.  Let me know if you see anything else you'd like to change.

vanroekel:
@mark-petersen this branch is ready for testing.  We'd like to move on this as soon as you have time.  I've tested with intel19 in debug on grizzly and the nightly passes for me.  We also have tested extensively in e3sm g-cases

mark-petersen:
@vanroekel will this be bfb with default settings in e3sm, or would you change GM settings on the E3SM PR that make it non bfb?

vanroekel:
It won't be BFB even with no settings changed.  So I think this would be a good time to add the flags to enable Redi in this PR to E3SM

mark-petersen:
Rebased on head, retested.

mark-petersen:
@maltrud @milenaveneziani @lconlon, I would like to merge this late Tuesday if possible. Please make any comments if you have time, or tell us if we need to wait for your review.

@jonbob this will be a non-bfb E3SM PR later this week.

vanroekel:
@mark-petersen not sure what the merge plan is on this, but can I make a quick update to registry to reflect our new tuned parameters form G-case testing?  I should be able to get this soon (less than 1 day)

vanroekel:
just missed the merge.  I'll add the defaults in the E3SM PR, putting the changes here in case that helps you -- proposed changes to namelist for GM/Redi
```
config_use_Redi = .true. -- For low res and RRM only
config_Redi_kappa = 400.
config_Redi_set_RediKappa_to_GMKappa = .false.
config_Redi_maximum_slope = 0.01
config_GM_closure = 'visbeck'
config_GM_baroclinic_mode = 3.0
config_GM_visbeck_alpha = 0.005
config_GM_visbeck_max_kappa = 1200.
```

PR #538, opened 2020-04-27
Optimizations for tracer horizontal mixing

mattdturner:
Optimizations to tracer horizontal mixing that include

- Removing pointer retrievals
- Trimming argument lists to remove Pool variables (no longer used in these routines)
- Replace mesh variables (previously retrieved from meshPool) with module-level variables in `ocn_mesh` module (relies on PR #496 being merged)

This PR requires the following PRs be merged first:

-  #496
-  #457
-  #513

Some of the changes in those PRs were either git cherry-pick'ed or manually added. As a result, once those PRs are merged there might need to be some commits squashed or changes reverted in this PR.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_hmix.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_hmix.F
@@ -149,10 +149,10 @@ contains
 
       call mpas_timer_start("tracer hmix")
 
-      call ocn_tracer_hmix_del2_tend(meshPool, layerThicknessEdge, tracers, tend, err1)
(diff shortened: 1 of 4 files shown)


mattdturner:
This was tested using a QU240 test case, and a EC60to30 test case (both generated via Compass). The results are bit-for-bit with Intel compiler on Cori.

As a result of the optimizations in this PR, there is a reduction of the time spent in tracer hmix by approx. 5%.  Unfortunately, all of this runtime reduction came from the `Redi` code and there was no speedup to the `del4` code.

mattdturner:
Rebased to current `ocean/develop`.

PR #539, opened 2020-04-27
Remove scratch allocates and config. Part 1: analysis members

mark-petersen:
Part 1 of #457.

### DIFF
--- a/src/core_ocean/analysis_members/mpas_ocn_eliassen_palm.F
+++ b/src/core_ocean/analysis_members/mpas_ocn_eliassen_palm.F
@@ -439,7 +439,6 @@ contains
          call mpas_pool_get_subpool(block % structs, 'eliassenPalmAM', am_epftPool)
          call mpas_pool_get_subpool(block % structs, 'state', statePool)
          call mpas_pool_get_subpool(block % structs, 'mesh', meshPool)
-         call mpas_pool_get_subpool(block % structs, 'eliassenPalmAMPKGScratch', scratchPool)
          call mpas_pool_get_subpool(block % structs, 'forcing', forcingPool)
          call mpas_pool_get_subpool(block % structs, 'diagnostics', diagnosticsPool)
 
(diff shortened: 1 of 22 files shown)


mark-petersen:
Passes nightly regression suite on grizzly with gnu and intel, optimized and debug.

@mattdturner, the first commit 2fda23a is exactly your changes. There were a few corrections for the Lagrangian particles, which I put in a second commit. You can review by inspection, since I tested. This will be the first performance PR to be merged, but E3SM repo is closed right now, and we wait for that.

src/core_ocean/analysis_members/mpas_ocn_eliassen_palm.F
```diff
@@ -358,93 +339,49 @@ subroutine ocn_compute_eliassen_palm(domain, timeLevel, err)!{{{
...
+      real(KIND=RKIND), dimension(:,:), allocatable :: array3_3D
+      real(KIND=RKIND), dimension(:,:), allocatable :: array1_3Dbuoy
+      real(KIND=RKIND), dimension(:,:), allocatable :: array2_3Dbuoy
+      real(KIND=RKIND), dimension(:,:), allocatable :: PVMidBuoyCoor
+      real(KIND=RKIND), dimension(:,:), allocatable :: PVMidBuoyCoorEA
+      real(KIND=RKIND), dimension(:,:), allocatable :: uMidBuoyCoorEA
+      real(KIND=RKIND), dimension(:,:), allocatable :: vMidBuoyCoorEA
+      real(KIND=RKIND), dimension(:,:), allocatable :: uPVMidBuoyCoorEA
+      real(KIND=RKIND), dimension(:,:), allocatable :: vPVMidBuoyCoorEA
+      real(KIND=RKIND), dimension(:,:,:), allocatable :: PVFluxTest
```
mattdturner, line 384: The allocatable statements might actually need to be module-level instead of subroutine-level.  This is a non-issue once #513 is merged, but having the global `!$omp parallel` implementation could cause the subroutine-level allocatable definitions to result in segfaults if running on multiple threads.

mark-petersen:
rebased, tested.

mattdturner:
I just rebased this PR.

mark-petersen:
On grizzly, passes nightly regression suite with gnu, intel17, intel19 with PIO2. Tested both debug and optimized. BFB with previous on all tests.

PR #543, opened 2020-05-01
fixes for redi to work for BGC

vanroekel:
redi_term{1-3} were declared in the activeTracers and debugTracers pool for output while testing.  These were never added to the BGC pools.  This PR makes those allocatable arrays in the Redi routine instead of in the pools

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
@@ -73,7 +73,6 @@ contains
 !-----------------------------------------------------------------------
 
    subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThickness, layerThicknessEdge, zMid, tracers, &
-                                        redi_term1, redi_term2, redi_term3, &
                                         RediKappa, dt, isActiveTracer, slopeTriadUp, slopeTriadDown, &
                                         kappaRediSfcTaper, kappaGMCell, rediLimiterCount, tend, err)!{{{
 
@@ -113,7 +112,7 @@ contains
       !-----------------------------------------------------------------
 
       real(kind=RKIND), dimension(:, :, :), intent(inout) :: &
-         tend, redi_term1, redi_term2, redi_term3     !< Input/Output: velocity tendency
+         tend
 
       !-----------------------------------------------------------------
       !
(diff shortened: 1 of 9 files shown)


mark-petersen:
I created a new test for the MPAS stand-alone nightly regression suite that runs all BGC tracer groups and Redi and GM simultaneously.  It failed with a seg fault on ocean/develop, but this passes, so it looks like it is fixed. The addition of this commit changes the thread test to fail. Looking into that now...

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -143,6 +142,7 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThickness, laye
       real(kind=RKIND), dimension(:), pointer :: areaCell, dvEdge, dcEdge
       real(kind=RKIND), dimension(:), allocatable :: minimumVal
       real(kind=RKIND), dimension(:, :), allocatable :: fluxRediZTop
+      real(kind=RKIND), dimension(:,:,:), allocatable :: redi_term1, redi_term2, redi_term3
```
mark-petersen, line 145: @vanroekel for allocated arrays that have a shared dimension amongst threads (iCell, iEdge, iVertex), the declaration needs to be put at the top of the module. Now passes bfb thread test. I'll push up the change.

mark-petersen:
@vanroekel I rebased at the end of the previous Redi PR, as I've merged COMPASS PRs in between and this will be more clear for the history on the E3SM submodule.

This PR includes the new BGC+ Redi test. Passes nightly regression suite, both gnu/opt and intel/debug, bfb with ocean/develop.

The Fortran code is identical to your changes, other than the one allocate line.

mark-petersen:
@jonbob by email communication, it sounds like this worked in E3SM with BGC on. Could you approve here, and I will then make a PR for E3SM?

mark-petersen:
@mattdturner FYI, the test of the BGC tracer group was merged in with this PR. Here are instructions.

```
# rebase your branch on ocean/develop. Then:
cd testing_and_setup/compass/
./list_testcases.py |grep QU240

  72: -o ocean -c global_ocean -r QU240 -t bgc_ecosys_test
  74: -o ocean -c global_ocean -r QU240 -t init

# You need those two. You can do this, like usual:
./setup_testcase.py -f general.config.ocean -m runtime_definitions/mpirun.xml --work_dir $DIR -n 72
./setup_testcase.py -f general.config.ocean -m runtime_definitions/mpirun.xml --work_dir $DIR -n 74
```
You have to run the init first. Then bgc_ecosys_test creates an initial_state.nc with the BGC fields in them, and runs a test with the forward model.

PR #544, opened 2020-05-01
Adds normalVelocity arrays to eddy products

vanroekel:
This adds two new fields to the eddy products AM -- normalVelocityTimesTemperature and normalVelocitSquared to the eddy products AM to allow for EKE and eddy heat transport calculations.

### DIFF
--- a/src/core_ocean/analysis_members/Registry_eddy_product_variables.xml
+++ b/src/core_ocean/analysis_members/Registry_eddy_product_variables.xml
@@ -68,7 +68,9 @@
 			<var name="velocityMeridionalSquared"/>
 			<var name="velocityZonalTimesTemperature"/>
 			<var name="velocityMeridionalTimesTemperature"/>
-                        <var name="velocityZonalTimesTemperature_GM"/>
-                        <var name="velocityMeridionalTimesTemperature_GM"/>
+      <var name="velocityZonalTimesTemperature_GM"/>
+      <var name="velocityMeridionalTimesTemperature_GM"/>
+      <var name="normalVelocityTimesTemperature"/>
+      <var name="normalVelocitySquared"/>
 		</stream>
 	</streams>
--- a/src/core_ocean/analysis_members/Registry_eddy_product_variables.xml
+++ b/src/core_ocean/analysis_members/Registry_eddy_product_variables.xml
@@ -44,14 +44,20 @@
                 />
                 <var name="velocityMeridionalTimesTemperature_GM" type="real" dimensions="nVertLevels nCells Time" units="m s^{-1} C"
                         description="cell-wise product of component of horizontal bolus velocity in the northward direction and temperature"
-                />
-    <var name="normalVelocitySquared" type="real" dimensions="nVertLevels nEdges Time" units="m^2 s^{-2}"
+                        />
+     <var name="normalGMBolusVelocitySquared" type="real" dimensions="nVertLevels nEdges Time" units="m^2 s^{-2}"
+      description="edge based square of normal velocity"
+    />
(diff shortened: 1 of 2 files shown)


vanroekel:
@mark-petersen here is the PR we discussed into eddy stats that would be great to add to amrapalli's simulation.   I've tested this in a short G-case and it looks like it is working

mark-petersen:
@vanroekel when we add this to E3SM, do you want all four variables in the monthly time average output?
```
normalGMBolusVelocityTimesTemperature
normalGMBolusVelocitySquared
normalVelocityTimesTemperature
normalVelocitySquared
```

vanroekel:
@mark-petersen yes it would be good to have all 4 variables, but if we could wrap the GM ones such that they aren't output for RRS meshes that would be great.

maltrud:
are we sure this works?  last time i tried it i got seg faults even though visually it seemed fine.

vanroekel:
@maltrud it is on for the redi branch in my home directory on compy so your recent tests should have picked up this new code.  Also my visbeck tests on anvil ran with this successfully.  Could you confirm though that it is working?  I forgot to include it in the default timeMonthly, but should show up in your bld directory

src/core_ocean/analysis_members/Registry_eddy_product_variables.xml
```diff
@@ -39,12 +39,28 @@
...
-                />
+		<var name="normalVelocitySquared" type="real" dimensions="nVertLevels nEdges Time" units="m^2 s^{-2}"
+			description="edge based square of normal velocity"
+		/>
+		<var name="normalVelocityTimesTemperature" type="real" dimensions="nVertLevels nEdges Time" units="m s^{-1} C"
+			description="edge based product of normal velocity and temperature"
+		/>
+		<var name="velocityZonalTimesTemperature_GM" type="real" dimensions="nVertLevels nCells Time" units="m s^{-1} C"
+			description="cell-wise product of component of horizontal bolus velocity in the eastward direction and temperature"
+			 packages="gm"
```
mark-petersen, line 50: @luke, I added all the GM variables to the gm package, so they do not use memory space for RRS runs.

src/core_ocean/analysis_members/mpas_ocn_eddy_product_variables.F
```diff
@@ -202,37 +205,107 @@ subroutine ocn_compute_eddy_product_variables(domain, timeLevel, err)!{{{
...
-               velocityMeridionalTimesTemperature_GM(k,iCell) = GMBolusVelocityMeridional(k,iCell)*activeTracers(index_temperature,k,iCell)
-             end do
-         end do
+         call mpas_pool_get_array(eddyProductVariablesAMPool,'normalVelocityTimesTemperature', normalVelocityTimesTemperature)
+         call mpas_pool_get_array(eddyProductVariablesAMPool,'normalVelocitySquared', normalVelocitySquared)
+
+         ! if GM is active, include the GM variables.
+         ! I repeated the block of code here for better performance, rather than
+         ! split the GM and non-GM variables into separate loops.
+         if (config_use_GM) then
```
mark-petersen, line 233: I also have two separate sections, one including GM variables and the other without.

mark-petersen:
I also added OpenMP directives. I think this is now ready to merge. @vanroekel take a quick look at my recent changes and let me know.

This passes the nightly regression suite.

mark-petersen:
> are we sure this works? last time i tried it i got seg faults even though visually it seemed fine.

@maltrud I don't see any problems now with either gnu or intel in stand alone. Once it is "next in line" I will be sure to test it with the E3SM PR.

vanroekel:
@mark-petersen changes look good to me.

PR #545, opened 2020-05-01
Revert some COMPASS changes intended to support conda MPI

xylar:
It turns out not to be necessary to run python scripts with `mpirun` as long as we don't use the MPI version of `netcdf4`.  This is possible even if we want to use the MPI version of `esmf` (which we do because `ESMF_RegridWeightGen` is way too slow in serial for certain mesh combinations).

This PR reverts:
* 8217097, which modified `setup_testcase.py` to add `mpirun` to python scripts in certain circumstances
* 190338d, which made sure `planar_hex` (which wasn't obviously a python script) also got run in this way.

This merge also updates to `compass` env. 0.1.5, which makes sure we always get the `nompi` variant of `netcdf4`, so that the shenanigans with `mpirun` isn't needed.

### DIFF
--- a/testing_and_setup/compass/README_ocean.md
+++ b/testing_and_setup/compass/README_ocean.md
@@ -6,11 +6,11 @@ To set up and run ocean test cases from COMPASS, you will need a conda
 environment.  First, install Miniconda3 (if miniconda is not already
 installed), then create a new conda environment as follows:
 ``` bash
-conda create -n compass_0.1.4 -c conda-forge -c e3sm python=3.7 compass=0.1.4
+conda create -n compass_0.1.5 -c conda-forge -c e3sm python=3.7 compass=0.1.5
 ```
 Each time you want to work with COMPASS, you will need to run:
 ```
-conda activate compass_0.1.4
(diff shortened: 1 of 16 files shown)


xylar:
I just need to deploy `compass 0.1.5` before this can get merged...

xylar:
Okay, `compass 0.1.5` is deployed on LANL IC, Cori, Compy and Anvil.

mark-petersen:
I'll merge in after Redi is completely settled on E3SM - probably in a few days.

mark-petersen:
Retested on IC. Passes nightly regression suite using both compass 0.1.4 and 0.1.5. Runs e3sm_coupling successfully with compass 0.1.4. The e3sm_coupling steps with `ESMF_RegridWeightGen` fail using `compass_0.1.5`, as described in #564 and MPAS-Dev/pyremap#15.

Merging in, as that specific problem is not caused by this PR.

xylar:
Thanks for merging these 3, @mark-petersen.  I'm working on a fix for #564 as you see.

PR #553, opened 2020-05-09
Remove scratch allocates and config. Part 2: Init mode

mark-petersen:
Part 2 of #457.


mattdturner:
Just rebased to the current `ocean/develop`.

mark-petersen:
On grizzly, tested nightly regression suite with gnu, intel17, intel19 with PIO2. Tested both debug and optimized. Passes all tests except there is a problem with the `bgc_ecosys_test` initialization using intel. Looking into that now. Otherwise, bfb with ocean/develop.

PR #555, opened 2020-05-11
New Mesh: WC12to60kmL60E3SMv2r01

mark-petersen:
Design discussion Water Cycle mesh, 12km high resolution region.


mark-petersen:
Instructions to create CUSP mesh on LANL IC. See [general MPAS instructions](https://acme-climate.atlassian.net/wiki/spaces/ECG/pages/1054113803/MPAS-Ocean+Initialization+and+Testing#Setting-up-a-test-case) for other machines and configurations.
1. Clone this branch
<details>
<summary>2. Create the test case:</summary> <p>

```
cd testing_and_setup/compass
./list_testcases.py | grep CUSP
  65: -o ocean -c global_ocean -r CUSP12 -t init
  66: -o ocean -c global_ocean -r CUSP12 -t spin_up
./setup_testcase.py \
  --config_file general.config.ocean \
  --work_dir $WORKDIR \
  --model_runtime runtime_definitions/mpirun.xml \
  -o ocean -c global_ocean -r CUSP12 -t init
```
where `$WORKDIR` is a path, usually to your scratch space. For example,
`--work_dir /lustre/scratch4/turquoise/mpeterse/runs/200429_test_new_branch`
and general.config.ocean is a file that specifies directory and file paths. Here is an example for LANL IC: [general.config.ocean](https://gist.github.com/mark-petersen/b0360e9fdb17a00669768c188b373883)
</details>
<details>
<summary>3. Create mesh </summary> <p>

On LANL IC, you can do this interactively:
```
salloc -N 1 -t 4:0:0 --qos=interactive
cd $WORKDIR
cd ocean/global_ocean/CUSP12/init
./run.py
```
That will proceed through the `base_mesh` and `culled_mesh` steps. It creates these files:
```
base_mesh: base_mesh.nc cellWidthGlobal.png   mesh_construction.png
culled_mesh:  culled_mesh.nc
```
You should look at the png files, and can also explore the netcdf files.
</details>
<details>
<summary>4. Create the initial state</summary> <p>

You need a compiled version of MPAS for this, as it uses the MPAS init mode. See [machine specific instructions](https://acme-climate.atlassian.net/wiki/spaces/ECG/pages/1054113803/MPAS-Ocean+Initialization+and+Testing#grizzly%2C-gnu). Then:
```
salloc -N 1 -t 4:0:0 --qos=interactive
cd $WORKDIR
cd ocean/global_ocean/CUSP12/init/initial_state
./run.py
```
where there is a link to your compiled `ocean_model` there, and you can increase the core count in `run.py` if needed. You will then see:
```
initial_state:
initial_state.nc initial_state.png vertical_grid.png
```
and you should look at those files as well. 
</details>
<details>
<summary>5. E3SM coupling files</summary> <p>

```
cd $WORKDIR
cd ocean/global_ocean/CUSP12/init/e3sm_coupling/
./run.py
```
</details>
<details>
<summary>6. spin up</summary> <p>

This is the same as step 2 above, but use the spin-up:
```
./setup_testcase.py \
  --config_file general.config.ocean \
  --work_dir $WORKDIR \
  --model_runtime runtime_definitions/mpirun.xml \
  -o ocean -c global_ocean -r CUSP12 -t spin_up
```
</details>

mark-petersen:
Based on the conversations the last few months, I made a best guess at the CUSP mesh using the signed distance function (PR #523). I created the mesh today with this PR and compass 0.1.5. It produces this mesh:
cellWidthGlobal
and these auxiliary images:

<details>
<summary> mesh construction steps </summary> <p>

mesh_construction
</details>
<details>
<summary> vertical grid </summary> <p>

vertical_grid
</details>
<details>
<summary> initial state pdfs </summary> <p>

initial_state
</details>

mark-petersen:
Here are a few questions for discussion for this mesh design:

- [ ] 1. What should the minimum resolution be? Currently 12km
That is set [here](https://github.com/mark-petersen/MPAS-Model/blob/6251855234e204a0a2710702198573539200ad9b/testing_and_setup/compass/ocean/global_ocean/CUSP12/init/define_base_mesh.py#L58). I created these:
12km min: 485k cells. 
14km min: 405k cells. 
For reference, EC60to30 has 265k cells.
On IC, meshes are available at: 
```
/lustre/scratch4/turquoise/mpeterse/runs/200511_cusp14_cusp_coastal_US_plus_mesh/ocean/global_ocean/CUSP14/init
/lustre/scratch4/turquoise/mpeterse/runs/200511_cusp12_cusp_coastal_US_plus_mesh/ocean/global_ocean/CUSP12/init
```

- [ ] 2. How many vertical levels?  Currently 60 and 64 layers are generated in COMPASS.

- [ ] 3. Should the high-resolution region be revised? In particular,
  a. Coastal width is 600km, [set here](https://github.com/mark-petersen/MPAS-Model/blob/ocean/cusp_coastal_US_plus_mesh/testing_and_setup/compass/ocean/global_ocean/CUSP12/init/define_base_mesh.py#L69)
  b. North Atlantic transition is [set here](https://github.com/mark-petersen/MPAS-Model/blob/6251855234e204a0a2710702198573539200ad9b/testing_and_setup/compass/ocean/global_ocean/CUSP12/init/define_base_mesh.py#L124) with the file [region_Gulf_Stream_extension.geojson](https://github.com/mark-petersen/MPAS-Model/blob/6251855234e204a0a2710702198573539200ad9b/testing_and_setup/compass/ocean/global_ocean/CUSP12/init/region_Gulf_Stream_extension.geojson), and can be easily changed.

- [ ] 4. Minor items
  a. Black and Capain seas are still there.
  b. At -180, high-res. region slightly degraded near Russia.

xylar:
@mark-petersen, your instructions involve `e3sm_coupling` before `spin_up` but it needs to be the opposite.  Though I wouldn't worry too much until after Thursday's meeting since we will likely need to do some reorganization anyway following that discussion.

xylar:
> 4. Minor items
> Black and Caspian Seas are still there.

Are you sure this is still true once you make the mesh?  There should be a critical land blockage to remove the Black Sea and the Caspian sea has no connection to the global ocean and should be culled by the normal flood-fill procedure.  If you are going on the image above, it's just a cartopy map and has no relevance for the MPAS mesh.

> At -180, high-res. region slightly degraded near Russia.

If I had a guess, it's that your `region_Bering_Sea` feature isn't periodic at +/-180.

pwolfram:
Following on from our discussion at https://acme-climate.atlassian.net/wiki/spaces/ECG/pages/1410270496/05-14-2020+Special+Topics+MPAS+E3SM+initialization, would recommend capturing physical basis or reasoning (several bullets, 30s-1min view) of design of mesh.

xylar:
@pwolfram, I think this is at least one point where the info you requested on why the resolution is distributed where it is could go here:
https://github.com/mark-petersen/MPAS-Model/blob/ocean/cusp_coastal_US_plus_mesh/testing_and_setup/compass/ocean/global_ocean/CUSP12/init/config_E3SM_coupling_files.ini#L14-L16

xylar:
> capturing physical basis or reasoning (several bullets, 30s-1min view) of design of mesh.

Would you do that here in the PR?  In the metadata I pointed to above?  Somewhere else?

pwolfram:
@xylar, I'd do that in the PR description and copy/paste these top points into the field you point to above.

proteanplanet:
@xylar and @mark-petersen.  I've been running the ruler over this mesh over the past few days, and am close to sharing a host of plots and checks on the mesh.  As I finish off my work, I want to supply you with a file of Arctic Shipping Assessment ship tracks through the Arctic that you may or may not wish to include in forced-open channels. With very minor changes, the 14km mesh would accommodate this.  What is your preferred format for that - Would a KML file work, or a Geoshape, or netcdf trace, or GeoJSON?

proteanplanet:
Of course, with my new-found knowledge after the tutorial, I could even venture to do this myself....  with your supervision.

xylar:
@proteanplanet, it would be great if you want to add some new critical passages and critical blockages!  Eventually, I will add the following to [future COMPASS documentation](https://github.com/MPAS-Dev/MPAS-Model/pull/472).

Adding new transects (critical blockages or passages) can be done through `geometric_features` but currently not directly in COMPASS in `MPAS-Model`.  Adding new features to `geometric_features` not quite trivial but it shouldn't be too hard.  

* First, go to http://geojson.io and make your transects.  It's fine to have multiple transects in the same file for now
* Add the required properties (name, component='ocean', object='transect', author=<you> and tags), see [Geometric Data](http://mpas-dev.github.io/geometric_features/stable/geometric_data.html) for details
* Make sure the list of tags includes `Critical_Passage` or `Critical_Land_Blockage`.  Other tags are up to you.  Tags are separated by semicolons
* Copy the geojson data on the right and paste it into a file in the root of the branch of `geometric_features` for your modification.  
* Make a copy of the example script for [Adding New Features](http://mpas-dev.github.io/geometric_features/stable/adding_features.html), change `my_new_features.geojson` to the name of your new geojson file, and run the script.
* Commit the changes to the repo and make a pull request to `geometric_features`, which I will review.

These changes will make their way into the next `geometric_features` release and the next `compass` conda environment.  So they're not immediately available for testing.  If you want me to set up a test `compass` environment, though, it's not too hard and we can talk about that.  Eventually, that process will also be in the documentation, too...

xylar:
@maltrud, @katsmith133 and possibly others had mentioned not being big fans of the name "CUSP" for this mesh.  I think this PR is a reasonable place to discuss alternatives.  @katsmith133, what was your suggestion again?

katsmith133:
@xylar not something that should actually be considered. It was very much a joke, but I will think about real names for the mesh and post any that I think are good.

philipwjones:
CONUS-Coast (like Kona coast and with our own coffee blend), crush (coastal refined US...), lots of other cr- words work.  Also, cousin, Custer, couscous, cuscus, cuss, cussed, custody...just a quick brain dump of c-words...can fill in an acronym for any of those and inverting word order opens up more...

pwolfram:
How about CONA (Coastal Ocean North America)?

vanroekel:
Now that we resolve a good chunk of the N. Atlantic and all the arctic, Coastal no longer seems like the most important item to include.  Not sure what the best choice is though.

maltrud:
North America Shelf Coastal and Arctic Refined
though we might have some copyright issues with this....

philipwjones:
Excellent! But I only watch it for the crashes...

proteanplanet:
I am close to publishing quality control on CUSP12 and CUSP14 for this PR.  I see on Confluence that WC12 and WC14 may be the new names for these meshes per Luke.   If so, I very much like this convention - shorter and focused on the simulation campaign at which they are targeted.  The rest of the information is now in the metadata.

pwolfram:
Sounds good-- happy with nice concise meta data description at https://github.com/mark-petersen/MPAS-Model/blob/ocean/cusp_coastal_US_plus_mesh/testing_and_setup/compass/ocean/global_ocean/CUSP12/init/config_E3SM_coupling_files.ini#L14-L16 and name is good too.

proteanplanet:
Quality control of mesh for global and polar performance: The first stage is to share the mesh characteristics for 42 designated areas of oceanographic or coupled importance, including a comparison of the coastline with the E3SM-HR V1 mesh.  These will be shared in two separate posts following this one for WC12 and WC14, respectively.   Further quality control posts will follow these two with comments on parts of the mesh that need to be addressed.

proteanplanet:
### WC12 sector plots - Indicating zoomed-in regions for the next post

WC12_sector_1
WC12_sector_2
WC12_sector_3
WC12_sector_4
WC12_sector_5
WC12_sector_6

proteanplanet:
WC12 zoomed plots - Numbers correspond to the figure in the previous post

WC12_zoom_1
WC12_zoom_2
WC12_zoom_3
WC12_zoom_4
WC12_zoom_5
WC12_zoom_6
WC12_zoom_7
WC12_zoom_8
WC12_zoom_9
WC12_zoom_10
WC12_zoom_11
WC12_zoom_12
WC12_zoom_13
WC12_zoom_14
WC12_zoom_15
WC12_zoom_16
WC12_zoom_17
WC12_zoom_18
WC12_zoom_19
WC12_zoom_20
WC12_zoom_21
WC12_zoom_22
WC12_zoom_23
WC12_zoom_24
WC12_zoom_25
WC12_zoom_26
WC12_zoom_27
WC12_zoom_28
WC12_zoom_29
WC12_zoom_30
WC12_zoom_31
WC12_zoom_32
WC12_zoom_33
WC12_zoom_34
WC12_zoom_35
WC12_zoom_36
WC12_zoom_37
WC12_zoom_38
WC12_zoom_39
WC12_zoom_40
WC12_zoom_41
WC12_zoom_42

proteanplanet:
### WC14 sector plots - Indicating zoomed-in regions for the next post

WC12_sector_1
WC12_sector_2
WC12_sector_3
WC12_sector_4
WC12_sector_5
WC12_sector_6

xylar:
@proteanplanet, those are going to be some very useful plots!  I guess we want to move toward having them be part of some standard output in the long run?

xylar:
A couple of comments on this PR and how it's evolving.  I think we need separate PRs for the 12 km and 14km versions of the proposed mesh.  Each PR would need to have the mesh long name (including revision number) in the title for us to find it more easily in the future.  We definitely need to fix the periodicity issue that I pointed out above before we proceed any further.

xylar:
@proteanplanet, the red meshes are the dual, correct?

proteanplanet:
### WC14 zoomed plots - Numbers correspond to the figure in the previous post

WC14_zoom_1
WC14_zoom_2
WC14_zoom_3
WC14_zoom_4
WC14_zoom_5
WC14_zoom_6
WC14_zoom_7
WC14_zoom_8
WC14_zoom_9
WC14_zoom_10
WC14_zoom_11
WC14_zoom_12
WC14_zoom_13
WC14_zoom_14
WC14_zoom_15
WC14_zoom_16
WC14_zoom_17
WC14_zoom_18
WC14_zoom_19
WC14_zoom_20
WC14_zoom_21
WC14_zoom_22
WC14_zoom_23
WC14_zoom_24
WC14_zoom_25
WC14_zoom_26
WC14_zoom_27
WC14_zoom_28
WC14_zoom_29
WC14_zoom_30
WC14_zoom_31
WC14_zoom_32
WC14_zoom_33
WC14_zoom_34
WC14_zoom_35
WC14_zoom_36
WC14_zoom_37
WC14_zoom_38
WC14_zoom_39
WC14_zoom_40
WC14_zoom_41
WC14_zoom_42

milenaveneziani:
Very nice plots, @proteanplanet.
Should we include comments here about possible passages to add? I can think of 3 for the Canadian Archipelago/North Passages region (and 1 in the Baltic).

vanroekel:
I definitely agree with @xylar we should separate out WC12 / WC14.  It is not clear or clean. to have both discussed here.  I would like to rename this PR to indicate WC12, using the long name -- that is what I think this one started as.  Do we have a PR for the 14?  Or do we need to open one?   -- @proteanplanet and @milenaveneziani please wait for comments pending a reorganization of the PRs

proteanplanet:
I have a synopsis summary of mesh quality, and am holding off until reorganization, however, it is actually beneficial to include both WC12 and WC14 in the same synopsis,

xylar:
> it is actually beneficial to include both WC12 and WC14 in the same synopsis,

I think there's no problem with making a comment on one PR and referring to that comment (you can get a link by clicking on the time since the comment was posted, oddly enough) in the other PR.  But I think it's really critical that the PR be named for the mesh that will be added to `ocean/develop` once it is merged.  And the associated metadata needs to be included.

proteanplanet:
### Synopsis on mesh quality for WC12 and WC14: 
 
**WC12**
 
1. The northwest passage is fully open, meaning that both the B- and C-grid are not encumbered by the narrow passage - hence no enforced landfast ice (Ref Zoom 1 and 37)
2. Mesh quality deteriorates at the international date line, and needs to be fixed (Ref Zoom 5 and 12)
3. The Strait of Georgia is not open. For the Water Cycle campaign, a rudimentary representation of it, as least at 12km resolution, seems desirable if not essential (Ref Zoom 14)
4. Italy’s boot is missing!  (Ref Zoom 19)
5. Give consideration to increasing resolution in parts of the Indonesian Throughflow to 12 km – Critical channels are one grid cell wide (Ref Zoom 28)
6. As with the Indonesian Throughflow,  serious consideration needs to be given to a slight resolution enhancement of the Agulhas current, since this is a critical region for global water properties that is highly scale dependent. (Ref Zoom 29)
7. Increase the resolution around the Hawaiian Islands so that they are separated by more than one grid cell, also ensuring that Kauai exists. It is essential as a U.S. State.  (Ref Zoom 41)
8. The Diomede Islands are missing from this mesh and the E3SM-HR mesh, but not the 14km mesh.  This is surprising. (Ref Zoom 5)
 
**WC14**
 
1. The northwest passage is fully closed, meaning that there is no ocean or sea ice throughflow.  Can this passage be made conditionally open, dependent on mesh resolution?  For example, it is fair and reasonable that it’s closed at 30km resolution, but perhaps we can decide that below 15km resolution, it is forced open? (Ref Zoom 1 and 37)
2. The Diomede Islands surprisingly show up in this mesh, but not the 12km mesh, suggesting an algorithmic sensitivity. (Ref Zoom 5)
3. As with the WC12 mesh, there is a grid density problem at the international date line in the Bering Sea, and this must be fixed.  This suggest that a non-periodic coordinate system may be better for smoothing the mesh density. (Ref Zoom 5 and 12)
4. As with WC12, we should give consideration to forcing a representation of the Strait of Georgia at the 14km scale.  This, of course, it not desirable at, say, 30km resolution.  (Ref Zoom 14)
5. Contrary to WC12, Italy has a boot!  We need to enforce Italy’s boot, in combination with , if only out of respect for Milena! (Ref Zoom 19)
6. As with WC12, give consideration to increasing resolution in parts of the Indonesian Throughflow to 14 km – Critical channels are one grid cell wide (Ref Zoom 28)
7. As with WC12 and the Indonesian Throughflow,  serious consideration needs to be given to a slight resolution enhancement of the Agulhas current, since this is a critical region for global water properties that is highly scale dependent. (Ref Zoom 29)
8. In the Hawaiian Islands so that they are separated by more than one grid cell – at 14km, Kauai exists, contrary to WC12.  This is not something we should leave to chance (Ref Zoom 41)
9. The Diomede Islands exist for this mesh and the E3SM-HR mesh, but not the 12km mesh.  This is surprising. (Ref Zoom 5)

xylar:
@proteanplanet and @qingli411, these are some really helpful observations about the meshes.

Regarding points 8 and 9, for the 12 and 14 km meshes, respectively, about islands being present sometimes and not other times: Because of the way that JIGSAW and MPAS coastlines work, even small changes to the mesh specification (e.g. fixing the periodicity bug or adding new higher resolution regions) are going to lead to grid-scale features appearing and disappearing, seemingly at random.  In some cases, it takes only small changes for a cell center to move from ocean to land, since we don't take an area average over a cell, but rather just a sampling at the cell center, to determine if a cell is land or ocean.  A different, less sensitive algorithm is something I'm keen to discuss but probably this PR isn't the place.

I will defer to @mark-petersen since he's "driving" here but I think the above points will somehow need to get translated into one or more of the following type of modifications:
* Here's a geojson shape where we need XX km resolution, decaying to the background over YY km.
* Here's a geojson transect for a new critical passage or land blockage

I wonder if the quickest way to do that might be to have a zoom-type meeting between @proteanplanet, @mark-petersen and any other interested party to make the appropriate shapes and get them added to this PR.

milenaveneziani:
Thanks @proteanplanet andd @qingli411, that's a nice summary.
And thank you for taking Italy's shape in high regards, my friends from Puglia would be very grateful: perhaps you can consider it a lower priority though ;-)

I would like to add a few more comments about passages:
1) we probably want to enforce two passages in the Canadian Archipelago: one is Hell Gate/Cardigan Strait (between Devon and Ellesmere Islands), which connects pretty directly with the Arctic; and one is Fury Sound, which connects the Archipelago with the upper part of Hudson Bay (Foxe Basin, I think it's called). These two passages account for ~0.3 Sv each of volume transport, so combined they amount to a similar number as the Nares St transport. They are ~10-20 km across, with depths of up to 150-200 m.
2) I see that the Baltic Sea is connected to the North Sea in wc12: do you know if it is also connected in wc14? I ask because that passage was actually closed in the ARRM60to10 mesh and the resolution in that area should be around 12-15 km in the ARRM mesh, so it is something we want to make sure about through an enforced channel. We definitely want the low-salinity water of the Baltic to connect to the open ocean, I think.

proteanplanet:
> @proteanplanet, those are going to be some very useful plots! I guess we want to move toward having them be part of some standard output in the long run?

@xylar - Yes, precisely.   It seems from the response that this is useful, so I'll keep moving toward this.  For V2 we'll do it this way (me sharing via GitHub), but the aim is for all of my code to be in Python, and for these to be produced automatically.  It took some time to work out the regions to focus on, and these can be expanded.  Please let me know (including @milenaveneziani, @qingli411, @maltrud, @vanroekel, @pwolfram et al.) if there are more regions we should enhance and check.  I've tried to consider as many of the E3SM ecosystem projects as possible in the regions, and they are easy to add (I'll add parts of Australia and New Zealand for global tides)

Following from this, I'm producing similar figures with bathymetry, with atmospheric resolution overlayed, and with automated flagging of one-cell-wide or one-level-deep passages as part of the analysis, which will also show up in these figures as a red cell.

proteanplanet:
> A different, less sensitive algorithm is something I'm keen to discuss but probably this PR isn't the place.

@xylar As always, right on the button.  Let's create an issue to worry about later. 

> I wonder if the quickest way to do that might be to have a zoom-type meeting between @proteanplanet, @mark-petersen and any other interested party to make the appropriate shapes and get them added to this PR. 

Yes, we can do that.  The bit I'm not sure about is how hard it would be to make land/passage enforcement scale-dependent.  I can help on the input side, but I'm not sure how hard it is from the Compass side?  This would be good to discuss via WebEx/Zoom.

proteanplanet:
@milenaveneziani 

>    1. we probably want to enforce two passages in the Canadian Archipelago: one is Hell Gate/Cardigan Strait (between Devon and Ellesmere Islands), which connects pretty directly with the Arctic; and one is Fury Sound, which connects the Archipelago with the upper part of Hudson Bay (Foxe Basin, I think it's called). These two passages account for ~0.3 Sv each of volume transport, so combined they amount to a similar number as the Nares St transport. They are ~10-20 km across, with depths of up to 150-200 m.

Nice catch.  Hells Gate is closed in both WC12 and WC14, but is nearly open in WC12 (Ref Zoom Number 1). Fury Strait is closed in both.  Whether or not we can enforce their opening perhaps depends on whether we can make this scale-dependent in the workflow.  There are some other candidates that jump out int he Archipelago region.

>     2. I see that the Baltic Sea is connected to the North Sea in wc12: do you know if it is also connected in wc14? I ask because that passage was actually closed in the ARRM60to10 mesh and the resolution in that area should be around 12-15 km in the ARRM mesh, so it is something we want to make sure about through an enforced channel. We definitely want the low-salinity water of the Baltic to connect to the open ocean, I think.

@xylar took are of this, and it is connected (see Zoom #2)

proteanplanet:
> @proteanplanet, the red meshes are the dual, correct?

Yes, sorry for not making that clear (and for only just replying now)

xylar:
> The bit I'm not sure about is how hard it would be to make land/passage enforcement scale-dependent. I can help on the input side, but I'm not sure how hard it is from the Compass side?

@proteanplanet, we would need a new capability to have a different set of critical passages (or land blockages) some meshes than in others.  But that shouldn't be too hard to put together.  I wouldn't worry about the logistics there.  What we really need is for someone to take a few minutes to hours to get on http://geojson.io and draw the transects, then give them meaningful names by editing the python dictionary of `"properties"`:
```
      "properties": {
        "name": "Blah Blah Passage"
      },
```
If these transects wouldn't be used in every mesh but they *would* conceivably be used in multiple meshes, we should put them in `geometric_features`.  If any of them are clearly just a one-off for this mesh, we should probably put them in the COMPASS directory instead.

The process for adding more high-res regions would be similar.  We would want someone to go on http://geojson.io and define a set of polygons that outline the higher resolution regions, giving each a name.  Then, we would need to edit the script that add these regions to set the corresponding resolution and transition width for each region.  This is likely to be an iterative process but it would be helpful for someone to go in and make the first pass at the new shapes to be added.  This is the process that I though might work well on a WebEx call.

xylar:
@proteanplanet, some testing and planning is definitely needed but https://github.com/MPAS-Dev/MPAS-Model/pull/586 should introduce the changes needed have "custom" critical passages and land blockages for this and other meshes as needed.  As I said in my previous comment, we could either include these in the COMPASS test case if they're one-offs or put them in `geometric_features` if they're useful for other meshes.

xylar:
@mark-petersen, can I push a fix for the periodicity of `region_Bering_Sea.geojson` to this branch or would you like to handle that some other way?

milenaveneziani:
May I ask why you think it is necessary to make critical passages resolution-dependent? Would it cause problems if a certain passage is forced to be open for a low-res mesh? Would it be too narrow and/or shallow, or is there a constraint in its width/depth so that for example it is forced to be at least two grid cells wide? If the latter, I wonder if we should take the trouble to make the passages mesh-dependent.

proteanplanet:
Hi @milenaveneziani,

> May I ask why you think it is necessary to make critical passages resolution-dependent?

To illustrate why this is necessary, I'll use the northwest passage as an example, but there are other places, such as the Strait of Georgia, for which this argument is also applicable.

In the 30-60km DECK mesh, the southernmost arm of the northwest passage around Victoria Island does not exist, and nor should it; model resolution has to count for something, and if we did force this open, it would be one cell wide for the entire channel. Aside from only one degree of freedom possible in ocean cells, the entire channel would be locked up with predestined landfast ice on the B-grid.   15 km is at the cusp of the realistic resolution for that passage, and it makes sense to force open the straits past Murray Island in the 12km and 14km meshes, but not at 30km.  See the attached two equivalent figures for the 12km and 30km meshes. 

DECK_zoom_1

WC12_zoom_1

xylar:
Thanks @milenaveneziani for your question and @proteanplanet for an elegant answer.   I was imagining something along these lines and it's nice to have it clearly illustrated.

xylar:
> I volunteer to do this by next Monday. First, I'll make sure everyone has contributed, so I'll ping all in the mesh review group, and get them to add to the confluence list if they would like to.

@proteanplanet, that's okay with me but it does mean this process is going to drag on for awhile.  As I'm sure you know, I prefer to work faster than the group and the project as a whole seems to like to.  I will be gone next week so it's either now or in 2 weeks.

maltrud:
i am not really convinced that opening these passages even for the 60to30km case is a bad idea.  perhaps the flow to the south of Victoria Island won't be particularly realistic, but no flow is less realistic.  and we may have to make things 2 cells wide for the B-grid issue.  i just can't imagine that it matters.  in addition, it seems unlikely that any of our future production grids will have resolution greater than 15-20km in the Arctic.  i'm worried that making things resolution dependent will unnecessarily complicate the process for essentially no reason.

proteanplanet:
@maltrud @xylar @milenaveneziani 
OK. I hear you, let's cut the scale dependency as a step too far, and put it off for Phase 3 mesh generation activities.  There remain inherent scale assumptions in what we are doing, so some requests for straits will need to take the cut.

milenaveneziani:
@proteanplanet: I just finished going through all the plots you posted for both meshes. There is only one other thing that came to my attention: for the wc12 mesh, an island formed a bit downstream of Nares St (see region/plot 16), with 2 grid cells separating it from (the real) Ellesmere Island. Maybe it's OK, but wanted to double check with you.

eclare108213:
Since I am coming very late to this conversation, I have a very basic question:  why are 2 grids that are seemingly so similar needed - nearly the same resolution although one of them has 80,000 more cells?
Otherwise, this all looks great to me.  Re the passage south of Ellesmere Island, I'd recommend that it be included unless the resolution would make it ridiculously wide at low res.  I don't know about the ocean flow through there, but a sea ice consideration for high res is whether the ice would be able to break out in the summer, particularly if it gets locked in as fast ice and then grows too thick through the 'ice pump' mechanism or weird temperature/salinity characteristics in the ocean.  Likewise, I think Nares Strait ought to also be included as a matter of course.

vanroekel:
@eclare108213 both meshes are under consideration for Water cycle v2 but we will not support both meshes.  In my view it makes most sense to go with the 14 km version unless we find that there is a compelling science reason to use the slightly higher resolution version, which is why we are keeping both for now.

mark-petersen:
> @mark-petersen, can I push a fix for the periodicity of region_Bering_Sea.geojson to this branch or would you like to handle that some other way?

@xylar and all, please push any changes to this branch. Be sure to change them in WC14 in #584 as well if they apply to both.

vanroekel:
@proteanplanet , @qingli411 , @milenaveneziani , @maltrud , and @xylar I've been having some discussions with Xylar and Andrew about the mesh design and I'd like to request that we focus this initial PR on resolution enhancement regions only.  Looking back through the discussion I think the only point to discuss is whether or not we want to enhance resolution for Indonesian through flow and/or Agulhas.  I would appreciate your thoughts on this.  My feeling is that we can't sufficiently refine those regions without becoming too expensive for use in WC production runs.

Also, I would also like your opinions on the inclusion of WC12 at all?  It is very unclear to me that going from 14 to 12 will improve the climate.  It is fine to merge this in, but I'm not convinced it is worth testing in e3sm.

vanroekel:
an important clarification -- the discussion on channels will not be lost.  We will return to this after finalizing the resolution regions.

proteanplanet:
@vanroekel  Happy to proceed with WC14 as is, excepting the possible addition mentioned in https://github.com/MPAS-Dev/MPAS-Model/pull/584#issuecomment-638594717.
Pending response from @maltrud, we may proceed with WC12 for InteRFACE, possibly adjusted for ICoM.

pwolfram:
Thanks everyone-- my plan is to take the same one as used for InteRFACE and low-res WC and then composite a higher resolution patch for mid-Atlantic hurricane / estuary simulation on top.  Have a bit of this workflow ready to go and can share once this mesh is accepted.

vanroekel:
@proteanplanet and @milenaveneziani just a note that I would like to approve this today based on the regions of resolution only.  Also with full understanding this won't be the final mesh for E3SM.  We will verify channels and islands in a later revision.  Please post here if you approve the mesh based on high res regions only

pwolfram:
A question, it doesn't look like we are constant high-res near Norway-- seems like we don't actually want to have any gradation in resolution there-- is this just a bug in production of the mesh?  I'm assuming resolution in this subset of the earth (shown below) should be constant 12 or 14km, right?  There is a similar issue near the Bearing Strait / Korea.

image

image

image

xylar:
@pwolfram, the second image you posted shows the periodicity artifact that my last commit should fix.  @mark-petersen and @pwolfram, can the two of your work together to change the region shapes to address the concerns raised by the other 2 images and repost a new resolution image today?

vanroekel:
@pwolfram do you want to push a fix for the regions you noted in the other two regions?  I'd like to do a spin up today so @jonbob can make E3SM files tomorrow.  I think it also may be okay to wait for a later revision, what do you think?

milenaveneziani:
same here

pwolfram:
@vanroekel, since I didn't make the original files, I wouldn't be exactly sure why they went wrong so I don't think this will be fast for me to push for today because of other commitments.  

I'm guessing this may be more involved since it already has slipped through testing.  I could help later in another revision.  I would guess there may be additional tweaks needed and putting this on the next iteration list seems the best approach to not slow things down.

Key thing is to keep track of these things-- what is the best way to do this?

pwolfram:
@vanroekel, could we just move over @xylar's fixes in #584 here?  It seems like we want the same geojson files between each of these cases, just with different resolution values.

vanroekel:
@xylar I think from conversations with @pwolfram and @proteanplanet they would like to keep the WC12 for ICoM/InterFACE, so I think we should keep both, but WC will only use the WC14. I will let them comment if they want to keep pushing on this mesh.  For me the priority is WC14

xylar:
@pwolfram and @proteanplanet, if we want to maintain two meshes with a lot of redundancy in the long run, it would be good to reorganize things so they share the shapes used to construct them to avoid the current redundancy (in both shapes and commits).

xylar:
@pwolfram, you were concerned about the metadata describing the choices in mesh resolution.  The current statement is:
```
description = MPAS Contiguous United States Plus mesh for E3SM version
              ${e3sm_version}, with a focused ${min_res}-km resolution
              around North America and ${levels} vertical levels
```
This clearly needs revision.  Would you be willing to make a first attempt that this?  I will use the same for the WC14 (since the `${min_res}` gets replaced automatically by the config parser).  Thanks in advance!

pwolfram:
@xylar, I propose the following subject to discussion.  I strongly argue for inclusion of our PR# in the string because we are using the github discussion to document.

```
description = MPAS North America and Arctic Focused Water Cycle mesh for E3SM version
              ${e3sm_version}, with a focused ${min_res}-km resolution
              around North America and ${levels} vertical levels as documented in PR#555.
```

xylar:
That works for me.  I will use that for now.  I will use the full URL for the PR because the PR number won't be helpful for an E3SM user on its own.

pwolfram:
Thanks @xylar!

xylar:
Adding the "Don't merge" tag until changes to #584 get brought in and metadata gets updated.

proteanplanet:
> @pwolfram and @proteanplanet, if we want to maintain two meshes with a lot of redundancy in the long run, it would be good to reorganize things so they share the shapes used to construct them to avoid the current redundancy (in both shapes and commits).

Thanks @xylar.  The critical deciding factor will likely be the be timing statistics for WC14 for InteRFACE.   @pwolfram and I will discuss this in the coming week.

vanroekel:
@xylar I've rebased and tested this.  I think it is ready to merge, but want to check with you as you have a comment above about metadata.  That PR is merged but want to make sure you are good with this going in.

xylar:
@mark-petersen, is the timing right for @vanroekel to merge this or should we wait for an `e3sm/develop` merge before that?

mark-petersen:
Should this still be merged in, or was it superseded by the 14km version in https://github.com/MPAS-Dev/MPAS-Model/pull/628 or others?

xylar:
@mark-petersen, my feeling is that all revisions need to be in the commit history for `ocean/develop` even if we don't end up using them.  If we close the PR and delete the branch, all info on this mesh is lost.  @vanroekel, what is your feeling?

mark-petersen:
I agree. Let's merge in, so we have both 14 and 12km options.

PR #556, opened 2020-05-13
Fixes cvmix driver area weighted average bugs

vanroekel:
Fixes a bug in the area weighted averaging for 3 CVMix quantities

Currently there are three edge based quantities: surfaceFrictionVelocity, RiTopOfCell, bulkRichardsonNumberShear.  The averaging to cell centers is resulting in a value 2X what is expected.  The averaging is dcEdge/2 * dvEdge times the value on the edge divided by areaCell.  However, the area of each triangle should be dcEdge / 2 * dvEdge / 2.  This PR fixes these bugs in each place.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -291,6 +291,10 @@
 					description="If true, the N2 limiting is applied to the horizontal diffusion term"
 					possible_values=".true. or .false."
 		/>
+		<nml_option name="config_Redi_apply_flux_limiter" type="logical" default_value=".true." units="unitless"
+					description="If true, apply a flux limiter to guarantee physical bounds on T&S"
+					possible_values=".true. or .false."
+		/>
 	</nml_record>
 	<nml_record name="GM_eddy_parameterization" mode="forward">
 		<nml_option name="config_use_GM" type="logical" default_value=".false." units="NA"
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -291,10 +291,6 @@
 					description="If true, the N2 limiting is applied to the horizontal diffusion term"
 					possible_values=".true. or .false."
 		/>
-		<nml_option name="config_Redi_apply_flux_limiter" type="logical" default_value=".true." units="unitless"
-					description="If true, apply a flux limiter to guarantee physical bounds on T&S"
-					possible_values=".true. or .false."
-		/>
 	</nml_record>
 	<nml_record name="GM_eddy_parameterization" mode="forward">
 		<nml_option name="config_use_GM" type="logical" default_value=".false." units="NA"
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -1987,7 +1987,7 @@
 			 description="List of edges that border each of the cells that straddle each edge."
 		/>
 		<var name="weightsOnEdge" type="real" dimensions="maxEdges2 nEdges" units="unitless"
-			 description="Reconstruction weights associated with each of the edgesOnEdge."
+			 description="Reconstruction weights associated with each of the edgesOnEdge, used to reconstruct the tangentialVelocity from normalVelocities on neighboring edges."
 		/>
 		<var name="dvEdge" type="real" dimensions="nEdges" units="m"
 			 description="Length of each edge, computed as the distance between verticesOnEdge."
@@ -2184,6 +2184,9 @@
 		<var name="daysSinceStartOfSim" type="real" dimensions="Time" units="days"
 			 description="Time since simulationStartTime, for plotting"
 	 	/>
+		<var name="edgeAreaFractionOfCell" type="real" dimensions="maxEdges nCells" units="unitless"
+			 description="edgeAreaFractionOfCell(iEdge, iCell) is the fractional area of cell iCell encompassed by the triangle with edge iEdge connected to the cell center of cell iCell. On a perfect planar hex mesh this is always 1/6. This weighting is used to interpolate scalars from edges to cell centers. Use 2*edgeAreaFractionOfCell to interpolate normal vectors at edges to vector norms at cell centers."
+		/>
 		<var name="salinitySurfaceRestoringTendency" type="real" dimensions="nCells Time" units="m PSU/s" packages="activeTracersSurfaceRestoringPKG"
 			 description="salinity tendency due to surface restoring"
 		/>
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -1382,7 +1382,7 @@ contains
       integer :: iCell, iEdge, i, k, err, timeLevel
       integer, pointer :: indexTempFlux, indexSaltFlux
       real (kind=RKIND) :: numerator, denominator, turbulentVelocitySquared, fracAbsorbed, fracAbsorbedRunoff
-      real (kind=RKIND) :: deltaVelocitySquared, delU2, invAreaCell
+      real (kind=RKIND) :: sumSurfaceStressSquared
 
       type (field2DReal), pointer :: densitySurfaceDisplacedField, thermalExpansionCoeffField, salineContractionCoeffField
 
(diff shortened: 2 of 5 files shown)


vanroekel:
Tested in 30 year G-case here

https://portal.nersc.gov/project/e3sm/lvroekel/20200509_interpBuGTest.GMPAS-IAF.T62_oECv3_anvil_yr_10_30/

Comparison run is here

https://portal.nersc.gov/project/e3sm/lvroekel/20200403_GMPAS-IAF_noRedi/

There are numerous changes, but notably MLD biases are reduced and AMOC is improved 1.5-2Sv.

vanroekel:
@mark-petersen this bug was discovered in testing of new changes to Redi.  I would appreciate if you can evaluate this as soon as you can.

Also, I've fixed this same issue in redi and I think it may allow us to get rid of the flux limiter code we had to add, but will have to test this.

xylar:
A bit of detective work suggests this was introduced 7 years (!!) ago in this commit: https://github.com/MPAS-Dev/MPAS-Model/commit/2f0bb15bf161d5c0b5178e7917aad96c3923406a

@vanroekel, perhaps you already did this but it would be worth searching for similar mistakes elsewhere.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -205,6 +207,9 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThickness, laye
                iCellSelf = 2
             endif
 
+            areaSum = areaSum + 0.5_RKIND * dvEdge(iEdge) * dcEdge(iEdge)
+      !      rediKappaCell(iCell) = rediKappaCell(iCell) + 0.5_RKIND*RediKappa(iEdge) * invAreaCell * dvEdge(iEdge) * dcEdge(iEdge)
```
xylar, line 211: @vanroekel, is this for debugging, and should it be removed?

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -261,10 +266,10 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThickness, laye
                   redi_term2(iTr, k, iCell) = redi_term2(iTr, k, iCell) - flux_term2*invAreaCell
                   redi_term2_edge(iTr, k, iEdge) = -flux_term2
 
-                  dTracerDx = (tracers(iTr, k, cell2) - tracers(iTr, k, cell1))*dvEdge(iEdge)*0.25_RKIND
+                  dTracerDx = (tracers(iTr, k, cell2) - tracers(iTr, k, cell1))/dcEdge(iEdge)*0.5_RKIND
                   fluxRediZTop(iTr, k) = fluxRediZTop(iTr, k) + &
                                          0.5_RKIND*(slopeTriadUp(k, iCellSelf, iEdge) + slopeTriadDown(k - 1, iCellSelf, iEdge)) &
-                                         *dTracerDx*invAreaCell
+                                         *dTracerDx
```
xylar, line 272: This seems like a different bug fix, not mentioned in the PR itself, where the gradient was not being computed correctly?

xylar:
A quick grep suggests this miscalculation is happening many other places, but the 3 @vanroekel found are the only ones likely impacting our simulation results.  Still, these others should get attention in the future.

These don't matter because the factor of 2 error is in both the numerator and the denominator, but it's confusing:
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/analysis_members/mpas_ocn_high_frequency_output.F#L518
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/analysis_members/mpas_ocn_high_frequency_output.F#L520
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/analysis_members/mpas_ocn_high_frequency_output.F#L524
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/analysis_members/mpas_ocn_high_frequency_output.F#L528

In global stats, `areaEdge` is meant to be the sum of the two triangles for a given edge (`0.5*dcEdge*dvEdge`) but it is missing the factor of 0.5.  This means that global inetgrals of edge quantities are a factor of 2 too large.  Averages will be correct because the factor is again in numerator and denominator:
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/analysis_members/mpas_ocn_global_stats.F#L406
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/analysis_members/mpas_ocn_global_stats.F#L1071

It looks like an `areaEdge` that is 4 times too large is getting used to compute `k33`, but this same area is used to compute the total volume used to normalize `k33`, so the error doesn't matter:
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_gm.F#L303

Later in the same routine, `areaEdge` is computed correctly, which makes the first instance all the more confusing:
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_gm.F#L560

In diagnostics, I'm going to assume that several confusing factor cancel out and the calculation of KE on cell and vertex centers is right:
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_diagnostics.F#L372-L377
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_diagnostics.F#L436
* There's a factor of 0.25 to convert `dcEdge*dvEdge` into the area of an edge triangle
* There's a factor of 0.5 for the definition of KE
* There's a factor of 2 because we are only adding the normal velocity, not the tangential velocity (and I vaguely recall from one of the Ringler papers that this just magically works out to a factor of 2).

The factor of 2 definitely matters here, though I don't think we run with the MPAS (as opposed to the CVMix) Richardson mixing anymore.
* https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_vmix_coefs_rich.F#L565-L566

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
vanroekel: yes, thanks for catching that I will remove.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
vanroekel: This is not correct as is.  It should be changed as well, but looking at this again it is not right. the top change here should be undone, but the second one is correct.  the top one is 
```
dTracerDx = (tracers(iTr, k, cell2) - tracers(iTr, k, cell1))/dcEdge(iEdge)**dvEdge(iEdge)*dcEdge(iEdge)*0.25_RKIND
```
So it does have the same averaging as the CVMix issues.  I will clean this file up

vanroekel:
@xylar thanks for the review! and for finding the many other places this occurs.  I had done a grep for invAreaCell to find other spots, but obviously that was woefully inadequate.

xylar:
> I had done a grep for invAreaCell to find other spots, but obviously that was woefully inadequate.

I grepped for `dcEdge` and `dvEdge` on the same line but I suspect that's also going to miss some spots.  (For example, I saw the `kineticEnergyCell` computation only by coincidence that it looks so similar to the `kineticEnergyVertex`.)

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -730,16 +730,17 @@ subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnostic
...
       do iCell=1,nCells
          RiTopOfCell(:,iCell) = 100.0_RKIND
-         invAreaCell1 = 1.0_RKIND / areaCell(iCell)
+         areaSum = 0.0_RKIND 
          do k=2,maxLevelCell(iCell)
            shearSquared = 0.0_RKIND
            do i = 1, nEdgesOnCell(iCell)
              iEdge = edgesOnCell(i, iCell)
-             factor = 0.5_RKIND * dcEdge(iEdge) * dvEdge(iEdge) * invAreaCell1
+             factor = 0.5_RKIND * dcEdge(iEdge) * dvEdge(iEdge)
```
mark-petersen, line 738: @vanroekel thanks for the detective work here. Summing the factors and dividing by the sum guarantees that we are not off by a factor, so that is a nice choice. I will add a commit where we remove all these factors of 0.5, which are not needed when we normalize by the sum anyway.

xylar: @mark-petersen, please, please don't do that! The efficiency gain (if any) is not worth the confusion this causes about what the real area is. This confusion can then propagate to other code, as I believe happened here.

I would like us to do precisely the opposite and make sure areas are correct even when normalization makes this unnecessary. This confusion is just pointless. I can't imagine it saves any meaningful flops.

vanroekel:
@mark-petersen please hold off merging this.  @qingli411 pointed out a factor of two may be needed in the velocity shear average.  I think this may be related to the factor of 2 we needed for the Redi mixing term 3.  I need to think about this a bit.  It is different from this bug, but I don't want people to use these changes if merged right now.

mark-petersen:
OK. Everything is on hold right now because E3SM testing has been off-line for the past week due to stopped Sandia testing. I will touch base before scheduling this merge.

Now that you say this, I remember working through the interpolation from an edge normal vector to cell centers, and being surprised by a factor of two. I'll review that again see what applies here.

vanroekel:
@qingli411 is right.  there is a need for a factor of 2.  I'd still like to make the area fixes, but include an explicit factor of two with a comment so there isn't future confusion.  I'll make the change.

mark-petersen:
Yes, I agree. I had to convince myself again, but the formula for the magnitude of a vector at the cell center is:
```
 |v| = sqrt( 2 * sum_e ( a_e / A * v_e ^2 ) )
```
where v_e is the normal vector at each edge e, a_e is the triangle area, A is the polynomial area. The factor of 2 is what is in question here. If we were just taking a scalar RMS from edge to cell center, the formula would be 
```
T_{RMS} = sqrt( sum_e ( a_e / A * T_e^2  ) )
```
which is what we were expecting when this PR was made. The factor of 2 is because on a C-grid, you only get the normal component at each edge, so it's not a full representation of the magnitude at each edge.

To convince yourself, consider a hexagon and a vector field v=(1,0). We know that a_e/A = 1/6, and the answer should be |v|=1.
mesh hexagon
For Case 1 with v=(1,0) we have:
```
|v| = sqrt( 2 * 1/6 * ( 2*0 + 4*(sqrt(3)/2 )^2 ) = 1
```
For Case 2 with v=(0,1) we have 
```
|v| = sqrt( 2 * 1/6 * ( 2*1 + 4*(1/2 )^2 ) = 1
```
The 1/2 and sqrt(3)/2 come from the projections on 60 degree and 30 degree angles.

xylar:
@mark-petersen, I think the general version wouldn't have equal-sized triangles but the analysis should generalize.

In [Ringler et al. 2010](https://doi.org/10.1016/j.jcp.2009.12.007) Eqs. (63) and (64), KE is derived with the assumption that a sum over triangle areas of the normal component of velocity gives exactly half of the vector magnitude (with the other half presumed to come from the tangential component) but this is not proved.  A similar assertion is made in [Thuburn et al. 2009](https://doi.org/10.1016/j.jcp.2009.08.006) in the footnote at the bottom of page 8328 but again isn't proven.

In any case, I think we've caused ourselves a significant headache by not commenting on these various factors in these computations of vector magnitude at cell centers from edge normals (including the computation of KE in diagnostics).

mark-petersen:
> I think the general version wouldn't have equal-sized triangles but the analysis should generalize.

In my equation above `a_e` is the area of each triangle associated with edge e. I assumed equal-sized of areas in my example of a hexagon, where `a_e/A = 1/6`.

@xylar, thanks for looking through the papers. I agree that we caused ourselves a headache by not commenting properly on a confusing weighting scheme.

xylar:
@vanroekel, you had found what you thought were factor-of-two errors in 3 places related to CVMix.  Then, we realized a factor of 2 is needed for vector quantities.  Were all 3 places related to vector quantities and they're all okay?  Or are at least some scalars, and the factor of 2 is a problem?

vanroekel:
@xylar most spots were vectors that needed the factor of two, but there is one spot in redi mixing that should not have the factor of two.  My hunch is that this was causing some of the unphysical temperatures with redi mixing.  It's on my todo list to verify that.  But, I would still like to make the changes here for area averaging even though the factor of two was needed.  If we left the averaging as it is now, there would be smaller errors for all of our variable resolution regions.

xylar:
Thanks @vanroekel.  That confirms my impression that there are still some bugs to work out here.

> But, I would still like to make the changes here for area averaging even though the factor of two was needed. If we left the averaging as it is now, there would be smaller errors for all of our variable resolution regions.

I'm not quite sure what the concern is here.  The area of a cell should be the sum of the areas of the triangles associated with edges (to round-off).  And since edges are exactly half way between cell centers and exactly perpendicular to the line between cell centers according to the mesh spec,  the kite associated with an edge should be made up of two triangles that are exactly the same area.  So there should be no inexactness associated with different resolutions of the adjacent cells as far as I understand things.

vanroekel:
thanks for the explanation @xylar I think I understand what you are saying here.  I'm not nearly as familiar with the mesh spec.  If dcEdge is indeed perpendicular to dvEdge than I agree that the sum or the triangles would be areaCell.  however, when I was debugging this I was printing out the averaging of a constant redikappa to cell centers from edges and was not getting the same constant back.  That's why I was concerned with the current implementation with areaCell.  Let me recheck that diagnostic and post again later.  It's quite likely I had a conceptual bug there.

xylar:
@vanroekel, That's supposed to be the case.  I seem to recall in discussions with @dengwirda that JIGSAW can generate meshes where that's not the case and he thought we should relax that assumption to get some other desirable properties.  But I think the current assumption in the code is that the lines connecting cells on and edge is exactly perpendicular to linse connecting vertices on an edge.

dengwirda:
@xylar, @vanroekel: You should always be able to rely on orthogonality of edges with JIGSAW meshes, but (depending on the "kind" of mesh requested) the position of edge intersection may be different. The two cases:

**Voronoi / Delaunay**: polygon edges pass through the midpoints of triangle edges. (Triangle edges don't necessarily pass through polygon edge midpoints though).

**Power / Laguerre**: None of the edges necessarily pass through midpoints.

Currently, MPAS uses only the standard Voronoi mesh type, as per the original TRiSK papers, so none of this is an issue. If Power meshes were to be used in the future some of this may need revisiting.

vanroekel:
@xylar I just confirmed the concern was misplaced, I must have done something wrong in my initial rework of area averaging for the redi coefficient.  I checked again in python and found the sum of the triangles equal to areaCell.  I will update the code in this PR to use areaCell in the averaging.  Thanks for the help in better understanding the mesh.

vanroekel:
sure, not a problem.  I still may also remove that if.  It seems the limiting may be necessary even with the fixes here.  But if I keep it, will certainly fix indentation.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1513,14 +1512,15 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
...
           delU2 =  (surfaceStress(iEdge))**2
           deltaVelocitySquared = deltaVelocitySquared + factor * delU2
         enddo
-        surfaceStressMagnitude(iCell) = sqrt(deltaVelocitySquared)
+        surfaceStressMagnitude(iCell) = sqrt(deltaVelocitySquared / areaCell(iCell))
 
        ! compute surface friction velocity
-         surfaceFrictionVelocity(iCell) = sqrt(surfaceStressMagnitude(iCell) / rho_sw)
+       ! NOTE that the factor of 2 is from averaging dot products to cell centers on a C-grid
+         surfaceFrictionVelocity(iCell) = sqrt(2.0_RKIND*surfaceStressMagnitude(iCell) / rho_sw)
```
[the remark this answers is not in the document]
vanroekel: @mark-petersen thanks for catching this.  That is a bad mistake.  What you have is what the code should be.  The third line is correct.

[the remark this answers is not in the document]
mark-petersen: Thanks. Changed those lines in 6f7e27f9a.

mark-petersen:
Rebased, corrected `surfaceStressMagnitude` and `surfaceFrictionVelocity`.

src/core_ocean/shared/mpas_ocn_init_routines.F
```diff
@@ -308,6 +308,54 @@ subroutine ocn_init_routines_setup_sign_and_index_fields(meshPool)!{{{
...
+         do i = 1, nEdgesOnCell(iCell)
+            iEdge = edgesOnCell(i, iCell)
+            ! edgeAreaFractionOfCell is the fractional area of cell iCell
+            ! encompassed by the triangle with edge iEdge connected to the cell
+            ! center of cell iCell. On a perfect planar hex mesh this is always
+            ! 1/6. This weighting is used to interpolate scalars from edges to
+            ! cell centers. Use 2*edgeAreaFractionOfCell to interpolate normal
+            ! vectors at edges to vector norms at cell centers."
+            edgeAreaFractionOfCell(i, iCell) = &
+               0.25_RKIND * dcEdge(iEdge) * dvEdge(iEdge) / areaCell(iCell)
```
mark-petersen, line 353: I've introduced a new variable, `edgeAreaFractionOfCell`, which is the fractional area of the triangle on that edge. I hope this name is clear, and I'm open to suggestions. Triangle is already used for the dual cells. Hopefully this will clarify the confusion we had on this PR, and may same some compute time, as this exact sequence appears many times per timestep for every cell.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -727,19 +727,18 @@ subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnostic
...
          RiTopOfCell(:,iCell) = 100.0_RKIND
-         invAreaCell1 = 1.0_RKIND / areaCell(iCell)
          do k=2,maxLevelCell(iCell)
            shearSquared = 0.0_RKIND
            do i = 1, nEdgesOnCell(iCell)
              iEdge = edgesOnCell(i, iCell)
-             factor = 0.5_RKIND * dcEdge(iEdge) * dvEdge(iEdge) * invAreaCell1
              delU2 = (normalVelocity(k-1,iEdge) - normalVelocity(k,iEdge))**2
-             shearSquared = shearSquared + factor * delU2
+             shearSquared = shearSquared + edgeAreaFractionOfCell(i,iCell) * delU2
```
mark-petersen, line 738: In this PR I am only changing from `factor` to `edgeAreaFractionOfCell` in what was originally changed in this PR, since these are corrections. I tested before versus after this substitution, and it is not bfb but machine precision (1e-13 or less). I will change them in the rest of the code in a different PR.

mark-petersen:
@vanroekel and @xylar let me know if you think the addition of `edgeAreaFractionOfCell` is helpful, and if you like that name.

src/core_ocean/shared/mpas_ocn_init_routines.F
xylar: Thanks @mark-petersen.  I think this is a good variable name and good documentation of how to use it.

src/core_ocean/shared/mpas_ocn_diagnostics.F
xylar: Probably not surprising that the code wouldn't be bit-for-bit given potential changes in order of operations. That sounds reasonable to me but we would also need to make sure the results aren't climate changing, unless the bug fixes themselves are climate changing (entirely possible).

vanroekel: @mark-petersen this is consistent with what I saw in testing, the sum of the triangles does not quite equal areaCell, it is off at about the level you show here (1e-13).  Xylar could be right here too, that it isn't the triangle area itself but order of operations.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1513,15 +1513,12 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
...
+          deltaVelocitySquared = deltaVelocitySquared + edgeAreaFractionOfCell(i,iCell) * delU2
         enddo
-        surfaceStressMagnitude(iCell) = sqrt(deltaVelocitySquared)
-
-       ! compute surface friction velocity
-         surfaceFrictionVelocity(iCell) = sqrt(surfaceStressMagnitude(iCell) / rho_sw)
-
+       ! NOTE that the factor of 2 is from averaging dot products to cell centers on a C-grid
+        surfaceStressMagnitude(iCell) = sqrt(2.0_RKIND * deltaVelocitySquared )
+        surfaceFrictionVelocity(iCell) = surfaceStressMagnitude(iCell) / rho_sw
```
vanroekel, line 1521: I don't think either this version or as originally listed in the code is right.  Let me look at this more deeply.

src/core_ocean/shared/mpas_ocn_init_routines.F
vanroekel: I like this as well.  Thanks for adding this!

src/core_ocean/Registry.xml
```diff
@@ -291,6 +291,10 @@
 					description="If true, the N2 limiting is applied to the horizontal diffusion term"
 					possible_values=".true. or .false."
 		/>
+		<nml_option name="config_Redi_apply_flux_limiter" type="logical" default_value=".true." units="unitless"
+					description="If true, apply a flux limiter to guarantee physical bounds on T&S"
+					possible_values=".true. or .false."
+		/>
```
vanroekel, line 297: I meant to remove this functionality before merge.  Let me do that and push again

vanroekel:
@mark-petersen I've just pushed some changes that should fix the issue in surface friction velocity.  The original code was correct and we needed a square root twice.  I also removed the Redi flux limit check, e3sm testing shows this is still necessary.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1513,15 +1513,12 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
...
+          deltaVelocitySquared = deltaVelocitySquared + edgeAreaFractionOfCell(i,iCell) * delU2
         enddo
-        surfaceStressMagnitude(iCell) = sqrt(deltaVelocitySquared)
-
-       ! compute surface friction velocity
-         surfaceFrictionVelocity(iCell) = sqrt(surfaceStressMagnitude(iCell) / rho_sw)
-
+       ! NOTE that the factor of 2 is from averaging dot products to cell centers on a C-grid
+        surfaceStressMagnitude(iCell) = sqrt(2.0_RKIND * deltaVelocitySquared)
+        surfaceFrictionVelocity(iCell) = sqrt(surfaceStressMagnitude(iCell) / rho_sw)
```
mark-petersen, line 1521: @vanroekel Looking at units:
```
surfaceStressMagnitude N m^{-2} = kg m s^{-2} m^{-2} = kg m^{-1} s^{-1}
surfaceFrictionVelocity = sqrt(surfaceStressMagnitude(iCell) / rho_sw)
 sqrt( kg m^{-1} s^{-2} / (kg/m^3) ) = sqrt( m^2 / s^2 )  = m/s
```
Yes, that is right and I agree with your formulation. I was tricked by the summation variable deltaVelocitySquared, which has units of `(N m^{-2})^2)`, not `( m/s)^2`. I'll change that to `sumSurfaceStressSquared` for clarity.

PR #559, opened 2020-05-14
Add support for a prerequisite tag in COMPASS

xylar:
Several ocean test cases depend on other ocean test cases.  Currently, this is used to create a mesh and/or initial condition that are then reused by one or more other tests.

Currently, the user simply must make sure the prerequisites are present when setting up a test case.  However, this merge makes sure that prerequisites are always present a regression suite and that they run before the dependent test case.  The main motivation for this change is future support for running multiple test cases from the nightly test suite in parallel.

This is done by parsing a tag like the following example from a `config_driver.xml` for the dependent test case:
```
<prerequisite core="ocean" configuration="global_ocean" resolution="QU240" test="init"/>
```

These tags will be added to individual test cases in separate PRs to the individual `<core>/develop` branches.

### DIFF
--- a/testing_and_setup/compass/manage_regression_suite.py
+++ b/testing_and_setup/compass/manage_regression_suite.py
@@ -147,7 +147,7 @@ def process_test_setup(test_tag, config_file, work_dir, model_runtime,
 # }}}
 
 
-def process_test_clean(test_tag, work_dir, suite_script):  # {{{
+def process_test_clean(test_tag, work_dir):  # {{{
     dev_null = open('/dev/null', 'a')
 
     # Process test attributes
@@ -304,7 +304,7 @@ def clean_suite(suite_tag, work_dir):  # {{{
     for child in suite_tag:
         # Process <test> children within the <regression_suite>
         if child.tag == 'test':
-            process_test_clean(child, work_dir, regression_script)
+            process_test_clean(child, work_dir)
 # }}}
 
 
--- a/testing_and_setup/compass/manage_regression_suite.py
+++ b/testing_and_setup/compass/manage_regression_suite.py
@@ -365,6 +365,31 @@ def get_test_case_procs(suite_tag):  # {{{
                 name = case.attrib['name']
                 cases.append(name)
 
+            prereqs = list()
+            for config_prereq in config_root.iter('prerequisite'):
+                prereq = dict()
+                for tag in ['core', 'configuration', 'resolution', 'test']:
+                    prereq[tag] = config_prereq.attrib[tag]
+
+                # Make sure the prerequisite is already in the test suite
+                found = False
+                for other_name, other_test in testcases.items():
+                    match = [prereq[tag] == other_test[tag] for tag in
+                             ['core', 'configuration', 'resolution', 'test']]
+                    if all(match):
+                        found = True
+                        prereq['name'] = other_name
+                        break
+
+                if not found:
+                    raise ValueError(
+                        'Prerequisite of {} does not precede it in the test '
+                        'suite: {} {} {} {}'.format(
+                            test_name, prereq['core'], prereq['configuration'],
+                            prereq['resolution'], prereq['test']))
+
+                prereqs.append(prereq)
+
             del config_root
             del config_tree
 
@@ -403,7 +428,9 @@ def get_test_case_procs(suite_tag):  # {{{
                                     'test': test_test,
                                     'path': test_path,
                                     'procs': procs,
-                                    'threads': threads}
+                                    'threads': threads,
+                                    'prereqs': prereqs}
+
     return testcases  # }}}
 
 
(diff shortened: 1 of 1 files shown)


testing_and_setup/compass/manage_regression_suite.py
```diff
@@ -147,7 +147,7 @@ def process_test_setup(test_tag, config_file, work_dir, model_runtime,
 # }}}
 
 
-def process_test_clean(test_tag, work_dir, suite_script):  # {{{
+def process_test_clean(test_tag, work_dir):  # {{{
```
xylar, line 150: This is just clean-up of an unused argument

testing_and_setup/compass/manage_regression_suite.py
```diff
@@ -417,7 +456,7 @@ def summarize_suite(suite_tag):  # {{{
 # }}}
 
 
-if __name__ == "__main__":
+def main ():  # {{{
```
xylar, line 459: Making `main()` a function to avoid having module-level variables (sometimes prevents bugs).

testing_and_setup/compass/manage_regression_suite.py
```diff
@@ -495,14 +534,15 @@ def summarize_suite(suite_tag):  # {{{
...
             setup_suite(suite_root, args.work_dir, args.model_runtime,
                         args.config_file, args.baseline_dir, args.verbose)
-            summarize_suite(suite_root)
+            summarize_suite(testcases)
             if args.verbose:
                 cmd = ['cat',
                        args.work_dir + '/manage_regression_suite.py.out']
                 print('\nCase setup output:')
-                print(subprocess.check_output(cmd))
+                print(subprocess.check_output(cmd).decode('utf-8'))
```
xylar, line 545: This looks like garbage in python 3 without the decode

testing_and_setup/compass/manage_regression_suite.py
```diff
@@ -495,14 +534,15 @@ def summarize_suite(suite_tag):  # {{{
         if args.setup:
             print("\n")
             print("Setting Up Test Cases:")
+            testcases = get_test_case_procs(suite_root)
             setup_suite(suite_root, args.work_dir, args.model_runtime,
                         args.config_file, args.baseline_dir, args.verbose)
-            summarize_suite(suite_root)
+            summarize_suite(testcases)
```
xylar, line 540: The changes require doing the parsing that was previously in `summarize_suite()` before doing `setup_suite()` so we can error out if the prerequisites are not present.  The info from this parsing is stored in nested `dict`s that will likely be useful for future efforts for parallelization.

xylar:
@mark-petersen, @matthewhoffman, and @pwolfram, this is work that feeds into work that  @jamilgafur is doing, supervised by @vanroekel and me, toward paralleling regression test suites.  For that work, it will be important to know which test cases must run before which others, and which can run in parallel.

xylar:
@matthewhoffman, I'm mostly wanting to make sure you are okay with these changes, and to have you run a quick test of your regression suite(s) with these changes to make sure nothing breaks.

xylar:
@mark-petersen, the easiest way to test this would be to copy this version of `manage_regression_suite.py` into the right place in #560 and run the nightly regression suite (and any others that are affected by prerequisites).  This worked well for me on my laptop, but I will also test on LANL IC as soon as I can.

xylar:
@jamilgafur and @vanroekel, I would definitely appreciate your feedback, too, particularly since this work is mainly supposed to help with the parallelization that @jamilgafur is working on.

pwolfram:
@xylar, just to be clear here-- you need this tag to build a dependency graph so you can execute in parallel, right?  Basically, something like if there is a prerequisite and it isn't fulfilled, move case to bottom of queue and if only one item in queue report error.

xylar:
@pwolfram, yep, that's the idea.  Since we don't typically run in parallel, I'm just checking to see that any prerequisites are already in the list of test cases to run when a new test case gets added, because otherwise serial execution gets broken.  In parallel, one approach would be what you suggest.

vanroekel:
@xylar Just a FYI that @jamilgafur has merged this to his local branch for testing the parallel nightly regression.  After that test,  I will approve.  If you'd prefer me to approve sooner, let me know and I can look more closely at the code itself to approve.

xylar:
@vanroekel, no rush.  My guess is that this won't go in particularly soon, so @jamilgafur should test as much as needed.

xylar:
@vanroekel, what is the order of merging this vs #593?  I had thought this would get approved and merged before too much work went into #593.

vanroekel:
sorry @xylar I meant to come back and approve this.  I will approve now as it works great in our testing.  then we'll rebase #593.

xylar:
Great, thanks @vanroekel!

PR #485, opened 2020-03-24
Merge coastal to ocean/develop: COMPASS and sediment additions

mark-petersen:
Changes in coastal for COMPASS python scripts are needed in ocean/develop.

PR #560, opened 2020-05-14
Add prerequisite tags to all necessary test cases

xylar:
These test cases depend on other test cases to run properly. Before this change, the only way to know that is via symlinks to directories outside of this test case.  With this change, it will be easier for regression test suites to make sure prerequisites are present and that they always run before the test cases that depend on them (even if we run tests cases in parallel).

This PR should be tested in conjunction with #559 but can be merged independently of that PR without doing any harm.  The new `prerequisite` tags simply will be ignored until the changes to `manage_regression_suite.py` make their way into `ocean/develop`.

### DIFF
--- a/testing_and_setup/compass/ocean/Gaussian_hump/USDEQU120cr10rr2/delaware/config_driver.xml
+++ b/testing_and_setup/compass/ocean/Gaussian_hump/USDEQU120cr10rr2/delaware/config_driver.xml
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="Gaussian_hump" resolution="USDEQU120cr10rr2" test="build_mesh"/>
 	<case name="init">
 		<step executable="./run.py" quiet="true" pre_message=" * Running init" post_message="     Complete"/>
 	</case>
--- a/testing_and_setup/compass/ocean/drying_slope/hybrid/1km/config_driver.xml
+++ b/testing_and_setup/compass/ocean/drying_slope/hybrid/1km/config_driver.xml
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="1km"/>
 	<case name="init_step">
 		<step executable="./run.py" quiet="true" pre_message=" * Running init_step" post_message="     Complete"/>
 	</case>
--- a/testing_and_setup/compass/ocean/drying_slope/hybrid/250m/config_driver.xml
+++ b/testing_and_setup/compass/ocean/drying_slope/hybrid/250m/config_driver.xml
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="250m"/>
 	<case name="init_step">
 		<step executable="./run.py" quiet="true" pre_message=" * Running init_step" post_message="     Complete"/>
 	</case>
--- a/testing_and_setup/compass/ocean/drying_slope/sigma/1km/config_driver.xml
+++ b/testing_and_setup/compass/ocean/drying_slope/sigma/1km/config_driver.xml
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="1km"/>
 	<case name="init_step">
 		<step executable="./run.py" quiet="true" pre_message=" * Running init_step" post_message="     Complete"/>
 	</case>
--- a/testing_and_setup/compass/ocean/drying_slope/sigma/250m/config_driver.xml
+++ b/testing_and_setup/compass/ocean/drying_slope/sigma/250m/config_driver.xml
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="250m"/>
 	<case name="init_step">
 		<step executable="./run.py" quiet="true" pre_message=" * Running init_step" post_message="     Complete"/>
 	</case>
--- a/testing_and_setup/compass/ocean/drying_slope/zstar/1km/config_driver.xml
+++ b/testing_and_setup/compass/ocean/drying_slope/zstar/1km/config_driver.xml
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="1km"/>
 	<case name="init_step">
 		<step executable="./run.py" quiet="true" pre_message=" * Running init_step" post_message="     Complete"/>
 	</case>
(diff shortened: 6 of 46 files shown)


xylar:
@mark-petersen, see https://github.com/MPAS-Dev/MPAS-Model/pull/559#issuecomment-628497165 for comments about testing this PR.

xylar:
@pwolfram, you are welcome to test this out but mostly I'd appreciate you taking a look at the tags I added to test cases that you manage and make sure they seem correct.  And, of course, that you're okay with these tags being added.  It should be harmless to *not* include the tags if you would rather not, it just might mean you can't safely include your test cases in a regression test suite running in parallel, down the road when we hope to have that capability.

xylar:
@jamilgafur and @vanroekel, this PR is less critical for you to have a look at, but you will eventually want the version of the QU240 test case with the `prerequisite` tags for the parallel version of the nightly regression test suite.

xylar:
## Testing

On Grizzly:
* I ran the nightly regression test suite with this PR and `manage_regression_suite.py` from #559.  
* I ran the LIGHT regression suite and it had an error in the SOMA 32km test case related to a missing local file `check_particle_sampling.py` that I have previously mentioned (https://github.com/MPAS-Dev/MPAS-Model/pull/480#issuecomment-611407161).
* I discovered that the land-ice fluxes regression suite is broken, so I will fix that separately and won't try to test it here.
* I ran the RPE test suite for 1 hour, and all but the last test completed successfully.  The baroclinic_channel/1km test case started fine but timed out.

I didn't try to check bit-for-bit against the current `ocean/develop`, so @mark-petersen, that might be a good test with the nightly regression suite.

pwolfram:
@xylar, is this the file you are looking for: `https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/coastal/testing_and_setup/compass/ocean/soma/analysis/check_particle_sampling.py`?  Sounds like the link was just broken?

testing_and_setup/compass/ocean/Gaussian_hump/USDEQU120cr10rr2/delaware/config_driver.xml
```diff
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="Gaussian_hump" resolution="USDEQU120cr10rr2" test="build_mesh"/>
```
pwolfram, line 2: 👍

testing_and_setup/compass/ocean/drying_slope/hybrid/1km/config_driver.xml
```diff
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="1km"/>
```
pwolfram, line 2: 👍

testing_and_setup/compass/ocean/drying_slope/hybrid/250m/config_driver.xml
```diff
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="250m"/>
```
pwolfram, line 2: 👍

testing_and_setup/compass/ocean/drying_slope/sigma/1km/config_driver.xml
```diff
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="1km"/>
```
pwolfram, line 2: 👍

testing_and_setup/compass/ocean/drying_slope/sigma/250m/config_driver.xml
```diff
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="250m"/>
```
pwolfram, line 2: 👍

testing_and_setup/compass/ocean/drying_slope/zstar/1km/config_driver.xml
```diff
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="1km"/>
```
pwolfram, line 2: 👍

testing_and_setup/compass/ocean/drying_slope/zstar/250m/config_driver.xml
```diff
@@ -1,4 +1,5 @@
 <driver_script name="run_test.py">
+	<prerequisite core="ocean" configuration="drying_slope" resolution="meshes" test="250m"/>
```
pwolfram, line 2: 👍

xylar:
>  is this the file you are looking for: https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/coastal/testing_and_setup/compass/ocean/soma/analysis/check_particle_sampling.py? Sounds like the link was just broken?

It looks likely.  It seems like the test case isn't making the local link it should be.  If you want that test case to get stress tested, it would be worth fixing.

pwolfram:
@xylar, should this be bundled into some existing PR or its own PR?

xylar:
@pwolfram, it's not related to any existing PRs I'm aware of so I think it should be its own.

xylar:
@mark-petersen, thanks for the review.  I agree, let's wait until #559 has made it to `ocean/develop` before merging this.

mark-petersen:
Adding don't merge label until #559 has made it's way to ocean/develop.

xylar:
@mark-petersen, I rebased this PR now that #559 and `develop` have been merged to `ocean/develop`.  I also added the `prerequisite` tag to the `WC14` test case.  I believe this is ready to merge when you are accepting COMPASS merges to `ocean/develop`.

xylar:
@sbrus89, I'm trying to get this work in following #485.  It seems like there are tons of new test cases from `ocean/coastal` that need prerequisite tags.  Could we have a conversation about this sometime soon?  The `hurricane` tests are a particular sticking point because you currently link to the same `config_driver.xml` for many tests, but each of these has a different prerequisite.

xylar:
@sbrus89, and update: I went ahead and added prerequisite tags to all the `hurricane` tests, which required me to make copies of several `config_driver.xml` files.  I hope that won't be a problem.  Please take a look and make sure I didn't make any obvious mistakes.  Currently, the `prerequisite` tags don't get used so there's not a great way to test if they are correct. We can test more extensively once we have parallel execution of the regression suite.

xylar:
I believe we will go with a different approach to address this problem.  It will be part of development of COMPASS v. 1.0 on the new COMPASS repo: https://github.com/MPAS-Dev/compass

PR #561, opened 2020-05-14
Fix land-ice fluxes regression suite

xylar:
It previously pointed to the old (E3SM v1) location for the QU240 test case with ice-shelf cavities.

Also, fix the name of the LIGHT regression suite.


xylar:
## Testing

I ran the land-ice regression suite successfully to completion on Grizzly.

I verified that the name of the resulting python script is not correct for the LIGHT regression suite but I didn't run it.  (One of its test cases is broken anyway, see #560.)

PR #562, opened 2020-05-14
Enable missing_value to override default_value for fields

MiCurry:
This commit allows missing_value to override the default value of a field if
default_value is not specified for a field in a core's Registry.xml file. If
missing_value and default_value are both present for a field, then the field
will be initialized to default_value.

Specifying missing_value for a field in the Registry.xml will still create a
missing_value NetCDF attribute if that field is written (even if default_value
is also specified).

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -1063,6 +1063,12 @@ int parse_var_array(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t var
 	// Determine field type and default value.
 	get_field_information(vararrtype, vararrdefaultval, default_value, vararrmissingval, missing_value, &type);
 
+	// If a default_value is not specified, but a missing_value is, then set the
+	// default_value (defaultValue) to missing_value
+	if(!vararrdefaultval && vararrmissingval) {
+		snprintf(default_value, 1024, "%s ! defaultValue taking specified missing_value", missing_value);
+	}
+
 	// Determine ndims, hasTime, and decomp type
 	build_dimension_information(registry, var_arr_xml, &ndims, &hasTime, &decomp);
 	ndims++; // Add a dimension for constituents in var_array
@@ -1488,6 +1494,12 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
 
 	// Determine field type and default value.
 	get_field_information(vartype, vardefaultval, default_value, varmissingval, missing_value, &type);
+
+	// If a default_value is not specified, but a missing_value is, then set the
+	// default_value (defaultValue) to missing_value
+	if(!vardefaultval && varmissingval) {
+		snprintf(default_value, 1024, "%s ! defaultValue taking specified missing_value", missing_value);
+	}
 
 	// Determine ndims, hasTime, and decomp type
 	build_dimension_information(registry, var_xml, &ndims, &hasTime, &decomp);


MiCurry:
This PR closes issue #562.

As mentioned in the initial comment of this PR, if both `missing_value` and `default_value` are specified, the field will be initialized with `default_value`. If some other action should occur, I'm happy to update this PR.

mgduda:
@MiCurry This looks good so far. I think the same logic as you've implemented for `var`s is also needed for `var_array`s.

MiCurry:
> @MiCurry This looks good so far. I think the same logic as you've implemented for `var`s is also needed for `var_array`s.

Ah, I forgot about `var_array`s. Thanks @mgduda.

I just pushed an update that contains the same changes for `var_array`s. I also moved the change for `var`s to below the call to `get_field_information`. This is consistent with the `var_array` change and seems a bit more organized/logical.

MiCurry:
All, I have updated the commit and PR message for this PR to be more explicit as to what occurs when `missing_value` and `default_value` are specified. From our telecom on 5/26 I believe this implementation is correct, but please let me know if it isn't!

mgduda:
@MiCurry Is it possible that your last force-push of 6fc82f0 removed the changes for `var_array`s?

MiCurry:
> @MiCurry Is it possible that your last force-push of [6fc82f0](https://github.com/MPAS-Dev/MPAS-Model/commit/6fc82f0fab7191398afa4b08e033a192afee6263) removed the changes for `var_array`s?

@mgduda You are correct. The last force push just re-added the var array changes!

xylar:
As part of testing this PR, I tried to assign a missing value to one `var` in a `var_array`:
``` xml
                <var name="testMissingVar" type="real" dimensions="nVertLevels nCells Time" units="unitless" missing_value="9.969209968386869e36"
                        description="A variable that is zero where valid and missing_value where not"
                />
                <var_array name="testMissingArray" type="real" dimensions="nVertLevels nCells Time">
                        <var name="testMissingArrayVar" array_group="testMissingGroup" units="unitless" missing_value="9.969209968386869e36"
                                description="A variable that is one where valid and missing_value where not"
                        />
                </var_array>
```
This didn't work as expected -- when I look in `inc/structs_and_variables.inc`, only `testMissingVar` has the `missing_value` attribute.  I realize this goes beyond this PR but @mgduda and @MiCurry, is that the expected behavior?  Are `default_value` and `missing_value` only available at the `var_array` level, not at the individual `var` level for a `var_array`?

xylar:
Thanks for this great work, @MiCurry!  I tested using a branch (which I will keep around only until this PR gets merged into `ocean/develop`): https://github.com/xylar/MPAS-Model/pull/new/ocean/test_missing_value

I found that everything worked as expected if I assigned a real to `missing_value`.  But If I assigned `FILLVAL`, I got a compile-time error:
```
../inc/structs_and_variables.inc:8479:39:

       r3Ptr(1) % defaultValue = FILLVAL ! defaultValue taking specified missing_value
                                       1
Error: Symbol 'fillval' at (1) has no IMPLICIT type
```
This is because `FILLVAL` isn't getting detected and replaced with the appropriate `MPAS_*_FILLVAL` in `default_value` in the same way as it is in `missing_value`.  See:
https://github.com/MiCurry/MPAS-Model/blob/registry/missing_value/src/tools/registry/gen_inc.c#L451-L472
To fix this, these lines:
https://github.com/MiCurry/MPAS-Model/blob/registry/missing_value/src/tools/registry/gen_inc.c#L428-L449
would need to be something like:
``` c
	if (strcmp(vartype, "real") == 0){
		(*type) = REAL;
		if(!varval){
			snprintf(default_value, 1024, "0.0");
		} else if(strcmp(varmissval, "FILLVAL") == 0) {
			snprintf(missing_value, 1024, "MPAS_REAL_FILLVAL");
		} else {
			snprintf(default_value, 1024, "%s", varval);
		}
	} else if (strcmp(vartype, "integer") == 0){
		(*type) = INTEGER;
		if(!varval){
			snprintf(default_value, 1024, "0");
		} else if(strcmp(varmissval, "FILLVAL") == 0) {
			snprintf(missing_value, 1024, "MPAS_INT_FILLVAL");
		} else {
			snprintf(default_value, 1024, "%s", varval);
		}
	} else if (strcmp(vartype, "text") == 0){
		(*type) = CHARACTER;
		if(!varval){
			snprintf(default_value, 1024, "''");
		} else if(strcmp(varmissval, "FILLVAL") == 0) {
			snprintf(missing_value, 1024, "MPAS_CHAR_FILLVAL");
		} else {
			snprintf(default_value, 1024, "%s", varval);
		}
	}
```
(Boy, would this code be more elegant and readable in python...)

MiCurry:
>  I realize this goes beyond this PR but @mgduda and @MiCurry, is that the expected behavior? Are default_value and missing_value only available at the var_array level, not at the individual var level for a var_array?

That was the expected behavior of my implementation and I think the current behavior of master. `default_value` and `missing_value` are only available at the `var_array` level. Here is a snippet of the atmosphere's core scalar var_array (with a few members removed) from core_atmosphere/inc/structs_and_variables.inc where the `defaultValue` and `missingValue` is set for a var_array.
``` Fortran
      r3Ptr(1) % dimNames(1) = 'num_scalars'
      r3Ptr(1) % dimNames(2) = 'nVertLevels'
      r3Ptr(1) % dimNames(3) = 'nCells'

      r3Ptr(1) % defaultValue = 0.0
      allocate(r3Ptr(1) % attLists(size(r3Ptr(1) % constituentNames, dim=1)))
      do index_counter = 1, size(r3Ptr(1) % constituentNames, dim=1)
         allocate(r3Ptr(1) % attLists(index_counter) % attList)
      end do
      if (associated(newSubPool)) then
         call mpas_pool_get_dimension(newSubPool, 'index_qv', const_index)
      end if
      if (const_index > 0) then
         call mpas_add_att(r3Ptr(1) % attLists(const_index) % attList, 'long_name', 'Water vapor mixing ratio')
         call mpas_add_att(r3Ptr(1) % attLists(const_index) % attList, 'units', 'kg kg^{-1}')
         r3Ptr(1) % missingValue = MPAS_REAL_FILLVAL
         r3Ptr(1) % constituentNames(const_index) = 'qv'
      end if
      if (associated(newSubPool)) then
         call mpas_pool_get_dimension(newSubPool, 'index_qc', const_index)
      end if
      if (const_index > 0) then
         call mpas_add_att(r3Ptr(1) % attLists(const_index) % attList, 'long_name', 'Cloud water mixing ratio')
         call mpas_add_att(r3Ptr(1) % attLists(const_index) % attList, 'units', 'kg kg^{-1}')
         r3Ptr(1) % missingValue = MPAS_REAL_FILLVAL
         r3Ptr(1) % constituentNames(const_index) = 'qc'
      end if
      if (associated(newSubPool)) then
      r3Ptr(1) % block => block
```

I believe `r3Ptr(1) % defaultValue` and `r3Ptr(1) % missingValue` effect all members of the array. In the code snippet above there is a single assignment for `r3ptr(1) % defaultValue`, but one for each member for the `r3Ptr(1) % missingValue` assignment. I think the assignments for `missingValue` are all identical statements, so even if the values assigned to them would be different, the entire array would get the last one that was executed (I think).

MiCurry:
> I found that everything worked as expected if I assigned a real to `missing_value`. But If I assigned `FILLVAL`

@xylar, ah I did not realize one could use `FILLVAL` to set `missing_value` and `default_value`. Thanks for pointing that out, and thanks for giving the code to fix it. I'll get to adding that!

xylar:
>  thanks for giving the code to fix it.

You're welcome, with the caveat that I didn't test that code *at all* so it could be completely useless.

mgduda:
I think the fix may be as simple as changing `vararrmissingval` to `missing_value` and similarly for the code in `parse_var` in the current PR:

```Diff.patch
@@ -1066,7 +1066,7 @@ int parse_var_array(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t var
        // If a default_value is not specified, but a missing_value is, then set the
        // default_value (defaultValue) to missing_value
        if(!vararrdefaultval && vararrmissingval) {
-               snprintf(default_value, 1024, "%s ! defaultValue taking specified missing_value", vararrmissingval);
+               snprintf(default_value, 1024, "%s ! defaultValue taking specified missing_value", missing_value);
        }
```

xylar:
> I think the fix may be as simple as changing `vararrmissingval` to `missing_value` and similarly for the code in `parse_var` in the current PR

@mgduda, ah, you're exactly right!  Thanks for catching this.  

My suggestion is wrong because `default_value` would be the expected thing coming out of `get_field_information()` but then it would get replaced in the new lines  @MiCurry has added.

xylar:
Thanks @MiCurry.  I'll test this again later tonight.

MiCurry:
Okay @mgduda, @xylar and everyone else, I have just pushed changes that changes `vararrmissingval` to `missing_value`. So now adding `FILLVAL` to `missing_value` now successfully compiles.

mark-petersen:
@matthewhoffman is my ocean test sufficient for approval, or would you like to test as well?

mark-petersen:
@MiCurry thanks for your work on this. Please merge, or I can if you prefer.

mgduda:
@mark-petersen I don't think @MiCurry is able to push to the `develop` branch, so I'll merge this today.

PR #552, opened 2020-05-08
Adds a marsh flooding coastal case

caozd999:
This PR adds two test cases that analyze the effects of salt marsh on coastal flooding. 
The cases implement marsh as a bottom drag in idealized land-coast-ocean transects. Each case includes two comparison experiments to show the effects of marsh width on coastal flooding.

PR #563, opened 2020-05-18
Move jigsaw_to_MPAS to mpas_tools

### LINKED ISSUE #550: Move jigsaw_to_MPAS to MPAS-Tools
@dengwirda has expressed interest in using [`jigsaw_to_MPAS`](https://github.com/MPAS-Dev/MPAS-Model/tree/ocean/develop/testing_and_setup/compass/ocean/jigsaw_to_MPAS) outside of COMPASS.

This would be useful for making new MPAS meshes for potential river and land components to come.

It would also mean that changes to `jigsaw_to_MPAS` would be more challenging, since they would involve:
* changing a branch on MPAS-Tools
* building a test conda package
* building a test conda environment
* running COMPASS with the test environment
This sounds daunting but it can be done without too much difficulty as long as the process is documented.

It could also have many advantages:
* obviously, other infrastructures and workflows besides COMPASS could use it
* `jigsaw_to_MPAS` would be a proper python module, rather than a "fake" one that we have to link to in every test case
* dependencies could be maintained as part of releases of the `mpas_tools` conda package, rather than through the separate `compass` metapackage
* ...

xylar:
This merge uses `mpas_tools.mesh.creation` instead of `jigsaw_to_MPAS`.  These tools are being moved to MPAS-Tools to allow them to be used by other workflows, see #550.

This merge includes:
* removing the `jigsaw_to_MPAS` module from COMPASS
* Updating the `compass` environment version to the one that includes `mpas_tools` with `mesh.creation`.
* Removing links to `jigsaw_to_MPAS` from several test cases
* Calling `build_mesh` and `inject_bathymetry` from `mpas_tools` in test cases that use these
* Switching from `mpas_tools.conversion` to `mpas_tools.mesh.conversion` (though the former still exists for backward compatibility)
* A small fix to some cartopy-related axis calls that seem to be depricated

closes #550

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/ARM60to10/init/define_base_mesh.py
+++ b/testing_and_setup/compass/ocean/global_ocean/ARM60to10/init/define_base_mesh.py
@@ -5,6 +5,7 @@ import numpy as np
 import mpas_tools.mesh.creation.mesh_definition_tools as mdt
 from mpas_tools.mesh.creation.coastal_tools import \
     signed_distance_from_geojson, mask_from_geojson
+from mpas_tools.viz.colormaps import register_sci_viz_colormaps
 from geometric_features import read_feature_collection
 import xarray
 import matplotlib
@@ -40,7 +41,7 @@ def cellWidthVsLatLon():
     fig = plt.figure()
     plt.clf()
     fig.set_size_inches(10.0, 10.0)
-    mdt.register_sci_viz_colormaps()
+    register_sci_viz_colormaps()
 
     # Create cell width vs latitude for Atlantic and Pacific basins
     QU1 = np.ones(lat.size)
--- a/testing_and_setup/compass/ocean/global_ocean/WC12/init/define_base_mesh.py
+++ b/testing_and_setup/compass/ocean/global_ocean/WC12/init/define_base_mesh.py
@@ -5,6 +5,7 @@ import numpy as np
 import mpas_tools.mesh.creation.mesh_definition_tools as mdt
 from mpas_tools.mesh.creation.coastal_tools import \
     signed_distance_from_geojson, mask_from_geojson
+from mpas_tools.viz.colormaps import register_sci_viz_colormaps
 from geometric_features import read_feature_collection
 import xarray
 import matplotlib
@@ -41,7 +42,7 @@ def cellWidthVsLatLon():
     fig = plt.figure()
     plt.clf()
     fig.set_size_inches(10.0, 14.0)
-    mdt.register_sci_viz_colormaps()
+    register_sci_viz_colormaps()
     #cmapBluesHalf = truncate_colormap(cmapIn='Blues', minval=0.0, maxval=0.7)
 
     # Create cell width vs latitude for Atlantic and Pacific basins
--- a/testing_and_setup/compass/ocean/global_ocean/WC14/init/define_base_mesh.py
+++ b/testing_and_setup/compass/ocean/global_ocean/WC14/init/define_base_mesh.py
@@ -5,6 +5,7 @@ import numpy as np
 import mpas_tools.mesh.creation.mesh_definition_tools as mdt
 from mpas_tools.mesh.creation.coastal_tools import \
     signed_distance_from_geojson, mask_from_geojson, distance_from_geojson
+from mpas_tools.viz.colormaps import register_sci_viz_colormaps
 from geometric_features import read_feature_collection
 import xarray
 import matplotlib
@@ -41,7 +42,7 @@ def cellWidthVsLatLon():
     fig = plt.figure()
     plt.clf()
     fig.set_size_inches(10.0, 14.0)
-    mdt.register_sci_viz_colormaps()
+    register_sci_viz_colormaps()
     #cmapBluesHalf = truncate_colormap(cmapIn='Blues', minval=0.0, maxval=0.7)
 
     # Create cell width vs latitude for Atlantic and Pacific basins


xylar:
This PR requires the changes to MPAS-Tools in https://github.com/MPAS-Dev/MPAS-Tools/pull/311

xylar:
## Testing

So far, I successfully tested `QU240/init` and `QU240wISC/init` test cases on my laptop, but I didn't check for bit-for-bit and such.  Much more testing is obviously needed.

xylar:
@mark-petersen and @pwolfram, this PR likely needs to go in at a point when `ocean/develop` and `ocean/coastal` are in sync with one another.  Can we try to coordinate that sometime soon?  Obviously, I can update any test cases on `ocean/coastal` that have been missed here, but I want to avoid any unnecessary frustration from this transition.

pwolfram:
@xylar, can you please clarify what you are hoping to have in terms of coordination?  This is obviously a big change.  My preference would be for us to migrate all test cases from `ocean/coastal` over before doing this change.  @caozd999 has one (#552) and I have a few too that aren't in PR stage yet.

testing_and_setup/compass/ocean/coastal/USDEQU120cr10rr2/build_mesh/config_culled_mesh.xml
```diff
@@ -14,8 +13,7 @@
 			<argument flag="--with_critical_passages"></argument>
 			<argument flag="--preserve_floodplain"></argument>
 		</step>
-		<step executable="python">
-			<argument flag="-m">jigsaw_to_MPAS.inject_bathymetry</argument>
+		<step executable="inject_bathymetry">
```
pwolfram, line 16: Less is more-- nice!

testing_and_setup/compass/ocean/global_ocean/config_files/config_mesh_metrics.xml
```diff
@@ -6,15 +6,14 @@
...
 	<add_link source_path="bathymetry_database" source="SRTM15_plus_earth_relief_15s.nc" dest="earth_relief_15s.nc"/>
-	<add_link source_path="mpas_model" source="testing_and_setup/compass/ocean/jigsaw_to_MPAS/inject_bathymetry.py" dest="inject_bathymetry.py"/>
 
 	<run_script name="run.py">
 		<step executable="MpasMeshConverter.x">
 			<argument flag="mesh.nc"></argument>
 			<argument flag="mesh_with_metrics.nc"></argument>
 		</step>
 
-		<step executable="./inject_bathymetry.py">
```
pwolfram, line 17: Is this correct?  This appears to be a local script and sometimes these were not local scripts.  May need to be careful here.

xylar:
>  My preference would be for us to migrate all test cases from ocean/coastal over before doing this change. 

Yep, that's exactly what I mean.  We should only merge this at a point where `ocean/develop` and `ocean/coastal` are identical and all test cases then get updated with these changes.  I hope that can happen in the not too distant future so there isn't too much that has to be redone, but I realize this might be a bit tricky to coordinate.

pwolfram:
Thanks @xylar, I understand now and agree.

pwolfram:
@mark-petersen-- thoughts on when you want to do the `ocean/coastal` and `ocean/develop` merge?

testing_and_setup/compass/ocean/global_ocean/config_files/config_mesh_metrics.xml
xylar: It is just different syntax for the same call.  In some cases, a local link was being made to `inject_bathymetry.py` itself (as here) and in other cases the link is to the `jigsaw_to_MPAS` module, so you need to call the same script with `python -m jigsaw_to_MPAS.inject_bathymetry` instead of `./inject_bathymetry.py`.  It's always the same script, which will now be an entry point instead.

testing_and_setup/compass/ocean/coastal/USDEQU120cr10rr2/build_mesh/config_culled_mesh.xml
xylar: A happy side-effect rather than something I was explicitly doing to clean things up.

xylar:
> Go ahead and merge when you are ready, and we can make additional changes if needed in another PR.

I don't think that's a safe attitude for this PR.  I think every affected test case needs to be checked to make sure it is bit-for-bit before and after these changes.  Otherwise, we will not have confidence in such an extensive set of changes, and it will likely affect everyone working with non-idealized test cases for awhile going forward.

xylar:
@mark-petersen and @sbrus89, I'm done with my testing and I'm confident that this PR and https://github.com/MPAS-Dev/MPAS-Tools/pull/311 don't change any test cases.  The `hurricane/USDEQU60at15cr5rr100WD/build_mesh` takes *for* *ever* so I haven't tested it completely but will see if it finishes in 12 hours or less.  But I'll take that risk.

Please do at least some testing before approving the PR.

mark-petersen:
@xylar, I'm trying to test this. I need to load compass 0.1.9 on grizzly. The load latest is 0.1.8. I hate to say it, but I don't know how to create compass 0.1.9 locally myself.  Are there instructions? I should really know how to do it.

xylar:
@mark-petersen, please use `test_compass_0.1.9`.

xylar:
> I hate to say it, but I don't know how to create compass 0.1.9 locally myself. Are there instructions? I should really know how to do it.

It's quite complicated.  I have to make test builds of both `mpas_tools` and `compass`.  I then upload them to my own `anaconda` channel and then install them on `grizzly`.  Nothing I want you to worry about for now.

xylar:
Just FYI, I do plan to spend a considerable amount of time in the next 3-4 weeks writing documentation, and this will be part of the process I will be documenting.

mark-petersen:
Thanks @xylar . I set up and ran test cases on grizzly with `test_compass_0.1.9` (I see how to query the versions now). I used the new compass to set up and run the nightly regression suite, which passed a bfb comparison with the previous ocean/develop. I also used the setup command for an EC60to30 mesh init case, which is taking a long time in the mesh culler right now.

xylar:
> which is taking a long time in the mesh culler right now

Like I've said, the mesh conversion tools really need to be replaced with xarray/dask versions that are faster and parallel.  That would be a great project for the right student or postdoc...

xylar:
And thanks very much for testing.  Let me know how it goes with `EC60to30`.

xylar:
@sbrus89, is this something you will have time to test today?

mark-petersen:
> Let me know how it goes with EC60to30.
It ran through init and began spin-up successfully. Also just ran QU240wISC successfully.

sbrus89:
@xylar, I'll try to run some tests this afternoon

xylar:
@sbrus89, wonderful!  That would be much appreciated!

mark-petersen:
This one worked: `USDEQU240at60cr20rr4/build_mesh`. I'm done testing. Will merge after approval from @sbrus89.

xylar:
> Will merge after approval from @sbrus89.

The safer order of operations would be for me to create `compass_0.1.9` officially and then we merge this.  That would be my preference, since folks trying to use test cases after this merge will need to know that they need to load the `test_compass_0.1.9` environment and I don't want people to get in the habit of using the test environments for anything but testing the PRs they were created for.

mark-petersen:
OK. What if you do what's needed tomorrow morning Berlin time for compass 0.1.9, and I'll merge this tomorrow morning NM time?

xylar:
Perfect!  (Assuming @sbrus89's testing goes well.)

xylar:
@mark-petersen, `compass_0.1.9` is now in place on LANL IC, Cori, Compy, Cooley and Anvil.  You may merge this at your convenience.

PR #565, opened 2020-05-19
Fixes B(N) computation bugs in mpas_ocn_vmix.F

caozd999:
This PR fixes a series of bugs in the computation of B(N) in mpas_ocn_vmix.F.
As an example, the original code computes B(N) in the 
**subroutine ocn_vel_vmix_tend_implicit_spatially_variable_mannings**  (as well as other subroutines for vertical mixing) by

>`B(N) = 1.0_RKIND - A(N) + dt*implicitCd &
                * sqrt(kineticEnergyCell(k,cell1) + kineticEnergyCell(k,cell2)) / layerThicknessEdge(k,iEdge)`

which should be
>`B(N) = 1.0_RKIND - A(N) + dt*implicitCd &
               * sqrt(kineticEnergyCell(N,cell1) + kineticEnergyCell(N,cell2)) / layerThicknessEdge(N,iEdge)`

PR #567, opened 2020-05-19
Fix B(N) computation bugs in shared/mpas_ocn_vmix.F

caozd999:
Currently, the implicit vertical mixing routine applies the bottom drag with kinetic energy referenced to `k`, but outside a `k` loop. This is a simple typo and should be `N`. This should be fixed although the `k` out from the loop has the same value as `N`.

Same bug PR has been approved in ocean/coastal.  See https://github.com/MPAS-Dev/MPAS-Model/pull/565

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix.F
@@ -319,7 +319,7 @@ contains
          ! Apply bottom drag boundary condition on the viscous term
          ! second line uses sqrt(2.0*kineticEnergyEdge(k,iEdge))
          B(N) = 1.0_RKIND - A(N) + dt*implicitBottomDragCoef &
-              * sqrt(kineticEnergyCell(k,cell1) + kineticEnergyCell(k,cell2)) / layerThicknessEdge(k,iEdge) &
+              * sqrt(kineticEnergyCell(N,cell1) + kineticEnergyCell(N,cell2)) / layerThicknessEdge(N,iEdge) &
               ! added Rayleigh terms
               + dt*(rayleighBottomDampingCoef + &
                     rayleighDampingCoef /((1.0_RKIND - rayleighDepthVariable) + rayleighDepthVariable*edgeThicknessTotal))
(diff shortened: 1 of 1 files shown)


pwolfram:
@mark-petersen, here is the `ocean/develop` PR as requested in our discussion with @caozd999 .  Thanks!

pwolfram:
@mark-petersen,  is there something here @caozd999 and I need to do to move this forward?  Happy to do so-- thanks!

mark-petersen:
Thanks for the reminder. I agree I need to get this in. I'll make it a higher priority.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -319,7 +319,7 @@ subroutine ocn_vel_vmix_tend_implicit_rayleigh(meshPool, dt, kineticEnergyCell,
          ! Apply bottom drag boundary condition on the viscous term
          ! second line uses sqrt(2.0*kineticEnergyEdge(k,iEdge))
          B(N) = 1.0_RKIND - A(N) + dt*implicitBottomDragCoef &
-              * sqrt(kineticEnergyCell(k,cell1) + kineticEnergyCell(k,cell2)) / layerThicknessEdge(k,iEdge) &
+              * sqrt(kineticEnergyCell(N,cell1) + kineticEnergyCell(N,cell2)) / layerThicknessEdge(N,iEdge) &
```
mark-petersen, line 322: @vanroekel and @jonbob, this is an important correction, where the index for kinetic energy in bottom drag was effectively using `N-1` rather than the bottom layer `N`.  It will be a non-bfb change in E3SM. It should be a small change, but will need to be tested.

vanroekel: @mark-petersen and @jonbob once anvil is back today I can easily test this in a G-case

vanroekel:
@mark-petersen I implemented this fix into e3sm and my testing shows this is BFB with previous.  To test why I see this, I made a super simple code something like

```
program loopTest

integer :: k

do k=1,10
enddo

print *, k

end program loopTest
```

if I run this code I get ```11``` in the print statement, suggesting the 'k' values in the code snippets changed is indeed equal to N and not N-1, I still think this should be fixed, but appears BFB to me.  Could you share your testing that shows non-BFB, maybe I'm trying the wrong compiler combination.  I tried intel and gfortran, but not debug mode.  

Also pinging @caozd999 for any info on his testing.

caozd999:
@vanroekel You are right. Actually `k `after the loop equals to `N`, I've checked that just now: I added `if (k==N) call mpas_log_write('k equals to N after the do loop')` after the `Do k` loop, and the output proved that. Although this won't influence the model results, I still think this should be fixed. BTW, I have modified the original comment accordingly.

vanroekel:
@caozd999 definitely agree we should merge this, the code as is it stands is very confusing.   I just wanted to make sure I wasn't testing this wrong since I was seeing BFB.

mark-petersen:
@vanroekel thanks for the help. I said it was non-bfb due to my read of the code, not due to testing. That is good to know. I'll merge it when the E3SM repo is open.

mark-petersen:
Tested locally. It is indeed bit-for-bit with the previous nightly regression on `ocean/develop`. This is still a good cleanup.

PR #568, opened 2020-05-21
Adds Mixed layer heat and salinity budgets

vanroekel:
This commit adds an AM to compute mixed layer integrated heat and salinity budgets, it also adds a budget term to capture Redi and horizontal mixing on temperature and salinity.

### DIFF
--- a/src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
+++ b/src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
@@ -101,5 +101,7 @@
 			<var_array name="activeTracerVertAdvectionMLTend"/>
 			<var_array name="activeTracerHorAdvectionMLTend"/>
 			<var_array name="activeTracerVertMixMLTend"/>
+			<var_array name="activeTracersML"/>
+			<var_array name="bruntVaisalaFreqML"/>
 		</stream>
 	</streams>
--- a/src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
+++ b/src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
@@ -76,11 +76,11 @@
 			<var name="temperatureML" array_group="activeTracerFTendGroup" units="degrees Celsius"
 				description="ML average temperature"
 			/>
-			<var name="salinityML" array_group="activeTracerFTendGroup" units="PSU per second"
+			<var name="salinityML" array_group="activeTracerFTendGroup" units="PSU"
 				description="ML average salinity"
 			/>
 		</var_array>
-		<var name="bruntVaisalaFreqML" type="real" dimensions="nCells Time"
+    <var name="bruntVaisalaFreqML" type="real" dimensions="nCells Time" units="s^{-2}"
 			description="ML average BVF"
 		/>
 	</var_struct>
--- a/src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
+++ b/src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
@@ -24,63 +24,63 @@
 		<package name="mixedLayerHeatBudgetAMPKG" description="This package includes variables required for the mixedLayerHeatBudget analysis member."/>
 	</packages>
 	<var_struct name="mixedLayerHeatBudgetAM" time_levs="1" packages="mixedLayerHeatBudgetAMPKG">
-		<var_array name="activeTracerHorAdvectionMLTend" type="real" dimensions="nCells Time" packages="mixedLayerHeatBudgetAMPKG">
-			<var name="temperatureHorAdvectionMLTend" array_group="activeTracerHAdvTendGroup" units="degrees Celsius per second" name_in_code="temperatureHorAdvectionMLTend"
+		<var_array name="activeTracerHorAdvectionMLTend" type="real" dimensions="nCells Time">
+			<var name="temperatureHorAdvectionMLTend" array_group="activeTracerHAdvTendGroup" units="degrees Celsius per second"
 				description="ML average horizontal advection tendency for temperature"
 			/>
-			<var name="salinityHorAdvectionMLTend" array_group="activeTracerHAdvTendGroup" units="PSU per second" name_in_code="salinityHorAdvectionMLTend"
+			<var name="salinityHorAdvectionMLTend" array_group="activeTracerHAdvTendGroup" units="PSU per second"
 				description="ML average horizontal advection tendency for salinity"
 			/>
 		</var_array>
-		<var_array name="activeTracerVertAdvectionMLTend" type="real" dimensions="nCells Time" packages="mixedLayerHeatBudgetAMPKG">
-			<var name="temperatureVertAdvectionMLTend" array_group="activeTracerVAdvTendGroup" units="degrees Celsius per second" name_in_code="temperatureVertAdvectionMLTend"
+		<var_array name="activeTracerVertAdvectionMLTend" type="real" dimensions="nCells Time">
+			<var name="temperatureVertAdvectionMLTend" array_group="activeTracerVAdvTendGroup" units="degrees Celsius per second"
 				description="ML average vertical advection tendency for temperature"
 			/>
-			<var name="salinityVertAdvectionMLTend" array_group="activeTracerVAdvTendGroup" units="PSU per second" name_in_code="salinityVertAdvectionMLTend"
+			<var name="salinityVertAdvectionMLTend" array_group="activeTracerVAdvTendGroup" units="PSU per second"
 				description="ML average vertical advection tendency for salinity"
 			/>
 		</var_array>
-		<var_array name="activeTracerVertMixMLTend" type="real" dimensions="nCells Time" packages="mixedLayerHeatBudgetAMPKG">
-			<var name="temperatureVertMixMLTend" array_group="activeTracerVMixTendGroup" units="degrees Celsius per second" name_in_code="temperatureVertAdvectionMLTend"
+		<var_array name="activeTracerVertMixMLTend" type="real" dimensions="nCells Time">
+			<var name="temperatureVertMixMLTend" array_group="activeTracerVMixTendGroup" units="degrees Celsius per second"
 				description="ML average vertical mixing tendency for temperature"
 			/>
-			<var name="salinityVertMixMLTend" array_group="activeTracerVMixTendGroup" units="PSU per second" name_in_code="salinityVertMixMLTend"
+			<var name="salinityVertMixMLTend" array_group="activeTracerVMixTendGroup" units="PSU per second"
 				description="ML average vertical mxiing tendency for salinity"
 			/>
 		</var_array>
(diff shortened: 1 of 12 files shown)


vanroekel:
@maltrud and @milenaveneziani I'd really appreciate your comments on this PR.  I've tested this in QU240 and it looks okay.  I'm going to test in E3SM next.

vanroekel:
This has now been tested in a number of E3SM simulations and seems to be working.  I think this is ready for review.

mark-petersen:
@milenaveneziani and @maltrud, I'm trying to work through these analysis PRs this week. Do you have any comments on this one?

mark-petersen:
Rebased, added to COMPASS. @vanroekel Here are a few notes for your future reference, but I already took care of them in this PR.
1. When adding a new file, add it to `src/core_ocean/ocean.cmake`. We have both a Makefile system and cmake (for E3SM) right now.
2. I try not to use `name_in_code` in Registry. It leads to typos and bugs that are hard to find. In this case your `name_in_code` was the same as `var name`, so I just removed them.
3. Individual `var` and `var_array` do not need to be labelled with packages if they are within a `var_struct` that has that package. For analysis members, the whole `var_struct` is in the package for that AM.
4. The xml files are formatted with tab indents, not spaces. See if you can copy the lines with your editor to keep the tab format.  You might be highlighting with the mouse and pasting with command-v, which will paste spaces. In vim the visual mode copy and paste preserves the tab format.

mark-petersen:
@vanroekel if you are happy with this PR, it can be merged, probably today, and we will make an E3SM PR. I'm guessing the output should be similar to the ocean heat content for E3SM? Please list which variables you want added to the monthly time average output from this AM, and if it should be default on in E3SM for all cases (I think so).

src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
```diff
@@ -0,0 +1,107 @@
...
+			/>
+			<var name="salinityForcingMLTend" array_group="activeTracerFTendGroup" units="PSU per second"
+				description="ML average forcing tendency (non thickness flux changes) for salinity"
+			/>
+		</var_array>
+		<var_array name="activeTracersML" type="real" dimensions="nCells Time">
+			<var name="temperatureML" array_group="activeTracerFTendGroup" units="degrees Celsius"
+				description="ML average temperature"
+			/>
+			<var name="salinityML" array_group="activeTracerFTendGroup" units="PSU per second"
```
milenaveneziani, line 79: units above should be "PSU"

src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
```diff
@@ -0,0 +1,107 @@
...
+				description="ML average salinity"
+			/>
+		</var_array>
+		<var name="bruntVaisalaFreqML" type="real" dimensions="nCells Time"
```
milenaveneziani, line 83: we could add units in the var above (I assume "seconds^-2"?)

src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
vanroekel: thanks @milenaveneziani copy paste error here, I just fixed it

src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
vanroekel: fixed

vanroekel:
@mark-petersen thanks for the list of pointers, good to know.  The PR looks good.  given that all of these are 2D fields, I would like all added to the monthly average AM AND the monthly min / max AMs as well.  Here is the list
```
<var_array name="activeTracerForcingMLTend"/>
 <var_array name="activeTracerHorMixMLTend"/>
 <var_array name="activeTracerNonLocalMLTend"/>
 <var_array name="activeTracerVertAdvectionMLTend"/>
 <var_array name="activeTracerHorAdvectionMLTend"/>
 <var_array name="activeTracerVertMixMLTend"/>
  <var_array name="activeTracersML"/>
 <var_array name="bruntVaisalaFreqML"/>
```

milenaveneziani:
@vanroekel: I left a similar comment (without reading yours here :) in  the PR about monthly output.

I haven't gone through the  code that actually computes the budget terms, at least not very carefully: would you like me to do that, or just keep it a visual inspection at this time?

vanroekel:
@milenaveneziani if you don't mind taking a close look at the computation I would appreciate it, but if you don't have time, I'm fairly confident in the calculations, but would appreciate the second look.

src/core_ocean/analysis_members/Registry_mixed_layer_heat_budget.xml
```diff
@@ -0,0 +1,107 @@
...
+		/>
+		<nml_option name="config_AM_mixedLayerHeatBudget_write_on_startup" type="logical" default_value=".true." units="unitless"
+			description="Logical flag determining if an analysis member write occurs on start-up."
+			possible_values=".true. or .false."
+		/>
+	</nml_record>
+	<packages>
+		<package name="mixedLayerHeatBudgetAMPKG" description="This package includes variables required for the mixedLayerHeatBudget analysis member."/>
+	</packages>
+	<var_struct name="mixedLayerHeatBudgetAM" time_levs="1" packages="mixedLayerHeatBudgetAMPKG">
```
milenaveneziani, line 26: The temporal tendency is not included in  the output stream? Is there something  that we want, for closure?

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_heat_budget.F
```diff
@@ -0,0 +1,417 @@
...
+            k = 1
+            depth = layerThickness(k, iCell)
+            do while (depth <= dThreshMLD(iCell) .and. k < maxLevelCell(iCell))
+               do iTracer = 1, num_tracers
+                  activeTracerHorAdvectionMLTend(iTracer, iCell) = activeTracerHorAdvectionMLTend(iTracer, iCell) + &
+                                                 activeTracerHorizontalAdvectionTendency(iTracer, k, iCell)*layerThickness(k, iCell)
+                  activeTracerHorMixMLTend(iTracer, iCell) = activeTracerHorMixMLTend(iTracer, iCell) + &
+                                                             activeTracerHorMixTendency(iTracer, k, iCell)*layerThickness(k, iCell)
+                  activeTracerForcingMLTend(iTracer, iCell) = activeTracerForcingMLTend(iTracer, iCell) + &
+                                                         activeTracerSurfaceFluxTendency(iTracer, k, iCell)*layerThickness(k, iCell)
```
milenaveneziani, line 263: just a  style comment, feel free to ignore: I'd put the forcing terms following each other, like you did for HorAdv/HorMix, VertAdv/VertMix.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_heat_budget.F
```diff
@@ -0,0 +1,417 @@
...
+            activeTracerHorAdvectionMLTend(:, iCell) = 0.0_RKIND
+            activeTracerNonLocalMLTend(:, iCell) = 0.0_RKIND
+            activeTracerHorMixMLTend(:, iCell) = 0.0_RKIND
+            activeTracerVertMixMLTend(:, iCell) = 0.0_RKIND
+            activeTracerVertAdvectionMLTend(:, iCell) = 0.0_RKIND
+            activeTracerForcingMLTend(:, iCell) = 0.0_RKIND
+            activeTracersML(:, iCell) = 0.0_RKIND
+            k = 1
+            depth = layerThickness(k, iCell)
+            do while (depth <= dThreshMLD(iCell) .and. k < maxLevelCell(iCell))
```
milenaveneziani, line 256: quick question: can MLD ever be smaller than layerThickness(1)? or is it set to 0 in that case?

milenaveneziani:
@vanroekel: I went through the code more carefully. I don't have any more questions at the moment.

src/core_ocean/analysis_members/mpas_ocn_analysis_driver.F
```diff
@@ -178,6 +179,7 @@ subroutine ocn_analysis_setup_packages(configPool, packagePool, iocontext, err)!
       call mpas_pool_add_config(analysisMemberList, 'lagrPartTrack', 1)
       call mpas_pool_add_config(analysisMemberList, 'eliassenPalm', 1)
       call mpas_pool_add_config(analysisMemberList, 'mixedLayerDepths', 1)
+      call mpas_pool_add_config(analysisMemberList, 'mixesLayerHeatBudget', 1)
```
milenaveneziani, line 182: typo above: 'mixesLayer'

vanroekel:
@milenaveneziani thanks for the catches!  I'll work on these.  @mark-petersen please hold off on merging until I can address these issues.

mark-petersen:
No rush. I have others in line, so this can proceed when you are ready.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_heat_budget.F
vanroekel: the mixed layer will always be above 10m + the next layer thickness so this shouldn't be a problem, but is probably wise to fix just in case.

vanroekel: actually looking at the code again, I think it is okay.  If layer dThreshMLD < layerTHickness(k,ICell) we will just pick up the partial integral after this do loop

vanroekel:
@mark-petersen since this runs at the end of the timestep, should I pull in tracers at time level 1 or 2?  I can't figure which is the most up to date timestamp.  I think it is two.

src/core_ocean/analysis_members/mpas_ocn_mixed_layer_heat_budget.F
milenaveneziani: good point.

mark-petersen:
All the AMs have the time level passed in. For example:
```
mpas_ocn_meridional_heat_transport.F:294:   subroutine ocn_compute_meridional_heat_transport(domain, timeLevel, err)!{{{
mpas_ocn_meridional_heat_transport.F:302:      integer, intent(in) :: timeLevel
mpas_ocn_meridional_heat_transport.F:466:         call mpas_pool_get_array(tracersPool, 'activeTracers', activeTracers, timeLevel)
```
so generally that is the best method. If you are something beyond that, it would be time level 2, since it is at the end of the time step.

vanroekel:
it looks like the analysis driver is sending in time level 1, I'll be consistent with that for now, but wonder if we should change that to timeLevel=2

https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/analysis_members/mpas_ocn_analysis_driver.F#L692

vanroekel:
I think this is ready to go now, I just added a ML average tendency term, so when this goes to E3SM we also need to include 

```
<var_array name="activeTracersTendML"/>
```

vanroekel:
This would be for the timeMontlyavg, max, min streams.

PR #569, opened 2020-05-22
Ocean/remove unused modules

### LINKED ISSUE #557: Remove unused vertical mixing schemes
Since we now only use CVMix, the `rich` vertical mixing scheme and perhaps others should be removed so we don't have to maintain code we don't use.

mark-petersen:
Removes these modules:
```
vmix
vmix_const
vmix_rich
vmix_tanh
hmix_del2_tensor
hmix_del4_tensor
```
The vmix modules are replaced by cvmix. Redundant code is poor practice and leads to errors. The hmix tensor routines were experimental. They caused noisy fields and were never used in the end.

All the COMPASS tests that used `vmix_const` were changed over to cvmix with constant background viscosity. The others were not used by COMPASS tests.

Fixes #557


mark-petersen:
passes nightly regression suite.

mark-petersen:
@vanroekel @pwolfram or @xylar 
I might have a window to merge this in, if any of you could review soon. It is very simple, just removes the unused modules we had discussed. I ran the nightly regression suite, so I think visual inspection is sufficient.

src/core_ocean/mode_init/Registry.xml
```diff
@@ -22,6 +22,20 @@
 	<!--**********************************-->
 	<!-- Namelists for init run mode only -->
 	<!--**********************************-->
+	<nml_record name="partial_bottom_cells" mode="init">
+		<nml_option name="config_alter_ICs_for_pbcs" type="logical" default_value=".false." units="unitless"
+					description="If true, initial conditions are altered according to the config_pbc_alteration_type flag."
+					possible_values=".true. or .false."
+		/>
+		<nml_option name="config_pbc_alteration_type" type="character" default_value="full_cell" units="unitless"
```
pwolfram, line 30: I'm confused as to why these additions for removal of code... can you please clarify?

src/core_ocean/Registry.xml
```diff
@@ -190,80 +190,6 @@
 					possible_values="'split_explicit', 'RK4', 'unsplit_explicit'"
 		/>
 	</nml_record>
-	<nml_record name="ALE_vertical_grid" mode="forward">
```
pwolfram, line 193: Why is this option disappearing?  We need this for a variety of coastal applications.

src/core_ocean/Registry.xml
```diff
@@ -190,80 +190,6 @@
...
-		<nml_option name="config_use_highFreqThick_del2" type="logical" default_value=".false." units="unitless"
-					description="If true, high frequency thickness prognostic equation ($h^{hf}$) includes term 3 on the RHS, the diffusion term."
-					possible_values=".true. or .false."
-		/>
-		<nml_option name="config_highFreqThick_del2" type="real" default_value="100.0" units="m^2 s^{-1}"
-					description="Horizonal high frequency thickness diffusion, $\kappa_{hhf}$."
-					possible_values="any positive real"
-		/>
-	</nml_record>
-	<nml_record name="partial_bottom_cells" mode="init">
```
pwolfram, line 253: I see now why the addition elsewhere...

src/core_ocean/Registry.xml
```diff
@@ -1174,18 +996,64 @@
...
+		<nml_option name="config_vert_taper_weight_depth_1" type="real" default_value="250.0" units="m"
+					description="Vertical coordinate taper weight is one above this depth, linearly decreases to zero below."
+					possible_values="any positive real value, but typically 100 to 1000 m."
+		/>
+		<nml_option name="config_vert_taper_weight_depth_2" type="real" default_value="500.0" units="m"
+					description="Vertical coordinate taper weight is zero below this depth, linearly increases to one above."
+					possible_values="any positive real value, but typically 100 to 1000 m and greater than config_vert_taper_weight_depth_1."
+		/>
+		<nml_option name="config_use_min_max_thickness" type="logical" default_value=".false." units="unitless"
+					description="If true, a minimum and maximum thicknesses are enforced to prevent massless and very thick layers."
```
pwolfram, line 1017: I'm assuming these just were a direct copy/paste without any loss of information?

src/core_ocean/mode_init/Registry.xml
pwolfram: I see where they were removed above-- thanks.

testing_and_setup/compass/ocean/baroclinic_channel/template_forward.xml
```diff
@@ -1,21 +1,12 @@
...
 	<namelist>
-		<option name="config_ocean_run_mode">'forward'</option>
-		<option name="config_dt">'00:00:30'</option>
-		<option name="config_btr_dt">'00:00:02'</option>
-		<option name="config_run_duration">'0000_00:15:00'</option>
-		<option name="config_block_decomp_file_prefix">'graph.info.part.'</option>
-		<option name="config_pio_num_iotasks">1</option>
-		<option name="config_pio_stride">4</option>
 		<option name="config_write_output_on_startup">.false.</option>
-		<option name="config_time_integrator">'split_explicit'</option>
```
pwolfram, line 11: Why remove?  This helps make things clearer.

testing_and_setup/compass/ocean/hurricane/forward_template.xml
```diff
@@ -15,8 +15,8 @@
 		<option name="config_ALE_thickness_proportionality">'weights_only'</option>
 		<option name="config_vert_coord_movement">'uniform_stretching'</option>
 
-                <option name="config_use_const_visc">.true.</option>
-                <option name="config_vert_visc">1.0e4</option>
```
pwolfram, line 19: Thanks

testing_and_setup/compass/ocean/overflow/1km/rpe_test/config_rpe_test_6_nu_1000.xml
```diff
@@ -9,7 +9,8 @@
 		<option name="config_dt">'0000_00:03:00'</option>
 		<option name="config_btr_dt">'0000_00:00:09'</option>
 		<option name="config_run_duration">'0000_40:00:00'</option>
-		<option name="config_use_const_visc">.true.</option>
```
pwolfram, line 12: Too bad this all can't be lumped together in a template file-- don't think it is worth effort here for now.

testing_and_setup/compass/ocean/soma/soma_run.xml
```diff
@@ -4,10 +4,6 @@
 		<option name="config_ocean_run_mode">'forward'</option>
 		<option name="config_use_mom_del4">.true.</option>
 		<option name="config_mom_del4_div_factor">10.0</option>
-		<option name="config_convective_visc">1.0</option>
-		<option name="config_convective_diff">1.0</option>
-		<option name="config_use_rich_visc">.true.</option>
-		<option name="config_use_rich_diff">.true.</option>
```
pwolfram, line 10: No addition to compensate for SOMA?

testing_and_setup/compass/ocean/surface_waves/thickness_source/1km/config_forward.xml
```diff
@@ -18,7 +18,7 @@
 		<option name="config_dt">'0000_00:01:00'</option>
 		<option name="config_time_integrator">'RK4'</option>
 		<option name="config_run_duration">'0001_00:00:01'</option>
-		<option name="config_use_const_visc">.false.</option>
+		<option name="config_use_cvmix_background">.false.</option>
```
pwolfram, line 21: Thanks

testing_and_setup/compass/ocean/ziso/ziso_run.xml
```diff
@@ -4,8 +4,6 @@
 		<option name="config_ocean_run_mode">'forward'</option>
 		<option name="config_use_mom_del4">.true.</option>
 		<option name="config_mom_del4_div_factor">10.0</option>
-		<option name="config_convective_visc">1.0</option>
-		<option name="config_convective_diff">1.0</option>
```
pwolfram, line 8: Same thing-- no addition here?  I'm a little confused and am missing something.

pwolfram:
@mark-petersen, did a quick visual inspection review with comments / questions above.

xylar:
> why removal of the tensor operations?
This was discussed in these comments:
https://github.com/MPAS-Dev/MPAS-Model/issues/557#issuecomment-630278347
https://github.com/MPAS-Dev/MPAS-Model/issues/557#issuecomment-630281274

testing_and_setup/compass/ocean/baroclinic_channel/template_forward.xml
xylar: Why does this help to make things clearer in this one test case, when we just assume it's the default in nearly all other test cases?

testing_and_setup/compass/ocean/soma/soma_run.xml
xylar: This looks like it was a really odd configuration with CVMix running with KPP and yes also with Richardson vertical mixing also turned on.  That makes no sense and isn't something we can duplicate with CVMix on its own.

testing_and_setup/compass/ocean/ziso/ziso_run.xml
xylar: These are already the default values (both with and without CVMix):
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/Registry.xml#L471-L478
They could be added explicitly, though, if we think it's meaningful.  I personally don't.

testing_and_setup/compass/ocean/baroclinic_channel/template_forward.xml
pwolfram: Because we will have multiple time stepping options in the future and this is more explicit.  We are already running with RK4 for the hurricane cases, for instance, and the new implicit barotropic solve is in progress, not to mention the spatially variable time stepping work starting in the next week or so.

testing_and_setup/compass/ocean/soma/soma_run.xml
pwolfram: CVmix should have been turned off, which is a bug then.  From the SOMA paper (Wolfram et al 2015) the specification is (noting the committed code was done post-facto because SOMA predates COMPASS):

We use the Richardson number–based parameterization of Pacanowski and Philander (1981) as our ver- tical mixing model. Background values of nu_v = 106-4 m^2/s and kappa_v 10^-5 m^2/s are used for viscosity and diffusivity away from the mixed layer and convective mixing. The coefficient for Richardson number vertical mixing function is that suggested by Pacanowski and Philander (1981), namely, C_Ri = 0.005 m^2/s.  Vertical mixing values of nu_v=kappa_v=1 m^2/s are used to remove gravitationally unstable profiles.

testing_and_setup/compass/ocean/ziso/ziso_run.xml
pwolfram: Agreed

testing_and_setup/compass/ocean/baroclinic_channel/template_forward.xml
xylar: Hmm, okay, I get your point.  I still think this is of negligible importance.  Adding these tags will be quite a chore and one test case fewer likely won't make a difference.  But also won't do any harm.

src/core_ocean/Registry.xml
pwolfram: (I did not verify this myself)

testing_and_setup/compass/ocean/soma/soma_run.xml
xylar: Good to know.  Could you make suggested edits to the configuration, then?

pwolfram:
@mark-petersen, cleaned up comments above-- only real issue left is potential concern over mixing used in SOMA, especially since @lizcarlson needs it for her work.

testing_and_setup/compass/ocean/soma/soma_run.xml
pwolfram: The case is in production by @lizcarlson -- is this useful @lizcarlson?  Don't want to change anything out from underneath your work, but also don't want you to use a case that isn't as promised in the paper (because of the basins to COMPASS transition).

testing_and_setup/compass/ocean/baroclinic_channel/template_forward.xml
pwolfram: Maybe we just leave them out but make sure as they are changed to something else they are specified.  It is hard for me to foresee what will be clearer.

testing_and_setup/compass/ocean/soma/soma_run.xml
mark-petersen: That file had both cvmix and the old rich mixing on. I removed the old rich flags, but these remain:
```
  7         <option name="config_use_cvmix">.true.</option>
  8         <option name="config_use_cvmix_background">.true.</option>
  9         <option name="config_use_cvmix_convection">.true.</option>
 10         <option name="config_cvmix_background_diffusion">5.0e-6</option>
 11         <option name="config_cvmix_convective_basedOnBVF">.true.</option>
 12         <option name="config_use_cvmix_shear">.true.</option>
 13         <option name="config_cvmix_shear_mixing_scheme">'KPP'</option>
 14         <option name="config_use_cvmix_kpp">.true.</option>
```
They were added in 2d229ed6db (by Wolfram 2016-12-07)

Based on your paragraph above, I think you want:
```
config_use_cvmix = .true.     (this is now default true)
config_use_cvmix_background = .true.
config_cvmix_background_diffusion = 1.0e-5
config_cvmix_background_viscosity = 1.0e-4
config_use_cvmix_convection = .true.

config_use_cvmix_shear = .true.
config_cvmix_num_ri_smooth_loops = 2
config_cvmix_use_BLD_smoothing = .true.
config_cvmix_shear_mixing_scheme = 'PP'
config_cvmix_shear_PP_nu_zero = 0.005
```
If you agree, I can update the soma file. The main change is from KPP to PP. That looks like a mistake in the original.

testing_and_setup/compass/ocean/baroclinic_channel/template_forward.xml
mark-petersen: My philosophy here is to have template and config file contain only the flags that change at that level. So the 1km/5km/10km forward templates contains the time step and viscosity change only. 

In addition, even at the configuration level (baroclinic channel), I would prefer not to specify namelist options that are both the default setting and uniform for all cases within that configuration. For example, all baroclinic channel cases use `split_explicit`, the default, so it's not needed in the template. I agree that if you have a set of cases like your hurricanes, some with the default and some with non-default settings, it makes sense to explicitly state the default setting.

testing_and_setup/compass/ocean/drying_slope/forward_template.xml
```diff
@@ -21,8 +21,8 @@
 		<option name="config_time_integrator">'RK4'</option>
 		<option name="config_run_duration">'0000_12:00:01'</option>
 		<option name="config_use_mom_del2">.false.</option>
-		<option name="config_use_const_visc">.true.</option>
-		<option name="config_vert_visc">1e4</option>
+		<option name="config_use_cvmix_background">.true.</option>
+		<option name="config_cvmix_background_viscosity">1e4</option>
```
vanroekel, line 25: Is this value correct?  It is massive for a vertical viscosity.

testing_and_setup/compass/ocean/hurricane/forward_template.xml
```diff
@@ -15,8 +15,8 @@
...
+                <option name="config_use_cvmix_background">.true.</option>
+                <option name="config_cvmix_background_viscosity">1.0e4</option>
```
vanroekel, line 19: same question here, seems like a really big number

testing_and_setup/compass/ocean/internal_waves/5km/internal_waves_5km_template.xml
```diff
@@ -11,11 +11,10 @@
...
 		<option name="config_use_mom_del2">.true.</option>
 		<option name="config_mom_del2">10.0</option>
-		<option name="config_convective_visc">1.0</option>
-		<option name="config_convective_diff">1.0</option>
 		<option name="config_implicit_bottom_drag_coeff">1.0e-2</option>
-		<option name="config_use_const_visc">.true.</option>
-		<option name="config_vert_visc">1.0e-4</option>
+		<option name="config_use_cvmix_background">.true.</option>
+		<option name="config_use_cvmix_convection">.true.</option>
+		<option name="config_cvmix_background_viscosity">1.0e-4</option>
```
vanroekel, line 17: this is the default, does it need to be explicitly specified?

testing_and_setup/compass/ocean/hurricane/forward_template.xml
vanroekel: perhaps these two are supposed to be 1e-4?

testing_and_setup/compass/ocean/soma/soma_run.xml
vanroekel: @mark-petersen the specification you say here is consistent with what @pwolfram writes above.  It turns on the default PP behavior.  This shouldn't impact @lizcarlson whose simulations are complete.  I've verified with her mixing was configured appropriately

src/core_ocean/Registry.xml
```diff
@@ -444,7 +350,7 @@
 		/>
 	</nml_record>
 	<nml_record name="cvmix" mode="forward">
-		<nml_option name="config_use_cvmix" type="logical" default_value=".false." units="NA"
+		<nml_option name="config_use_cvmix" type="logical" default_value=".true." units="NA"
```
vanroekel, line 353: @mark-petersen note that turning this on will not actually enable any mixing schemes.  I don't know if this is the intention for test cases.  To get PP by default you also need use ```config_use_cvmix_shear = .true.```

vanroekel: I actually think the test cases are consistent with this setting.

lizcarlson:
@vanroekel is correct, my main simulations are complete and I run them from my own branch anyway, so any changes made here will not affect the results.  Thanks for checking @pwolfram!

testing_and_setup/compass/ocean/soma/soma_run.xml
pwolfram: Awesome, so very glad to hear!!!  Thanks @vanroekel

testing_and_setup/compass/ocean/drying_slope/forward_template.xml
pwolfram: We can't run single layer so this is a "hack" to get the behavior to match theoretical results in our V&V.

testing_and_setup/compass/ocean/hurricane/forward_template.xml
pwolfram: Pseudo-2D approach...

pwolfram:
@lizcarlson, so very happy too hear-- thanks for confirming so we can move forward!

pwolfram:
@mark-petersen, as far as I can tell @xylar, @vanroekel, @lizcarlson and myself worked through all issues and you can merge now..

pwolfram:
PS-- congrats on a -1700 or so line PR, code will be better for it!

PR #572, opened 2020-05-26
GPU implementation of ocean baroclinic velocity tendencies

philipwjones:
This PR is a substantial modification to all the ocean baroclinic velocity tendencies and includes:
- a complete GPU implementation in which all tendencies - except tidal for now - are computed on the accelerator (using OpenACC) and all data is transfered at the top level driver (ocn_tend_vel)
-   a number of CPU optimizations performed along the way
-   elimination of meshPool, configPool
-   misc cleanup of comments and stuff

This PR contains similar changes that are in #513 #536 #569 at least so will need to modify/rebase once those are merged.

Performance speedup for this part of the code was 2.8x using 2 GPUs on Summit compared with an 8-rank MPI only case. Details of performance will depend on configuration. CPU performance improved by ~20% in the same 8-rank QU240 test. More speedup is expected as we migrate more data to the device elsewhere. The computational part excluding data transfer showed a 10x speedup.

This is not quite b4b due to changes in order of operations in a couple of routines and not b4b using the accelerator (different chip architecture). But in both cases, the differences are at roundoff level. Tested most of the options (eg for hmix, pgrad), though since it was tested in standalone QU240, the forcing routines really didn't get much of a workout.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vel_forcing_surface_stress.F
+++ b/src/core_ocean/shared/mpas_ocn_vel_forcing_surface_stress.F
@@ -5,223 +5,261 @@
 ! Additional copyright and license information can be found in the LICENSE file
 ! distributed with this code, or at http://mpas-dev.github.com/license.html
 !
-!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
 !
 !  ocn_vel_forcing_surface_stress
 !
 !> \brief MPAS ocean surface stress
-!> \author Doug Jacobsen, Mark Petersen, Todd Ringler
-!> \date   September 2011
+!> \author Doug Jacobsen, Mark Petersen, Todd Ringler, Phil Jones
+!> \date   September 2011, updated May 2020
 !> \details
 !>  This module contains the routine for computing
 !>  tendencies from surface stress.
 !
-!-----------------------------------------------------------------------
+!-------------------------------------------------------------------------------
 
 module ocn_vel_forcing_surface_stress
 
-   use mpas_derived_types
-   use mpas_pool_routines
    use mpas_timer
 
    use ocn_constants
-   use ocn_forcing
+   use ocn_config
+   use ocn_mesh
 
    implicit none
    private
    save
 
-   !--------------------------------------------------------------------
-   !
+   !----------------------------------------------------------------------------
    ! Public parameters
-   !
-   !--------------------------------------------------------------------
+   !----------------------------------------------------------------------------
 
-   !--------------------------------------------------------------------
-   !
+   !----------------------------------------------------------------------------
    ! Public member functions
-   !
-   !--------------------------------------------------------------------
+   !----------------------------------------------------------------------------
 
    public :: ocn_vel_forcing_surface_stress_tend, &
              ocn_vel_forcing_surface_stress_init
 
-   !--------------------------------------------------------------------
-   !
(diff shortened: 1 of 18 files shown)


pwolfram:
Quick questions @philipwjones:

1. This still works with tidal tendencies, correct?
2. Is there a plan to move over the tidal tendencies?

pwolfram:
cc @sbrus89 and Nairita

mattdturner:
> This still works with tidal tendencies, correct?

Unless I'm mistaken, if running with GPUs a info message will be printed and then the tidal tendency code will be calculated in the CPU instead of the GPU.  Notice how the only code in `#ifdef MPAS_OPENACC` is the call to `mpas_log_write`, and the rest of the code is run no matter what pre-compiler flags are used.

```
#ifdef MPAS_OPENACC
         call mpas_log_write( &
            'vel tendency: tidal forcing not supported for GPU', &
            MPAS_LOG_CRIT)
#endif
```

I'll let @philipwjones answer your 2nd question.

src/core_ocean/shared/mpas_ocn_vel_forcing_surface_stress.F
```diff
@@ -5,223 +5,261 @@
...
-        enddo
-
-        if ( maxLevelEdgeTop(iEdge) > 0 .and. remainingStress > 0.0_RKIND) then
-           tend(maxLevelEdgeTop(iEdge), iEdge) = tend(maxLevelEdgeTop(iEdge), iEdge) &
-                         + edgeMask(maxLevelEdgeTop(iEdge), iEdge) * surfaceStress(iEdge) * remainingStress &
-                         / rho_sw / layerThicknessEdge(maxLevelEdgeTop(iEdge), iEdge)
-        end if
+      ! allocate space and transfer data
+      allocate(absorbFrac (nVertLevels,nEdgesOwned))
+      !$acc enter data create(absorbFrac)
```
mattdturner, line 128: Should this `!$acc` be wrapped in an `#ifdef MPAS_OPENACC` as well?

philipwjones:
@pwolfram  If you are running on GPUs (OPENACC enabled) and with tidal forcing, this will exit with an error. It still works for CPU only runs. This is temporary - trying to get a lot of GPU code integrated before the end of the month and an ECP deliverable. But the way the tidal forcing modifies zMid and moves pointers back and forth interferes with the copies on the GPU and it was going to take some thought on how to manage that in an efficient way. Sorry - will get back to it once I finish integrating other stuff

pwolfram:
@philipwjones, we should have removed changes to `zMid` in the new version but it may not yet be on `ocean/develop` (cc @mark-petersen because we are in discussions about this merge).  Can you please make an issue of potential concerns so we can avoid them or refractor here?  Don't want this to not be included in the GPU-enabled version because of some programming issue on our side.  Also know we all want to write better code moving forward too-- thanks!

philipwjones:
Tried to rebase this yesterday, but there have been some reorganization of code since this was first submitted and the rebase got a little ugly and I don't have high confidence. So...will probably do a fresh checkout of a more recent version and re-implement the GPU mods, maybe incorporating some new ideas from Az first.  Will update the branch when this is done.

mattdturner:
> So...will probably do a fresh checkout of a more recent version and re-implement the GPU mods, maybe incorporating some new ideas from Az first

That's probably a good idea.  We just had an issue where the merge conflict resolution in one of my PRs reverted prior changes (bugfix in https://github.com/MPAS-Dev/MPAS-Model/pull/672), so starting over with a fresh checkout in a few subroutines might be better.

philipwjones:
Replaced by #772 and another future PR.

PR #573, opened 2020-05-26
Make timekeeping module private by default

akturner:
Made timekeeping module private with explicit public statements for public members
Fixed bug in mpas_forcing which used mpas_timekeeping rather than mpas_derived_types in interfaces

### DIFF
--- a/src/framework/mpas_forcing.F
+++ b/src/framework/mpas_forcing.F
@@ -17,6 +17,7 @@ module mpas_forcing
 #define FORCING_WARNING_WRITE(M) call mpas_log_write( M , messageType=MPAS_LOG_WARN)
 #define FORCING_ERROR_WRITE(M) call mpas_log_write( M , messageType=MPAS_LOG_ERR)
 
+  use mpas_kind_types
   use mpas_derived_types
   use mpas_field_routines
   use mpas_pool_routines
@@ -250,7 +251,7 @@ contains
     ! forward variable interval function
     interface
        function variable_interval_forward(currentTime) result(variableInterval)
-         use mpas_timekeeping
+         use mpas_derived_types
          type(MPAS_Time_type), intent(in) :: currentTime
          type(MPAS_TimeInterval_type) :: variableInterval
        end function variable_interval_forward
@@ -259,7 +260,7 @@ contains
     ! backward variable interval function
     interface
        function variable_interval_backward(currentTime) result(variableInterval)
-         use mpas_timekeeping
+         use mpas_derived_types
          type(MPAS_Time_type), intent(in) :: currentTime
          type(MPAS_TimeInterval_type) :: variableInterval
        end function variable_interval_backward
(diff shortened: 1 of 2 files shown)


matthewhoffman:
@akturner , this looks like helpful cleanup.  I have a few questions about certain member of mpas_timekeeping being public (which I guess points to the need for this sort of cleanup).

1. I'm not sure if `daysInMonth` is meant to be public.  I see there is an identically named variable in `src/core_ocean/mode_init/mpas_ocn_init_global_ocean.F`.
2. I'm not sure what `abs` is and if it should be public.  It seems a bad idea for a routine name to match that of an intrinsic Fortran routine.
3. I'm not sure if `isLeapYear` is meant to be public.  I see it is used in `src/core_atmosphere/physics/mpas_atmphys_manager.F` and `src/core_seaice/shared/mpas_seaice_column.F`, but it does not follow MPAS convention of having public routines start with `mpas_`.

Apart from understanding those issues (and maybe @mgduda knows the answers), I'm good with this PR.

mgduda:
@akturner Having taken a quick scan through the changes, this all looks good. I'll do more thorough testing with a broader range of compilers soon.

I did find that we will need to fix a few module `use` statements in the atmosphere core in order to get it to compile after merging this PR. What I'm proposing is that I (or @MiCurry) make a PR to address these `use` statements in the atmosphere core, and to merge that PR before we merge this one. That way, there will not be a point on the `develop` branch where the atmosphere core fails to build.

mgduda:
@matthewhoffman The `abs` definition in `mpas_timekeeping` just overloads the `abs` intrinsic so that it also works with an `MPAS_TimeInterval_type`, in much the same way that we define the `+` and `-` operators for time intervals. So, I think we do want `abs` to be public.

Regarding the use of `isLeapYear` in `mpas_atmphys_manager.F`, it may be that we can rework whatever logic is currently using `isLeapYear` to make use of more abstract entities like time instants and time intervals that will automatically account for leap years when arithmetic on them is performed. I'll see what I can do in a reasonably short time frame.

xylar:
> Regarding the use of `isLeapYear` in `mpas_atmphys_manager.F`, it may be that we can rework whatever logic is currently using `isLeapYear` to make use of more abstract entities like time instants and time intervals that will automatically account for leap years when arithmetic on them is performed.

It seems like `isLeapYear` should probably just remain public if the atmosphere core is using it.

matthewhoffman:
I'm certainly not opposed to `isLeapYear` remaining public, I just wanted to ensure that it was meant to be used in that way before officially supporting it.  I like Xylar's idea of adding the `mpas_` prefix and calling it good.

mgduda:
As of PR #582, we've addressed issues in the atmosphere core that would lead to build errors after the merge of this PR.

I like the idea of adding an `mpas_` prefix to `isLeapYear` to maintain the convention that public entities begin with `mpas_`. I'd imagine a single additional commit that makes this change across the entire code base would be appropriate, as that would ensure there is no commit that doesn't compile.

matthewhoffman:
@akturner , it looks like this PR has fallen through the cracks.  Is this something that we should still try to merge?  Or should it be closed?

PR #574, opened 2020-05-27
Update compass env to v0.1.6

xylar:
This version of compass fixes an issue on LANL IC compute nodes that was occurring when `ESMF_RegridWeightGen` wasn't getting called with the conda version of `mpirun`.  (This bug was uncovered right before the COMPASS tutorial last week.)

### DIFF
--- a/testing_and_setup/compass/README_ocean.md
+++ b/testing_and_setup/compass/README_ocean.md
@@ -6,11 +6,11 @@ To set up and run ocean test cases from COMPASS, you will need a conda
 environment.  First, install Miniconda3 (if miniconda is not already
 installed), then create a new conda environment as follows:
 ``` bash
-conda create -n compass_0.1.5 -c conda-forge -c e3sm python=3.7 compass=0.1.5
+conda create -n compass_0.1.6 -c conda-forge -c e3sm python=3.7 compass=0.1.6
 ```
 Each time you want to work with COMPASS, you will need to run:
 ```
-conda activate compass_0.1.5
+conda activate compass_0.1.6
 ```
 
 An appropriate conda environment is already available on Los Alamos National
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I ran the nightly and land-ice-fluxes regression suites on Grizzly with a test version of the `compass_0.1.6` environment.  I also ran the `e3sm_coupling` steps for the QU240 and QU240wISC test cases.  Everything ran successfully, whereas the `e3sm_coupling` stage had been failing on 1 core in previous tests.

xylar:
@mark-petersen, do you want to test this yourself?  If so, I have a `test_compass_0.1.6` environment on Grizzly.  If not, I can go ahead and merge https://github.com/MPAS-Dev/pyremap/pull/16 and https://github.com/MPAS-Dev/MPAS-Tools/pull/312, and make `compass_0.1.6`.  Then, I would still ask you to run some tests with that before this PR gets merged.

mark-petersen:
Passes nightly regression suite on grizzly with `test_compass_0.1.6`. Thanks! I think you can merge your other two PRs. I'll merge this one when timing works for E3SM. 

The critical part is when you change the version in `load_latest_compass.sh`, right? Are the versions in the files in this PR just for documentation, or do they actually set what compass version is run?

xylar:
> The critical part is when you change the version in `load_latest_compass.sh`, right? Are the versions in the files in this PR just for documentation, or do they actually set what compass version is run?

Ideally, you should be using the local `load_compass_env.sh` instead of `load_latest_compass.sh`, since the former ensures that the compatible `compass` env. gets loaded.  So we don't want to merge this until the environment exists.

Part of my script for creating `compass` environments is to update `load_latest_compass.sh` automatically.  That is, it always happens before we merge PRs like this one that update the version in COMPASS on ocean/develop.  It is not necessarily true that a given ocean/develop will work fine with the latest `compass`, so, again, it is better to get used to using the `load_compass_env.sh`.

xylar:
Thanks, @mark-petersen!  I really appreciate you keeping PRs rolling through.

PR #575, opened 2020-05-27
Fix bad docs path in Travis CI

xylar:
A glitch became apparent when CI ran on 51d5624709f75182130a2fd94715e1bb13310172.


xylar:
@mark-petersen, this might be the first of several minor PRs before I get CI working seamlessly.

xylar:
What I'm going to do just this once for testing is have it build the documentation in the PR rather than after merging.  That way, we should find out if everything works.  We wouldn't want this to happen normally because it would upload documentation before the PR has actually been merged, which is obviously problematic.

xylar:
I don't know if this is going to work.  I'm not sure if my testing failed because you can only push to the repo as part of merging a PR or making a tag (this would seem like a good safety measure if that's the case) or if I just messed something up.  Let's find out...

xylar:
Thanks @mark-petersen!

PR #583, opened 2020-06-01
Add advection operator convergence test

vanroekel:
Adds the Skamarock and Gassmann cosine bell test case and a regression
suite across resolutions to test operator convergence

Reference: Section 2a in Skamarock, W.C. and A. Gassmann, 2011: Conservative Transport Schemes for Spherical Geodesic Grids: High-Order Flux Operators for ODE-Based Time Integration. Mon. Wea. Rev., 139, 2962–2975, [https://doi.org/10.1175/MWR-D-10-05056.1](https://doi.org/10.1175/MWR-D-10-05056.1)

### DIFF
--- a/src/core_ocean/mode_init/Makefile
+++ b/src/core_ocean/mode_init/Makefile
@@ -26,7 +26,8 @@ TEST_CASES = mpas_ocn_init_baroclinic_channel.o \
              mpas_ocn_init_isomip.o \
              mpas_ocn_init_isomip_plus.o \
              mpas_ocn_init_hurricane.o \
-             mpas_ocn_init_tidal_boundary.o
+             mpas_ocn_init_tidal_boundary.o \
+             mpas_ocn_init_cosine_bell.o
              #mpas_ocn_init_TEMPLATE.o
 
 all: init_mode
--- a/src/core_ocean/mode_init/Registry.xml
+++ b/src/core_ocean/mode_init/Registry.xml
@@ -16,6 +16,7 @@
 #include "Registry_isomip_plus.xml"
 #include "Registry_hurricane.xml"
 #include "Registry_tidal_boundary.xml"
+#include "Registry_cosine_bell.xml"
 // #include "Registry_TEMPLATE.xml"
 
 
--- a/src/core_ocean/mode_init/Registry_cosine_bell.xml
+++ b/src/core_ocean/mode_init/Registry_cosine_bell.xml
@@ -1,0 +1,30 @@
+	<nml_record name="cosine_bell" mode="init" configuration="cvmix_WSwSBF">
+		<nml_option name="config_cosine_bell_temperature" type="real" default_value="15.0" units="deg C"
+			description="Temperature of the ocean."
+			possible_values="Any real number"
+		/>
+		<nml_option name="config_cosine_bell_salinity" type="real" default_value="35.0" units="PSU"
+			description="Salinity of the ocean."
+			possible_values="Any real number"
+		/>
+		<nml_option name="config_cosine_bell_lat_center" type="real" default_value="0.0" units="radians"
+			description="latitude center of cosine bell"
+      possible_values="Any real number between -pi/2 and pi/2"
+		/>
+		<nml_option name="config_cosine_bell_lon_center" type="real" default_value="3.141592" units="radians"
+			description="longitude center of cosine bell"
+			possible_values="Any non-negative real number between 0 and 2pi"
+		/>
+	 <nml_option name="config_cosine_bell_psi0" type="real" default_value="1.0" units="unitless"
+			description="hill max of tracer"
+      possible_values="Any real number"
+		/>
+		<nml_option name="config_cosine_bell_radius" type="real" default_value="2123666.667" units="m"
+			description="radius of cosine bell"
+			possible_values="Any non-negative real number between 0 and 2pi"
+		/>
+	<nml_option name="config_cosine_bell_vel_pd" type="real" default_value="24.0" units="days"
+			description="time for bell to transit equator once"
+			possible_values="Any non-negative real number"
+		/>
+	</nml_record>
--- a/src/core_ocean/mode_init/Registry_cosine_bell.xml
+++ b/src/core_ocean/mode_init/Registry_cosine_bell.xml
@@ -1,4 +1,4 @@
-	<nml_record name="cosine_bell" mode="init" configuration="cvmix_WSwSBF">
+	<nml_record name="cosine_bell" mode="init" configuration="cosine_bell">
 		<nml_option name="config_cosine_bell_temperature" type="real" default_value="15.0" units="deg C"
 			description="Temperature of the ocean."
 			possible_values="Any real number"
@@ -9,21 +9,21 @@
 		/>
 		<nml_option name="config_cosine_bell_lat_center" type="real" default_value="0.0" units="radians"
 			description="latitude center of cosine bell"
-      possible_values="Any real number between -pi/2 and pi/2"
+			possible_values="Any real number between -pi/2 and pi/2"
 		/>
 		<nml_option name="config_cosine_bell_lon_center" type="real" default_value="3.141592" units="radians"
 			description="longitude center of cosine bell"
 			possible_values="Any non-negative real number between 0 and 2pi"
 		/>
-	 <nml_option name="config_cosine_bell_psi0" type="real" default_value="1.0" units="unitless"
+		<nml_option name="config_cosine_bell_psi0" type="real" default_value="1.0" units="unitless"
 			description="hill max of tracer"
-      possible_values="Any real number"
+			possible_values="Any real number"
 		/>
 		<nml_option name="config_cosine_bell_radius" type="real" default_value="2123666.667" units="m"
 			description="radius of cosine bell"
 			possible_values="Any non-negative real number between 0 and 2pi"
 		/>
-	<nml_option name="config_cosine_bell_vel_pd" type="real" default_value="24.0" units="days"
+		<nml_option name="config_cosine_bell_vel_pd" type="real" default_value="24.0" units="days"
 			description="time for bell to transit equator once"
 			possible_values="Any non-negative real number"
 		/>
--- a/src/core_ocean/mode_init/mpas_ocn_init_cosine_bell.F
+++ b/src/core_ocean/mode_init/mpas_ocn_init_cosine_bell.F
@@ -1,0 +1,300 @@
+! Copyright (c) 2013,  Los Alamos National Security, LLC (LANS)
+! and the University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
+! Additional copyright and license information can be found in the
+! LICENSE file
+! distributed with this code, or at
+! http://mpas-dev.github.com/license.html
+!
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+!
+!  ocn_init_cosine_bell
+!
+!> \brief MPAS ocean initialize case -- CVMix Unit Test
+!> WSwSBF means Wind Stress with Surface Buoyancy Forcing
+!> \author Todd Ringler
+!> \date   04/23/2015
+!> \details
+!>  This module contains the routines for initializing the
+!>  the cvmix WSwSBF unit test configuration. This in a
+!>  single column configuration
+!
+!-----------------------------------------------------------------------
+
+module ocn_init_cosine_bell
+
+   use mpas_kind_types
+   use mpas_io_units
+   use mpas_derived_types
+   use mpas_pool_routines
+   use mpas_constants
+   use mpas_stream_manager
+   use mpas_dmpar
+
+   use ocn_init_cell_markers
+   use ocn_init_vertical_grids
+
+   implicit none
+   private
+   save
+
+   !--------------------------------------------------------------------
+   !
+   ! Public parameters
+   !
+   !--------------------------------------------------------------------
+
+   !--------------------------------------------------------------------
+   !
+   ! Public member functions
+   !
+   !--------------------------------------------------------------------
+
+   public :: ocn_init_setup_cosine_bell, &
+             ocn_init_validate_cosine_bell
+
+   !--------------------------------------------------------------------
+   !
+   ! Private module variables
+   !
+   !--------------------------------------------------------------------
+
+!***********************************************************************
+
+contains
+
+!***********************************************************************
+!
+!  routine ocn_init_setup_cosine_bell
+!
+!> \brief   Setup for cvmix WSwSBF unit test configuration
+!> \author  Todd Ringler
+!> \date    04/23/2015
+!> \details
+!>  This routine sets up the initial conditions for the cvmix WSwSBF
+!unit test configuration.
+!
+!-----------------------------------------------------------------------
+
+   subroutine ocn_init_setup_cosine_bell(domain, iErr)!{{{
+
+   !--------------------------------------------------------------------
+
+      type (domain_type), intent(inout) :: domain
+      integer, intent(out) :: iErr
+      real (kind=RKIND) :: temperature, salinity
+
+      type (block_type), pointer :: block_ptr
+
+      type (mpas_pool_type), pointer :: meshPool, verticalMeshPool,statePool
+      type (mpas_pool_type), pointer :: diagnosticsPool, forcingPool
+
+      type (mpas_pool_type), pointer :: tracersPool
+
+      integer, pointer :: nVertLevels, nVertLevelsP1, nCellsSolve,nEdgesSolve, nEdge, nVerticesSolve
+      integer, pointer :: index_temperature, index_salinity,index_tracer1
+
+      integer, dimension(:), pointer :: maxLevelCell
+      real (kind=RKIND), dimension(:), pointer :: refBottomDepth,refZMid
+      real (kind=RKIND), dimension(:), pointer :: lonCell, latEdge, latCell, latVertex 
+      real (kind=RKIND), dimension(:), pointer :: bottomDepth, angleEdge
+      real (kind=RKIND), dimension(:), pointer :: fCell, fEdge, fVertex,xCell, yCell
+      real (kind=RKIND), dimension(:, :), pointer :: layerThickness,restingThickness,normalVelocity
+      real (kind=RKIND), dimension(:, :, :), pointer :: activeTracers,debugTracers
+
+      real (kind=RKIND), dimension(:), pointer :: interfaceLocations
+
+      integer :: iCell, iEdge, iVertex, k, kML
+
+      real (kind=RKIND) :: temp, velocity
+
+      character (len=StrKIND), pointer :: config_init_configuration
+
+      real (kind=RKIND), pointer ::config_cosine_bell_temperature, &
+                                   config_cosine_bell_salinity,    &
+                                   config_cosine_bell_radius,  &
+                                   config_cosine_bell_lat_center, &
+                                   config_cosine_bell_lon_center, &
+                                   config_cosine_bell_psi0,  &
+                                   config_cosine_bell_vel_pd, &
+                                   sphere_radius
+      ! assume no error
+      iErr = 0
+
+      ! get and test if this is the configuration specified
+      call mpas_pool_get_config(domain % configs,'config_init_configuration', config_init_configuration)
+      if(config_init_configuration .ne. trim('cosine_bell')) return
+
+      ! load the remaining configuration parameters
+      call mpas_pool_get_config(domain % configs,'config_cosine_bell_temperature', &
+                                config_cosine_bell_temperature)
+      call mpas_pool_get_config(domain % configs,'config_cosine_bell_salinity', &
+                    config_cosine_bell_salinity)
+      call mpas_pool_get_config(domain % configs,'config_cosine_bell_radius', &
+                                config_cosine_bell_radius)
+      call mpas_pool_get_config(domain % configs,'config_cosine_bell_psi0', &
+                                config_cosine_bell_psi0)
+      call mpas_pool_get_config(domain % configs,'config_cosine_bell_vel_pd', &
+                                config_cosine_bell_vel_pd)
+      call mpas_pool_get_config(domain % configs,'config_cosine_bell_lat_center', &
+                                config_cosine_bell_lat_center)
+      call mpas_pool_get_config(domain % configs,'config_cosine_bell_lon_center', &
+                                config_cosine_bell_lon_center)
+      !load data that required to initialize the ocean simulation
+      block_ptr => domain % blocklist
+      do while(associated(block_ptr))
+        call mpas_pool_get_subpool(block_ptr % structs, 'mesh',meshPool)
+        call mpas_pool_get_subpool(block_ptr % structs, 'verticalMesh',verticalMeshPool)
+        call mpas_pool_get_subpool(block_ptr % structs, 'state',statePool)
+        call mpas_pool_get_subpool(block_ptr % structs, 'diagnostics',diagnosticsPool)
+        call mpas_pool_get_subpool(block_ptr % structs, 'forcing',forcingPool)
+
+        call mpas_pool_get_subpool(statePool, 'tracers', tracersPool)
+      
+        call mpas_pool_get_dimension(meshPool, 'nVertLevels',nVertLevels)
+        call mpas_pool_get_dimension(meshPool, 'nCellsSolve',nCellsSolve)
+        call mpas_pool_get_dimension(meshPool, 'nEdge', nEdge)
+        call mpas_pool_get_dimension(meshPool, 'nEdgesSolve',nEdgesSolve)
+        call mpas_pool_get_dimension(meshPool, 'nVerticesSolve',nVerticesSolve)
+        call mpas_pool_get_config(meshPool, 'sphere_radius', sphere_radius)
+
+        call mpas_pool_get_dimension(tracersPool, 'index_temperature',index_temperature)
+        call mpas_pool_get_dimension(tracersPool, 'index_salinity',index_salinity)
+        call mpas_pool_get_dimension(tracersPool, 'index_tracer1',index_tracer1)
+
+        call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
+        call mpas_pool_get_array(meshPool, 'bottomDepth', bottomDepth)
+        call mpas_pool_get_array(meshPool, 'angleEdge', angleEdge)
+        call mpas_pool_get_array(meshPool, 'latCell', latCell)
+        call mpas_pool_get_array(meshPool, 'latEdge', latEdge)
+        call mpas_pool_get_array(meshPool, 'latVertex', latVertex)
+        call mpas_pool_get_array(meshPool, 'lonCell', lonCell)
+        call mpas_pool_get_array(meshPool, 'refBottomDepth', refBottomDepth)
+        call mpas_pool_get_array(meshPool, 'fCell', fCell)
+        call mpas_pool_get_array(meshPool, 'fEdge', fEdge)
+        call mpas_pool_get_array(meshPool, 'fVertex', fVertex)
+        call mpas_pool_get_array(meshPool, 'xCell', xCell)
+        call mpas_pool_get_array(meshPool, 'yCell', yCell)
+
+
+        call mpas_pool_get_array(verticalMeshPool, 'refZMid', refZMid)
+        call mpas_pool_get_array(verticalMeshPool, 'restingThickness',restingThickness)
+
+        call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocity, 1)
+        call mpas_pool_get_array(tracersPool, 'activeTracers',activeTracers, 1)
+        call mpas_pool_get_array(tracersPool, 'debugTracers',debugTracers, 1)
+        call mpas_pool_get_array(statePool, 'layerThickness',layerThickness, 1)
+
+        ! Set refBottomDepth and refBottomDepthTopOfCell
+        do k = 1, nVertLevels
+           refBottomDepth(k) = 100.0_RKIND
+           refZMid(k) = 50.0_RKIND
+        end do
+
+        do iCell = 1, nCellsSolve
+           if(associated(activeTracers) ) then
+
+           ! Loop from surface through surface layer depth
+        
+           activeTracers(index_temperature, :, iCell) = config_cosine_bell_temperature
+           activeTracers(index_salinity, :, iCell) = config_cosine_bell_salinity
+          endif
+
+        if(associated(debugTracers)) then
+          temp = sphere_radius*acos(sin(config_cosine_bell_lat_center)*sin(latCell(iCell)) + &
+                  cos(config_cosine_bell_lat_center)*cos(latCell(iCell))*cos(lonCell(iCell) - &
+                  config_cosine_bell_lon_center))
+
+          if( temp < config_cosine_bell_radius ) then
+            debugTracers(:,:,iCell) = config_cosine_bell_psi0/2.0_RKIND * ( 1.0_RKIND + &
+              cos(3.1415926_RKIND*temp/config_cosine_bell_radius))
+          else
+            debugTracers(:,:,iCell) = 0.0_RKIND
+          endif
+        endif
+        ! Set layerThickness
+        do k = 1, nVertLevels
+           layerThickness(k, iCell) = 100.0_RKIND 
+           restingThickness(k, iCell) = layerThickness(k, iCell)
+        end do
+
+        ! Set Coriolis parameter
+        fCell(iCell) = 0.0_RKIND*omega*sin(latCell(iCell))
+
+        ! Set bottomDepth
+        bottomDepth(iCell) = 100.0_RKIND
+
+        ! Set maxLevelCell
+        maxLevelCell(iCell) = nVertLevels
+
+     end do  ! do iCell
+
+     do iEdge = 1, nEdgesSolve
+        fEdge(iEdge) = 0.0_RKIND*omega*sin(latEdge(iEdge))
+     end do
+
+     do iVertex=1, nVerticesSolve
+        fVertex(iVertex) = 0.0_RKIND*omega*sin(latVertex(iVertex))
+     end do
+
+     !set normal velocity
+     do iEdge = 1,nEdgesSolve
+       velocity = 2.0*3.14159265*sphere_radius*cos(latEdge(iEdge)) / &
+         (86400.0_RKIND*config_cosine_bell_vel_pd)
+       normalVelocity(:,iEdge) = velocity*cos(angleEdge(iEdge))
+     enddo
+
+     block_ptr => block_ptr % next
+   end do
+
+  !--------------------------------------------------------------------
+
+   end subroutine ocn_init_setup_cosine_bell!}}}
+
+!***********************************************************************
+!
+!  routine ocn_init_validate_cosine_bell
+!
+!> \brief   Validation for CVMix WSwSBF mixing unit test case
+!> \author  Doug Jacobsen
+!> \date    04/01/2015
+!> \details
+!>  This routine validates the configuration options for the CVMix
+!WSwSBF mixing unit test configuration.
+!
+!-----------------------------------------------------------------------
+
+   subroutine ocn_init_validate_cosine_bell(configPool, packagePool,iocontext, iErr)!{{{
+
+   !--------------------------------------------------------------------
+
+      type (mpas_pool_type), intent(inout) :: configPool
+      type (mpas_pool_type), intent(inout) :: packagePool
+      type (mpas_io_context_type), intent(inout) :: iocontext
+
+      integer, intent(out) :: iErr
+
+      character (len=StrKIND), pointer :: config_init_configuration
+      integer, pointer :: config_vert_levels
+
+      iErr = 0
+
+      call mpas_pool_get_config(configPool, 'config_init_configuration',config_init_configuration)
+
+      if(config_init_configuration .ne. trim('cosine_bell')) return
+
+      call mpas_pool_get_config(configPool, 'config_vert_levels',config_vert_levels)
+
+      config_vert_levels = 3 
+
+   !--------------------------------------------------------------------
+
+   end subroutine ocn_init_validate_cosine_bell!}}}
+
+!***********************************************************************
+
+end module ocn_init_cosine_bell
+
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+! vim: foldmethod=marker
--- a/src/core_ocean/mode_init/mpas_ocn_init_cosine_bell.F
+++ b/src/core_ocean/mode_init/mpas_ocn_init_cosine_bell.F
@@ -11,14 +11,16 @@
 !
 !  ocn_init_cosine_bell
 !
-!> \brief MPAS ocean initialize case -- CVMix Unit Test
-!> WSwSBF means Wind Stress with Surface Buoyancy Forcing
-!> \author Todd Ringler
-!> \date   04/23/2015
+!> \brief MPAS ocean initialize case -- Cosine Bell
+!> \author Luke Van Roekel
+!> \date   09/01/2020
 !> \details
 !>  This module contains the routines for initializing the
-!>  the cvmix WSwSBF unit test configuration. This in a
-!>  single column configuration
+!>  the cosine bell test case
+!>  Reference: Section 2a in Skamarock, W.C. and A. Gassmann, 2011: Conservative
+!>  Transport Schemes for Spherical Geodesic Grids: High-Order Flux Operators for
+!>  ODE-Based Time Integration. Mon. Wea. Rev., 139, 2962–2975,
+!>  https://doi.org/10.1175/MWR-D-10-05056.1
 !
 !-----------------------------------------------------------------------
 
@@ -68,12 +70,14 @@ contains
 !
 !  routine ocn_init_setup_cosine_bell
 !
-!> \brief   Setup for cvmix WSwSBF unit test configuration
-!> \author  Todd Ringler
-!> \date    04/23/2015
+!> \brief   Set-up for test case
+!> \author  Luke Van Roekel
+!> \date    09/01/2020
 !> \details
-!>  This routine sets up the initial conditions for the cvmix WSwSBF
-!unit test configuration.
+!>  Reference: Section 2a in Skamarock, W.C. and A. Gassmann, 2011: Conservative
+!>  Transport Schemes for Spherical Geodesic Grids: High-Order Flux Operators for
+!>  ODE-Based Time Integration. Mon. Wea. Rev., 139, 2962–2975,
+!>  https://doi.org/10.1175/MWR-D-10-05056.1
 !
 !-----------------------------------------------------------------------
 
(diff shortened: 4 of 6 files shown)


vanroekel:
@mark-petersen  and @xylar here is my WIP advection operator convergence for the global advective test case.  If you have some time, I'd greatly appreciate your comments.

vanroekel:
also tagging @jamilgafur

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
```diff
@@ -110,8 +110,8 @@ def build_mesh(
                      output_name='mesh_triangles.nc', on_sphere=on_sphere)
 
     print('Step 4. Convert from triangles to MPAS mesh')
-    write_netcdf(convert(xarray.open_dataset('mesh_triangles.nc')),
-                 'base_mesh.nc')
+    write_netcdf(convert(xarray.open_dataset('mesh_triangles.nc'),
+                    graphInfoFileName='graph.info'),'base_mesh.nc')
```
xylar, line 114: Perfect!  My only comment here would be to indent another 5 spaces so the formatting follows the PEP8 convention.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/config_driver.xml
```diff
@@ -0,0 +1,5 @@
+<driver_script name="run.py">
+  <case name="rmse">
```
xylar, line 2: Eventually, we'd want to add some `prerequisite` tags here.  It would be harmless to do so now if you want to.  See https://github.com/MPAS-Dev/MPAS-Model/pull/560

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
...
+            ii = line.find('=')+1
+            radius = float(line[ii:])
+        if "config_cosine_bell_psi0" in line: 
+            ii = line.find('=')+1
+            psi0 = float(line[ii:])
+        if "config_cosine_bell_vel_pd" in line: 
+            ii = line.find('=')+1
+            pd = float(line[ii:])
+
+    init = xr.open_dataset('../../'+resTag+'/default/init_step/initial_state.nc')
```
xylar, line 40: I really prefer people alter strings with `.format()` instead of string "addition".  This may be may own prejudice but I don't see string addition used by seasoned python programmers.
```suggestion
    init = xr.open_dataset('../../{}/default/init_step/initial_state.nc'.format(resTag))
```

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
...
+    #find time since the beginning of run
+    ds = xr.open_dataset('../../'+resTag+'/default/forward/output/output.0001-01-01_00.00.00.nc')
```
xylar, line 42: Similar to above, maybe use `.format()`.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
...
+        newLon -= 2.0*np.pi
+
+    #construct analytic tracer
+    tracer = np.zeros_like(init.tracer1[0,:,0].values)
+    latC = init.latCell.values
+    lonC = init.lonCell.values
+    for i in range(len(latC)):
+        temp = R*np.arccos(np.sin(latCent)*np.sin(latC[i]) + np.cos(latCent)*np.cos(latC[i])*np.cos(lonC[i] - newLon))
+        if temp < radius:
+            tracer[i] = psi0 / 2.0 * (1.0 + np.cos(3.1415926*temp/radius))
```
xylar, line 67: Never loop if you don't have to in python.

```suggestion
    temp = R*np.arccos(np.sin(latCent)*np.sin(latC) + np.cos(latCent)*np.cos(latC)*np.cos(lonC - newLon))
    mask = temp < radius
    tracer[mask] = psi0 / 2.0 * (1.0 + np.cos(numpy.py*temp[mask]/radius))
```

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
+import numpy as np
+import xarray as xr
+import matplotlib
+import matplotlib.pyplot as plt
+import os 
+
+%matplotlib inline
```
xylar, line 7:
```suggestion
# matplotlib inline
```

xylar, line 7: I haven't seen this one before.

vanroekel: this was an error in importing from jupyter notebook to python script.  It's ipython magic.  I've removed this.

xylar:
@vanroekel, most of my comments are just little things that you're free to take or leave.  Looks like a decent approach to me.  Happy to do some testing once the time comes.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
vanroekel: that's nice! thanks for the suggestion

xylar:
> Thanks @xylar I will make the changes. As I test this, I'm having a hard time controlling the plot size and position of the convergence label (annotate) command. Do you have any suggestions? Here is what a plot looks like right now
> 

Oh, yeah, htat doesn't look great.  Let me see what I can figure out.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
+import numpy as np
+import xarray as xr
+import matplotlib
```
xylar, line 3: Might want to switch to the `Agg` backend here so it works on compute nodes, too, and without x-forwarding.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
...
+import os 
+
+%matplotlib inline
+matplotlib.rcParams['figure.figsize'] = (16, 16) # Large figures
+dpi=200;
+
+### axis_font = {'fontname':'Arial', 'size':'18'}    
+title_font = {'fontname':'Arial', 'size':'32', 'color':'black', 'weight':'normal'}
+matplotlib.rc('xtick', labelsize=28)
+matplotlib.rc('ytick', labelsize=28)
```
xylar, line 14: I think this is the stuff that's causing your figure to look bad.  Try taking it out.  I'll make a few recommendations below once you do that.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
...
+    exec('rmse'+res[i]+',nCells'+res[i]+' = rmse(res[i],24)')
+    exec('xtemp.append(nCells'+res[i]+')')
+    exec('ytemp.append(rmse'+res[i]+')')
+xdata = np.asarray(xtemp)
+ydata = np.asarray(ytemp)
+
+p = np.polyfit(np.log10(xdata),np.log10(ydata),1)
+conv = abs(p[0])
+
+plt.loglog(xdata,ydata,'k')
```
xylar, line 90: If the figure isn't the size you want, set the size here:
``` python
plt.figure(figsize=[<width>, <height>])
```
where width and height are in inches.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
...
+plt.scatter(xdata,ydata,s=40,c='r')
+plt.annotate('Order of Convergence = {}'.format(np.round(conv,2)),xy=((xdata[-1]+xdata[0])*0.4,ydata[0]),fontsize=28)
+plt.savefig('convergence.png')
```
xylar, line 93: If the fig is low resolution here, set `dpi=200` here.  I wouldn't bother doing that earlier unless you're going to use `plt.show()`.

testing_and_setup/compass/ocean/cosine_bell/RMSE/default/compute_rmse.py
```diff
@@ -0,0 +1,94 @@
...
+    exec('ytemp.append(rmse'+res[i]+')')
+xdata = np.asarray(xtemp)
+ydata = np.asarray(ytemp)
+
+p = np.polyfit(np.log10(xdata),np.log10(ydata),1)
+conv = abs(p[0])
+
+plt.loglog(xdata,ydata,'k')
+plt.scatter(xdata,ydata,s=40,c='r')
+plt.annotate('Order of Convergence = {}'.format(np.round(conv,2)),xy=((xdata[-1]+xdata[0])*0.4,ydata[0]),fontsize=28)
```
xylar, line 92: 28 is a big font size. And it might work better to just put it in the upper left corner or something.  There are probably other tricks for putting it "where it fits bets" like is done for legends.  If you want that, I can look into it.

vanroekel:
thanks @xylar!  Your comments have been extremely helpful.  Here is an updated plot

convergence

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
vanroekel: Just so I'm clear, indent the second line 5 more spaces?  Is this so it is easier to see it is part of the convert function?

vanroekel:
@mark-petersen I've run through the advection test and here is the result

convergence

The convergence order is much lower than expected.  When you have a chance could you look through this test and see if I have an error?

testing_and_setup/compass/ocean/jigsaw_to_MPAS/build_mesh.py
mark-petersen: If you load `e3sm_unified` you can run
```
autopep8 --in-place --aggressive --aggressive foo.py
```
on your python files for automatic pep8 spacing.

mark-petersen:
@vanroekel thanks for your work on this. The low order of convergence is certainly surprising. The direct comparison is [Skamarock and Gassmann](https://journals.ametsoc.org/doi/full/10.1175/MWR-D-10-05056.1) [figure 4](https://journals.ametsoc.org/na101/home/literatum/publisher/ams/journals/content/mwre/2011/15200493-139.9/mwr-d-10-05056.1/production/images/large/mwr-d-10-05056.1-f4.jpeg):
image
They don't give numbers for the computed order on this plot, but everything is at least 2. MPAS-Ocean should be using the fourth order scheme. I would expect that limiting is not turning on, but we could check that.

vanroekel:
@mark-petersen monotone is on here.  We also are using the blended 3rd/4th order scheme (coeff3rdOrder =0.25).  I've tried for various values of that coefficient and it doesn't help convergence order.  I've also tried with non monotone and the order is roughly the same.

vanroekel:
One other thing, there is a pretty strong time step sensitivity.  Cutting dt to 1 min changes convergence strongly (not for the better).  I'll add some plots of the tracer vs exact solution shortly

convergence

vanroekel:
a quick update, using RK4 improves convergence, but only slightly

convergence

src/core_ocean/mode_init/Makefile
```diff
@@ -26,7 +26,8 @@ TEST_CASES = mpas_ocn_init_baroclinic_channel.o \
              mpas_ocn_init_isomip.o \
              mpas_ocn_init_isomip_plus.o \
              mpas_ocn_init_hurricane.o \
-             mpas_ocn_init_tidal_boundary.o
+             mpas_ocn_init_tidal_boundary.o \
+						 mpas_ocn_init_cosine_bell.o
```
jamilgafur, line 30: Small edit, but spacing here is weird.

ambrad:
I have a preliminary observation to make. For a number of reasons, I speculated that perhaps the issue in this test is the quality of the meshes rather than the quality of the advection algorithm. Thus, my first focus was to study the effect of grid.

I put together a simple standalone advection tester. It reads in the grid file and then computes advection using the simple 2nd-order flux-form method eqs 4 and 5 in Skamarock and Gassmann MWR 2011 [SG11]. So for now I'm ignoring the high-order terms and the FCT limiter.

The standalone tester uses the same solid-body rotation flow that Luke uses, and the comparison is for a smooth function after one full rotation.

I ran it on two different grid sequences:
- (A) this PR's base_mesh.nc files for QU240, QU120, QU60 (I'm guessing QU means "quasi-uniform");
- (B) [meshes from this page](https://mpas-dev.github.io/atmosphere/atmosphere_meshes.html), in particular the 480, 240, 120, 60 km ones.

I find that for sequence B, I get order of accuracy ~2 but for sequence (A) the results are poor, as this PR reports.

In detail, for sequence B, I get these relative errors:
```
Sequence B:
>> m-480-x1.2562.grid.nc
nstep 3333 cfl 12445.115345 dt 311.071107 R 1.000000
re 7.885604094598288e-02  // l2 relative error
>> m-240-x1.10242.grid.nc
nstep 6664 cfl 6223.805186 dt 155.582233 R 1.000000
re 2.079732736169380e-02
>> m-120-x1.40962.grid.nc
nstep 13327 cfl 3112.049415 dt 77.796954 R 1.000000
re 5.516979485210611e-03
>> m-060-x1.163842.grid.nc
nstep 26653 cfl 1556.042011 dt 38.899936 R 1.000000
re 1.456514346074740e-03
```
In fact, the order of accuracy is not _quite_ 2; it's empirically ~1.94. For example, from the finest two runs, sqrt(5.517e-03/1.457e-03) = 1.946. But my understanding from SG11 is that for a nonuniform mesh -- even quasi-uniform -- the formal OOA is in fact just 1, but in practice we see closer to 2 for the 2nd-order method if the mesh is quasi-uniform and of high quality.

In contrast, for sequence A, running the same standalone code, I get this:
```
Sequence A
>> QU240
nstep 6722 cfl 6170.272881 dt 154.239810 R 6371000.000000
re 2.837935619617233e-02
>> QU120
nstep 13410 cfl 3092.770919 dt 77.315436 R 6371000.000000
re 1.238791742755619e-02
>> QU60
nstep 26785 cfl 1548.329830 dt 38.708232 R 6371000.000000
re 6.347095107476773e-03
```
Here we see for the finest two meshes sqrt(1.239e-02/6.347e-03) = 1.397.

ambrad:
I ran with the Cosine Bells initial condition -- here two bells, which tends to be used more often than one -- and got the following results:
```
Cosine Bells

>> m-480-x1.2562.grid.nc
nstep 834 cfl 12445.115345 dt 1243.165468 R 1.000000
re 5.268225679573880e-01
>> m-240-x1.10242.grid.nc
nstep 1666 cfl 6223.805186 dt 622.328932 R 1.000000
re 1.755478424166668e-01
>> m-120-x1.40962.grid.nc
nstep 3332 cfl 3112.049415 dt 311.164466 R 1.000000
re 5.090638329330135e-02
>> m-060-x1.163842.grid.nc
nstep 6664 cfl 1556.042011 dt 155.582233 R 1.000000
re 1.476058078890171e-02  OOA 1.857

>> QU240
nstep 1681 cfl 6170.272881 dt 616.775729 R 6371000.000000
re 1.800198720122974e-01
>> QU120
nstep 3353 cfl 3092.770919 dt 309.215628 R 6371000.000000
re 6.099823027247491e-02
>> QU60
nstep 6697 cfl 1548.329830 dt 154.815589 R 6371000.000000
re 2.299687683198164e-02  OOA 1.629
```
So mesh quality still has an impact. Here are images for the QU240 resolution:
m-240-x1.10242.grid.nc
ff-cosinebells-m-240
QU240
ff-cosinebells-QU240

ambrad:
I ended up implementing the nondivergent flow test case before moving on. Here I'll just show two 120km results for nondivergent flow with cosine bells. The runs use my own angleEdge calculation. In each figure, top-left is initial, top-right is midpoint, bottom-left is final, and bottom-right is final-initial. This is with the 2nd-order method, so we shouldn't expect great results. The main thing of interest is the QU120 field has the mesh imprint in it that we've observed previously.

m-120-x1.40962.grid.nc
ff-nondiv-cb-m-120
QU120
ff-nondiv-cb-QU120

ambrad:
It may actually be pretty easy to see that the QU* meshes have poor quality just by looking at the meshes. Here I show areaCell as a function of (lon,lat). To make it easy to compare, divide all areas by the maximum one, so the max value in the image is 1.
m-120-x1.40962.grid.nc
areaCell-m-240
QU240
areaCell-QU240

vanroekel:
@dengwirda could you take a look through this comment history and offer any thoughts?   In particular would you expect the seams in jigsaw meshes to generate a poorer advection convergence and generate noise relative to the standard Lloyd's algorithm SCVT code used by mpas-atmosphere?

dengwirda:
@vanroekel @ambrad Just trying to come up to speed here, but my initial thoughts are that, yes --- this seems to be a comparison between quasi-structured meshes (the sequence B icosahedrons) and unstructured meshes (the sequence A QU meshes via jigsaw), which shows the unstructured cases have somewhat lower L2 convergence rates. Approx 1.86 vs 1.63 in the cosine bell config. if I'm reading things correctly.
- My understanding of TRiSK is that it's indeed only 2nd-order accurate given "perfect" meshes (i.e. structured regular hexagons + equilateral triangles), so one wouldn't expect full 2nd-order convergence in general unstructured cases.
- I think I'm a little surprised by the solid-body rotation results. In theory, this should be simpler than the cosine bell / nondivergent cases, and yet seems to give worse results?
- Just as an aside, jigsaw does do Lloyd-like mesh optimisation too, but it doesn't start from a structured icosahedral config. 
- The issue wrt. `angleEdge` is interesting --- if any of the terms/variables in TRiSK etc don't fully account for mesh unstructured-ness there'll be extra error introduced for sure!
- What's the overall goal you're aiming for in this work? On one hand, it'd be great to understand why TRiSK doesn't do so well as soon as the meshes become non-perfect. On the other hand, you could look into some newer jigsaw settings that create "less unstructured" meshes.

Happy to help with this, just let me know.

ambrad:
Regarding angleEdge, this should be an easy and local fix to an MPAS tool. The current code is [here](https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp#L2162).

Its method is to difference the latitude values of the two vertices of an edge, then divide that by the spherical arc length of edge, and finally take acos of that. This approach has an error for finite edge length that grows with the length of the edge. For example, consider an edge that has length pi and is oriented at angle pi/2. Then the latitude difference is pi/2, and acos((pi/2)/pi) = acos(1/2) = pi/3, whereas we want pi/4.

A better approach is as follows, recalling that we can use cross products of Euclidean quantities and still get the angle on the sphere. Let the edge center be ec, and let its supporting vertices be v0 and v1. Then do this:
```
! ec is the edge center on the unit sphere, with vertices v0 and v1
vdif = v1 - v0                        ! edge vector
nml = normalize(cross(vdif, ec))      ! edge normal
east = normalize(cross([0,0,1], ec))  ! local east at edge center
north = normalize(cross(ec, east))    ! local north at edge center
angleEdge = atan2(dot(nml, north), dot(nml, east))
```
This calculation still has the problem that tends to occur for all completely unstructured meshes: vdif loses accuracy as the mesh is refined. This is true in the latitude difference calculation above, too. But this is a small detail that becomes relevant at only extremely high resolution, and likely it affects many calculations in addition to this one.

vanroekel:
@dengwirda thanks for chiming in.  This work came out of a larger model verification effort.  We have been trying to put in place operator testing into MPAS.  Horizontal advection was our first, but have been surprised by the results.  Also in E3SM simulations advective tendencies have always been very noisy (near grid scale) and I had wondered if this poor convergence played a role there.  However, if this is a jigsaw parameter issue, that may not be the case as we are using the non jigsaw generation for our simulations right now.  So perhaps those are unrelated issues.

vanroekel:
@ambrad thanks for the angleEdge explanation, do you think this inaccuracy could cause observable noise in E3SM simulations?

ambrad:
@vanroekel that's a really good point, and one that I can test. I used the m-* sequence B meshes and swapped in the latitude-difference calculation summarized above to get angleEdge. Thus, I'm testing the simulation on a mesh that we know can give smooth results with a slightly bad angleEdge. I run with an infinitely differentiable field and an off-axis solid body rotation. And indeed we see grid-scale signal that is clearly distinct from the dispersion error. I'm a little concerned here about confirmation bias, so (as I always do) I'll be sure to update this comment if I suddenly realize that there's a flaw in this experiment.

m-240-x1.10242.grid.nc with cross/dot-produce-based angleEdge
ff-angleEdge-good-rt-m-240
m-240-x1.10242.grid.nc with lattidue-difference-based angleEdge
ff-angleEdge-bad-rt-m-240

pbosler:
@ambrad @vanroekel One odd thing I notice about the Sequence A meshes may be a significant digits issue.   The sequence B meshes appear to be on the unit sphere (R = 1.00000), while the sequence B meshes appear to have an earth-sized sphere with R=6371000, which would have units of meters, but seems wrong to me.

For example, in `mpas/src/framework/mpas_constants.F`, the radius of the sphere is defined to be 6371229.0 meters

ambrad:
@pbosler that's a good point and could be a source of error. However, if it is, it isn't in a straightforward way; the treatment of the Earth radius is separate from that of the mesh-dependent sphere_radius. In any case, my test code uses the radius as given.

I implemented the high-order terms today. I haven't fully verified my implementation yet, but it's good enough to see that the 3rd-order term with beta = 0.25 improves solutions a lot. In particular, it substantially improves the dispersion error. First, for a quasi-uniform mesh, compare orders 2 and 3, and notice that the dispersion error is substantially less with order 3:
order 2, m-120:
ff-nc2-m-120
order 3, m-120:
ff-nc3-m-120

For the sequence A meshes, order 3 also dissipates the errors due to the less uniform cells:
order 2, QU120
ff-nc2-QU120
order 3, QU120
ff-nc3-QU120

However, the error due to the mesh is not entirely gone. This is apparent when using solid-body rotation with a smooth field:
order 3, m-120
ff-rt3-m-120
order 3, QU120
ff-rt3-QU120
Notice that in the sequence-B mesh, the effect of the 10 pentagons is visible in the error field, so even this mesh type has associated systematic error.

dengwirda:
Hi all, I've been trying to understand if / why there may be a difference between the shallow-water test cases and the solid-body results --- a few quick thoughts / questions:
- Is `angleEdge` used in both cases? I can understand that it's used in the solid-body runs (to build the `u_nrm` edge velocities from the full vector field), but I'd expect that in the SWE `u_nrm` is determined via TRiSK directly, and that perhaps `angleEdge` is not used at all. Is this the case? If there are indeed issues with `angleEdge` do you think this may explain the differences in behaviour?
- As @pbosler says, there's currently some issues with `mpas_mesh_converter.cpp` and `sphereRadius` etc --- we're fixing this (for other reasons) at present.
- In terms of the 2nd and 3rd order schemes, are they both upwinded in some fashion? If not (if the 2nd order scheme is centred) does this explain the dispersive vs diffusive errors? (even-ordered centred schemes would have leading dispersive error terms, while upwinded stencils should have diffusive leading errors, as I understand it).
- For me, the fact that the few pentagons in the icosahedral runs are causing obvious artifacts (in the solid-body tests) is a bit worrying. I understand these cells are not perfect hexagons, but they're hardly "poor quality" cells.

ambrad:
The 2nd-order scheme is centered and the 3rd-order scheme has upwinded hyperdiffusivity.

The pentagon result is from my own and quite independent standalone code, so we shouldn't conclude anything from it yet. This code is not yet achieving the full-order accuracy that I'm looking for, so it probably still has bugs.

dengwirda:
Thanks @ambrad. 
- I actually think the full-sphere tracer distribution you're using is perhaps better in this case than the more compact functions, since it ensures all of the pentagons are "hit" by the advection. I definitely also appreciate that the "streak" features are showing up in the anomaly rather than the actual tracer distribution, but I think it still shows something slightly strange is perhaps happening in the solid-body tests.
- Is it worth checking whether discrete `div(u_nrm) = 0` (by setting a unit tracer globally)?

ambrad:
@dengwirda, I may be mistaken, but in the case of solid-body rotation on the sphere, I don't think the discrete divergence is generally everywhere 0 unless the specified velocity normal is the analytical mean over an edge. I believe but am not positive that this edge mean is not the intended imposed velocity; rather, I think the intended imposed velocity is the point value at the center.

So that we have some data to discuss regarding this point, here's an image for a constant field set to 0.42 from my standalone tester. I've run with the order-2 method for simplicity:
c-QU240, order 2, rotation
ff-r12-c240

dengwirda:
Thanks for including the plots @ambrad. I think I'm on the other side of the fence on this issue though --- as I currently understand it, I think a discrete div-free velocity is what's needed for these type of conservative transport problems. I agree that designing such a `u_nrm` field is not necessarily straightforward, especially for the general unstructured case, but perhaps the following may be useful?
- In [Blossey et al](https://atmos.uw.edu/~durrand/pdfs/J%20Comput%20Phys%202008%20Blossey.pdf), the velocity components are derived using a discrete streamfunction. There's also some very brief discussion included in the appendix + footnotes re: ensuring discrete `div(u) = 0`.
- Previously I've had some success essentially doing as you suggest --- approximating the edge integral averages by sampling the analytical vector field along each edge using a higher-order quadrature rule. For me, this quadrature approximation has often seemed simpler than trying to derive a streamfunction.

Overall, I too, have had some questions / doubts about the way some of the numerics are structured, so very happy to discuss further. I understand what you're saying --- that the edge velocities are often thought of edge midpoint "point" values --- but I've never been able to quite square this with my understanding of finite-volumes, where at higher-orders, or for general unstructured meshes you'd expect to approximate the flux integrals at multiple quadrature points on each edge / face...

ambrad:
A few loose ends:
- Re: an earlier question, angleEdge is used only for imposed-flow tests.
- I'm still not getting the order of accuracy I expect for the nondivergent flow test with smooth field. But in the same standalone code I've implemented another method, a cell-integrated semi-Lagrangian method, that uses much of the same code and has no problem getting this test to > 2nd order. Currently it advects the cell vertices essentially exactly. But the main point is the method uses all of the same infrastructure and meshes, increasing my confidence in my standalone code. Here's example output from this method on the sequence-C 120km mesh:
ir-48-ng2-c120
Compare with an example from the standalone flux-form Eulerian method:
ff-ng3-c120

vanroekel:
@ambrad @dengwirda @mark-petersen I wanted to check in if there are any updates/final conclusions on the advection operator in MPAS.  I still think this is a pretty high priority issue.

ambrad:
@vanroekel I'll summarize my findings as follows:

1. MPAS-Ocean's Eulerian flux-form (EFF) implementation doesn't appear to have any bugs causing the problems above.
2. EFF is formally first-order accurate but on a quasiuniform mesh, the first-order term is small.
3. On the unstructured grids used in the convergence test this PR introduces, the first-order terms become dominant.
4. I have a standalone EFF implementation alongside a standalone semi-Lagrangian incremental remap (IR) implementation using a common code base, and in these implementations, IR gets second-order accuracy or better regardless of grids while EFF does not.

I can give a presentation on findings if others would like. [This confluence page](https://acme-climate.atlassian.net/wiki/spaces/COM/pages/1621396160/RFCISL) summarizes work on a new incremental remap method, and I can add convergence plots for (i) the current MPAS-Ocean transport method and (ii) the unstructured grid sequence.

ambrad:
@vanroekel I gave this problem some more thought tonight. In the convergence figure for my standalone EFF implementation I posted on confluence, the solid-body rotation curves for EFF have these characteristics:

- On a QU grid, they are each 2nd-order accurate.
- On a nonuniform grid, the 2nd-order EFF method degrades to OOA ~1, while the 3rd-order EFF degrades only a small amount below OOA 2.

This plus old runs of MPAS-Ocean with this PR's code made me think that there's something amiss in MPAS-Ocean's 3rd-order flux function. So I started visually looking over the calculation of the coefficients tonight. I believe I have found at least one bug, corresponding to a sign error in the implementation of the beta term in eq 7 of SG11, with the fix here:
```
diff --git a/src/operators/mpas_tracer_advection_helpers.F b/src/operators/mpas_tracer_advection_helpers.F                                                                                    
index f18570b..8a2e9a4 100644                                                                                                                                                                 
--- a/src/operators/mpas_tracer_advection_helpers.F                                                                                                                                           
+++ b/src/operators/mpas_tracer_advection_helpers.F                                                                                                                                           
@@ -229,14 +229,14 @@ module mpas_tracer_advection_helpers                                                                                                                                    
            k = mpas_binary_search(sorted_cell_indices, 2, 1, nAdvCellsForEdge(iEdge), indexToCellID(cell2))                                                                                  
            if(k <= nAdvCellsForEdge(iEdge)) then                                                                                                                                             
              adv_coefs(k, iEdge) = adv_coefs(k, iEdge) + deriv_two(1,2,iEdge)                                                                                                                
-             adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(1,2,iEdge)                                                                                                        
+             adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(1,2,iEdge)                                                                                                        
            end if                                                                                                                                                                            
                                                                                                                                                                                              
            do iCell = 1, nEdgesOnCell(cell2)                                                                                                                                                 
              k = mpas_binary_search(sorted_cell_indices, 2, 1, nAdvCellsForEdge(iEdge), indexToCellID(cellsOnCell(iCell,cell2)))                                                             
              if(k <= nAdvCellsForEdge(iEdge)) then                                                                                                                                           
                adv_coefs(k, iEdge) = adv_coefs(k, iEdge) + deriv_two(iCell+1, 2, iEdge)                                                                                                      
-               adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(iCell+1, 2, iEdge)                                                                                              
+               adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(iCell+1, 2, iEdge)                                                                                              
              end if                                                                                                                                                                          
            end do ! loop over iCell 
```
A minor issue I noticed is that the time step is set to 5min for all three grids in this PR. To do a proper convergence analysis, it should in general decrease by the mesh refinement factor, unless one is doing an analysis with an essential exactness in time. I switched to 30min, 15min, 7.5min, and get these results:
```
2nd-order flux
[ 10409  41402 165068]  # nbr of cells
[0.00431186 0.00190345 0.00099809]  # l2 errr
                                                                                                                                                                                              
3rd-order flux, before fix
[0.00338049 0.00181098 0.00158864]

3rd-order flux, after fix
[ 10409  41402 165068]
[0.00279333 0.00143265 0.00073552]
```
I'm still not seeing OOA almost 2 as I do in my standalone impl, but this looks like this could be progress.

ambrad:
I'm still hunting for the source of the OOA ~1 convergence; my standalone impl indicates we should see OOA ~1.75 on this grid for cosine bells rotated around the equator. But a small point worth making here is that for nonsmooth grids, cranking up beta can be useful, confirmed in both my standalone code and with MPAS-Ocean. With the latter, I see
```
[240km 120km 60km]
beta=0.25 from above
[0.00279333 0.00143265 0.00073552]
beta=1
[0.00225356 0.00130337 0.00068451]
```

ambrad:
To get closer to what my standalone impl does, I switched to RK4 (from split_explicit) and ran 1 full rotation. With the fix above, I'm seeing good behavior. Details:
```
switch to RK4
full rotation in 1 day by multiplying velocity by 24
resolutions 240km,120km, 60km
time steps   2min, 1min,30sec
use mpas_ocn_tracer_advection_std.F
order-2 flux               [0.04306733 0.0151201  0.00483401] OOA ~ 1.6
order-3 flux beta=1        [0.01030493 0.00226823 0.00050179] OOA > 2
order-3 flux beta=1/4      [0.00471197 0.00106989 0.00028895] OOA 1.9-2.1
order-4 flux               [0.0071832  0.00447862 0.00253661] OOA < 1
order-3 flux beta=1/4 mono [0.00513074 0.00109597 0.00027202] OOA > 2
order-3 flux beta=1/4, wrong sign in helper.F
                           [0.01326037 0.01270969 0.03431614]
```

vanroekel:
@ambrad Thanks this is great!  Just one question, in your latest tests (last comment) are you using the QU meshes or the newer meshes?  I'm mostly wondering if we should also put your angleEdge fixes in as well or if you think the 3rd order coefficient fixes are sufficient?

ambrad:
@vanroekel:

1. I'm using the meshes from this PR, the Jigsaw-created ones.
2. I'm using angleEdges as is. It should be fixed at some point, but it's a low priority since the error affects only these imposed-flow problems, not actual simulations. But, again, to do really good convergence testing all over the sphere, it should eventually be fixed.
3. In contrast, the sign error in the third-order flux affects any simulation for which the high-order flux is used with beta > 0.
4. Repeating a point above, I find that the third-order flux is useful to improve solution quality on nonsmooth meshes. If beta > 0 has been avoided in the past because of (correctly) perceived error, you might reexamine that choice once the sign fix goes in.

mark-petersen:
@ambrad Thanks for your thorough work on this. I am looking carefully at your proposed sign correction 
```
src/operators/mpas_tracer_advection_coefficients
Contribution from cell1: (remains the same)
218              adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(1,1,iEdge)
225              adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(iCell+1, 1, iEdge)
Contribution from cell2: (minus corrected to plus)
232              adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(1,2,iEdge)
239              adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(iCell+1, 2, iEdge)
```
right now, trying to understand the implications of it. It looks like only the ocean core calls `mpas_tracer_advection_coefficients`, so this error would not have occurred the atmosphere, sea ice, land ice, or shallow water cores. 

Indeed, the atmosphere core has its own version of initializing this variable, and I see that  it includes your negative sign:
```
src/core_atmosphere/mpas_atm_core.F
Contribution from cell1:
1198             adv_coefs_3rd(j_in,iEdge) = adv_coefs_3rd(j_in,iEdge) + deriv_two(1,1,iEdge)
1206             adv_coefs_3rd(j_in,iEdge) = adv_coefs_3rd(j_in,iEdge) + deriv_two(iCell+1,1,iEdge)
Contribution from cell2:
1217             adv_coefs_3rd(j_in,iEdge) = adv_coefs_3rd(j_in,iEdge) - deriv_two(1,2,iEdge)
1225             adv_coefs_3rd(j_in,iEdge) = adv_coefs_3rd(j_in,iEdge) - deriv_two(iCell+1,2,iEdge)
```

Thanks again. This might solve problems with grid-scale noise that we've been seeing for a long time. I'm testing with this change now in a variety of settings. 

On the topic of time stepping, RK4 should certainly be more trusted for order of convergence. In normal multi-layer ocean flows with both baroclinic and barotropic dynamics, split explicit timestepping is stable with a longer timestep, by a factor of 5 to 10. But you need to set the barotropic timestep, `config_btr_dt` appropriately, for the surface gravity wave speed.

vanroekel:
This has now been updated on to ocean/develop.  I've also think I've addressed the comments of @ambrad, but if you could take a look again to verify I'd appreciate it.  I'm testing this now.

ambrad:
@vanroekel, great, I see these things:
- `conv = abs(p[0])*2.0` to get OOA for this refinement sequence;
- `<option name="config_time_integrator">'RK4'</option>`;
- `<option name="config_run_duration">'0024_00:00:00'</option>` for one full rotation;
- 30, 15, 7.5 min time steps;
- nice changes to the RMSE script for more flexibility.

I'm not going to run these changes myself unless what you run ends up not looking as we expect: OOA ~2 with sign-fix changes.

In this comment above -- https://github.com/MPAS-Dev/MPAS-Model/pull/583#issuecomment-675862111 -- I ran with the equivalent of 48, 24, 12 min time steps, so if you end up wanting to speed up this test a bit, there's a bit of wiggle room. I believe you can even get to 60, 30, 15min, though I've tested these time steps only in my standalone tester (but with RK4). Sorry for quoting 30,15,7.5 earlier; I forgot that I pushed on these time steps a little to decrease the run time.

vanroekel:
@ambrad here is the result with current MPAS, convergence plot
convergence
Convergence is still a bit low relative to what you have above, but I'll run with your fix now

ambrad:
@vanroekel so this is with or without the sign fix?

vanroekel:
Sorry this is without the fix.  I'm running with the fix now.

vanroekel:
@ambrad here is with the fix.   OOA says 1, I wonder, am I missing a factor of 2 still?

convergenceNew

ambrad:
@vanroekel No, you have the right OOA for that plot. I'll pull the branch and see what's up.

ambrad:
@vanroekel any idea why I'm seeing build errors of this sort on this branch?
```
mpas_ocn_gm.F(271): error #6404: This name does not have a type, and must have an explicit type.   [CONFIG_REDI_USE_N2_BASED_TAPER]
      if ( config_Redi_use_N2_based_taper .and. .not. local_config_GM_kappa_lat_depth_variable ) then
```
I ran submodule sync and submodule --init --recursive; no luck.

ambrad:
@vanroekel: One possible issue I see from visual inspection (so my confidence in this is low; I'm just mentioning it while I wait for build-error suggestions): In `compute_rmse.py`, I see you're still opening `output.0001-01-01_00.00.00.nc`, but the test now runs for 24 days. So I think it's possible you're looking at convergence at 1/24 of a rotation rather than after one full rotation. My experience, documented above, is this leads to OOA 1, quite likely because the measurement isn't done in the asymptotic regime and other things (possibly bad, but let's focus on one issue at a time) are dominating the error.

ambrad:
@vanroekel, now that I got the branch to build, I did the following:

1. Modify time steps to 60,30,15min.
2. Modify output frequency to 24 hours.

Then I ran without the fix and with it, and I get the following plots:

Before fix
convergence-before-sign-fix
After fix
convergence-after-sign-fix
So this makes sense.

Next, I looked at what's going wrong in your figure. Turns out it's close to what I guessed earlier, but not quite. The key problem is this line in the rmse script:
```
    exec('rmse'+res[i]+',nCells'+res[i]+' = rmse(res[i],24)') 
```
With the new 24-day run and no modification to the write interval, the 24 is no longer correct. So you were seeing results from day 1, or 1/24 of a rotation. You can keep this line if you make modification 2 above, which is useful anyway to keep the file sizes down.

vanroekel:
ah of course, thanks for catching the silly mistake @ambrad I've made the script a bit more robust to find the time associated with the period to rotate around once (set in namelist).  I now get the same 2.12 OOA you do.  Thanks again!

vanroekel:
This has now been rebased on ocean/develop and seems to be ready for review

xylar:
Sorry @vanroekel.  The recent reorganization of COMPASS and moving some of the infrastructure to MPAS-Tools means that there are going to have to be some changes to this test case to make it compatible.  I'll put together a branch on my fork with the necessary fixes and we can decide how to bring it over to your branch.

xylar:
@vanroekel, my PR into your branch is basically in lieu of a review of the code itself: https://github.com/vanroekel/MPAS-Model/pull/3

You can cherry-pick my commits, reset o my branch, or handle that PR however you like.

After bringing in my changes, this branch needs to be rebased onto `ocean/develop` again to bring in recent changes.  I tried to test on Grizzly but got stuck in the queue.  I'll check back later.

xylar (approved):
This has been a really amazing test that has taught us a lot!  Thanks @vanroekel and everyone else involved in getting it working and debugging the problems it exposed!

With the fixes I put in related to the COMPASS reorg, it worked great for me.  Attached is the convergence plot I got.
convergence

xylar:
I am seeing "errors" in the forward runs:
```
ERROR:  Warning: Sea surface height is outside of acceptable physical range, i.e. abs(sum(h)-bottomDepth)>20m.
```
It's probably worth setting:
```
&debug
...
    config_check_ssh_consistency = .false.
```
to disable this check.  Otherwise, folks might get confused by the fact that there are error logs.

xylar:
@vanroekel, the merge commit from my PR might get removed in a rebase, which is fine.  If it prevents you from rebasing, just cherry-pick the other commits after a rebase.

vanroekel:
@xylar just rebased on ocean/develop it went smoothly and it looks like your commits were preserved.

xylar:
But the merge commit was removed, as I suspected.

xylar:
@vanroekel, are you going to add the modification to set `config_check_ssh_consistency = .false.` or should I do that?

vanroekel:
@xylar working on adding this now.  Will push the changes shortly

mark-petersen:
@xylar are you done with your recent changes? If so, I'll test and prep it for the next E3SM slot. This has Fortran + COMPASS changes, so I didn't want to do it with today's batch.

xylar:
Yes, I’m done. The last commit takes care of my only remaining concern.

mark-petersen:
Rebased, compiled and tested with gnu and intel debug for initial condition. Ran QU240 test through forward.

mark-petersen (approved):
This looks great! Thanks for your work @vanroekel. I cleaned up the COMPASS files so they are linked to templates rather than a bunch of individual files. Then it was very easy to add more resolutions in between. This still spans QU60 to QU240:
convergence
This is beautiful! The extra test cases give us more confidence in the slope.

I also was able to clean up the simulations. It had been writing output and restart files very frequently. It now can run all seven cases in 12 minutes on 4 nodes on grizzly with gnu.

mark-petersen:
@jonbob This one is ready to merge when we have an open E3SM slot. It changes both COMPASS and init mode Fortran code.

xylar:
I have a draft PR for moving the COMPASS changes from this PR to the new COMPASS repo: https://github.com/MPAS-Dev/compass/pull/8

Once https://github.com/MPAS-Dev/compass/pull/2 has been merged, we need to:

- [ ] rebase https://github.com/MPAS-Dev/compass/pull/8
- [ ] remove COMPASS changes from this PR
- [ ] Re-test within the new COMPASS branch, pointing to a built of MPAS-Ocean from this PR

xylar:
> remove COMPASS changes from this PR

@vanroekel, would you like to remove the COMPASS changes from this PR, or would you prefer I do it?  Here is a link that I **think** describes the approach we would use: https://stackoverflow.com/a/17824718/7728169

xylar:
@vanroekel, I went ahead and tried the rebase.  The link I posted above wasn't much help, I just needed to keep deleting modified contents in `testing_and_setup/compass`.  Here's my version of the branch, which just has 2 commits:
https://github.com/xylar/MPAS-Model/tree/ocean/addCosineBellTest
I think the second commit could be squashed with the first.

vanroekel:
@xylar thanks for taking the time to pull the COMPASS changes out of that branch.  It looks good to me.

xylar:
@vanroekel, I'm going to take that as tacit permission to alter your branch to update this PR.

xylar:
@mark-petersen, please merge this when it is convenient for E3SM.  Then, we can update the `ocean/develop` submodule in https://github.com/MPAS-Dev/compass/pull/8 and merge that PR.

mark-petersen:
@xylar, thanks for rebasing. Hopefully we can continue to merge these peripheral PRs during the mushy freeze.

mark-petersen:
Rebased. Corrected author and test case names. Tested successfully using COMPASS case, which is here: https://github.com/MPAS-Dev/compass/pull/8, with gnu optimized and debug.

mark-petersen:
Rebased version passes nightly regression suite with gnu debug. Ran convergence test here: https://github.com/MPAS-Dev/compass/pull/8 with gnu debug an optimized, produces same plot as above. Takes 12 minutes on grizzly with 4 nodes.

Issue #525, opened 2020-04-18
Errors in CUSP 12 spin-up


xylar:
CUSP8 and CUSP12 stage `spin_up/test_final_settings` is misconfigured.:
```
config_run_duration = '00-01-00_00:00:00'
```
But the last restart file is:
```
restart.0001-01-07_00.00.00.nc
```
This is not compatible with how MPAS handles restarts: there would need to be a restart file at an even month boundary for this to work.

xylar:
I'm going to see if similar issues emerge elsewhere and then fix this in #523

xylar:
Update: this is really beyond #523 and needs someone to figure out what is actually useful for the CUSP spin-up.  For example, the `timeSeriesMonthlyOutput` analysis is turned on but the output is also not being produced.  And monthly output is only going to be useful if this is running for a second month, following a month of spin-up.  @mark-petersen, I'm going to assign this to you but please let me know who "owns" this test case and should fix this.

xylar:
I suspect CUSP8 will have the same issues, since the various `prep_spin_up` stages have been modified in #445 but the final stage of that test case was also not updated.

mark-petersen:
Thanks for pointing this out. I will make modifications in https://github.com/MPAS-Dev/MPAS-Model/pull/523.

xylar:
Renaming since CUSP 8 has been removed.

mark-petersen:
This now applies to `WC12` and `WC14`. I double checked, and in both the run duration is divisible by the restart frequency. For example, here
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/ocean/global_ocean/WC14/spin_up/config_test_final_settings.xml#L14
and here
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/ocean/global_ocean/WC14/spin_up/config_test_final_settings.xml#L32

PR #584, opened 2020-06-01
New Mesh: WC14to60kmL60E3SMv2r02

mark-petersen:
Design discussion Water Cycle mesh, 14km high resolution region.


testing_and_setup/compass/ocean/global_ocean/WC14/init/define_base_mesh.py
```diff
@@ -0,0 +1,190 @@
...
+    EC60to30 = mdt.EC_CellWidthVsLat(lat)
+    EC60to30Narrow = mdt.EC_CellWidthVsLat(lat, latPosEq = 8.0, latWidthEq = 3.0)
+
+    # Expand from 1D to 2D
+    _, cellWidth = np.meshgrid(lon, EC60to30Narrow)
+    plot_cartopy(2, 'narrow EC60to30', cellWidth, '3Wbgy5')
+    plotFrame = 3
+
+    # global settings for regionally refines mesh
+    highRes = 14.0 #[km]
```
mark-petersen, line 58: This design is currently identical to the 12km WC12to60kmL60E3SMv2r01 in #555, except that the high resolution is set to 14km here.

testing_and_setup/compass/ocean/global_ocean/WC14/init/config_E3SM_coupling_files.ini
```diff
@@ -0,0 +1,73 @@
...
+long_name = ${prefix}${min_res}to${max_res}kmL${levels}E3SMv${e3sm_version}r${mesh_version}
+prefix = CUSP
+description = MPAS Contiguous United States Plus mesh for E3SM version
+              ${e3sm_version}, with a focused ${min_res}-km resolution
+              around North America and ${levels} vertical levels
+e3sm_version = 2
+mesh_version = 01
+creation_date = autodetect
+min_res = 12
+max_res = 60
```
xylar, line 21: This metadata needs to be updated before this PR gets merged.

mark-petersen: Updated. Thanks for pointing it out.

milenaveneziani:
All, I was thinking: since the mesh creation process is BFB for a specific machine (right?), would it make sense to add an entry in the metadata, something like `creation_machine`, or something like that?

proteanplanet:
@milenaveneziani 

> All, I was thinking: since the mesh creation process is BFB for a specific machine (right?), would it make sense to add an entry in the metadata, something like `creation_machine`, or something like that?

I like this idea.

vanroekel:
All as mentioned in #555 I would like to move this forward and focus our discussion / acceptance on the regions of high resolution only.  @proteanplanet, @milenaveneziani, @maltrud, @lconlon, and @qingli411 please offer your thoughts on whether these regions are acceptable or if you'd like changes.  I would like to move toward testing a first revision by the weekend.

proteanplanet:
I vote to move ahead with WC14. I'll try to slip in the Victoria Island GeoJSON by the weekend, but we should proceed regardless since that's not going to change much.  Aside from that, I think we should proceed with all other passages and seas as they are.

xylar:
> I'll try to slip in the Victoria Island GeoJSON by the weekend

@proteanplanet, I don't see a feasible way of getting that transect into the mesh in time for a spin-up starting this weekend.  Even if you create the geojson file, it would either need to be put in "properly" through geometric_features, requiring a new release of that software and an updated compass environment with that new version of geometric_features or it would need to be included in this mesh as a one-off using https://github.com/MPAS-Dev/MPAS-Model/pull/586, which hasn't been reviewed or tested yet by anyone but me.

So my vote is to hold off on transects and include all modifications together in the next iteration of the mesh.

vanroekel:
@proteanplanet I agree with @xylar here.  I would like to approve and merge this mesh today based on the regions of high resolution only.  This will definitely not be the final e3sm WC mesh and we will revisit transects, passages, islands, in a later revision.  Please approve based on the resolution distribution of the mesh alone today.  Also cc'ing @milenaveneziani for her approval.

milenaveneziani:
Resolution looks good to me.

vanroekel:
Thanks all!  I'm going to grab this and do the MPAS spin up shortly.  @pwolfram  and @mark-petersen do you want to push a fix for the mesh gradations noted by @pwolfram in #555?  I'm also okay waiting for that for the next revision.  I doubt it will make much impact on the large scale flow.

xylar:
@vanroekel, wait just a bit.  I pushed the resolution around Norway further north east and I also realized that my fix of the Bering Sea region only went into the 12 km mesh, not the WC14 version.  (I didn't realize @mark-petersen made a copy -- I don't think this is what we want to do in the future.)

vanroekel:
@xylar no problem.  Let me know when ready.

xylar:
The latest images showing resolution and mesh construction steps:
cellWidthGlobal
mesh_construction

xylar:
@vanroekel, hold off.  The problem with the international dateline is still there.  I can see it in my plot.

xylar:
Once again, the latest images showing resolution and mesh construction steps:
cellWidthGlobal
mesh_construction

xylar:
## Some lessons for COMPASS meshes

@mark-petersen and others who will be working on meshes with geojson regions in the future, here are a few point to keep in mind:
* The geojson files need to have points that are strictly between -180 and 180 longitude and -90 and 90 latitude.
* Shapes that cross the international date line must have exactly the same points at -180 and at +180 longitude to ensure they are periodic
* A given "feature collection" from `geometric_features` can't have two features with the same name.  So the "name" property really should contain something unique and meaningful, not just "region for CUSP mesh".  A shape that crosses the international date line will need an "east" and "west" piece with unique names.
* A transition in resolution must either be smooth or take place on land.  The transition at the Korean Peninsula (once I fixed the Kamchatka region, which was past -180) was not smooth in the ocean and will need to be revisited.

xylar:
@vanroekel, I am finally done for now.  I don't think a re-review by any of the others should be needed because the changes I've made were either minor or related to the lack of periodicity at the date line.

vanroekel:
thanks @xylar I'll take this and move toward E3SM testing.

Related to your points above, do you have any suggestions on where we could post those lessons so we don't have to remember they are in this PR later?

xylar:
> Related to your points above, do you have any suggestions on where we could post those lessons so we don't have to remember they are in this PR later?

Once #472 gets merged to `ocean/develop`, I'll be putting together COMPASS documentation.  That's the proper place.  Until then, I'm just trying to make people aware who happen to be paying attention.

xylar:
@vanroekel and @pwolfram, just FYI I'm still working on getting the Korean Peninsula to look "right" but that will definitely be something for the next iteration.

vanroekel:
Thanks @xylar, that makes perfect sense.  I'm starting the spin up of this revision now.

xylar:
I was able to get the Kamchatka Peninsula to transition more abruptly than beofre:
cellWidthGlobal

@vanroekel, I'm not going to push any new commits to this branch unless you give the go-ahead.  So let me know if/when you would like these changes.

vanroekel:
@xylar if it is easy for you to push now (I know it's getting very late for you) I can pull it in.  I was just about to start the spin up

vanroekel:
But I should say I'm fine waiting for the next revision too.  I'm happy to go with what's easier for you

xylar:
It's easy to push.  Give me a couple minutes to commit.

xylar:
@vanroekel, okay, I'm done and calling it a night.  Thanks for letting me sneak this last bit in.

vanroekel:
Thanks to you as well for the hard work @xylar!  I'm going to run the spin up then merge this we can then move deeper analysis to an r03 PR to be opened.

testing_and_setup/compass/ocean/global_ocean/WC14/spin_up/config_test_final_settings.xml
```diff
@@ -0,0 +1,46 @@
...
+	<add_link source="../../init/initial_state/graph.info" dest="graph.info"/>
+	<add_link source="../../init/initial_state/init_mode_forcing_data.nc" dest="forcing_data.nc"/>
+	<add_link source="../spin_up1/Restart_timestamp" dest="Restart_timestamp"/>
+	<add_link source="../spin_up1/restarts" dest="restarts"/>
+
+	<add_executable source="model" dest="ocean_model"/>
+
+	<namelist name="namelist.ocean" mode="forward">
+		<template file="template_forward.xml" path_base="script_configuration_dir"/>
+		<template file="template_monthly_output.xml" path_base="script_configuration_dir"/>
```
vanroekel, line 13: @xylar and @mark-petersen this is causing a failure at this stage of the spin up because this line wasn't included in streams below as well, but even doing that ```template_monthly_output.xml``` did not include ```timeSeriesStatsMonthlyRestart```.  I can make these changes, but my question is if we should do this here, since the xml template doesn't appear to be part of this PR.

xylar: @vanroekel, this was part of the reason for an issue I posted in April: https://github.com/MPAS-Dev/MPAS-Model/issues/525

In my view, any changes you have to make to `config_final_settings.xml` are fair game for this pull request and hopefully would let us close #525.  I would have like to have this addressed in a separate PR but that obviously didn't happen and now is the time.  Sorry for the mess!

xylar: My suggestion would be to remove `timeSeriesStatsMonthlyOutput` because:
* it doesn't work well unless you run for a full month
* it doesn't work well if you don't start at the beginning of a month
* spin-up isn't always compatible with running for exactly 1 month prior the `test_final_settings` stage.

vanroekel: I agree I would prefer to just disable ```timeSeriesMonthly```. I will make the change.

PR #586, opened 2020-06-02
Add support for custom critical passages and land blockages

xylar:
This is added to support potential addition of "custom" passages or blockages in #555 and #584.  The custom transects could come from a local file or it could be pulled in from `geometric_features`, e.g. with the `merge_features` command.

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/scripts/cull_mesh.py
+++ b/testing_and_setup/compass/ocean/global_ocean/scripts/cull_mesh.py
@@ -24,11 +24,11 @@ to be opened. Otherwise, steps 2, 5 and 9 are skipped
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
-import os
 from optparse import OptionParser
 import xarray
 
-from geometric_features import GeometricFeatures
+from geometric_features import GeometricFeatures, FeatureCollection, \
+    read_feature_collection
 from mpas_tools import conversion
 from mpas_tools.io import write_netcdf
 from mpas_tools.ocean.coastline_alteration import widen_transect_edge_masks, \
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I tested this by adding a `transect.geojson` file that put a canal through the US.  I modified `config_culled_mesh.xml` in the `QU240/init` test case to use this new file as a custom critical passage and the result is a mesh with a horrible canal through the US (as expected):

US canal

xylar:
@pwolfram, I simplified the logic of `cull_mesh.py` a bit to accommodate my changes here.  This included a modification to code involving preserving floodplain.  I don't anticipate any problems but it wouldn't hurt to have you run a test case of yours that preserves the floodplain using this branch and let me know if you have any trouble.

testing_and_setup/compass/ocean/global_ocean/scripts/cull_mesh.py
```diff
@@ -37,16 +37,37 @@
...
+                       "the default passages (--with_critical_passages)")
+parser.add_option("--custom_land_blockages", dest="custom_land_blockages",
+                  help="A geojson file with critical land blockages to close. "
+                       "This file may be supplied in addition to or instead of "
+                       "the default blockages (--with_critical_passages)")
 parser.add_option("--preserve_floodplain", action="store_true",
-                  dest="preserve_floodplain", default=False)
+                  dest="preserve_floodplain", default=False,
+                  help="Whether to to use the cellSeedMask field in the base "
+                       "mesh to preserve a floodplain at elevations above z=0")
```
xylar, line 59: @pwolfram, can you tell me if this is correct?  I think it is based on my recent work on `inject_preserve_floodpalin`.

mark-petersen: @pwolfram any comment? It is just a help text, so I may merge to get this in.

PR #587, opened 2020-06-03
Fixes bugs in tracer_advection_std

vanroekel:
Currently there are calls in tracer_advection_std to scratch pool arrays
that have been removed from registry.  This commit makes std advection
functional again.


vanroekel:
also tagging @jamilgafur as this might be good to test in your horizontal advection test.  Merge this in and set ```config_monotonic = .false.```

mark-petersen:
@vanroekel thanks for posting. @mattdturner removed all the scratch `var_struct` variables in #457 and replaced them with local old-fashioned allocatable arrays. That PR is on schedule to merge next week. I think that will take care of this problem, but we will make sure to test the `tracer_advection_std` routine, and leave this PR open until that is done.

PR #588, opened 2020-06-03
Added Parallel processing support with prereq for manage_regression_s…

jamilgafur:
…uite

Incorporated two arguments for parallel support -lp (local_parallel) and -N (Nodes) to
allow for parallel stream processing of regression suites. Default of N is 1 (one); first
processes all tests labeled prereq and then processed all other tests.


jamilgafur:
@vanroekel

vanroekel:
@jamilgafur there are still a bunch of merge conflicts.  Could you look through and see if you have any signature of the conflicts?  e.g. grep for HEAD again in the files listed here?

vanroekel:
@xylar or @mark-petersen -- @jamilgafur and I did a rebase onto ocean/develop, but the commit history is still really wild.  Do you have any git suggestions on what we should do?

xylar:
@jamilgafur and @vanroekel, it doesn't seem like the rebase was done correctly.  There are hundreds of files modified in the one commit you have.

Also, it seems you are rebasing onto @jamilgafur's version of ocean/develop, not MPAS-Dev/MPAS-Model/ocean/develop, but that's a minor thing.

Also also, I think you may be modifying files directly in the `testing_and_setup/compass` directory, in which case they need to be merged into `develop`, not `ocean_develop`.

My suggestion would be to copy the files you actually meant to modify to a temporary location (or tar them up or something like that), then reset to MPAS-Dev/MPAS-Model/ocean/develop, then copy back or untar the files, then redo the commit.

vanroekel:
Thanks for the suggestion @xylar we'll close this PR and try your suggestion (including a PR into develop instead).

jamilgafur:
I will fail miserably or it will work fine. Either way, I am on it!

xylar:
@vanroekel and @jamilgafur, if there are files in the `testing_and_setup/ocean` directory that have changed, they need to go into `ocean/develop`.  I think that's not the case but wanted to make sure.  Is #593 the new PR?  If so, we're good.

vanroekel:
yes #593 is the replacement to this PR.

Issue #500, opened 2020-04-01
Automatic population of submodules


pwolfram:
As part of the build process, I would propose we automatically initialize submodules, e.g., for CVmix to not require an explicit call of the form: `git submodule update --init --recursive` in `ocean\develop`.  This will enable similar behavior as to what we had previously and make builds simpler.

philipwjones:
One of the reasons we had the earlier hand-rolled scripts was to avoid any unnecessary git updates in the middle of the build since it can require a password entry, either for git itself or on some machines, to access the ssh key file. That will stall the build if you're running in the background or walk away for a bit and the prompt can get buried in all the build output.

The current build does something similar to the old scripts. There is a check to see if the update has happened and prints a pretty informative error msg if there has been no download. You could always insert the git command at that point instead of printing an error message. But again, it requires some attentiveness during the build to notice a password prompt if needed or in the case of git failure to understand what happened.

Don't have a strong opinion either way. For E3SM builds, you're still going to need the explicit git command since there are many other submodules scattered about.

pwolfram:
Also, when you are developing you'll get a meaningless git status notation of 
`modified:   src/core_ocean/cvmix (untracked content)`
which is completely unnecessary because no editing of note was done to cvmix.

Minimally, we should modify `.gitignore` so we don't have this problem because cvmix is its own repo.

xylar:
@pwolfram, I have no objection to you making these changes.

xylar:
To address @philipwjones' comment about having to enter passwords, it would be best if the `git submodule` call happens immediately at the beginning of the build and only after checking if it is really necessary.

pwolfram:
@philipwjones, am I understanding that the key thing here is to have a clean crash as opposed to a hang when building automatically to prevent issues with needing a password, etc?  This would be for testing, production systems, etc.

If there isn't a clean way too work around this in software, which isn't currently obvious to me if it exists, I can just close the issue.

philipwjones:
@pwolfram Yes, that's correct. Even if there's away for it to time out and bail, that would be preferable to an indefinite hang. I'm still open to solutions to make this work, but I don't currently have any good ideas either.

philipwjones:
And we shouldn't forget your other point - we need to add the right info to gitignore for cvmix. Still reporting as untracked...

pwolfram:
[...]

PR #591, opened 2020-06-03
Updates build process by automating submodules

pwolfram:
Automatically downloads submodule code bases and has an auto-timeout to prevent a hang.

### DIFF
--- a/src/core_ocean/Makefile
+++ b/src/core_ocean/Makefile
@@ -44,8 +44,16 @@ post_build:
 	( cp $(ROOT_DIR)/default_inputs/streams.ocean.forward $(ROOT_DIR)/streams.ocean.forward )
 	( cp $(ROOT_DIR)/default_inputs/streams.ocean.analysis $(ROOT_DIR)/streams.ocean.analysis )
 	( cp $(ROOT_DIR)/default_inputs/streams.ocean.init $(ROOT_DIR)/streams.ocean.init )
+# timeout switch to prevent build from hanging
+GITTIMEOUTSEC=60
 
 libcvmix:
+	if [ -e cvmix/.git ]; then \
+		(echo "Found cvmix repo") \
+	else \
+		(echo "Attempting retrieval of cvmix repo"; cd ../../; timeout $(GITTIMEOUTSEC) git submodule update --init --recursive) \
+	fi
+# recheck success of downloaded repo
 	if [ -e cvmix/.git ]; then \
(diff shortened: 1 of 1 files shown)


pwolfram:
Will close #500

philipwjones:
Unfortunately, this didn't work as expected in the failure mode. Tried on NERSC where I usually get a pwd prompt. First it suppressed the pwd prompt itself, so it's not clear that it's expecting a pwd entry for the key. Then after timing out, it just errors out with no message and doesn't proceed to the useful next step where it would have output the usual submodule error message.   

When it was successful (eg on Summit), it worked fine. So it's just the failure mode that fails.

pwolfram:
@philipwjones, thanks for testing!  Not sure I completely understand how you want it to behave.  The goals I'm hearing need to be accomplished are:

 - [ ]  pwd prompt passes through 
 - [ ] better error message following time-out (including prior submodule error message)
 - [ ] verify failure mode capabilities (above) on NERSC

Is this consistent with what you'd expect or desire it to do?

xylar:
> verify failure mode capabilities (above) on NERSC

@pwolfram, my guess is that @philipwjones just happens to have his GitHub account configured on NERSC in a way that it prompts for a password, not that this is an issue specific to NERSC.  That is, you testing on NERSC might not reveal this issue if your GitHub account isn't configured to prompt for a password there.

Other than that, it looks like your check list is right.  My request in addition to that from https://github.com/MPAS-Dev/MPAS-Model/issues/500#issuecomment-607764771 still holds.  It seems really important to me that the make process triggers the calls to `git submodule` within the first few seconds of the `make` process.  As a user, I typically start a make and go do something else while it builds.  If I were to have to come back later, enter a password, and go off again before the build finished, that would be really annoying (just as annoying as the current process).

philipwjones:
@pwolfram  Yes the checklist is correct. A couple of quick points. The password prompt is not a github password but the password to unlock the ssh key file. Second, the appropriate error message would get triggered if the timeout didn’t error, but control just passed on to the next if block. 

@xylar ’s request is likely not possible unless you created an artificial dependence in the top level makefile. All the framework and other primary dependencies will always take precedence before it can get to this core-level makefile.

xylar:
> @xylar ’s request is likely not possible unless you created an artificial dependence in the top level makefile. All the framework and other primary dependencies will always take precedence before it can get to this core-level makefile.

@philipwjones, okay, thanks for letting me know.  I'll let that request go, since that sounds like too much trouble.

pwolfram:
What is everyones' thoughts on the value of this PR?  If the consensus is it is of limited value I will close.  I'm thinking it may not be worth everyones' trouble.

PR #592, opened 2020-06-03
Adds support for lib vs lib64

pwolfram:
This generalizes library searches to include both `lib` and `lib64` for the `netCDF` and `parallel-netCDF` libraries where `lib64` takes precedence.   This is needed to ensure proper compilation when using the `PIO2` library.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -525,17 +525,23 @@ ifneq ($(wildcard $(PIO_LIB)/libgptl\.*), )
 endif
 
 ifneq "$(NETCDF)" ""
+ifneq ($(wildcard $(NETCDF)/lib), )
+	NETCDFLIBLOC = lib
+endif
+ifneq ($(wildcard $(NETCDF)/lib64), )
+	NETCDFLIBLOC = lib64
+endif
 	CPPINCLUDES += -I$(NETCDF)/include
 	FCINCLUDES += -I$(NETCDF)/include
-	LIBS += -L$(NETCDF)/lib
+	LIBS += -L$(NETCDF)/$(NETCDFLIBLOC)
 	NCLIB = -lnetcdf
 	NCLIBF = -lnetcdff
-	ifneq ($(wildcard $(NETCDF)/lib/libnetcdff.*), ) # CHECK FOR NETCDF4
+	ifneq ($(wildcard $(NETCDF)/$(NETCDFLIBLOC)/libnetcdff.*), ) # CHECK FOR NETCDF4
 		LIBS += $(NCLIBF)
 	endif # CHECK FOR NETCDF4
 	ifneq "$(NETCDFF)" ""
 		FCINCLUDES += -I$(NETCDFF)/include
-		LIBS += -L$(NETCDFF)/lib
+		LIBS += -L$(NETCDFF)/$(NETCDFLIBLOC)
 		LIBS += $(NCLIBF)
 	endif
 	LIBS += $(NCLIB)
@@ -543,9 +549,15 @@ endif
 
 
 ifneq "$(PNETCDF)" ""
+ifneq ($(wildcard $(PNETCDF)/lib), )
+	PNETCDFLIBLOC = lib
+endif
+ifneq ($(wildcard $(PNETCDF)/lib64), )
+	PNETCDFLIBLOC = lib64
+endif
 	CPPINCLUDES += -I$(PNETCDF)/include
 	FCINCLUDES += -I$(PNETCDF)/include
-	LIBS += -L$(PNETCDF)/lib -lpnetcdf
+	LIBS += -L$(PNETCDF)/$(PNETCDFLIBLOC) -lpnetcdf
 endif
 
 RM = rm -f


pwolfram:
Note, sets `"USE_PIO2=true"` for this case as target option.

cc @xylar

vanroekel:
@pwolfram it seems worrisome to have only one configuration with USE_PIO2=true, it may cause unexpected issues for other users.  We should make sure the broader team is on board with that change.

vanroekel:
looking closer it seems you are adding a new configuration right?  Then I'm fine with the PIO2 change.  Disregard the previous comment

pwolfram:
Yes, @vanroekel, only specifying it for that specific build, not general for everyone.

pwolfram:
ps @vanroekel , please leave a review.

vanroekel:
sure @pwolfram will test this stack in the advection convergence test case and chime in after

mgduda:
The new `intel-mpi` target appears identical to the existing `ifort` target in all but two respects:
1. The names of the MPI compiler wrappers are Intel-specific in the former; and
1. the `USE_PIO2` option is hard-wired to `true` in the former.

Regarding the hard-wiring of `USE_PIO2=true`, I don't really like the idea of having any target hard-wire any of the [options listed in the Makefile](https://github.com/MPAS-Dev/MPAS-Model/blob/v7.0/Makefile#L882-L894), as it defeats the purpose of options and makes the target less general than it needs to be. It also doesn't seem that difficult to ask users to specify `USE_PIO2=true` when compiling on a particular machine; we have decent checks in the `Makefile` that should even tell users if they need to add the `USE_PIO2=true` option.

Regarding the Intel-specific compiler wrappers, my general concern is that we're setting a precedent that will support the proliferation of build targets to support every possible variation in build environment. I'm open to arguments in favor of the `intel-mpi` target, but it's probably worth asking whether the modules on grizzly could be set up to let `mpif90` be a symlink or alias for `mpiifort`, and similarly for `mpicc` and `mpiicc`? Another option to consider is whether the Intel MPI implementation provides a Fortran wrapper named `mpifort` (which, I understand, is the more modern version of `mpif90`), and whether we could simply update the existing `ifort` target to use `mpifort` rather than `mpif90`?

vanroekel:
@pwolfram note that following the comment from @mgduda I tried building on grizzly with your stack and with 
```
make ifort CORE=ocean USE_PIO2=true
```
This did fail.  However, if I did
```
export I_MPI_F90=ifort
```
first then the model built fine without the new configuration.  I would suggest we remove intel-mpi and just add the export line to our individual scripts to load modules on grizzly.

vanroekel:
note the solution for building was found here -- https://stackoverflow.com/questions/32335519/gfortran-is-called-instead-of-mpif90

pwolfram:
The general version of it from https://software.intel.com/en-us/forums/intel-clusters-and-hpc-technology/topic/288354 is
```
Note that if you want to use mpif90/mpicc/etc (eg for existing configure scripts/Makefiles, etc) with the Intel compilers, one need only modify environment variables to point to the appropriate compilers: eg,
export I_MPI_CC=icc
 export I_MPI_CXX=icpc
 export I_MPI_F77=ifort
 export I_MPI_F90=ifort
for bash-type shells, or
setenv I_MPI_CC icc
setenv I_MPI_CXX icpc
setenv I_MPI_F77 ifort
setenv I_MPI_F90 ifort
for csh-type shells.
```

pwolfram:
In sum, consensus is I should remove the `intel-mpi` target (done).  I've tested this works by using these load-up settings on grizzly:

```
module purge                                                                                                                          
module load friendly-testing                                                                                                          
module load intel/19.0.4 intel-mpi/2019.4 hdf5-parallel/1.8.16 pnetcdf/1.11.2 netcdf-h5parallel/4.7.3 mkl/2019.0.4 scorpio/pio2/1.10.1
export I_MPI_CC=icc                                                                                                                   
export I_MPI_CXX=icpc                                                                                                                 
export I_MPI_F77=ifort                                                                                                                
export I_MPI_F90=ifort                                                                                                                
```

and build with 

`make ifort CORE=ocean USE_PIO2=true`

pwolfram:
Thanks @xylar!  Everyone, changes made and PR ready to move forward with other review acceptances.

pwolfram:
@mgduda, removed 

```
echo "loading modules or grizzly"
module purge 
# assumes python2, so this doesn't work                                                                                                                         
#source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_latest_compass.sh                                                      
module load friendly-testing                                                                                                          
module load intel/19.0.4 intel-mpi/2019.4 hdf5-parallel/1.8.16 pnetcdf/1.11.2 netcdf-h5parallel/4.7.3 mkl/2019.0.4 scorpio/pio2/1.10.1
export I_MPI_CC=icc                                                                                                                   
export I_MPI_CXX=icpc                                                                                                                 
export I_MPI_F77=ifort                                                                                                                
export I_MPI_F90=ifort 
```

to build with

`make ifort CORE=ocean USE_PIO2=true`

from PR descriptions and commit messages as requested.

mgduda:
@pwolfram Thanks for the updates to the PR description. It's not clear to me how the changes are related to `PIO2`, which is specifically mentioned in the PR description. It looks to me like the changes affect the netCDF and parallel-netCDF libraries without regard to which version of PIO is used. Or is it that only PIO2 can use netCDF libraries in `lib64`? It might also make the PR description more complete to mention that the `lib` versus `lib64` affects just the netCDF and parallel-netCDF libraries, and to point out which of `lib` or `lib64` will take precedence if both were to somehow be found. Of course, most people would have no trouble figuring this out by looking at the code diffs, but it seems not a bad practice to be thorough in descriptions of code changes, especially in terms of the _why_ (the _what_ is usually apparent from the diffs, but the motivation isn't always so clear).

pwolfram:
Thanks @mgduda, please see updated descriptions following your feedback.

mark-petersen:
@pwolfram thanks for your work on this.

PR #598, opened 2020-06-05
fixed bad array sizes in local ocean arrays

### LINKED ISSUE #597: local ocean allocatable arrays are wrong size
In some of the recently introduced ocean changes, local allocatable arrays were only allocated up to nCellsAll (formerly nCellsSolve). However, adjacency indices in cellsOnCell, etc. use nCellsAll+1 for non-existent neighbors. This results in out-of-bounds indexing and these arrays need to be resized.

So far, this is seen in the new ocean mesh module and the ocn_tracer_advection_mono module.

philipwjones:
Two ocean modules had local arrays that were not properly sized to accommodate the practice of using nCellsAll+1 (Edges, Vertices) as a location to represent non-existent cells, edges, vertices. This would result in out-of-bounds references. This fix just extends the dimension by one for the arrays that will potentially be  referenced in this way.

Fixes #597 

[b4b] in QU240 test on Summit - meaning out of bounds refs so far had fortunately no impact

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_mesh.F
+++ b/src/core_ocean/shared/mpas_ocn_mesh.F
@@ -459,15 +459,15 @@ contains
                                   boundaryCellTmp)
 
          allocate ( &
-            edgeMask(nVertLevels, nEdgesAll), &
-            cellMask(nVertLevels, nCellsAll), &
-            vertexMask(nVertLevels, nVerticesAll), &
-            boundaryEdge(nVertLevels, nEdgesAll), &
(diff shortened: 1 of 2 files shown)


philipwjones:
Sorry guys, got a bit sloppy.  When I fixed array sizes in the mesh module, I forgot to change subsequent initialization loop limits. Just pushed a fix (well, actually two commits because I missed a second set of loops in the update routine). Still b4b, and actually works in the branch where I triggered this bug.

mark-petersen:
Tested with gnu and intel, optimized and debug, is bfb with previous.

PR #599, opened 2020-06-08
Add reference cell width to del2 and del4 scaling

mark-petersen:
Currently, the values of horizontal viscosity and tracer diffusion are referenced to the smallest grid cell in the domain. This means that their values must be adjusted for every new resolution. 

This PR adds the flags
```
config_hmix_use_ref_cell_width (default false)
config_hmix_ref_cell_width
```
If `config_hmix_use_ref_cell_width` is false, there is no change, so we have backwards compatibility for MPAS and E3SM simulations. If `config_hmix_use_ref_cell_width` is true, then the values of
```
config_mom_del2
config_mom_del4
config_tracer_del2
config_tracer_del4
```
are now referenced to the length `config_hmix_ref_cell_width`, where del2 values are proportional to the cell width, and del4 values are proportional to cell width^3. With this new capability, the del2 and del4 flags will not need to change, except for fine tuning.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -204,7 +204,7 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_hmix_ref_cell_width" type="real" default_value="30.0e3" units="m"
-					description="Reference cell width. If config_hmix_use_ref_cell_width = .true., then hmix coefficients are set to be (for example) nu_2 = config_mom_del2*(cellWidth / config_hmix_use_ref_cell_width) where cellWidth is the effective cell width computed as 2*sqrt(areaCell/pi)"
+					description="Reference cell width. If config_hmix_use_ref_cell_width = .true., then hmix coefficients are set to be nu_{2h} = config_mom_del2*(cellWidth / config_hmix_use_ref_cell_width) and nu_{4h} = config_mom_del4*(cellWidth / config_hmix_use_ref_cell_width)^3 where cellWidth is the effective cell width computed as 2*sqrt(areaCell/pi). See Hoch et al 2020 JAMES eq 1,2. This relation applies within a simulation, but also generally among multiple simulations, so the parameters config_mom_del2, config_mom_del4, and config_hmix_use_ref_cell_width can generally remain at their standard values, and just be adjusted for fine tuning."
 					possible_values="Any positive real number, but typically a resolution number such as 30km."
 		/>
 		<nml_option name="config_apvm_scale_factor" type="real" default_value="0.0" units="unitless"
@@ -213,12 +213,12 @@
 		/>
 	</nml_record>
 	<nml_record name="hmix_del2" mode="forward">
-		<nml_option name="config_use_mom_del2" type="logical" default_value=".true." units="unitless"
+		<nml_option name="config_use_mom_del2" type="logical" default_value=".false." units="unitless"
 					description="If true, Laplacian horizontal mixing is used on the momentum equation."
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_mom_del2" type="real" default_value="1.0e3" units="m^2 s^{-1}"
-					description="Horizonal viscosity, $\nu_{2h}$. If config_hmix_use_ref_cell_width = .true. then $\nu_h$ = config_mom_del2*(cellWidth / config_hmix_use_ref_cell_width). If config_hmix_use_ref_cell_width = .false. then it is referenced to the smallest cell."
+					description="Horizontal viscosity, $\nu_{2h}$. If config_hmix_use_ref_cell_width = .true. then $\nu_h$ = config_mom_del2*(cellWidth / config_hmix_use_ref_cell_width). If config_hmix_use_ref_cell_width = .false. then it is referenced to the smallest cell."
 					possible_values="any positive real"
 		/>
 		<nml_option name="config_use_tracer_del2" type="logical" default_value=".false." units="unitless"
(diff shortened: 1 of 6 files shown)


mark-petersen:
Passed nightly regression suite. BFB with `ocean/develop` when `config_hmix_use_ref_cell_width=false`. Comparing the QU240 runs with `config_hmix_use_ref_cell_width=true` here versus the old settings, changes are not bit-for-bit but are machine precision.

src/core_ocean/shared/mpas_ocn_init_routines.F
```diff
@@ -316,50 +316,65 @@ end subroutine ocn_init_routines_setup_sign_and_index_fields!}}}
...
-
-      if (scaleHmixWithMesh) then
-         do iEdge = 1, nEdges
-            cell1 = cellsOnEdge(1,iEdge)
-            cell2 = cellsOnEdge(2,iEdge)
-            meshScalingDel2(iEdge) = 1.0_RKIND / ( ((meshDensity(cell1) + meshDensity(cell2) ) / 2.0_RKIND) &
-                                   / maxMeshDensity)**(1.0_RKIND / 4.0_RKIND)  ! goes as dc**1
-            meshScalingDel4(iEdge) = 1.0_RKIND / ( ((meshDensity(cell1) + meshDensity(cell2) ) / 2.0_RKIND) &
-                                   / maxMeshDensity)**(3.0_RKIND / 4.0_RKIND)  ! goes as dc**3
-            meshScaling(iEdge)     = 1.0_RKIND / ( ((meshDensity(cell1) + meshDensity(cell2) ) / 2.0_RKIND) &
```
mark-petersen, line 360: The `meshScaling` variable is identical to `meshScalingDel2`, so I removed `meshScaling`. It was only used in the Leith closure, and I changed it to `meshScalingDel2` there.

src/core_ocean/shared/mpas_ocn_init_routines.F
```diff
@@ -316,50 +316,65 @@ end subroutine ocn_init_routines_setup_sign_and_index_fields!}}}
...
+         if (config_hmix_use_ref_cell_width) then
+            ! Mesh scaling is set by areaCell and config_hmix_ref_cell_width
+            do iEdge = 1, nEdges
+               cell1 = cellsOnEdge(1,iEdge)
+               cell2 = cellsOnEdge(2,iEdge)
+               ! Effective cell width at edge iEdge, assuming neighboring cells
+               ! are circles for this calculation.
+               cellWidth = 2.0_RKIND * sqrt((areaCell(cell1) + areaCell(cell2) ) / 2.0_RKIND / pii)
+               meshScalingDel2(iEdge) = cellWidth / config_hmix_ref_cell_width
+               meshScalingDel4(iEdge) = ( cellWidth / config_hmix_ref_cell_width )**3
```
vanroekel, line 357: @mark-petersen I don't understand the change to a constant reference width, why not use a minimum value for the mesh here too?  This implementation is certainly cleaner and clearer, but it's not clear to me why we would want a constant parameter for config_hmix_ref_cell_width

mark-petersen: The viscosity follows this relation:
image
This applies within a simulation, but also generally among multiple simulations. With this PR, the flags can be left at:
```
config_hmix_use_ref_cell_width = 30e3
config_mom_del2 = 1000
config_mom_del4 = 1.2e11
```
for the above parameters, and then just adjusted slightly as needed. 

Pegging the del2 and del4 coefficients to the minimum grid cell is much less convenient and prone to error. For example, say you are spinning up a number of meshes with different minimum grid cell sizes. You would need to find the minimum grid cell size, plug it into the formula above for dx, and then put those del2 and del4 values in the namelist. That is what we do now.

Of course, in cases where you want to custom-specify the grid cell size and coefficient values, you can adjust these flags for that purpose.

vanroekel: This makes sense.  Could this be documented somehow?  Some way to say that if you use the ref_cell_width option you shouldn't move values too far from the defaults?  Isn't this relation from a paper as well?  Could we reference that in the registry?

testing_and_setup/compass/ocean/global_ocean/template_forward.xml
```diff
@@ -3,8 +3,6 @@
 		<option name="config_ocean_run_mode">'forward'</option>
 		<option name="config_block_decomp_file_prefix">'graph.info.part.'</option>
 		<option name="config_time_integrator">'split_explicit'</option>
-		<option name="config_use_mom_del2">.true.</option>
-		<option name="config_use_mom_del4">.true.</option>
```
pwolfram, line 7: Do these default true?  Not sure why they are being removed.  Do we really want the values to be defaulted to true?  Each simulation will require someone to consider this -- at small scale `del2` should be used and `del4` at large scale, for instance.  Having both default `.false` ensures this level of thought, but perhaps having both on does no harm in general.

src/core_ocean/Registry.xml
```diff
@@ -199,48 +199,56 @@
 					description="Global maximum of the mesh density"
 					possible_values="Any positive real number. If set any negative real number, config_maxMeshDensity is computed during the initialization of each simulation."
 		/>
+		<nml_option name="config_hmix_use_ref_cell_width" type="logical" default_value=".false." units="unitless"
+					description="If true, hmix coefficient values are set with reference to config_hmix_ref_cell_width. If false, hmix coefficient values are referenced to smallest gridcell in the mesh. The false setting is for backwards compatilibity. When false, hmix coefficient flags must be adjusted for every new mesh with a different minimum-sized cell."
+					possible_values=".true. or .false."
```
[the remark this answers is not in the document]
mark-petersen: Yes. I just added two sentences here explaining that. Thanks.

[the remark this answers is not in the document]
mark-petersen: Actually, not fix it to the minimum, but if `config_hmix_use_ref_cell_width=true` you can fix the coefficient to any `config_hmix_ref_cell_width` of your choice.

testing_and_setup/compass/ocean/global_ocean/QU240/template_forward.xml
```diff
@@ -3,8 +3,8 @@
 		<option name="config_dt">'02:00:00'</option>
 		<option name="config_btr_dt">'00:06:00'</option>
 		<option name="config_run_duration">'0000_06:00:00'</option>
+		<option name="config_hmix_use_ref_cell_width">.true.</option>
```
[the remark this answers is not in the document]
mark-petersen: I decided to change `config_hmix_use_ref_cell_width` to true in global sims one at a time by resolution. So I'm only changing it for the QU240 here.

testing_and_setup/compass/ocean/global_ocean/template_forward.xml
mark-petersen: Thanks. I changed this back.

PR #601, opened 2020-06-08
Move daysSinceStartOfSim calculation to master when threaded

jonbob:
This PR moves the calculation of daysSinceStartOfSim, which is done in three places, to inside a loop with "omp master" pragmas around it. Previously, one was inside an "omp single" pragma, but that did not guarantee that the master thread would get the correct result. While this does not impact the results at all, it was causing random problems in testing using threaded configurations where cprnc output from globalStats files would find an error, just in the daysSinceStartOfSim field. With thanks to @amametjanov 

[BFB]

### DIFF
--- a/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
+++ b/src/core_ocean/mode_analysis/mpas_ocn_analysis_mode.F
@@ -137,6 +137,15 @@ module ocn_analysis_mode
             simulationStartTime = startTimeStamp
          end if
 
+         block => block % next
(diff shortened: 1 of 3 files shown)


jonbob:
I tested this in E3SM by running SMS_P12x2.ne4_oQU240.A_WCYCL1850.anvil_intel.allactive-mach_mods six times and checking for the correct value of daysSinceStartOfSim in the globalStats file after each submission. So far they have all been correct, while before this change I saw bad results by the fourth test.

PR #603, opened 2020-06-17
Update flags in Makefile.in.E3SM

mark-petersen:
Add `-DUSE_PIO2` to override flags in E3SM makefile. See https://github.com/E3SM-Project/E3SM/issues/3646


mark-petersen:
Since this only involves an update to an E3SM file, @jonbob's approval is sufficient.

PR #606, opened 2020-06-20
Add NCOM bottom drag formulation

nairita87:
Add NCOM's formula for bottom drag coefficient calculation,
the formula is cb(i,j) = max(cbmin, (vonk/log(0.5*depth(i,j)/z0))**2 ))

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix.F
@@ -730,7 +730,7 @@ contains
 
       integer :: iEdge, k, cell1, cell2, N, Nsurf, nEdges
       integer, pointer :: nVertLevels
-      real (kind=RKIND) :: implicitCd
+      real (kind=RKIND) :: implicitCd, varmin
       integer, dimension(:), pointer :: nEdgesArray
 
       integer, dimension(:), pointer :: minLevelEdgeBot, maxLevelEdgeTop
@@ -821,7 +821,10 @@ contains
          end do
 
          ! average cell-based implicit bottom drag to edges and convert Mannings n to Cd
-         if (config_use_vegetation_drag .AND. config_use_vegetation_manning_equation) then
+        if (config_use_implicit_bottom_roughness) then
+           varmin=min(0.16_RKIND/(log(0.5_RKIND*0.5_RKIND*(bottomDepth(cell1)+bottomDepth(cell2))/0.001))**2.0,0.1)
+           implicitCd =max(0.0025_RKIND,varmin)
+         elseif (config_use_vegetation_drag .AND. config_use_vegetation_manning_equation) then
            implicitCd = gravity*(0.5_RKIND*(vegetationManning(cell1) + vegetationManning(cell2)))**2.0 * &
             (0.5_RKIND * (ssh(cell1) + ssh(cell2) + bottomDepth(cell1) + bottomDepth(cell2)))**(-1.0_RKIND/3.0_RKIND)
          else
@@ -1137,7 +1140,9 @@ contains
       if (config_use_implicit_bottom_drag_variable) then
         call ocn_vel_vmix_tend_implicit_spatially_variable(meshPool, bottomDrag, dt, kineticEnergyCell, &
           vertViscTopOfEdge, layerThickness, layerThickEdge, normalVelocity, err)
-      else if (config_use_implicit_bottom_drag_variable_mannings) then
+          !!vertViscTopOfEdge, layerThickness, layerThicknessEdge, normalVelocity, err)
+      else if (config_use_implicit_bottom_drag_variable_mannings.or. &
+        config_use_implicit_bottom_roughness) then
         ! update bottomDrag via Cd=g*n^2*h^(-1/3)
         call ocn_vel_vmix_tend_implicit_spatially_variable_mannings(meshPool, forcingPool, bottomDrag, &
           dt, kineticEnergyCell, &
--- a/src/core_ocean/shared/mpas_ocn_vmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix.F
@@ -730,7 +730,7 @@ contains
 
       integer :: iEdge, k, cell1, cell2, N, Nsurf, nEdges
       integer, pointer :: nVertLevels
-      real (kind=RKIND) :: implicitCd, varmin
+      real (kind=RKIND) :: implicitCd
       integer, dimension(:), pointer :: nEdgesArray
 
       integer, dimension(:), pointer :: minLevelEdgeBot, maxLevelEdgeTop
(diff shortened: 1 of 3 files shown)


philipwjones:
Hey all. Not a reviewer, but I did just refactor these impvmix routines for GPUs (see the ocean/gpuimpvmix branch in my philipwjones fork) and have a few comments. For both performance reasons and eliminating the need for all these separate routines, it's actually better to compute the bottom drag term as an (nEdges) array in a separate loop before computing the ABC coefficients.  

Second, as a general rule, don't use real numbers as exponents if they are in fact integers. The algorithm for computing real exponentiation is more expensive than the integer form so you want the compiler to choose the right form. This is especially true if the exponent is 2 or 3 where the compiler will often replace with a simple multiply (or even better, for a simple square, do the replacement yourself and avoid the exponentiation). 

For fixed constant real numbers, be sure to add the  _RKIND suffix (eg 0.5_RKIND) to ensure correct precision - you may only get single precision otherwise. Finally, try to combine/precompute some of these factors (especially divisions). Although flops are cheap, they are not free and precomputing common factors almost always wins.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -796,7 +796,146 @@ subroutine ocn_vel_vmix_tend_implicit_spatially_variable_mannings(meshPool, bott
...
+         cell1 = cellsOnEdge(1,iEdge)
+         cell2 = cellsOnEdge(2,iEdge)
+         do k = 1, N
+         layerThicknessEdge(k,iEdge) = 0.5_RKIND * (layerThickness(k,cell1) &
+         +layerThickness(k,cell2))
+         end do
+
+         ! average cell-based implicit bottom drag following the formulation
+         ! in HYCOM (reference : NCOM's formula for 2D case 
+         !cb(i,j) = max(cbmin, (vonk/log(0.5*depth(i,j)/z0))**2 ))
```
pwolfram, line 896: Minor tab formatting issue here.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -796,7 +796,146 @@ subroutine ocn_vel_vmix_tend_implicit_spatially_variable_mannings(meshPool, bott
...
+        implicitCd =max(0.0025,min(0.16/log(0.5*0.5*((bottomDepth(cell1)+bottomDepth(cell2))/0.001)**2.0),0.1))
```
pwolfram, line 897: Would be good to have these constants as variables defined at start of function, except for factors of 1/2.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -796,7 +796,146 @@ subroutine ocn_vel_vmix_tend_implicit_spatially_variable_mannings(meshPool, bott
...
+         +layerThickness(k,cell2))
+         end do
+
+         ! average cell-based implicit bottom drag following the formulation
+         ! in HYCOM (reference : NCOM's formula for 2D case 
+         !cb(i,j) = max(cbmin, (vonk/log(0.5*depth(i,j)/z0))**2 ))
+        implicitCd =max(0.0025,min(0.16/log(0.5*0.5*((bottomDepth(cell1)+bottomDepth(cell2))/0.001)**2.0),0.1))        
+        !! if((implicitCd.ge.0.1).or.(implicitCd.le.0.0001))then
+        !!implicitCd=0.0025
+        !!endif
```
pwolfram, line 900: May not need the commented lines above....

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -1078,6 +1221,13 @@ subroutine ocn_vmix_implicit(dt, meshPool, diagnosticsPool, statePool, forcingPo
             dt, kineticEnergyCell, &
             vertViscTopOfEdge, layerThickness, layerThicknessEdge, normalVelocity, &
             ssh, bottomDepth, err)
+        else if (config_use_implicit_bottom_roughness) then
```
pwolfram, line 1224: May want to be clear here that this is `config_use_implicit_bottom_roughness_z0` to make it crystal clear.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -1078,6 +1221,13 @@ subroutine ocn_vmix_implicit(dt, meshPool, diagnosticsPool, statePool, forcingPo
...
+          ! update bottomDrag via Cd=g*n^2*h^(-1/3)
```
pwolfram, line 1225: Is this formula used?   This comment seems like it was left over from a copy / paste but I may be missing something here.

src/core_ocean/shared/mpas_ocn_vmix.F
qingli411: @nairita87 In this line the square should be outside of the log, right? I think it should be
```
implicitCd = max(0.0025, min(0.16/log(0.5*0.5*(bottomDepth(cell1)+bottomDepth(cell2))/0.001)**2.0, 0.1 ) )
```

qingli411: Also do we want to save the drag coefficient, perhaps by updating `bottomDrag` so that it can be used elsewhere, e.g., in GOTM?

nairita87: @qingli411 yes I think you are right. I checked the paper by Oley. The square is over log(zb/z0).

I am confused because, in HYCOM, the code is as follows:
CBP(I,J) = MAX( CBMIN,
     &                      VONK**2/LOG(0.5*DEPTH(I,J)/Z0)**2 )

Yes I will save the drag coefficient

mark-petersen:
Rebased to head of `ocean/develop`. Passes nightly test suite.

nairita87:
The PR has been verified against previous studies (Pringle et al 2021) and also with TPXO8 data. 
Attached are plots to show such comparisons.
m2amp_ra4topo4

The following plot can be compared with Pringle et al 2021 (Figs 5 and 6)
standard_m2amp_ra4topo4

Below is the plot showing code agreement before and after rebasing
rebase

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -821,9 +821,10 @@ subroutine ocn_vel_vmix_tend_implicit_spatially_variable_mannings(meshPool, forc
          end do
 
          ! average cell-based implicit bottom drag to edges and convert Mannings n to Cd
-        if (config_use_implicit_bottom_roughness) then
-           varmin=min(0.16_RKIND/(log(0.5_RKIND*0.5_RKIND*(bottomDepth(cell1)+bottomDepth(cell2))/0.001))**2.0,0.1)
-           implicitCd =max(0.0025_RKIND,varmin)
+         if (config_use_implicit_bottom_roughness) then
+           implicitCd = max(0.0025_RKIND, &
+                        min(0.1_RKIND, &
+                          0.16_RKIND/log(250.0_RKIND*(bottomDepth(cell1)+bottomDepth(cell2)))**2 ))
```
mark-petersen, line 827: @nairita87 please check this. I consolidated these two lines, putting min and max together, and removing the varmin variable.

Also, 0.5*0.5/0.001 = 250, so I just pre-computed the math.

nairita87: @mark-petersen , Mark, I verified, passes the test nicely
commit_c7583d8

mark-petersen:
The changes here were absorbed into #606, which is also on drag, so they could be tested together.

PR #607, opened 2020-06-20
Internal wave drag implementation

nairita87:
Topographic wave drag term added to the momentum equation. 
When  `config_use_topographic_wave_drag = .true.` we add a drag term
`du/dt= -(1/rinv)*u`, where 'rinv' is the e-folding time-scale. The e-folding time-scale is read from a map (in this case from the HYCOM data)
When `config_use_implicit_bottom_roughness = .true` use the bottom drag formulation
`max(cbmin, (vonk/log(0.5*depth(i,j)/z0))**2 ))` from NCOM.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
+++ b/src/core_ocean/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
@@ -7,14 +7,14 @@
 !
 !|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
 !
-!  ocn_vel_forcing_explicit_bottom_drag
+!  ocn_vel_forcing_topographic_wave_drag
 !
-!> \brief MPAS ocean explicit bottom drag
+!> \brief MPAS ocean topographic wave drag
 !> \author Nairita Pal
 !> \date   Oct 2020
 !> \details
 !>  This module contains the routine for computing
-!>  tendencies from explicit bottom drag.
+!>  tendencies from topographic wave drag.
 !
 !-----------------------------------------------------------------------
 
@@ -53,8 +53,8 @@ module ocn_vel_forcing_topographic_wave_drag
    !
    !--------------------------------------------------------------------
 
-   logical :: topographicWaveDragOn!!explicitBottomDragOn
-   real (kind=RKIND) :: topographicWaveDragCoeff!!explicitBottomDragCoef
+   logical :: topographicWaveDragOn
+   real (kind=RKIND) :: topographicWaveDragCoeff
 
 !***********************************************************************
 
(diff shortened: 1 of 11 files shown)


proteanplanet:
This is going to be canceled following advice from @nairita87, a new PR will be submitted.

mark-petersen:
Rebased to head of ocean/develop. Passes nightly test suite.

proteanplanet:
@nairita87: I think this is ready to merge? We just need @mark-petersen to review?

nairita87:
Verified the rebased code with the latest commit #a645c40. Matches well with previous versions
rebase_crrent

src/core_ocean/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
```diff
@@ -0,0 +1,196 @@
...
+      !$omp do schedule(runtime) private(k, cell1, cell2)
+      do iEdge = 1, nEdgesOwned
+        cell1 = cellsOnEdge(1,iEdge)
+        cell2 = cellsOnEdge(2,iEdge)
+        k =  maxLevelEdgeTop(iEdge)
+        ! topographic wave drag term:
+        ! du/dt = ... (1.0/topographic_wave_drag_rinv) * u
+        ! topographic_wave_drag_rinv is the e-folding time which is read from a file
+        tend(k,iEdge) = tend(k,iEdge) - topographicWaveDragCoeff &
+           /(0.5_RKIND*(topographic_wave_drag_rinv(cell1)+topographic_wave_drag_rinv(cell2)))*normalVelocity(k,iEdge)
```
mark-petersen, line 132: @nairita87 on these lines we have a divide and average from cell to edge every time step. It makes a lot more sense compute rinv in your original python interpolation to edges, and save a field which is 1/rinv at the edges. Call it topographic_wave_drag. Then these lines would simply be
```
131         tend(k,iEdge) = tend(k,iEdge) - topographicWaveDragCoeff * &
               topographic_wave_drag(iEdge)*normalVelocity(k,iEdge)
```

src/core_ocean/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
```diff
@@ -0,0 +1,192 @@
...
+      call mpas_timer_start('vel topographic wave drag')
+
+      !$omp do schedule(runtime) private(k)
+      do iEdge = 1, nEdgesOwned
+        k =  maxLevelEdgeTop(iEdge)
+        ! topographic wave drag term:
+        ! du/dt = ... 1/rinv * u
+        ! rinv is the e-folding time use in HyCOM. Here
+        ! topographic_wave_drag = 1/rinv
+        if (k>0) then
```
mark-petersen, line 126: @philipwjones the boundary edges between water and land (i.e. no cells) have `maxLevelEdgeTop(iEdge)=0`, so this if statement was needed within the edge loop to prevent accessing k=0. Is there a more performant way to do this?

nairita87:
Code verification against old PR,
 1/(topographic_wave_drag(icell1)+topographic_wave_drag(icell2)) is now replaced with 1/(topographic_wave_drag(iEdge))
rms_kin
m2amp_ra4topo4_compare

src/core_ocean/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
philipwjones: Probably not worth worrying about for this small loop and the alternatives are probably less robust, so I think ok to keep here.

mark-petersen (approved):
Looks good. @nairita87 thanks for your work.

PR #608, opened 2020-06-22
Fixes CVMix surface layer averaging routine

vanroekel:
Currently when cvmix computes averages for the surface layer (needed for
finding boundary layer depth) it simply does an arithmetic mean.  This
is reasonable for the 60 layer mesh (with near constant 10m resolution
through 200m), but is not correct for stretched
grids.  Here the averaging is changed to a layerThickness weighted
average.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
@@ -126,7 +126,7 @@ contains
       type(cvmix_data_type) :: cvmix_variables
 
       integer, dimension(:), pointer :: &
-        maxLevelCell, nEdgesOnCell
+        maxLevelCell, nEdgesOnCell, maxLevelEdgeTop
 
       real (kind=RKIND), dimension(:), pointer :: &
         latCell, lonCell, bottomDepth, surfaceBuoyancyForcing, surfaceFrictionVelocity, fCell, &
@@ -151,7 +151,7 @@ contains
       character (len=StrKIND), pointer :: config_cvmix_shear_mixing_scheme, config_cvmix_kpp_matching
 
       integer :: k, i, iCell, jCell, iNeighbor, iter, timeLevel, kIndexOBL, kav, iEdge, nCells
-      integer :: edgeCount, nEdges, topIndex, nsmooth, kpp_stage, cell1, cell2
+      integer :: edgeCount, nEdges, topIndex, nsmooth, kpp_stage
       integer, pointer :: nVertLevels, nVertLevelsP1
       integer, dimension(:), pointer :: nCellsArray
       integer, dimension(:), allocatable :: surfaceAverageIndex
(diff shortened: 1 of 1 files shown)


vanroekel:
@mark-petersen and @qingli411 if you could take a look here, I'd appreciate it.  I think this fix will be necessary for the 64 and 80 Layer meshes.  However, I'm not sure how big the impact will be, I'm running a sensitivity test now and will post when it is complete.

mark-petersen:
@vanroekel This is needed for the upcoming water cycle simulations, correct? And this must be a non-bfb change to all E3SM simulations? Any comments on your sensitivity tests on this PR?

vanroekel:
I realize I forgot to finish analysis of the control case.  I'll post sensitivity results here once mpas-analysis runs.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -530,25 +534,37 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, diagnost
               ! Build deltaVelocitySquared
               do i = 1, nEdgesOnCell(iCell)
                  iEdge = edgesOnCell(i, iCell)
+                 cell1 = cellsOnEdge(1, iEdge)
+                 cell2 = cellsOnEdge(2, iEdge)
+
                  deltaVelocitySquared(1) = 0.0_RKIND
 
-                 normalVelocitySum(1) = normalVelocity(1, iEdge)
+                 layerThicknessEdge = 0.5_RKIND*(layerThickness(1, cell1) + layerThickness(1, cell2))
```
mark-petersen, line 542: We have a layerThicknessEdge variable in the Registry. Of course this is a little calculation, but `layerThicknessEdge(iEdge)` has a better memory access for an edge loop. I can change that.

vanroekel: Thanks for catching that.  Not sure why I didn't think of that array!  Thanks for making the change too

vanroekel:
testing in e3sm suggests this will be non-BFB but only slightly climate changing (at least for 6030 resolution).  Here are the analysis
```
Baseline (control) 64 layer case -- https://portal.nersc.gov/project/e3sm/lvroekel/20200618_64LayerNewRedi_GMPAS-IAF.T62_oECv3_anvil_21_30/

with the fix -- https://portal.nersc.gov/project/e3sm/lvroekel/20200618_64LayerNewRediKPPfix_GMPAS-IAF.T62_oECv3_anvil_10_30/
```

mark-petersen:
@vanroekel I made the change to use the `layerThicknessEdge` array, commit edc25c6. I tested before/after and was concerned that it was not bfb between the two. I suspect that `layerThicknessEdge` is computed at the end of the step, so there is a time offset that doesn't matter much. 

I also changed the max vertical index from maxLevelCell to maxLevelEdgeTop, which makes more sense when you are computing on an edge. The on an edge below maxLevelEdgeTop but above maxLevelCell (edge between water and land) the normalVelocity is zero and layerThicknessEdge is -1e34, so this should be the same because of the zero velocity.

There changes seem trivial, but since they are non-bfb, I think we should either retest in E3SM to be sure, or remove them, and stick with your original commit. You can do whichever you prefer.

PR #613, opened 2020-06-29
Add option for LAPACK library

mark-petersen:
The ocean core is adding an implicit time stepping method that uses Lapack. Linking to Lapack is optional, so this addition should not affect the other cores.


hyungyukang:
It looks good and is working in my test systems.

PR #614, opened 2020-06-29
Adds calving and submarine melting capabilities for grounded marine ice sheet margins. Landice/ismip6 melt and calving

trhille:
Adds calving and submarine melting capabilities for grounded marine ice sheet margins.

Adds calving routine based on von Mises stress parameterization (Morlighem et al., 2016) to mpas_li_calving.F and ISMIP6 submarine melting (Slater et al., 2020) routine to mpas_li_iceshelf_melt.F. Uses a flux distribution algorithm to ensure that calving and melting can cause retreat of the margin rather than just thinning at the marginal cell.


src/core_landice/mode_forward/mpas_li_velocity.F
```diff
@@ -784,7 +784,7 @@ end subroutine li_velocity_finalize
    subroutine calculate_beta(block, err)
 
       use li_mask
-
+      use li_subglacial_hydro
```
matthewhoffman, line 787: @trhille , do you remember why this was needed?  It should only be needed if the code is calling subroutines/functions from that module.  It's not needed to access variables that 'belong' to subglacial hydro.

trhille: I don't specifically remember, but it must have stemmed from mistakenly thinking I needed 'use li subglacial_hydro' to access effectivePressure from the subglacial hydro pool. And then I had to update the Makefile to get it to compile once I had added this line.

matthewhoffman:
This PR was replaced.

PR #617, opened 2020-07-01
Optionally add links to load_compass_env.sh in test cases

xylar:
This is specified either in the config file or at the command line.

### DIFF
--- a/testing_and_setup/compass/setup_testcase.py
+++ b/testing_and_setup/compass/setup_testcase.py
@@ -1556,7 +1556,9 @@ def get_case_name(config_file):  # {{{
 def link_load_compass_env(init_path, configs):  # {{{
 
     if configs.getboolean('conda', 'link_load_compass'):
-        target = configs.get('conda', 'load_compass_script')
+        target = '{}/{}/load_compass_env.sh'.format(
+            configs.get('script_paths', 'script_path'),
+            configs.get('script_paths', 'core_dir'))
 
         link_name = '{}/load_compass_env.sh'.format(init_path)
         try:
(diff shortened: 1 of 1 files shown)


xylar:
@matthewhoffman, this would add a capability we've already been using a little bit on `ocean/develop` to have a script that knows about machines we commonly use COMPASS on and which can be sourced to load the appropriate conda environment for COMPASS.  An example of such a script is [here](https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/load_compass_env.sh), though I will need to move it into the `ocean` subdirectory in a separate PR.

If this functionality sounds useful to you, you will likely want to add a copy of the script to `landice` and update the version number for the compass conda env as needed.  If it doesn't sound useful, that's fine, and the only review I would request from you is to make sure this change doesn't cause you any trouble (either by inspection or by running a `landice` test case).

xylar:
@mark-petersen, I am having trouble with other testing I'm doing because of conflicts between `develop` and `ocean/develop`.  I think our quick-and-dirty approach of modifying `setup_testcase.py` in `ocean/develop` has turned out to be a bad move and I will strongly advocate against doing that in the future.

I realized in the process of putting in the changes we have made (which, after reversion of some of my MPI work, is just linking `load_compass_env.sh`) that there will be issues if different cores try to update the same `load_compass_env.sh`.  Instead, we need one per core and it will be clear which core owns each file and which branch it should be modified in.  I will make a separate PR to move `load_compass_env.sh` into the `ocean` directory on `ocean/develop`.

xylar:
## Testing

I tested by setting up the `ocean/global_coean/QU240wISC/init` test case with and without the `--link_load_compass` flag (after merging this branch into #618).  Without the flag, there is no `load_compass_env.sh` link in the test case, as expected.  With the flag, the link is there and sourcing it loads the expected `compass_0.1.6` environment on Grizzly.

matthewhoffman:
@xylar , this sounds useful.  I will try it out.

xylar:
@matthewhoffman, keep in mind that the more recent `compass_*` environments include MPI, and that causes problems if you load things in the wrong order.  The order of operations would be to first source `load_compass_env.sh` and then load the modules for the system compilers and MPI (so they take priority).  I know this is a pain but I haven't figured out a better way.

PR #618, opened 2020-07-01
Move load_compass_env.sh to the ocean dir

xylar:
Each core will have its own version of this file going forward.


xylar:
@mark-petersen and @sbrus89, this should be merged after #617 is merged to `develop` and `develop` is merged to `ocean/develop`.

xylar:
## Testing

I tested by setting up the `ocean/global_coean/QU240wISC/init` test case with and without the `--link_load_compass` flag (after merging #617 into this branch).  Without the flag, there is no `load_compass_env.sh` link in the test case, as expected.  With the flag, the link is there and sourcing it loads the expected `compass_0.1.6` environment on Grizzly.

xylar:
@mark-petersen, and @sbrus89, this should be tested an merged soon, since the `load_compass_env.sh` script won't be found in its current location anymore.

xylar:
@sbrus89, can you verify that the `--link_load_compass` flag works for you on this branch?

xylar:
@mark-petersen, please wait until @sbrus89 has approved.

mark-petersen:
@xylar, I can't get it to work with the `--link_load_compass` flag, but maybe I misunderstand how to use it. On `ocean/develop` with this PR merged locally, I try:
```
cd testing_and_setup/compass
./setup_testcase.py -q -f config.ocean_turq --work_dir /lustre/scratch3/turquoise/mpeterse/runs/nightly/ocean_model_200816_f53d3f0d \
  -n 54 -m runtime_definitions/mpirun.xml --link_load_compass

  File "./setup_testcase.py", line 27, in <module>
    import netCDF4
ImportError: No module named netCDF4
```
Also tried `link_load_compass = True` in config file, and get same response. Is that how we are supposed to test?

Do we want a `--link_load_compass` on `manage_regression_suite.py` for consistency?

sbrus89:
@xylar, My link ends up pointing to the `compass` directory, not the `ocean` directory.

xylar:
@mark-petersen, I think you're misunderstanding the purpose.  You still have to load compass.  This just gives you a local link to the `load_compass_env.sh` so you can source it, e.g. in a job script.  There is no way for the script to load compass for you because it is a python script and needs a working python environment already loaded.

xylar:
@sbrus89, maybe we need to rebase, since these changes will only work if https://github.com/MPAS-Dev/MPAS-Model/pull/617 is included.  I'll try locally on my laptop.

mark-petersen:
@xylar I did misunderstand, thanks.

xylar:
@sbrus89, it worked for me after a rebase.  Could you try again?

sbrus89:
Thanks @xylar, that works for me now on Grizzly.

xylar:
> Do we want a --link_load_compass on manage_regression_suite.py for consistency?

@mark-petersen, yes, that's on my to-do list.  It would be modification to `develop`, so it will take some time to propagate even when it happens.

PR #620, opened 2020-07-02
Fixes decomposition error in k33

vanroekel:
Currently it is being assumed rediKappa lives at cell centers for the
k33 term in mpas_ocn_vmix_coefs_redi.F, however the rest of the model
assumes rediKappa lives on edges, this is causing a decomposition test
to fail. This PR fixes that bug. Also removes warnings from analysis
members, and improves variable names and descriptions.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
@@ -76,7 +76,7 @@ contains
 
    subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThickness, layerThicknessEdge, zMid, tracers, &
                                         RediKappa, dt, isActiveTracer, slopeTriadUp, slopeTriadDown, &
-                                        kappaRediSfcTaper, kappaRediCell, rediLimiterCount, tend, err)!{{{
+                                        RediKappaSfcTaper, RediKappaScaling, rediLimiterCount, tend, err)!{{{
 
       !-----------------------------------------------------------------
       !
@@ -93,8 +93,8 @@ contains
       real(kind=RKIND), dimension(:, :), intent(in) :: &
          layerThicknessEdge, &!< Input: thickness at edge
          zMid, &!< Input: Z coordinate at the center of a cell
-         kappaRediSfcTaper, &!< Input: vertical structure of GM/Redi limiter
-         kappaRediCell, &
+         RediKappaSfcTaper, &!< Input: vertical structure of GM/Redi limiter
+         RediKappaScaling, &
          layerThickness
 
       real(kind=RKIND), dimension(:, :, :), intent(in) :: &
@@ -222,10 +222,10 @@ contains
             coef = dvEdge(iEdge)
 
             k = 1
-            kappaRediEdgeVal = 0.25_RKIND*(kappaRediCell(k, cell1) + kappaRediCell(k, cell2) + &
-                                           kappaRediCell(k + 1, cell1) + kappaRediCell(k + 1, cell2))* &
-                               0.25_RKIND*(kappaRediSfcTaper(k, cell1) + kappaRediSfcTaper(k, cell2) + &
-                                           kappaRediSfcTaper(k + 1, cell1) + kappaRediSfcTaper(k + 1, cell2))
+            kappaRediEdgeVal = 0.25_RKIND*(RediKappaScaling(k, cell1) + RediKappaScaling(k, cell2) + &
+                                           RediKappaScaling(k + 1, cell1) + RediKappaScaling(k + 1, cell2))* &
+                               0.25_RKIND*(RediKappaSfcTaper(k, cell1) + RediKappaSfcTaper(k, cell2) + &
+                                           RediKappaSfcTaper(k + 1, cell1) + RediKappaSfcTaper(k + 1, cell2))
 
             do iTr = 1, nTracers
                tracer_turb_flux = tracers(iTr, k, cell2) - tracers(iTr, k, cell1)
@@ -246,10 +246,10 @@ contains
 
             do k = 2, maxLevelEdgeTop(iEdge) - 1
 
-               kappaRediEdgeVal = 0.25_RKIND*(kappaRediCell(k, cell1) + kappaRediCell(k, cell2) + &
-                                              kappaRediCell(k + 1, cell1) + kappaRediCell(k + 1, cell2))*0.25_RKIND* &
-                                  (kappaRediSfcTaper(k, cell1) + kappaRediSfcTaper(k, cell2) + &
-                                   kappaRediSfcTaper(k + 1, cell1) + kappaRediSfcTaper(k + 1, cell2))
+               kappaRediEdgeVal = 0.25_RKIND*(RediKappaScaling(k, cell1) + RediKappaScaling(k, cell2) + &
+                                              RediKappaScaling(k + 1, cell1) + RediKappaScaling(k + 1, cell2))*0.25_RKIND* &
+                                  (RediKappaSfcTaper(k, cell1) + RediKappaSfcTaper(k, cell2) + &
+                                   RediKappaSfcTaper(k + 1, cell1) + RediKappaSfcTaper(k + 1, cell2))
 
                do iTr = 1, nTracers
                   ! \kappa_2 \nabla \phi on edge
@@ -282,10 +282,10 @@ contains
             end do
 
             k = maxLevelEdgeTop(iEdge)
-            kappaRediEdgeVal = 0.25_RKIND*(kappaRediCell(k, cell1) + kappaRediCell(k, cell2) + &
-                                           kappaRediCell(k + 1, cell1) + kappaRediCell(k + 1, cell2))* &
-                               0.25_RKIND*(kappaRediSfcTaper(k, cell1) + kappaRediSfcTaper(k, cell2) + &
-                                           kappaRediSfcTaper(k + 1, cell1) + kappaRediSfcTaper(k + 1, cell2))
+            kappaRediEdgeVal = 0.25_RKIND*(RediKappaScaling(k, cell1) + RediKappaScaling(k, cell2) + &
+                                           RediKappaScaling(k + 1, cell1) + RediKappaScaling(k + 1, cell2))* &
+                               0.25_RKIND*(RediKappaSfcTaper(k, cell1) + RediKappaSfcTaper(k, cell2) + &
+                                           RediKappaSfcTaper(k + 1, cell1) + RediKappaSfcTaper(k + 1, cell2))
 
             do iTr = 1, nTracers
                tracer_turb_flux = tracers(iTr, k, cell2) - tracers(iTr, k, cell1)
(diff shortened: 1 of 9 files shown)


vanroekel:
@mark-petersen I think I found the redi issue.  It was in the k33 term.  It was a place I missed moving rediKappa from cells to edges.

vanroekel:
@jonbob could you please take a look at the changes and see if I fixed the threading issue correctly?

jonbob:
@vanroekel - that's perfect. I think the big problem was having the omp pragma before the "nCells =" line instead of the do loop. For what it's worth, I'm still in the queue on anvil and should get results soon -- but at least with this fix it does compile

jonbob:
Passed ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.compy_intel

jonbob:
But it turns out my test didn't have redi on. With redi on, the test failed to complete -- I'll try to understand why

jonbob:
I'm trying a PEM test to see if the failure is due to threading

mark-petersen:
With this addition and Redi on, all tests fail the run stage, and end after a few steps because ocean variables go out of bounds. This includes even the simplest test, `PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel`. With Redi off the tests pass. Also, `origin/master` with Redi on passes the run stage. Looking at the code, it's not clear what is happening. Perhaps the line
```
nCells = nCellsArray(1)
```
is too small.

vanroekel:
@mark-petersen prior to this PR that code used nCellsSolve, which is equivalent to nCellsArray(1), so that doesn't seem like the cause.  Isn't PET a threading test?  Wouldn't it be more likely a private is missing?  I thought I put them in there but may have messed something up.  Testing by @jonbob and I find that the code passes with no threading.

mark-petersen:
With this change, I can't even get to the threading comparison, because all the E3SM runs I try fail after a few steps, with both gnu and intel on cori, and with Redi on. With Redi on origin/master it fails the PEM test but at least completes successfully. @jonbob or @vanroekel could you run a few tests to see if this PR addition fails for you too? I tried a few things this past weekend, but didn't fix it.

jonbob:
My PEM test (PEM_Ln9.ne30_oECv3_ICG.A_WCYCL1850S.compy_intel) passed on come with redi on. I'll try now with threading...

vanroekel:
@mark-petersen can you try your testing again with 
```
config_disable_redi_k33 = .true.
```
?
This will help isolate the issue

vanroekel:
@mark-petersen disregard my last comment.  I just saw a pretty glaring bug.  I'll push up a fix ASAP

vanroekel:
@mark-petersen and @jonbob I just noticed the new interpolation of RediKappa from edges to centers was missing a normalization by areaCell making the new RediKappa much too large, easily explaining the failures seen in testing.  Could you please try this fix out?

jonbob:
I'll test with the new changes right away. For for what it's worth, I got warnings looking like this for my first test:
````
WARNING: Error: Field layerThickness has more than one time level, but no timeLevel argument given.

**Comment by jonbob** [comment]
It's still failing the PET test for me

**Comment by vanroekel** [comment]
but everything else passes?

**Comment by jonbob** [comment]
I didn't try anything else -- should I run the PEM one?

**Comment by vanroekel** [comment]
That would be great.  I'm hoping this will get @mark-petersen pass the forward model failures, but since I didn't change any pragmas I guess it makes sense PET still fails.

**Comment by vanroekel** [comment]
@jonbob could you also try the PET test with the k33 term disabled?  The disable flag is in a comment above.

**Comment by jonbob** [comment]
@vanroekel - OK, modified PET test is submitted, PEM test is building. I'll let you know...

**Comment by jonbob** [comment]
@vanroekel - the PEM test still passes with your latest update, the PET test fails to run with k33 disabled....

**Comment by jonbob** [comment]
@vanroekel - I'm trying a PET test with the big OMP loop in mpas_ocn_tracer_hmix_redi.F commented out

**Comment by mark-petersen** [comment]
I just added two commits. The first adds some missing k in private loops. That might help the PET pass, but I don't know yet. The second takes care of these warnings:
```
WARNING: Error: Field layerThickness has more than one time level, but no timeLevel argument given.
```
those are unrelated to our current problems, but we might as well fix that too.

**Comment by jonbob** [comment]
@mark-petersen - thanks for fixing those warnings. I wasn't sure they were meaningful or not, but always good t clear them up.

@vanroekel - the PET test still failed to run with the big loop pragmas commented out

**Comment by mark-petersen** [comment]
Not passing the MPAS-Ocean thread test, which has redi on:
```
```
I should be able to figure it out from here, it is a lot faster.

File: src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F [hunk 1]
```diff
@@ -188,6 +189,7 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThickness, laye
         enddo
         rediKappaCell(iCell) = rediKappaCell(iCell) / areaCell(iCell)
       enddo
+      !$omp end do
```
vanroekel, line 192: @mark-petersen testing by @jonbob confirmed by my own testing shows that if we comment out L183 and L192 PET passes and with the lines the run fails.  It doesn't make sense why.  Do you have any thoughts?

mark-petersen:
@vanroekel The `RediKappa(iCell)` here is an obvious error, since it is indexed by iEdge.
```
shared/mpas_ocn_tracer_hmix_redi.F
337                flux_term3 = RediKappa(iCell)*2.0_RKIND* &
338                             (kappaRediSfcTaper(k, iCell)*kappaRediCell(k, iCell)*fluxRediZTop(iTr, k)  &
339                             * invAreaCell - kappaRediSfcTaper(k + 1, iCell)*kappaRediCell(k + 1, iCell) &
340                             *fluxRediZTop(iTr, k + 1) * invAreaCell)
```
Is that value in `kappaRediCell`, so `RediKappa(iCell)` just shouldn't be there?

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
mark-petersen: I agree that doesn't make any sense. It must be something else, and this just reveals it.

vanroekel:
Thanks @mark-petersen I found that one too.  I just haven't pushed up yet.  I agree with you about the omp directives just revealing a different problem.  I was combing through the rest of the code to find other redikappacell issues.  I'll push up my changes soon.

vanroekel:
@jonbob and @mark-petersen I think found the issue.  The array ```rediKappaCell``` was being allocated on each thread and partially filled by each thread but then not synced back to the same array so since there is no guarantee of what thread gets what index we were getting uninitialized rediKappaCell values later in the routine.  I've confirmed the previous commit fails MPAS-O only thread test (QU-240) but with the latest commit it passes.  I'm running the QU240 E3SM g-case now PET.

vanroekel:
confirming the new changes pass ```PET_Ln9.T62_oQU240.GMPAS-IAF```

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -332,7 +335,7 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, scratchPool, layerThickness, laye
                ! Add tendency for Term 3: d/dz ( h S grad phi) = ( S grad phi) fluxes
                ! 2.0 in next line is because a dot product on a C-grid
                ! requires a factor of 1/2 to average to the cell center.
-               flux_term3 = RediKappa(iCell)*2.0_RKIND* &
+               flux_term3 = rediKappaCell(iCell)*2.0_RKIND* &
                             (kappaRediSfcTaper(k, iCell)*kappaRediCell(k, iCell)*fluxRediZTop(iTr, k)  &
```
mark-petersen, line 339: `rediKappaCell` needs a `k` index. But more generally, term 3:
image
should only have one factor of kappa, and this line has kappa^2.

vanroekel: actually it doesn't.  there are two different arrays, one is kappaRediCell (one line down) this is rediKappaCell, which is the constant value interpolated from edges to centers.  I agree it's a poor naming convention.  Any suggestions how I could rename?

mark-petersen:
@vanroekel how about this. These two variables are left as-is:
```
2640         <var name="gmBolusKappa" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
2641              description="GM Bolus Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."

2644         <var name="RediKappa" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
2645              description="Redi Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
```
On it's native location, we don't specify. But an interpolated variable includes 'Cell' or 'Edge'. So we keep the allocated variable `RediKappCell` as is. Then, we change kappaGMCell, kappaRediCell to:
```
2491         <var name="gmKappaScaling" type="real" dimensions="nVertLevelsP1 nCells Time" units="NA"
2492             description="Scaling coefficient for GM kappa. Varies from 0 to 1."

2495         <var name="RediKappaScaling" type="real" dimensions="nVertLevelsP1 nCells Time" units="NA"
2496             description="Scaling coefficient for Redi kappa. Varies from 0 to 1."
```
Check my description. These are unitless, so vary from 0 to 1?

jonbob:
A clean checkout passed PET_D.T62_oQU240.GMPAS-IAF.anvil_intel and has at least successfully completed the first run of PET_D.T62_oEC60to30v3.GMPAS-IAF.anvil_intel. Once the second part of that test gets through the queue and runs, I'll update and OK this PR

mark-petersen:
Yes, so far so good. I'm getting all passes on the MPAS nightly regression suite and all passes so far with PET tests in E3SM. Will post when complete.

vanroekel:
@mark-petersen I like your naming.  Would it be okay with you if I take this and do the changes in a separate PR?

mark-petersen:
Passes run and compare on:
```
PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu
PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_gnu
PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel
PEM_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel
```
Also tested MPAS nightly regression suite after last PR with name change, debug and optimized, passes all. I think this is ready to merge.

vanroekel:
@mark-petersen thanks for making the name changes.  I've added a little more explanation of the scaling arrays to clarify how they are set.  I think this is ready to go after @jonbob finishes testing.

jonbob:
my test may take some time -- anvil is swamped right now. But I'm comfortable approving this PR based on all the other testing. I'm not sure the timing will matter, since it may take a few days to get into the repo...

PR #621, opened 2020-07-02
Add tidal harmonic decomposition analysis member

sbrus89:
This pull request adds a new analysis member to preform a global harmonic decomposition for tides. It uses a module borrowed from the ADCIRC codebase which will be adapted to interface with MPAS-Ocean.

### DIFF
--- a/docs/ocean/design_docs/harmonic_analysis/main.tex
+++ b/docs/ocean/design_docs/harmonic_analysis/main.tex
@@ -1,0 +1,439 @@
+\documentclass[11pt]{report}
+
+\usepackage{epsf,amsmath,amsfonts}
+\usepackage{graphicx}
+\usepackage{booktabs}
+\usepackage{color}
+\usepackage{natbib}
+\usepackage{multirow}
+\usepackage{hyperref}
+
+\setlength{\textwidth}{6.5in}
+\setlength{\oddsidemargin}{0in}
+\setlength{\evensidemargin}{0in}
+\setlength{\textheight}{8in}
+\setlength{\topmargin}{0.5in}
+
+\newcommand{\ds}{\displaystyle}
+\setlength{\parskip}{1.2ex}
+\setlength{\parindent}{0mm}
+\newcommand{\bea}{\begin{eqnarray}}
+\newcommand{\eea}{\end{eqnarray}}
+\newcommand{\nn}{\nonumber}
+
+\def\Mark#1{{\textcolor{red}{\bf \textit{By Mark: } #1}}} % new text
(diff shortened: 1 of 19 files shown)


sbrus89:
Currently, this is in its very early stages and will requires a lot more work before it can be merged. However, by starting the PR now we can start keeping track of discussion etc.

mark-petersen:
Thanks. Early is always a good idea. Please take a look at this, will be merged soon:
https://github.com/MPAS-Dev/MPAS-Model/pull/539
We just simplified how to access config flags - now with a `use ocn_config` module, that's it. We also changed the scratch allocates from a scratch pool to just local allocates. You can see examples in that PR.

sbrus89:
Thanks for the heads-up, Mark.

sbrus89:
Here are some preliminary results on this mesh (120km global -> 10km Delaware region):

cellWidth

M2 Amplitude:
M2Amplitude_comparison

M2 Phase:
M2Phase_comparison

S2 Amplitude:
S2Amplitude_comparison

S2 Phase:
S2Phase_comparison

N2 Amplitude:
N2Amplitude_comparison

N2 Phase:
N2Phase_comparison

O1 Amplitude:
O1Amplitude_comparison

O1 Phase:
O1Phase_comparison

K1 Amplitude:
K1Amplitude_comparison

K1 Phase:
K1Phase_comparison

sbrus89:
I don't think the results are too bad given the coarse mesh and lack of tuning, but the accuracy does need to be improved. However, I think these comparisons help show that the harmonic analysis AM is working.

sbrus89:
The simulation takes about 45 minutes using 360 cores on Grizzly.

sbrus89:
@mark-petersen and @proteanplanet, let me know if either of you have any suggestions at this point. I don't know GitHub usernames for Nairita, Kristin, or Brian but it would be good to include them as reviews as well (if they are interested).

sbrus89:
Restart capabilities are now working. I have verified that restarting after the harmonic analysis period (solves least squares system with no updates) and mid-way through the harmonic analysis period (updates and solves least squares system) is BFB with the full run.

sbrus89:
@proteanplanet and @mark-petersen, here is a draft of the design document for this new capability:

[Harmonic_Analysis.pdf](https://github.com/MPAS-Dev/MPAS-Model/files/4975701/Harmonic_Analysis.pdf)

Let me know if you have any feedback and I'll make changes. I'm planning on including the latex files in the `MPAS-Model/docs/ocean/design_docs/` directory as I've done before with tidal potential and atmospheric forcing features.

proteanplanet:
Hi Steve,  The design document looks good, and has all the information that seems to be needed. - Andrew

PR #622, opened 2020-07-02
Adds EC60to30 decomposition test to nightly

vanroekel:
This adds a new processor decomposition test to nightly regression suite

The EC60to30 mesh is used with 64 and 128 processors to help test cases similar to E3SM testing/simulations

### DIFF
--- a/testing_and_setup/compass/ocean/global_ocean/EC60to30/decomposition_test/config_128pes_run.xml
+++ b/testing_and_setup/compass/ocean/global_ocean/EC60to30/decomposition_test/config_128pes_run.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0"?>
+<config case="128pes_run">
+	<add_link source="../../init/initial_state/initial_state.nc" dest="init.nc"/>
+	<add_link source="../../init/initial_state/graph.info" dest="graph.info"/>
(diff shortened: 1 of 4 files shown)


vanroekel:
@mark-petersen this test helped locate the redi issue.  I wasn't totally sure where to direct this, I went with ocean/develop as it only touches ocean cases.  Let me know if that is not correct.

Also, I noticed the EC60to30 init takes the longest time by far in running (21 minutes vs 1 min or less for all others).  Do you have any thoughts as to why?

xylar:
> I wasn't totally sure where to direct this, I went with ocean/develop as it only touches ocean cases. Let me know if that is not correct.

Yep, that's right.  `ocean/develop` is the right place for anything within `testing_and_setup/compass/ocean` or otherwise with `ocean` in the name.

> Also, I noticed the EC60to30 init takes the longest time by far in running (21 minutes vs 1 min or less for all others). Do you have any thoughts as to why?

That seems long but EC60to30 init does take awhile which is why we don't normally include it in init.  The mask creator is really slow lately and @mark-petersen and I haven't been able to figure out why or when it changed.

xylar:
@vanroekel, I moved this to https://github.com/MPAS-Dev/compass/pull/12.  We can hopefully discuss it further there, once some other merges happen and I can rebase it onto master.

PR #623, opened 2020-07-02
Add GPU compiler flags for OpenACC and OpenMP_offload

amametjanov:
This adds a new framework compiler entry `xlf-summit-omp-offload` that has an option to build with OpenMP offloading support. This also enables OpenACC and OpenMP_offload builds in E3SM.

[bit-for-bit]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ xlf:
 	"OPENMP = $(OPENMP)" \
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI" )
 
-xlf-summit:
+xlf-summit-omp-offload:
 	( $(MAKE) all \
 	"FC_PARALLEL = mpif90" \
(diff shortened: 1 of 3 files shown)


mgduda:
Since there is already an `xlf` target that works on Summit, perhaps choosing a name other than `xlf-summit` might help to avoid confusion among users?

mark-petersen:
@amametjanov what is your preference on the name? Would it be helpful to be more specific than `xlf-summit`?

amametjanov:
No preference. Please suggest a name. `xlf-summit-e3sm`, `xlf-summit-gpu`, `xlf-summit-omp-offload`?

mark-petersen:
According to your PR description, `xlf-summit-omp-offload` makes the most sense. If you agree, please change it.

mark-petersen:
@mgduda thanks for the naming suggestion. Could you approve?

Issue #625, opened 2020-07-03
Synthetic hurricane model bug


pwolfram:
Use of the `testing_and_setup/compass/ocean/hurricane_wind_pressure/main.py` file produces a faulty last time step for `testing_and_setup/compass/ocean/hurricane/synthetic_hurricanes/SANDY_017771.json` synthetic hurricane using input file `hurricane_inputs.txt`:
```
SANDY_017771.json                                                                                       
2▸        ▸ ! grid flag 1 = regular grid (need further inputs), 2 = nc file (need file name)
mesh.nc
1013                    ! mbar, ambient pressure
1.                      ! holland B parameter
```
                                                                                                        
where `output.nc` has the faulty field that is not define for the last time step, resulting in spurious large number in MPAS-O:
```python
ipython
In [7]: import xarray as xr                                                                                                                                                                                                                                                                                                  

In [8]: ds = xr.open_dataset('out.nc')                                                                                                                                                                                                                                                                       

In [9]: ds.windSpeedU.isel(Time=-1)                                                                                                                                                                                                                                                                                          
Out[9]: 
<xarray.DataArray 'windSpeedU' (nCells: 15404)>
array([9.96921e+36, 9.96921e+36, 9.96921e+36, ..., 9.96921e+36, 9.96921e+36,
       9.96921e+36])
Dimensions without coordinates: nCells

In [10]: ds.windSpeedU.isel(Time=-2)                                                                                                                                                                                                                                                                                         
Out[10]: 
<xarray.DataArray 'windSpeedU' (nCells: 15404)>
array([-3.588561,  1.057031,  0.461513, ..., -6.030833,  6.671099,  6.515482])
Dimensions without coordinates: nCells
In [11]: ds.xtime[-2:]                                                                                                                                                                                                                                                                                                       
Out[11]: 
<xarray.DataArray 'xtime' (Time: 2)>
array([b'2012-01-12_23:00:00                                             ',
       b'2012-01-13_00:00:00                                             '],
      dtype='|S64')
Dimensions without coordinates: Time
```

pwolfram:
Bug found in https://github.com/MPAS-Dev/MPAS-Model/pull/624

sbrus89:
FIxed by #626

PR #626, opened 2020-07-03
Ensures windSpeed(U,V) fields are defined

pwolfram:
Previously, index error in synthetic hurricane code
would allow undefined fields to be produced, which
are interpreted in MPAS-O as a larger number, which
would cause unrealistic forcing.

### DIFF
--- a/testing_and_setup/compass/ocean/hurricane/hurricane_wind_pressure/winds_io/output_data.py
+++ b/testing_and_setup/compass/ocean/hurricane/hurricane_wind_pressure/winds_io/output_data.py
@@ -26,7 +26,7 @@ def write_netcdf(filename: str, curr_hurricane: Hurricane, grid: Geogrid, winds:
     # Format time
     ref_date = curr_hurricane[0].ref_time 
     xtime = []
-    for it in range(0,len(curr_hurricane)):
+    for it in range(0,len(curr_hurricane)-1):
       t = curr_hurricane[it].time
       date = ref_date + datetime.timedelta(hours=np.float64(t))
       xtime.append(date.strftime('%Y-%m-%d_%H:%M:%S'+45*' '))


testing_and_setup/compass/ocean/hurricane/hurricane_wind_pressure/winds_io/output_data.py
```diff
@@ -26,7 +26,7 @@ def write_netcdf(filename: str, curr_hurricane: Hurricane, grid: Geogrid, winds:
     # Format time
     ref_date = curr_hurricane[0].ref_time 
     xtime = []
-    for it in range(0,len(curr_hurricane)):
+    for it in range(0,len(curr_hurricane)-1):
```
pwolfram, line 29: The change is needed to ensure defined consistency with line 42 below.

pwolfram:
# Testing
Now, last entry in output file is defined:
```
ncdump -v windSpeedU out.nc | tail                                            
0.139699132152386, 6.45812737267995, 0.714791425321646, 0.7018045522007,      
5.17754476583453, 5.11802122521514, 1.29302178414458, 7.01528829277794,       
1.55769141700045, 5.23614541855713, 1.60587435070129, 6.20990985819468,       
1.57272692454057, 1.65961506024466, -0.646329650719935, 5.37579310636597,     
0.925775043341724, 4.62658368878317, 5.16306890261164, 5.29352525066913,      
0.543297239119926, -0.529453255416567, 5.35144155501153,                      
1.13152511178428, 4.74053131082641, -0.129834501553986, 1.42961014654955,     
3.98238907108978, 7.05777542611746, 0.632375598224051, 0.443204383712118,     
1.07859781979354, -6.03083347560618, 6.67109942488312, 6.51548206708945 ;     
```

pwolfram:
Fixes #625

pwolfram:
There is a potential bigger problem here-- how should the hurricane be defined?  Either it doesn't have the last time step or the initial time step should be 0s.  The issue is that to compute the hurricane velocity requires two time steps, so excluding the last point, as is done here is the best simple approach.

PR #627, opened 2020-07-09
Add script to setup a run directory for the atmosphere cores

MiCurry:
This merge as a script inside the `testing_and_setup/atmosphere` directory that prepares a run directory for both the `init_atmosphere` and `atmosphere` cores.

Currently, the script needs to be run in the test_and_setup/atmosphere as it assumes it is within the desired model to be copied.

### DIFF
--- a/testing_and_setup/atmosphere/setup_atm_run_dir
+++ b/testing_and_setup/atmosphere/setup_atm_run_dir
@@ -1,0 +1,174 @@
+#! /bin/sh
+
+# Setup a run directory for the MPAS init_atmosphere, and atmosphere cores.
+
+######################################################################
+# usage() - Display the usage message
+######################################################################
+usage()
+{
+    printf "Usage: setup_atm_run_dir setup-dir\n"
+}
+
+######################################################################
+# init_atmosphere_setup()
+#
+# $1 - Run directory
+# $2 - MPAS-Model Source Code Directory
+# Setup the init_atmosphere in the run directory ($1) by linking the
+# init_atmosphere exetuable and copying the init_atmosphere namelist and
+# streams from the default_inputs/ directory.
+#
+# On error, the program will exit and will return 1, otherwise, 0 will be
+# returned.
+#
+######################################################################
+init_atmosphere_setup()
+{
+    printf "Setting up the init_atmosphere core ...\n"
+    rundir=$1
+    mpasdir=$2
+
+    # See if the init_amtosphere_model is compiled
+    if ! [ -f "${mpasdir}/init_atmosphere_model" ]; then
+        printf "The MPAS directory does not appear to have the init_atmosphere core compiled!\n"
+        return 1
+    fi
+
+    ln -s "${mpasdir}/init_atmosphere_model" $rundir
+    if [ $? -ne 0 ]; then
+        printf "Failed to link 'init_atmosphere_model' from %s\n" $mpasdir
+        return 1
+    fi
+
+    cp "${mpasdir}/default_inputs/namelist.init_atmosphere" $rundir
+    if [ $? -ne 0 ]; then
+        printf "Failed to copy 'namelist.init_atmosphere' from %s\n" "${mpasdir}/default_inputs"
+        return 1
+    fi
+
+    cp "${mpasdir}/default_inputs/streams.init_atmosphere" $rundir
+    if [ $? -ne 0 ]; then
+        printf "Failed to copy 'streams.init_atmosphere' from %s\n" "${mpasdir}/default_inputs"
+        return 1
+    fi
+
+    printf "Succesfully setup the run directory for the init_atmosphere model\n"
+    return 0
+}
+
+######################################################################
+# atmosphere_setup()
+#
+# $1 - Run directory
+# $2 - MPAS-Model Source Code Directory
+# Setup the atmosphere core in the run directory ($1) by linking the
+# atmosphere_model exeutable, physics lookup tables, and by copying
+# the needed namelist, streams, and stream_lists from the default_inputs/
+# directory.
+#
+# On error, this function 1 will be returned, otherwise, 0 will be
+# returned.
+#
+######################################################################
+atmosphere_setup()
+{
+    printf "Setting up the atmosphere core ...\n"
+    rundir=$1
+    mpasdir=$2
+
+    # See if the amtosphere core is compiled
+    if ! [ -f "${mpasdir}/atmosphere_model" ]; then
+        printf "The MPAS directory does not appear to have the atmosphere core compiled!\n"
+        return 1
+    fi
+
+    ln -s "${mpasdir}/atmosphere_model" $rundir
+    if [ $? -ne 0 ]; then
+        printf "Failed to link 'atmosphere_model' from %s\n" $mpasdir
+        return 1
+    fi
+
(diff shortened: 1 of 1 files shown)


mgduda:
It doesn't look like the symbolic links in the run directory are created correctly if the path to the script is given as a relative path. For example, if I run:
```
./MPAS-Model/testing_and_setup/atmosphere/setup_atm_run_dir.sh terrain
```
the `terrain` directory contains the following link to the `init_atmopshere_model` program:
```
init_atmosphere_model -> ./MPAS-Model/init_atmosphere_model
```

mgduda:
It appears that the script may be assuming that `/bin/sh` is the bash shell (e.g., in the use of the `$BASH_SOURCE` environment variable). As was pointed out in PR #594, Ubuntu systems may to alias `/bin/sh` to the dash shell, in which case I don't think this script will work.

Although [this bash FAQ](http://mywiki.wooledge.org/BashFAQ/028) warns against using `$0` ("$0 is NOT the answer to your problem. If you think it is, you can either stop reading and write more bugs, or you can accept this and read on."), I don't think we need to worry about someone running this script via SSH over a network, or removing the `MPAS-Model` directly between the time the script is run and the time we link files from that directory.

MiCurry:
> It doesn't look like the symbolic links in the run directory are created correctly if the path to the script is given as a relative path. For example, if I run:
> 
> ```
> ./MPAS-Model/testing_and_setup/atmosphere/setup_atm_run_dir.sh terrain
> ```
> 
> the `terrain` directory contains the following link to the `init_atmopshere_model` program:
> 
> ```
> init_atmosphere_model -> ./MPAS-Model/init_atmosphere_model
> ```

Hmm okay, I'll look into that.

MiCurry:
> It appears that the script may be assuming that `/bin/sh` is the bash shell (e.g., in the use of the `$BASH_SOURCE` environment variable). As was pointed out in PR #594, Ubuntu systems may to alias `/bin/sh` to the dash shell, in which case I don't think this script will work.
> 
> Although [this bash FAQ](http://mywiki.wooledge.org/BashFAQ/028) warns against using `$0` ("$0 is NOT the answer to your problem. If you think it is, you can either stop reading and write more bugs, or you can accept this and read on."), I don't think we need to worry about someone running this script via SSH over a network, or removing the `MPAS-Model` directly between the time the script is run and the time we link files from that directory.

Not using `$BASH_SOURCE` sounds good to me. Using `$0` seems safe enough for our usage, but it also might be easier to pass in the `MPAS_Model` source code as an argument and copy the needed files from that directory. @mgduda do you have any thoughts and opinions on that?

mgduda:
I think requiring the `MPAS-Model` source directory as a command-line argument would make the script more cumbersome to use. Ultimately, I think it would be good to make this script as easy to run as possible. For example, a common use for me would be to compile MPAS-A in my home directory, then add `${HOME}/MPAS-Model/testing_and_setup/atmosphere` to my `$PATH`, and then in scratch space, any time I need to fill out a run directory, I could just run `setup_atm_run_dir.sh whatever`.

There's also the danger that if a user runs the `setup_atm_run_dir.sh` script from one version of MPAS-A and points to the source directory from another version, the correct files might not be linked (e.g., because future releases of MPAS-A might add new look-up tables, remove separate stream_list files, etc.).

MiCurry:
@mgduda that sounds good and reasonable to me!

MiCurry:
@mgduda I have updated this PR to address the problem of using a relative execution (by using `readlink`) and replaced `$BASH_SOURCE` with `$0`.

mgduda:
It looks like the options for `readlink` may not be portable across different systems. Here's what I get when I test on my macOS system:
```
$ ./MPAS-Model/testing_and_setup/atmosphere/setup_atm_run_dir terrain
readlink: illegal option -- f
usage: readlink [-n] [file ...]

usage: dirname path
ERROR: Can't seem to locate MPAS-Model directory!
ERROR: Please ensure that this script is in the testing_and_setup/atmosphere directory of
ERROR: the MPAS-Model you want to setup

```

testing_and_setup/atmosphere/setup_atm_run_dir
```diff
@@ -0,0 +1,175 @@
...
+# setup_run_atm_run_dir.sh
+#
+# \brief Copy and link the needed files for running the init_atmosphere, and
+# atmosphere core.
+# \details
+# Given a directory, copy and/or link all the needed executables, namelist,
+# streams, stream_lists and physics lookup tables needed for both the
+# init_atmosphere and atmosphere core.
+#
+# Currently, this script will need to be called from the
```
mgduda, line 133: This is no longer true, right?

testing_and_setup/atmosphere/setup_atm_run_dir
```diff
@@ -0,0 +1,175 @@
...
+# atmosphere_setup()
+#
+# $1 - Run directory
+# $2 - MPAS-Model Source Code Directory
+# Setup the atmosphere core in the run directory ($1) by linking the
+# atmosphere_model exeutable, physics lookup tables, and by copying
+# the needed namelist, streams, and stream_lists from the default_inputs/
+# directory.
+#
+# On error, this function -1 will be returned, otherwise, 0 will be
```
mgduda, line 71: It looks like `atmosphere_setup()` will return 1 on error rather than -1.

testing_and_setup/atmosphere/setup_atm_run_dir
```diff
@@ -0,0 +1,175 @@
...
+######################################################################
+# init_atmosphere_setup()
+#
+# $1 - Run directory
+# $2 - MPAS-Model Source Code Directory
+# Setup the init_atmosphere in the run directory ($1) by linking the
+# init_atmosphere exetuable and copying the init_atmosphere namelist and
+# streams from the default_inputs/ directory.
+#
+# On error, the program will exit and will return -1, otherwise, 0 will be
```
mgduda, line 23: It looks like `init_atmosphere_setup()` will return 1 on error rather than -1.

testing_and_setup/atmosphere/setup_atm_run_dir
MiCurry: I think I had the wrong wording here to begin with. I think what I meant to say was that the script needed to be _in_ the `testing_and_setup/atmosphere/` directory, rather than _called_. I've updated this comment to be clearer.

MiCurry:
@mgduda I finally took a look over and cleaned up some of the comments in `setup_atm_run_dir` from your comments and some things I wanted to reword.

Unless you see something else that is missing or think this script could use something more I think this PR is good to go!

mgduda:
In the commit message, could you change "atmosph" to "atmosphere"?  I think changing "dir" to "directory" would still let the first line fit in 80 columns. The commit message also doesn't really describe what the script does. Rather than some of the discussion about directory structure, it might be more helpful to explain what preparation of a run directory actually entails, and the rationale for copying some files and linking others.

MiCurry:
@mgduda after a brief hiatus from this I have updated the commit message for this PR. I have also updated the PR message.

I'm never quite sure what to do for the merge message with a PR with one commit. Should I write a brief of the original commit message? Or copy the commit message? Do you have any advice/guidance @mgduda?

mgduda (approved):

PR #629, opened 2020-07-10
Update CMake builds for pre-built MPAS tools

amametjanov:
Avoid building `namelist_gen`, `streams_gen` and `parse` tools if
pre-built tools are available in MPAS_TOOL_DIR environment variable.
This helps building MPAS in E3SM with Cray compilers (`cce/9.0.2-classic`) on ALCF Theta.

[BFB]


mark-petersen:
@amametjanov, this looks pretty straightforward. Could you tell us which computers and compilers you've tested on? 

@jonbob could you run a few standard tests with this merged locally on anvil?

mark-petersen:
@jgfouca since you wrote these cmake files originally, do you have any comments in this addition?

mark-petersen:
Tested on anvil with `PEM_Ln9.T62_oQU240.GMPAS-IAF`. Compiled without trouble.

PR #630, opened 2020-07-13
Update mpas_kd_tree.F to break ties on equidistant queries

MiCurry:
Previous to this commit if a query in is equidistant from two points of the KD-Tree, the resulting 'closet' point (from mpas_kd_search) would be the last point that the KD-Tree visited in its search.

This is fine for single processors, but if two tasks share the same point (along the boundary of two decompositions) they each can decide that a single query lies closer to two different points. For the static field interpolation this would mean that a single data value can be counted twice, which produces differences static-fields across task counts.

This commit breaks these ties so that each task will choose the same node that a query belongs too.

### DIFF
--- a/src/core_init_atmosphere/mpas_kd_tree.F
+++ b/src/core_init_atmosphere/mpas_kd_tree.F
@@ -158,6 +158,45 @@ module mpas_kd_tree
 
    end function mpas_kd_construct
 
+   !***********************************************************************
+   !
+   !  routine break_tie
+   !
+   !> \brief   Break a tie for two n-dim points
+   !> \author  Miles A. Curry
+   !> \date    07/07/20
+   !> \details
+   !> Compare 1..n dimensions of p1 and p2 and return -1 if p1(i) is less than
+   !> p2(i) and return 1 if p1(i) is greater than p2(i). If p1(i) and p2(i) are
+   !> equal, then the same comparison will be done on p1(i+1) and p2(i+1) until
+   !> p1(n) and p2(n). If p1(:) and p2(:) are equal across all n, then 0 will
+   !> be returned.
+   !
+   !-----------------------------------------------------------------------
+   function break_tie(p1, p2) result(tie)
+
+      implicit none
+
+      ! Input Variables
+      type (mpas_kd_type), intent(in) :: p1
+      type (mpas_kd_type), intent(in) :: p2
+      integer :: tie
+
+      integer :: i
+
+      tie = 0
+      do i = 1, size(p1 % point(:))
+          if (p1 % point(i) < p2 % point(i)) then
+              tie = -1
+              return
+          else if (p1 % point(i) > p2 % point(i)) then
+              tie = 1
+              return
+          endif
+      enddo
+
+   end function break_tie
+
 
    !***********************************************************************
    !
@@ -192,6 +231,15 @@ module mpas_kd_tree
       if (current_distance < distance) then
          distance = current_distance
          res => kdtree
+      else if (current_distance == distance) then
+          !
+          ! Consistently break a tie if a query is equidistant from two points
+          !
+          if (associated(res)) then
+              if (break_tie(res, kdtree) == 1) then
+                 res => kdtree
+              endif
+          endif
       endif
 
       !


MiCurry:
I have added an additional test to the `mpas_init_atm_core.F` Github gist that I used to test the original KD-Tree, which can be found here: https://gist.github.com/MiCurry/b5744ef36eec3b3ba40a8de806201ed1

mgduda:
It seems like having a module variable, `kd_tree_ndims`, to store the number of dimensions in a kd-tree would prevent us from creating multiple trees of differing dimensionality. Could the dimensionality of a tree be moved to a member of the `mpas_kd_type`? Alternatively, where `kd_tree_ndims` is used, would it not be possible to instead use, e.g., `size(p1 % point(:))`?

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -187,11 +231,25 @@ recursive subroutine mpas_kd_search_internal(kdtree, query, res, distance)
...
+      else if (current_distance == distance) then
+          !
+          ! Queries that are equal-distant from two points may end up on inteior cells of
+          ! two decompositions. So ties must be broken arbitrarially so different tasks agree
+          ! on the correct location of of a query.
+          !
+          if (associated(res)) then
+              tie = break_tie(res, kdtree)
+              if (tie == 1) then
+                 distance = current_distance
```
mgduda, line 249: Is this assignment necessary? L.240 should guarantee that the l.h.s here is the same as the r.h.s.

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -187,11 +231,25 @@ recursive subroutine mpas_kd_search_internal(kdtree, query, res, distance)
...
          res => kdtree
+      else if (current_distance == distance) then
+          !
+          ! Queries that are equal-distant from two points may end up on inteior cells of
+          ! two decompositions. So ties must be broken arbitrarially so different tasks agree
+          ! on the correct location of of a query.
+          !
+          if (associated(res)) then
+              tie = break_tie(res, kdtree)
+              if (tie == 1) then
```
mgduda, line 248: Instead of having an intermediate variable, `tie`, an alternative would be to just test directly on the return value of `break_tie`:
```fortran
if (break_tie(res, kdtree) == 1)
```

MiCurry:
> It seems like having a module variable, `kd_tree_ndims`, to store the number of dimensions in a kd-tree would prevent us from creating multiple trees of differing dimensionality. Could the dimensionality of a tree be moved to a member of the `mpas_kd_type`? Alternatively, where `kd_tree_ndims` is used, would it not be possible to instead use, e.g., `size(p1 % point(:))`?

Ah that is right. I forgot that `mpas_kd_tree` is a module and not a class.  My thought, which was probably incorrect, was that using a module variable would be quicker to access than using `size`, but with more thought, I'm guessing that the `size` intrinsic is pretty efficient, since Fortran size and shape information is stored with the variable.

I'll update it to use `size(p1 % point(:))`.

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: You're correct! Good point!

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: That sounds good.

Originally I had an `else if` statement that caught when `break_tie` returned 0. Do you think its worth trying to catch when `break_tie` returns 0 and produce an error? i.e. when two points of the KD Tree are the same?

mgduda: I don't think we need to consider it an error if there are two identical points in the kd-tree. If the two points are actually identical, then it doesn't matter which of them we return.

In the longer term, if we decide that it is erroneous to have duplicate points in the kd-tree, then I think the place to check for that would be in the construction of the tree, rather than while searching the kd-tree.

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -158,6 +161,47 @@ function mpas_kd_construct(points, ndims) result(tree)
...
 
+   !***********************************************************************
+   !
+   !  routine break_tie
+   !
+   !> \brief   Break a tie for two n-dim points
+   !> \author  Miles A. Curry
+   !> \date    07/07/20
+   !> \details
+   !> Compare 1..n dimensions of p1 and p2 and return -1 if p1(n) is less then
```
mgduda, line 172: "then" -> "than"

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -158,6 +161,47 @@ function mpas_kd_construct(points, ndims) result(tree)
...
+   !> point), then 0 will be returned.
+   !
+   !-----------------------------------------------------------------------
+   function break_tie(p1, p2) result(tie)
+
+      implicit none
+
+      ! Input Variables
+      type (mpas_kd_type), target, intent(in) :: p1
+      type (mpas_kd_type), target, intent(in) :: p2
```
mgduda, line 185: Is the `target` attribute needed for `p1` and `p2`?

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: That sounds good to me.

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: It is not, I always forget that Fortran is pass by reference.

src/core_init_atmosphere/mpas_kd_tree.F
mgduda: Should we remove this assignment?

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -158,6 +158,47 @@ function mpas_kd_construct(points, ndims) result(tree)
...
+      type (mpas_kd_type), intent(in) :: p2
+      integer :: tie
+
+      integer :: i
+
+      tie = 0
+      do i = 1, size(p1 % point(:))
+          ! Go to the next dimension for comparision if the two points are equal on a dimension
+          if (p1 % point(i) == p2 % point(i)) then
+             cycle
```
mgduda, line 191: In my opinion, it's a little awkward to have a `cycle` statement here. Wouldn't the following loop without a `cycle` statement work?
```fortran
      do i = 1, size(p1 % point(:))                                                                                                  
          if (p1 % point(i) < p2 % point(i)) then                                                                                    
              tie = -1                                                                                                               
              return                                                                                                                 
          else if (p1 % point(i) > p2 % point(i)) then                                                                               
              tie = 1                                                                                                                
              return                                                                                                                 
          end if                                                                                                                     
      end do 
```

mgduda:
There are a couple of places where the commit messages or code comments make references to interior/exterior cells, or to different MPI tasks, and I think the problem that this PR is addressing is not inherently tied to either of these. For example, a serial program could construct two different kd-trees that share points and be affected by the issue of searches in one tree or the other returning different "nearest" points. As such, I'd suggest we keep the descriptions general.

src/core_init_atmosphere/mpas_kd_tree.F
mgduda: Also, I'm not sure whether the loop as written (with the `cycle` statement) works as expected. For example, if I define two points:
```fortran
      p1 % point(:) = [1.0, 1.0, 0.0]                                            
      p2 % point(:) = [0.0, 0.0, 1.0]    
```
the `break_tie` function returns -1 (i.e., p1 < p2), which isn't what I was expecting. However, it could be that I've misunderstood the expected behavior. Ultimately, as long as the function works deterministically, we should be fine; but, it's always nice when the function behavior matches what most people would intuitively expect.

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: Yes, looks like I missed it in the last update.

src/core_init_atmosphere/mpas_kd_tree.F
MiCurry: @mgduda I think the portion of code is also a bit awkward, particularly I don't like the double if statements. I like your implementation better and I'll switch it over to that.

Too your second comment, I think -1 was being returned as there were a few return statements missing. When I wrote my unit test, I wrote a test and figured it was enough, but in the back of my mind I knew I needed a few more, and it appears that thought was right, so I'll add a few more tests to make sure we've got good coverage.

MiCurry:
> There are a couple of places where the commit messages or code comments make references to interior/exterior cells, or to different MPI tasks, and I think the problem that this PR is addressing is not inherently tied to either of these. For example, a serial program could construct two different kd-trees that share points and be affected by the issue of searches in one tree or the other returning different "nearest" points. As such, I'd suggest we keep the descriptions general.

That sounds great. Thanks for the note on this. I was actually wondering how exactly the commit message should be worded. I agree, I'll update it to be a bit more general.

MiCurry:
Okay, @mgduda. I have updated the comparison loop (so that it no longer contains `cycle`), removed the `distance = current_distance` assignment and generalized the commit message, and comments. Let me know if you think they could still be improved at all.

mgduda:
The commit message mentions 'closet' in a couple of places; I think this should be 'closest'.

Also, in the commit message, it's stated that "the resulting 'closet' point (from mpas_kd_search) would be the last
point that the KD-Tree visited in its search." Is this correct? I admit that I haven't reviewed the full logic of the search function, but from this block it looks like the point that is returned would be the first one that was encountered:
```fortran
      if (current_distance < distance) then
         distance = current_distance
         res => kdtree
```

MiCurry:
> The commit message mentions 'closet' in a couple of places; I think this should be 'closest'.

That's embarrassing! Thanks for the catch.

> Also, in the commit message, it's stated that "the resulting 'closet' point (from mpas_kd_search) would be the last
> point that the KD-Tree visited in its search." Is this correct? I admit that I haven't reviewed the full logic of the search function, but from this block it looks like the point that is returned would be the first one that was encountered:

You are correct! I was wrong. I just checked with my small toy program and it point chosen (originally) on equidistant queries is the _first_ point encountered.

I have just pushed an update that fixes the spelling mistakes and the above error.

src/core_init_atmosphere/mpas_kd_tree.F
```diff
@@ -158,6 +158,45 @@ function mpas_kd_construct(points, ndims) result(tree)
...
+   !
+   !  routine break_tie
+   !
+   !> \brief   Break a tie for two n-dim points
+   !> \author  Miles A. Curry
+   !> \date    07/07/20
+   !> \details
+   !> Compare 1..n dimensions of p1 and p2 and return -1 if p1(i) is less than
+   !> p2(i) and return 1 if p1(i) is greater than p2(i). If p1(i) and p2(i) are
+   !> equal, than the same comparison will be done on p1(i+1) and p2(i+1) until
```
mgduda, line 171: "than" -> "then"

mgduda:
@MiCurry Other than one other minor typo, I think this PR is ready to be merged. Since we have the opportunity to easily do so, though, it seems worth changing "than" to "then" as noted in the preceding comment.

MiCurry:
Okay, @mgduda I have just pushed a fix for that typo, so we should be good to merge!

mgduda:
@MiCurry It seems GitHub doesn't allow me to add a review if I've already pushed the merge commit. Needless to say, I've approved this PR.

PR #635, opened 2020-07-17
Clone submodules over HTTPS, not SSH

Helumpago:
This PR updates the `.gitmodules` settings so that Git will clone the [CVMix](https://github.com/CVMix/CVMix-src) and [BGC](https://github.com/E3SM-Project/Ocean-BGC.git) repositories over HTTPS instead of SSH.

This is important for systems (e.g., build servers) that don't have an SSH key associated with GitHub. These systems need to clone public repositories using anonymous HTTPS.

Prior to this change, attempts to build this repository would emit the following error: `Missing core_ocean/cvmix/.git, did you forget to 'git submodule update --init --recursive' ?` Due to not having an SSH key registered with GitHub, running the requested `git submodule update --init --recursive` line fails due to insufficient access rights.

After this PR, the update command works correctly on anonymous systems.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "src/core_ocean/cvmix"]
 	path = src/core_ocean/cvmix
-	url = git@github.com:CVMix/CVMix-src.git
+	url = https://github.com/CVMix/CVMix-src.git
 [submodule "src/core_ocean/BGC"]
 	path = src/core_ocean/BGC
-	url = git@github.com:E3SM-Project/Ocean-BGC.git
+	url = https://github.com/E3SM-Project/Ocean-BGC.git


Helumpago:
Thanks for reviewing.

@philipwjones, will that cause issues if someone updates the references for your submodules? For people who have already built MPAS, the `.git` file in the submodules will already exist, which means the update won't be triggered. The old files will stick around for other people.

Would it be better to drop the conditional and always run the update as part of the `libcvmix` and `libBGC` targets?

Helumpago:
Actually, I would really appreciate it if the build system _didn't_ try to automatically update the submodules. Doing so will cause a couple issues in container-based environments:
1. `git` generally isn't installed in containers.
1. Most container build systems are configured to ignore `.git` directories. This means that, even if someone does install `git` into their container, they won't be able to run the submodule update.

So, for what it's worth, I'd much prefer the explicit error to implicitly trying to download the submodules.

philipwjones:
@Helumpago - that's fine with me. I don't have a problem with the current setup and don't think it's too much of a burden to either remember to update after checkout or get a reminder if you don't.  I think others have found it more of an annoyance, so we had been exploring alternatives.

But I do think we should make your change at least - that does save some extra password queries and allows you to walk away from a long build without it hanging up waiting for you to enter a password.

And yes, you're correct that the changes to the makefile would not pick up an update in the submodule reference so you might be running with an old submodule. But that's also true of the current situation.

PR #637, opened 2020-07-22
Add a class to help manage and interpolate static datasets

MiCurry:
This merge adds a new module to the init_atmosphere core, mpas_geotile_manager, which contains a class, mpas_geotile_mgr_type, which can be used to help read, store and interpolate static fields. Currently, it is only being compiled, but in the future will be used to parallelize the interpolation of static fields.

### DIFF
--- a/src/core_init_atmosphere/mpas_geotile_manager.F
+++ b/src/core_init_atmosphere/mpas_geotile_manager.F
@@ -1,0 +1,1060 @@
+module mpas_geotile_manager
+
+    use iso_c_binding, only : c_float, c_char
+
+    use mpas_constants, only : pii
+    use mpas_kind_types, only : RKIND, StrKIND
+    use mpas_log, only : mpas_log_write
+    use mpas_derived_types, only : MPAS_LOG_ERR, MPAS_POOL_SILENT
+    use mpas_pool_routines, only : mpas_pool_type, mpas_pool_destroy_pool, mpas_pool_create_pool
+    use mpas_pool_routines, only : mpas_pool_add_config, mpas_pool_get_config
+    use mpas_pool_routines, only : mpas_pool_get_error_level, mpas_pool_set_error_level
+    use mpas_stack
+
+    implicit none
+
+    public :: mpas_geotile_mgr_type
+    public :: mpas_geotile_type
+
+    private
+
+    type mpas_geotile_mgr_type
+        type (mpas_pool_type), pointer :: pool
+        type (tile_hash), dimension(:,:), pointer :: hash
+        type (mpas_stack_type), pointer :: stack
+
+        character (len=StrKIND) :: directory ! Path to the dataset directory
+        character (len=StrKIND) :: index     ! Path the index file of the dataset directory
+
+        integer :: nTileX ! Number of tiles in the X direction
+        integer :: nTileY ! Number of tiles in the Y direction
+        integer :: pixel_nx ! Total number of pixels in the x direction
+        integer :: pixel_ny ! Total number of pixels in the y direction
+    contains
+        ! Public Procedures
+        procedure, public :: init => mpas_geotile_mgr_init
+        procedure, public :: finalize => mpas_geotile_mgr_finalize
+        procedure, public :: get_tile => mpas_geotile_mgr_get_tile
+        procedure, public :: pixel_to_latlon => mpas_geotile_mgr_pixel_to_latlon
+        procedure, public :: latlon_to_pixel => mpas_geotile_mgr_latlon_to_pixel
+        procedure, public :: tile_to_latlon => mpas_geotile_mgr_tile_to_latlon
+        procedure, public :: push_neighbors => mpas_geotile_mgr_push_neighbors
+
+        ! Stack Procedures
+        procedure, public :: push_tile => mpas_geotile_mgr_push_tile
+        procedure, public :: pop_tile => mpas_geotile_mgr_pop_tile
+        procedure, public :: is_stack_empty => mpas_geotile_mgr_stack_is_empty
+
+        ! Private Procedures
+        procedure, private :: search_tile => mpas_geotile_mgr_search_tile
+        procedure, private :: add_tile => mpas_geotile_mgr_add_tile
+        procedure, private :: gen_filename => mpas_geotile_mgr_gen_tile_name
+        procedure, private :: hash_to_ll => mpas_geotile_mgr_hash_to_latlon
+    end type mpas_geotile_mgr_type
+
+
+    type, extends(mpas_stack_payload_type) :: mpas_geotile_type
+        real (c_float), dimension(:,:,:), pointer :: tile
+
+        character (len=StrKIND) :: fname ! Path to the file that contains the data for this tile
+        integer :: hash_x ! The x offset of this tile in the hash table
+        integer :: hash_y ! The y offset of this tile in the hash table
+
+        integer :: x, y ! The tiles range, in pixels
+        logical :: is_processed = .false.
+    end type mpas_geotile_type
+
+
+    type tile_hash
+        type(mpas_geotile_type), pointer :: ptr => null()
+    end type tile_hash
+
+
+    contains
+
+
+    !***********************************************************************
+    !
+    !  public function mpas_geotile_mgr_init => init
+    !
+    !> \brief   Initialize a mpas_geotile_mgr class
+    !> \author  Miles A. Curry
+    !> \date    02/20/2020
+    !> \details
+    !> Initialize a geotile manager class by parsing the index file located
+    !> within path and allocated needed data structures for static interpolation.
+    !> Init should be called before calling any other mpas_geotile_mgr_type
+    !> procedures. If path is not a directory or no index file is found in path,
+    !> 1 will be returned. Upon success 0 will be returned.
+    !>
+    !> This function will also allocate the following variables in the pool attribute
+    !> of this geotile manager instance if they are not found within the index file:
+    !>     * tile_bdr = 0
+    !>     * signed = 0 ! No
+    !>     * scalefactor = 1.0_RKIND
+    !>     * endian = "big"
+    !
+    !-----------------------------------------------------------------------
+    function mpas_geotile_mgr_init(mgr, path) result(ierr)
+
+        use mpas_parse_geoindex, only : mpas_parse_index
+
+        implicit none
+
+        ! Input variables
+        class (mpas_geotile_mgr_type) :: mgr
+        character (len=*), intent(in) :: path
+
+        ! Local variables
+        character (len=StrKIND), pointer :: endian
+        integer, pointer :: tile_nx ! Number of pixels in the x-direction for a single tile
+        integer, pointer :: tile_ny ! Number of pixels in the y-direction for a single tile
+        integer, pointer :: signed
+        integer, pointer :: tile_bdr
+        integer :: err_level
+        real (kind=RKIND), pointer :: dx ! Grid spacing in the x-direction
+        real (kind=RKIND), pointer :: dy ! Grid spacing in the y-direction
+        real (kind=RKIND), pointer :: scalefactor
+        logical :: res
+
+        ! Return variable
+        integer :: ierr
+
+        ierr = 0
+
+        mgr % directory = path
+
+        ! Check to see if the index file exists in the directory
+        inquire(file=trim(mgr % directory)//"index", exist=res)
+        if (.not. res) then
+            call mpas_log_write("Could not find an 'index' file in geotile directory: "//trim(path), messageType=MPAS_LOG_ERR)
+            ierr = 1
+            return
+        endif
+        mgr % index = trim(mgr % directory)//"index"
+
+        ! Create the pool for this geotile and call mpas_parse_index
+        call mpas_pool_create_pool(mgr % pool)
+        ierr = mpas_parse_index(mgr % index, mgr % pool)
+        if (ierr /= 0) then
+            call mpas_log_write("Error parsing geotile index file: "//trim(mgr % index), messageType=MPAS_LOG_ERR)
+            ierr = 1
+            return
+        endif
+
+        signed => null()
+        endian => null()
+        scalefactor => null()
+        tile_bdr => null()
+
+        err_level = mpas_pool_get_error_level()
+        call mpas_pool_set_error_level(MPAS_POOL_SILENT)
+
+        call mpas_pool_get_config(mgr % pool, 'signed', signed)
+        call mpas_pool_get_config(mgr % pool, 'endian', endian)
+        call mpas_pool_get_config(mgr % pool, 'scale_factor', scalefactor)
+        call mpas_pool_get_config(mgr % pool, 'tile_bdr', tile_bdr)
+
+
+        !
+        ! tile_bdr, signed, endian, and scale_factor all have default values, so if they
+        ! are not present in the index file then set them as the default values, as
+        ! reported in section 3-53 of the WRF-ARW User's Guide
+        !
+        if (.not. associated(endian)) then
+            allocate(endian)
+            endian = "big"
+            call mpas_pool_add_config(mgr % pool, 'endian', endian)
+        endif
+
+        if (.not. associated(scalefactor)) then
+            allocate(scalefactor)
+            scalefactor = 1.0_RKIND
+            call mpas_pool_add_config(mgr % pool, 'scale_factor', scalefactor)
+        endif
+
+        if (.not. associated(signed)) then
+            allocate(signed)
+            signed = 0 ! 0 = 'no', 1 = 'yes'
+            call mpas_pool_add_config(mgr % pool, 'signed', signed)
+        endif
+
+        if (.not. associated(tile_bdr)) then
+            allocate(tile_bdr)
+            tile_bdr = 0
+            call mpas_pool_add_config(mgr % pool, 'tile_bdr', tile_bdr)
+        endif
+
+        ! Reset the pool's error level
+        call mpas_pool_set_error_level(err_level)
+
+        call mpas_pool_get_config(mgr % pool, 'tile_x', tile_nx)
+        call mpas_pool_get_config(mgr % pool, 'tile_y', tile_ny)
+        call mpas_pool_get_config(mgr % pool, 'dx', dx)
+        call mpas_pool_get_config(mgr % pool, 'dy', dy)
+
+        ! Calculate the total number of pixels in x dir
+        ! NOTE: This calculation assumes that a dataset is a global dataset and may
+        ! not work correctly for non-global datasets
+        mgr % pixel_nx = nint(360.0_RKIND / dx)
+        mgr % pixel_ny = nint(180.0_RKIND / dy)
+
+        ! Calculate the number of tiles in the x, y directions
+        ! NOTE: This calculation assumes that a dataset is a global dataset and may
+        ! not work correctly for non-global datasets
+        mgr % nTileX = mgr % pixel_nx / tile_nx
+        mgr % nTileY = mgr % pixel_ny / tile_ny
+
+        ! Allocate hash table
+        allocate(mgr % hash(0: mgr % nTileX, 0: mgr % nTileY))
+
+        ! Mark the stack as empty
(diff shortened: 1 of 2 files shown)


src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1000 @@
...
+            call mpas_log_write("Error generating filename", messageType=MPAS_LOG_ERR)
+            return
+        endif
+
+        !
+        ! Access the tile in the hash table (-1 here as the hash table is from
+        ! 0:tile_nx, and 0:tile_ny).
+        !
+        x = (start_x - 1) / tile_nx
+        if (x < 0 .or. y > size(mgr % hash, 1)) then
```
mgduda, line 373: As pointed out by the gfortran compiler, it seems that `y` is indeed being used before it has been defined here.
```
mpas_geotile_manager.F:373:0:                                                                                                        
                                                                                                                                     
  373 |         if (x < 0 .or. y > size(mgr % hash, 1)) then                                                                         
      |                                                                                                                              
Warning: ‘y’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1000 @@
...
+            call mpas_log_write("There was a problem getting the right tile", messageType=MPAS_LOG_ERR)
+            ierr = 1
+            return
+        endif
+        ierr = mgr % push_tile(neighbor)
+
+        ! Left
+        neighbor => null()
+        if (tile % hash_x -1 < 0) then
+            xHash = nTileX - 1
```
mgduda, line 847: The gfortran compiler seems correct in identifying `nTileX` as being used here before it has been defined:
```
mpas_geotile_manager.F:847:0:                                                                                                        
                                                                                                                                     
  847 |             xHash = nTileX - 1                                                                                               
      |                                                                                                                              
Warning: ‘ntilex’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1000 @@
...
+        call mpas_pool_get_config(mgr % pool, 'tile_x', tile_nx)
+        call mpas_pool_get_config(mgr % pool, 'tile_y', tile_ny)
+        call mpas_pool_get_config(mgr % pool, 'dx', dx)
+        call mpas_pool_get_config(mgr % pool, 'dy', dy)
+
+        ! Calculate the total number of pixels in x dir
+        ! NOTE: This calculation assumes that a dataset is a global dataset and may
+        ! not work correctly for non-global datasets
+        mgr % pixel_nx = 360.0_RKIND / dx 
+        mgr % pixel_ny = 180.0_RKIND / dy
```
mgduda, line 192: When computing `mgr % pixel_nx` and `mgr % pixel_ny`, I think we should be careful to round to the nearest integer rather than truncating to an integer. For example, a 15-arc-second dataset would have dx ~= 0.00416666667, which would give a `mgr % pixel_nx` value of 86399 with the code as written, when the correct value is 86400.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1000 @@
...
+    !> \brief   Translate a tile indices to latitude and longitude,
+    !> \author  Miles A. Curry
+    !> \date    02/20/2020
+    !> \details
+    !> Given a tile, translate the relative tile coordinates, i, j, of that
+    !> tile to a latitude and longitude coordinate. Upon success, lat, lon
+    !> will be in the range of -1/2 * pi to 1/2 * pi and 0 to 2.0 * pi, respectively.
+    !
+    !-----------------------------------------------------------------------
+    subroutine mpas_geotile_mgr_tile_to_latlon(mgr, tile, j, i, lat, lon)
```
MiCurry, line 617: When testing this function, I was a bit confused by the range of latitudes that it produced. If I grabbed the most north east tile `mgr % get_tile(5. * pi, 2.0 * pi, tile)` some of the range of latitudes produced by `tile_to_latlon` were greater than .5 * pi, and I'm not sure exactly why that would be.

A first thought is that this is because `known_lat` is not equal to exactly 5. * -pi. @mgduda, is that correct?

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1000 @@
...
+        endif
+        ierr = mgr % push_tile(neighbor)
+
+        ! Down
+        neighbor => null()
+        if (tile % hash_y - 1 < 0) then
+            xHash = modulo(tile % hash_x + (mgr % nTileX / 2), mgr % nTileX)
+            yHash = tile % hash_y
+        else
+            xHash = tile % hash_x + 1
```
mgduda, line 812: Why is 1 being added to `tile % hash_x` here?

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1000 @@
...
+            ierr = 1
+            return
+        endif
+        ierr = mgr % push_tile(neighbor)
+
+        ! Right
+        neighbor => null()
+        if (tile % hash_x + 1 > mgr % nTileX - 1) then
+            yHash = 1
+            xHash = tile % hash_y
```
mgduda, line 829: I don't quite understand the logic for setting `yHash` and `xHash` here.

src/core_init_atmosphere/mpas_geotile_manager.F
mgduda: Typically, the `known_lat` and `known_lon` values for a dataset give the coordinates of the *center* of a pixel. So if the boundary of the pixel extends to, e.g., the South Pole, then the pixel's center coordinates will be greater than -Pi/2.

src/core_init_atmosphere/mpas_geotile_manager.F
MiCurry: nTileX and nTileY were both actually being defined here in the deceleration section, but not used as they are being replaced by `mgr % nTileX` and `mgr % ntileY`.

MiCurry:
@mgduda thanks for the primarily review. I just pushed an update fixing the warnings and other issues you found.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,998 @@
...
+        
+        ! Return variable
+        integer :: ierr
+
+        ierr = 0
+
+        ! Check to see if the directory exists
+        inquire(file=trim(path), exist=res)
+        if (.not. res) then
+            call mpas_log_write("Could not find geotile directory: "//path, messageType=MPAS_LOG_ERR)
```
MiCurry, line 121: `path` needs to be `trim(path)` here.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,998 @@
...
+            ierr = 1
+            return
+        endif
+        mgr % directory = path
+
+        ! Check to see if the index file exists in the directory
+        inquire(file=trim(mgr % directory)//"index", exist=res)
+        if (.not. res) then
+            call mpas_log_write("Could not find an 'index' file in geotile directory: "//path, messageType=MPAS_LOG_ERR)
```
MiCurry, line 130: Aswell, `path` needs to be `trim(path)` here.

MiCurry:
@mgduda, I have just pushed an update to this branch that includes the changes above. I have pulled them out of 469c271 and moved them into 9111cfb and I have added the `convert_lx` routine in fe0709a.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1025 @@
...
+        real (kind=RKIND), pointer :: scalefactor
+        logical :: res
+
+        ! Return variable
+        integer :: ierr
+
+        ierr = 0
+
+        ! Check to see if the directory exists
+        inquire(file=trim(path), exist=res)
```
MiCurry, line 118: It seems to appear that ifort will set `res` to `.false.` on the above `inquire` call, but gfortran does not.

mgduda: This check may not really be necessary. The check in the next code block on whether the `index` file exists is a sufficient condition for the directory existing.

MiCurry: Actually, taking a look at the standard, it doesn't appear that `inquire` is able to work on directories. From section 9.9 of the Fortran 2003 standard:

> The INQUIRE statement may be used to inquire about properties of a particular named file or of the
> 25 connection to a particular unit.

And then in 9.4:

> A unit, specified by an io-unit, provides a means for referring to a file.

So it looks like gfortran's ability to inquire on a directory is an extension, similarly to ifort's `directory` keyword.


I also don't think this `inquire` statement is really needed, as the next `inquire` (into the index file) will return `.false.` if the directory doesn't exist. So I think its safe to remove it.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1025 @@
...
+        call mpas_pool_get_config(mgr % pool, 'known_lon', known_lon)
+        call mpas_pool_get_config(mgr % pool, 'known_lat', known_lat)
+        call mpas_pool_get_config(mgr % pool, 'dx', dx)
+        call mpas_pool_get_config(mgr % pool, 'dy', dy)
+
+        call rad2Deg(lat)
+        call rad2Deg(lon)
+
+        pixel_x = int((lon - known_lon) / dx)
+        pixel_y = int((lat - known_lat) / dy)
```
mgduda, line 693: In lines 692 and 693, I think `int` should be `nint`.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1025 @@
...
+        ierr = 0
+
+        call mpas_pool_get_config(mgr % pool, 'tile_bdr', tile_bdr)
+        call mpas_pool_get_config(mgr % pool, 'known_lat', known_lat)
+        call mpas_pool_get_config(mgr % pool, 'known_lon', known_lon)
+        call mpas_pool_get_config(mgr % pool, 'dx', dx)
+        call mpas_pool_get_config(mgr % pool, 'dy', dy)
+
+        lat = real((j - (tile_bdr + 1) + tile % y - 1), kind=RKIND) * dy + known_lat + 1.0
+        lon = real((i - (tile_bdr + 1) + tile % x - 1), kind=RKIND) * dx + known_lon + 1.0
```
mgduda, line 643: Where does the `+ 1.0` at the end of the expressions for `lat` and `lon` come from?

src/core_init_atmosphere/mpas_geotile_manager.F
mgduda: I just left a comment for lines 642 and 643 that may be related.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1025 @@
...
+        if (y < 0 .or. y > size(mgr % hash, 2)) then
+            return
+        endif
+
+        tile => mgr % hash(x,y) % ptr
+        if (associated(tile)) then
+            loaded = .true.
+        endif
+
+    end function
```
mgduda, line 386: To match the prevailing style in MPAS, we could change this line to
```fortran
    end function mpas_geotile_mgr_search_tile
```

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1025 @@
...
+
+        call mpas_pool_get_config(mgr % pool, 'signed', signed)
+        call mpas_pool_get_config(mgr % pool, 'endian', endian)
+        call mpas_pool_get_config(mgr % pool, 'scale_factor', scalefactor)
+
+
+        !
+        ! signed, endian, and scale_factor all have default values, so if they are
+        ! not present in the index file then set them as the default values, as reported
+        ! in section 3-53 of the WRF-AWR User's Guide
```
mgduda, line 159: "AWR" -> "ARW"

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1025 @@
...
+        call mpas_pool_get_config(mgr % pool, 'dx', dx)
+        call mpas_pool_get_config(mgr % pool, 'dy', dy)
+
+        ! Find the global pixel location that contains lat, lon
+        call mgr % latlon_to_pixel(lat, lon, x(1), y(1))
+
+        ! Calculate the range of this tile, which will be its filename
+        x(1) = (x(1) - modulo(x(1), tile_nx)) + 1
+        x(2) = x(1) + tile_nx
+        x(2) = x(2) - modulo(x(2), tile_nx)
```
mgduda, line 585: I may be overlooking some special cases, but this logic seems more complicated than necessary. We know that `x(1)` is one more than a multiple of `tile_nx`, and by line 584, so is `x(2)`; so, couldn't we just subtract 1 from `x(2)`?

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1025 @@
...
+        ! NOTE: This calculation assumes that a dataset is a global dataset and may
+        ! not work correctly for non-global datasets
+        mgr % pixel_nx = nint(360.0_RKIND / dx)
+        mgr % pixel_ny = nint(180.0_RKIND / dy)
+
+        ! Calculate the number of tiles in the x, y directions
+        ! NOTE: This calculation assumes that a dataset is a global dataset and may
+        ! not work correctly for non-global datasets
+        mgr % nTileX = int((360.0_RKIND / dx)) / tile_nx
+        mgr % nTileY = int((180.0_RKIND / dy)) / tile_ny
```
mgduda, line 197: As with the calculation of `pixel_nx` and `pixel_ny`, I think we need to use `nint` here rather than `int`. Better still, we could consider just using `mgr % pixel_nx` and `mgr % pixel_ny` rather than duplicating the computation of those variables.

src/core_init_atmosphere/mpas_geotile_manager.F
MiCurry: Trying out `nint` (as well as `floor` and `ceiling`) will cause the geotile_manager to generate tile names that don't exist for the dataset. Although, I'm not sure if this behavior is correct or not. For instance the geotile_manager will generate the tile `42001-43200.21601-22800` for the point: ```(1.57080, 0.00000)```. The calculation is:

```
* lat = 90.0
* (lat - known_lat) / dx = 21599.5059
* nint(21599.5059) = 21600
```

However, I don't think this is correct, as the very last tile contains the `21600` pixel: `42001-43200.20401-21600`. 

So perhaps the tile filename generation is incorrect? Perhaps off by one?

src/core_init_atmosphere/mpas_geotile_manager.F
MiCurry: Using `mgr % pixel_nx/ny` is a great idea!

src/core_init_atmosphere/mpas_geotile_manager.F
MiCurry: I'm not sure where it comes from. Previous implementations didn't have it. On its own, it doesn't make much sense to add 1.0 to the latitude or longitude.

MiCurry:
Okay @mgduda, I just pushed an update that addresses your above comments, except for changes on line 692 and 693, which I left a comment on!

src/core_init_atmosphere/mpas_geotile_manager.F
mgduda: Are the pixel coordinates that are returned by `mpas_geotile_mgr_latlon_to_pixel` supposed to be 0-based or 1-based?

mgduda: Let's assume that every latitude value in the interval [-Pi/2, Pi/2] is uniquely mapped to a j-index. Then for the center latitude, *lat*, of a given pixel, we can either say that all latitudes in the interval [*lat* - Pi/2, *lat* + Pi/2) map to that pixel, or that all latitudes in the interval (*lat* - Pi/2, *lat* + Pi/2] map to that pixel. Regardless of which we choose, there will always be a pole, i.e., -Pi/2 or Pi/2 that is not mapped to some j-index and that therefore needs special treatment.

MiCurry:
@mgduda I have just pushed an update to this branch which adds a default value to `tile_bdr` and uses `nint` on the `latlon_to_pixel` calculation and solves problems at the poles (as we discussed above). It also adds a new commit which adds a `mpas_geotile_manager.F` target to the `Makefile`.

mgduda:
It looks like the changes to `mpas_kd_tree.F` are being included in this PR as commit 04e88f0, while these were previously included in the `develop` branch as ddd137d . Could you rework the commit history of this PR so the tie-breaking commit doesn't appear as a new commit?

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1038 @@
...
+        call rad2Deg(lon)
+
+        pixel_x = nint((lon - known_lon) / dx)
+        pixel_y = nint((lat - known_lat) / dy)
+
+        if (pixel_x < 0) then
+            pixel_x = pixel_x + mgr % pixel_nx
+        endif
+
+        if (pixel_y < 0.0) then
```
mgduda, line 706: Since `pixel_y` is declared as an integer, should `0.0` be changed to `0`?

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1038 @@
...
+            pixel_y = pixel_y + 1
```
mgduda, line 707: For negative `pixel_y` values, adding 1 doesn't seem like the correct thing to do. Is the general intent of this block of code to ensure that `pixel_y` is contained in the interval `[0, mgr %p ixel_ny - 1`]?

MiCurry: @mgduda, yeah the intention was to contain pixel_y in the interval of `[0, mgr % pixel_ny - 1]`. Instead of adding 1 to negative values should we just set `pixel_y` to 0. Likewise with values of `pixel_y == mgr % pixel_nx - 1` instead of subtracting 1 should we set `pixel_y` to `mgr % pixel_nx - 1`?

Did you perhaps have a better solution in mind?

mgduda: I agree that simply setting `pixel_y` to zero when its value is negative, and setting `pixel_y` to `mgr % pixel_ny - 1` when its value exceeds `mgr % pixel_ny - 1` would be the most robust solution.

MiCurry: Sounds good

MiCurry:
@mgduda okay, I have just pushed an update that resolves rounding pixel_y values, changes 0.0 -> 0 and reworked the commit history to remove commit 04e88f0.

Let me know if you see anything else.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -0,0 +1,1038 @@
...
+        pixel_y = nint((lat - known_lat) / dy)
+
+        if (pixel_x < 0) then
+            pixel_x = pixel_x + mgr % pixel_nx
+        endif
+
+        if (pixel_y < 0) then
+            pixel_y = 0
+        else if (pixel_y >= mgr % pixel_ny) then
+            pixel_y = pixel_y - 1
```
mgduda, line 709: Should `pixel_y` be `mgr % pixel_ny` on the r.h.s.?

MiCurry: I think you're right. I'll push an update.

MiCurry:
@mgduda Okay, I just pushed an update!

PR #641, opened 2020-07-27
Update Albany to fix reversed vertical layering, other issues

matthewhoffman:
This merge updates Albany and Trilinos in MALI to a new version that fixes a number of issues (described by @mperego):

1.  (a bit technical) The temperature in Albany is computed at the prisms (or tets) nodes. We then compute the average of the temperature over the prism, and that's what we use for computing the viscosity. When we call Albany from MALI, the temperature is interpolated at the prism barycenters. While the cell averaged temperature and the temperature at the cell center are the same for tets, this is not true for prisms. The difference is not much but has been fixed.

2. There was an error at the MPAS-Albany interface, on the Albany side, when we import the sigma coords from MPAS, when the layers thickness is not uniform. Typically the layers are thinner close to the bed.. because of the error in the interface, we had layers thinner at the surface and thicker at the bed. Not sure if it has been there forever.. according to git, it has been like that for at least 2 years.. it's crazy I never noticed this before. I think this explain the issue we had in MALI, with the enthalpy solver, when we choose non uniform layers. Well, I fixed it now.

3. I changed some of the default hard coded options  on the Albany interface side, regarding the number of quadrature points (I raised the order to 4 from 3 when glp is off but we zero the basal friction on shelves) and the lateral conditions (we let Albany compute the ratio of immersed ice, to compute the back pressure). These options can be overwritten in the albany_input.yaml file. Also, I now turned on memoization by default, which saves a lot of time in assembly.

4. Depending on whether glp is enabled, MPAS extrapolates beta or set it to 0. This has been removed so that the corresponding option in Albany is the way to accomplish this.

5. When writing Albany mesh from MPAS, we need to make sure that the global ids of the prisms (or tets) are sorted, otherwise we won't correctly import in Albany fields like temperature that are defined on prisms. This has not affected our workflow yet, but it should be fixed. 

6. Lateral b.c. in Albany is updated.    The new lateral b.c. are specified normal stress at all margins, with the value being determined by the fraction of the thickness in contact with the ocean (terrestrial and floating being the two end members).


matthewhoffman:
Passes all tests.  New version of Albany changes answers by small amounts for all tests.  For SSA-like tests, changes are very small, order 10^-20 m/s.  For SIA-like tests, changes are larger, as expected, order 10^-12 to 10^-9 m/s.

matthewhoffman:
All HO tests had slight answer changes.  Manually blessed each with @mperego.

PR #643, opened 2020-08-03
Ocean/gm ismf fixes

vanroekel:
Integrals for computing GM kappa and the phase speed have been reworked to mitigate failures in ISMF simulations without redi.  With this PR all configurations of GM work and the old 3DGM (N2_dependent) is recoverable to its original configuration.

This PR will be non BFB in WC and Cryo configs.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -259,7 +259,6 @@ contains
 
             maxLocation = k
             do k = maxLocation, maxLevelCell(iCell)
-            do k = 1,maxLevelCell(iCell)
                BruntVaisalaFreqTopEdge = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
 
                RediKappaScaling(k, iCell) = min(max(config_GM_min_stratification_ratio, &
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -49,7 +49,6 @@ module ocn_gm
    logical :: local_config_GM_compute_visbeck
    logical :: local_config_GM_lat_variable_c2
    logical :: local_config_GM_kappa_lat_depth_variable
-   real(kind=RKIND), parameter :: local_config_GM_min_phase_speed = 0.1_RKIND
    real(kind=RKIND) :: slopeTaperFactor, sfcTaperFactor, rediGMinitValue
 !***********************************************************************
 
@@ -247,24 +246,19 @@ contains
       end do
       !$omp end do
 
-      if ( config_Redi_use_N2_based_taper .or. local_config_GM_kappa_lat_depth_variable ) then
+      if ( config_Redi_use_N2_based_taper .and. .not. local_config_GM_kappa_lat_depth_variable ) then
 
          !$omp do schedule(runtime) private(maxLocation, cell1, cell2, k, BruntVaisalaFreqTopEdge, maxN)
          do iCell = 1, nCells
-!            k = min(maxLevelCell(iCell) - 1, max(1, indMLD(iCell)))
-            k = 1
+            k = min(maxLevelCell(iCell) - 1, max(1, indMLD(iCell)))
             maxN = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
-            do k = 1, maxLevelCell(iCell)
+            do while (BruntVaisalaFreqTop(k + 1, iCell) > maxN .and. k < maxLevelCell(iCell) - 1)
+              k = k + 1
               maxN = max(maxN,max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND))
             enddo
 
-!            do while (BruntVaisalaFreqTop(k + 1, iCell) > maxN .and. k < maxLevelCell(iCell) - 1)
-!               k = k + 1
-!               maxN = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
-!            end do
-
-!            maxLocation = k
-!            do k = maxLocation, maxLevelCell(iCell)
+            maxLocation = k
+            do k = maxLocation, maxLevelCell(iCell)
             do k = 1,maxLevelCell(iCell)
                BruntVaisalaFreqTopEdge = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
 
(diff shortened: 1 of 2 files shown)


vanroekel:
@mark-petersen this fixes the ISMF failures in my testing.  Could you please try these changes yourself?  Previous failures happen from E3SM master but with
```
config_use_redi=.false.
```

vanroekel:
@maltrud and @milenaveneziani I expect changes to water cycle sims to be minimal with these changes the runs will not be BFB.  I will rerun the baseline G-case with no redi for comparison.

vanroekel:
@jonbob could you post results of your ISMF PEM test here when it finishes?

jonbob:
With  config_use_redi = .false..

PASS PEM_P480_Ld5_D.T62_oEC60to30v3wLI.GMPAS-IAF-ISMF.anvil_intel RUN time=2538
PASS PEM_P480_Ld5_D.T62_oEC60to30v3wLI.GMPAS-IAF-ISMF.anvil_intel COMPARE_base_modpes

My test ran and passed the comparison as well

mark-petersen:
@vanroekel could you write a few sentences about what you changed here? Did the analytic equations change for GM kappa and the phase speed, or just the code method to solve it? Can you say what was causing the problem before?  I'll give the code a careful read at the same time.

vanroekel:
@mark-petersen sure.  Here it is

This PR does not change the analytic formulation for calculating c^2 or the Visbeck parameterization of Kappa_gm.  Instead if fixes a few things and changes one formulation

1) In the visbeck / redi PR the N2 limiting followed exactly from Danabasoglu and Marshall 2007, where you have
```
scaling = max(minValue, (N/N_ref)^2)
```
where minValue = 0.1.  in the latest master, N_ref is taken slightly below the ML, whereas in cryo simulations with maint-1.2 N_ref is the max in the entire column.  In this PR we keep the scaling above for redi, but when ```config_GM_closure = 'N2_dependent'``` the GM/Redi scaling is returned to the cryo main-1.2 formulation

2) when computing spatially variable c^2, the vertical integral was using the layer thickness from the cell while the BruntVaisala Frequency lived on the top of the cell, so the averaging was not quite right.  That is fixed here

3) similarly integrals for visbeck needed a fix to the layer thickness averaging.

4) Finally the N2 limiting in (1) above had an unexpected k=1 right below the correct value of k, this is removed.

5) there was an inconsistency between a local_gm_phase_speed min and the namelist causing issues.  That is also fixed here

Unfortunately it's not clear to me what was causing the issue, my guess is that the offsetting of layer thickness was causing the issue but I think it was (1) + (5) primarily, but I think the layer thickness integral played a role too.

Let me know if you'd like more.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -274,7 +272,20 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
       if (local_config_GM_kappa_lat_depth_variable) then
         !$omp do schedule(runtime)
```
mark-petersen, line 273: needs private (k, maxN, BruntVaisalaFreqTopEdge)

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -49,7 +49,6 @@ module ocn_gm
    logical :: local_config_GM_compute_visbeck
    logical :: local_config_GM_lat_variable_c2
```
mark-petersen, line 50: These local config variables don't have any explanation of what they do. Most are clear, but would be good to have a sentence on this one. Does it mean that c2 varies with latitude? You could also put the explanation just before this line:
```
500 if (local_config_GM_lat_variable_c2) then
```

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -865,7 +876,7 @@ subroutine ocn_GM_init(domain, err)!{{{
 
          RediGMinitValue = 1.0_RKIND
          if (config_GM_closure == 'constant') then
-            local_config_GM_lat_variable_c2 = .false.
+            local_config_GM_lat_variable_c2 = .true.
```
mark-petersen, line 879: For constant GM closure, c2 varies with latitude? @vanroekel I'm fine with what you found works best, but could you explain why?

vanroekel: this is a holdover from my testing.  I will revert. Thanks for the catch!

src/core_ocean/shared/mpas_ocn_gm.F
vanroekel: My recollection is you and I decided to put them in so we could just have GM configurations in the namelist and not lots of options.  Although that only explains the second, not sure why I put a local_config_GM-compute_visbeck in there though.  Let me clean this up

src/core_ocean/shared/mpas_ocn_gm.F
vanroekel: thanks I will add this

src/core_ocean/shared/mpas_ocn_gm.F
mark-petersen: I think it's fine to have `local_config` variables if they are not meant to be changed by the user. I was saying that there should be a comment explaining what it is, either here, or where it is used in the code.

vanroekel: @mark-petersen a question for you.  Does it matter performance wise if we do a logical comparison for if statements or a string compare?  We could remove these local_configs and just use things like
```
if (config_gm_closure == 'visbeck')
```
but not sure which is better practice

vanroekel: makes sense, I agree on that.  Let me know what you think about my question just above and I'll add comments and make the other fixes.

philipwjones: Logical comparisons always better. So either logical flags or if there are multiple choices, setting up enum-like integer parameters as choices are much preferred. Strings must compare every letter and deal with a variety of conditions, terminations, etc.

vanroekel:
@mark-petersen I've tried to make this much clearer and have addressed your comments.  Let me know if you have any other suggestions

src/core_ocean/shared/mpas_ocn_gm.F
mark-petersen: @philipwjones is a string compare actually slow enough to matter if it is outside of a cell or edge loop? If so, we can do the string compare on init and replace it with a local logical or index. Doug used to do that, but it leads to confusing code because you need to always trace through extra variables.

vanroekel: @mark-petersen I think the code right now does the string compare on init and then sets the logicals right?

mark-petersen:
compiled with gnu, intel17, intel19. All debug and optimized. Passes all MPAS-O nightly regression suites.

PR #644, opened 2020-08-04
Fix path to docs in Travis CI

xylar:
Replace space (not backslash) with underscore in the path.

Documentation for "latest coastal" and "lastest ocean" is being
placed in a directory that is not very convenient to reference.

### DIFF
--- a/travis_ci/test_docs.bash
+++ b/travis_ci/test_docs.bash
@@ -13,7 +13,7 @@ fi
 # MPAS-Model tag or release
 echo "Docs version: $DOCS_VERSION"
 
-DOCS_PATH="${DOCS_VERSION//\//_}"
+DOCS_PATH="${DOCS_VERSION// /_}"
 
 PUBLICATION_BRANCH=gh-pages
 # Checkout the branch


xylar:
@mark-petersen and @sbrus89, I noticed that recent merges put documentation here:
http://mpas-dev.github.io/MPAS-Model/latest%20coastal/
and here:
http://mpas-dev.github.io/MPAS-Model/latest%20ocean/

I made a small mistake and the space should have been replaced with an underscore.  I'm mostly asking for your review to confirm that you think this change will fix the problem, since it's not very easy to test.

xylar:
This fix won't propagate to ocean/develop or ocean/coastal for awhile so we'll just have to live with the bad path for awhile.

mark-petersen:
@mgduda @matthewhoffman @akturner: FYI, for PRs to develop that only involve Travis CI and documentation, I will not ask for your approval, as we are still getting the kinks out.

xylar:
@mark-petersen, the space comes from this file:
https://github.com/xylar/MPAS-Model/blob/fix_docs_ci/travis_ci/get_docs_version.bash
It is intentional.  However, I want to replace the space with an underscore in the URL because a %20 is an ugly thing to have in a URL even if it works.  Mere when you're ready.

PR #646, opened 2020-08-05
 Snow melt bug fix in MPAS-SeaIce Consistent with Icepack (for E3SM V2)

proteanplanet:
This PR fixes a snowmelt bug in MPAS-SeaIce that derives from CICE, and has recently been corrected in Icepack and needs to be included in E3SM V2 ahead of the switch to Icepack.

For more information, see: https://github.com/CICE-Consortium/Icepack/pull/328/files


proteanplanet:
There is a mistake in this one too.  I'm having a bad day.  Starting again.

jonbob:
Not a bad day -- just a learning curve!

Issue #416, opened 2019-12-13
Validity check of fields for the GWDO scheme fails for regions over the ocean


marta-gil:
Hello,
This error message for the atmosphere model appears whenever the input is a region set completely over the ocean:

----------------------------------------------------------------------
    Beginning MPAS-atmosphere Error Log File for task 0 of 1
        Opened at 2019/12/12 11:26:13
----------------------------------------------------------------------

    ERROR: *******************************************************************************
    ERROR: The GWDO scheme requires valid var2d, con, oa{1,2,3,4}, and ol{1,2,3,4} fields,
    ERROR: but these fields appear to be zero everywhere in the model input.
    ERROR: Either set config_gwdo_scheme = 'off' in the &physics namelist, or generate
    ERROR: the GWDO static fields with the init_atmosphere core.
    ERROR: *******************************************************************************
    ERROR: Please correct issues with the model input fields and/or namelist.
    CRITICAL ERROR: Core init failed for core atmosphere
    Logging complete.  Closing file at 2019/12/12 11:26:13 

It seems like the check for the validity of the fields (var2d, con, oa{1,2,3,4}, and ol{1,2,3,4} fields) consists in looking for non-zero values. However, these fields are indeed zero over water so the model should be able to proceed.

MiCurry:
Agreed! And can confirm the same behavior on our end for a region over the ocean that has zero terrain and GWDO fields. The above error is generated in the physics_compatibility_check subroutine in physics/mpas_atmphys_control.F:

https://github.com/MPAS-Dev/MPAS-Model/blob/51d5624709f75182130a2fd94715e1bb13310172/src/core_atmosphere/physics/mpas_atmphys_control.F#L476-L490

An easy fix, as provided by @mgduda, would be to check the landuse field is entirely water. If it is, we can let the model proceed and if not, raise this error.

Thanks for creating the issue @marta-gil!

PR #652, opened 2020-08-11
Fix GWDO check on all water regional simulations

MiCurry:
This merge corrects the atmosphere_model's GWDO check that would incorrectly be triggered when running a regional simulation entirely over water.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_control.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_control.F
@@ -448,6 +448,9 @@
  real (kind=RKIND) :: maxvar2d_local, maxvar2d_global
  real (kind=RKIND), dimension(:), pointer :: var2d
  integer, pointer :: nCellsSolve
+ integer, pointer :: iswater_lu
+ integer, pointer, dimension(:) :: ivgtyp
+ integer :: all_water, iall_water
  character (len=StrKIND), pointer :: gwdo_scheme
  type (block_type), pointer :: block
  type (mpas_pool_type), pointer :: meshPool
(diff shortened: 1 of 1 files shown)


MiCurry:
This PR addresses issue #416.

marta-gil:
Hello,

I found the issue while doing some preliminary tests, but I am not myself running simulations that may encounter this issue (at least for now). Thanks for fixing it!

De: Miles Curry
Enviat: dimarts, 11 d’agost de 2020 23:35
Per a: MPAS-Dev/MPAS-Model
A/c: Marta Gil Bardaji; Mention
Tema: Re: [MPAS-Dev/MPAS-Model] Fix GWDO check on all water regionalsimulations (#652)

This PR addresses issue #416. Thanks again for creating the issue for this @marta-gil and sorry about the delay to getting to it. I'm not sure if you're still running the same simulation, @marta-gil, when you found this issue, but it would be great if you wanted to test these changes on your use case.

src/core_atmosphere/physics/mpas_atmphys_control.F
```diff
@@ -473,7 +476,22 @@ subroutine physics_compatibility_check(dminfo, blockList, streamManager, ierr)
...
 
-     if (maxvar2d_global <= 0.0_RKIND) then
+     !
+     ! The GWDO check below can fail on regional simulations that are completely above
+     ! water. So, check to see if the simulation is completely above water and do not
+     ! throw the error if it is.
+     !
+     call mpas_pool_get_array(sfc_inputPool, 'iswater', iswater_lu)
+     call mpas_pool_get_array(sfc_inputPool, 'ivgtyp', ivgtyp)
+     if (all(ivgtyp(:) == iswater_lu)) then
```
mgduda, line 486: I think we need to restrict the range of cells that we're checking to `(1:nCells)`, or even better, to `(1:nCellsSolve)`, to avoid mis-interpreting the garbage cell as a non-water point.

@MiCurry Just in case you happen to see this, would you mind if I pushed an amended commit to this PR branch?

mgduda (changes requested):

src/core_atmosphere/physics/mpas_atmphys_control.F
MiCurry: @mgduda feel free! I've added you as a contributor on my fork, so all my branches are belong to you!

mgduda: Many thanks!

mgduda (approved):
All looks good to me!
Thanks for the PR, @MiCurry , and thanks for finding this issue, @marta-gil !

PR #653, opened 2020-08-12
Update ISOMIP+ docs, including tutorial for Ocean0 on LANL IC 

xylar:
This is updated in the documentation, rather than in the local README.

This merge also removes the README in ISOMIP+ COMPASS directory because it is now redundant (and out of date).


xylar:
I also just added documentation for the ISOMIP+ test cases more generally.  This is not for LANL IC, but rather it's intended to be a bit more general.  It's far from perfect but just needed to get something up there.  I'll move on to other needed docs and come back to this if and when it's needed.

xylar:
I did a local test build of the docs on my laptop and everything looked right (at a quick glance).  I think it should build okay once we merge.

mark-petersen:
@xylar and @cbegeman Thanks for working on the documentation. Will merge in a bit...

PR #654, opened 2020-08-14
Update MPAS framework for E3SM builds with machine- and/or compiler-specific flags

amametjanov:
Different machines and compilers may require adjusted compiler flags. In E3SM, Depends.$compiler, Depends.$machine and Depends.$machine.$compiler enable this. This PR adds these files by makefile includes and uses them for OpenACC-based GPU builds. Tested with PGI compiler on ORNL Summit and ANL Blues machines.

[bit-for-bit]

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,6 +9,16 @@
 
 # Source CIME-generated Macros
 include(${CASEROOT}/Macros.cmake)
+# Load machine/compiler specific settings
+set(COMPILER_SPECIFIC_DEPENDS ${CASEROOT}/Depends.${COMPILER}.cmake)
+set(MACHINE_SPECIFIC_DEPENDS ${CASEROOT}/Depends.${MACH}.cmake)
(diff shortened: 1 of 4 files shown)


mark-petersen:
@amametjanov could you describe how you tested this? Did you test on our standard systems as well, like cori, compy, or Anvil?  Is this ready, or still in process?

mark-petersen:
Whoops - just saw your testing description at the top. So this is ready? I'll need to bring the `ocean/cmake` changes in through the `ocean/develop` branch, but I can do that.

amametjanov:
Yes, this is ready. This just simplifies the workflow if compiler flags need to be adjusted. Otherwise, updates to `src/build_core.cmake` would need to go upstream through MPAS repo before getting used in E3SM repo. Porting to new machines and compilers needs a quicker workflow.

Tested with `SMS_P36x1_Ld1.T62_oEC60to30v3.CMPASO-NYF` on summit+pgigpu and blues+pgigpu. Checked with `nvprof` that OpenACC kernels executed on GPUs.

PR #655, opened 2020-08-17
Parallel interpolation of the terrain field

MiCurry:
This merge allows the ability the init_atmosphere model to safely interpolate the ter field in parallel. However, because other fields are yet to be updated to run in parallel, the check in init_atm_cases to prevent parallel interpolation is still present.

Before this merge, parallel interpolation of the ter field (or any static field) was possible, but would incorrectly interpolate datatile pixels on interpolating on convex decompositions (when running non-serial interpolations without a convext partition file) as the nearest_cell routine would incorrectly choose the wrong cell for interpolation point for a datatile pixel.

This merge updates the algorithm for the interpolation of the ter field so that it can be interpolated in parallel in the future.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -272,32 +272,14 @@
 !
 ! Interpolate HGT
 !
-!nx = 126
-!ny = 126
- nx = 1206
- ny = 1206
- nz = 1
- isigned  = 1
- endian   = 0
- wordsize = 2
- scalefactor = 1.0
- allocate(rarray(nx,ny,nz))
- allocate(nhs(nCells))
- nhs(:) = 0
- ter(:) = 0.0
 
- rarray_ptr = c_loc(rarray)
-
- start_lat = -89.99583
  select case(trim(config_topo_data))
     case('GTOPO30')
        call mpas_log_write('Using GTOPO30 terrain dataset')
        geog_sub_path = 'topo_30s/'
-       start_lon = -179.99583
     case('GMTED2010')
        call mpas_log_write('Using GMTED2010 terrain dataset')
        geog_sub_path = 'topo_gmted2010_30s/'
-       start_lon = 0.004166667
     case('default')
        call mpas_log_write('*****************************************************************', messageType=MPAS_LOG_ERR)
        call mpas_log_write('Invalid topography dataset '''//trim(config_topo_data) &
@@ -307,66 +289,7 @@
        call mpas_log_write('Please correct the namelist.', messageType=MPAS_LOG_CRIT)
  end select
 
- do jTileStart = 1,20401,ny-6
-    jTileEnd = jTileStart + ny - 1 - 6
-
-    do iTileStart=1,42001,nx-6
-       iTileEnd = iTileStart + nx - 1 - 6
-       write(fname,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(geog_data_path)//trim(geog_sub_path), &
-                    iTileStart,'-',iTileEnd,'.',jTileStart,'-',jTileEnd
-       call mpas_log_write(trim(fname))
-       call mpas_f_to_c_string(fname, c_fname)
-
-       call read_geogrid(c_fname,rarray_ptr,nx,ny,nz,isigned,endian, &
-                         scalefactor,wordsize,istatus)
-       call init_atm_check_read_error(istatus, fname)
-
-       iPoint = 1
-       do j=4,ny-3
-       do i=4,nx-3
-          lat_pt = start_lat + (jTileStart + j - 5) * 0.0083333333
-          lon_pt = start_lon + (iTileStart + i - 5) * 0.0083333333
-          lat_pt = lat_pt * PI / 180.0
-          lon_pt = lon_pt * PI / 180.0
-
-          iPoint = nearest_cell(lat_pt,lon_pt,iPoint,nCells,maxEdges, &
-                                nEdgesOnCell,cellsOnCell, &
-                                latCell,lonCell)
-
-          !
-          ! For all but the outermost boundary cells, we can safely assume that the nearest
-          ! model grid cell contains the pixel (else, a different cell would be nearest)
-          !
-          if (bdyMaskCell(iPoint) < nBdyLayers) then
-             ter(iPoint) = ter(iPoint) + rarray(i,j,1)
-             nhs(iPoint) = nhs(iPoint) + 1
-
-          ! For outermost boundary cells, additional work is needed to verify that the pixel
-          ! actually lies within the nearest cell
-          else
-             zPixel = sphere_radius * sin(lat_pt)                  ! Model cell coordinates assume a "full" sphere radius
-             xPixel = sphere_radius * cos(lon_pt) * cos(lat_pt)    ! at this point, so we need to ues the same radius
-             yPixel = sphere_radius * sin(lon_pt) * cos(lat_pt)    ! for source pixel coordinates
-
-             if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(iPoint), yCell(iPoint), zCell(iPoint), &
-                         nEdgesOnCell(iPoint), verticesOnCell(:,iPoint), xVertex, yVertex, zVertex)) then
-
-                ter(iPoint) = ter(iPoint) + rarray(i,j,1)
-                nhs(iPoint) = nhs(iPoint) + 1
-
-             end if
-          end if
-       end do
-       end do
-
-    end do
- end do
-
- do iCell = 1,nCells
-    ter(iCell) = ter(iCell) / real(nhs(iCell))
- end do
- deallocate(rarray)
- deallocate(nhs)
+ ter(:) = 0.0_RKIND
  call mpas_log_write('--- end interpolate TER')
 
 
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -20,8 +20,8 @@
  use mpas_geometry_utils, only : mpas_in_cell
  use mpas_atmphys_utilities
 
- use mpas_kd_tree, only : mpas_kd_type, mpas_kd_construct, mpas_kd_free
- use mpas_geotile_manager, only : mpas_geotile_mgr_type
+ use mpas_kd_tree, only : mpas_kd_type, mpas_kd_construct, mpas_kd_free, mpas_kd_search
+ use mpas_geotile_manager, only : mpas_geotile_mgr_type, mpas_geotile_type, mpas_latlon_to_xyz
 
  use iso_c_binding, only : c_char, c_int, c_float, c_loc, c_ptr
 
@@ -106,7 +106,7 @@
  real(kind=RKIND),dimension(:,:,:),allocatable:: vegfra
 
  integer, pointer :: isice_lu, iswater_lu
- integer, pointer :: nCells, nEdges, nVertices, maxEdges
+ integer, pointer :: nCells, nCellsSolve, nEdges, nVertices, maxEdges
  logical, pointer :: on_a_sphere
  real (kind=RKIND), pointer :: sphere_radius
  
@@ -125,6 +125,7 @@
  real (kind=RKIND), dimension(:), pointer :: fEdge, fVertex
 
  real (kind=RKIND), dimension(:), pointer :: ter
+ integer (kind=I8KIND), dimension(:), pointer :: ter_integer
  real (kind=RKIND), dimension(:), pointer :: soiltemp
  real (kind=RKIND), dimension(:), pointer :: snoalb
  real (kind=RKIND), dimension(:), pointer :: shdmin, shdmax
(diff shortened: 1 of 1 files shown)


MiCurry:
I expect this PR will take a fair bit of time to be ready to merged.

I have tested these changes with GNU and Intel for bit-for-bit on 4, 8, 16 and 36 processor counts with both terrain datasets GTOPO30 and GMTED2010 and got identical results. 

We should test this branch on higher processor counts and meshes with smaller grid spacing. @mgduda do you have any suggestion for what meshes we should test?

@mgduda I haven't checked these commits for new warnings, but I'll go ahead and do so and I'll push an update with updates/clean.

MiCurry:
I was able to compile this branch without warnings with the PGI, GNU and Intel compilers!

MiCurry:
@mgduda as suggested I decided to run some parallel terrain interpolation on limited-area regions. I was able to get correct results, however, I'm not getting bit-for-bit results across different task counts. As well, I'm getting significantly **_slower_** interpolation times on a LA mesh. 

For reference running the terrain interpolation on the global x1.10242.grid.nc with 12 tasks took 142.72 s, while running the terrain interpolation on the [temperate.channel.pts](https://gist.github.com/MiCurry/8f5853d74ab99bc6a6ef9909a85ed738) region with 12 tasks took 1526.76 s. Significantly slower..

I'm not exactly why bit-for-bit results are not happening yet, but I think I have an idea why the LA interpolation is significantly slower. I think because there are more boundary cells, the `mpas_in_cell` routine is called more frequently the more `mpas_rotate_about_vector`, and `mpas_sphere_angle` are called, which contain a significant number of trigonometry (and square root) functions are caused, causing the slowdown. At least, that's my currently explanation.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,132 @@ subroutine init_atm_static(mesh, dims, configs)
...
+
+               ! For outermost boundary cells, additional work is needed to verify that the pixel
+               ! actually lies within the nearest cell
+               else
+                  ! Model cell coordinates assume a "full" sphere radius
+                  ! at this point, so we need to use the same radius
+                  ! for source pixel coordinates
+                  call mpas_latlon_to_xyz(x, y, z, sphere_radius, lat_pt, lon_pt)
+
+                  if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(res % id), yCell(res % id), zCell(res % id), &
```
MiCurry, line 409: @mgduda The use of xPixel`, `yPixel`, and `zPixel` here was the cause of our problems. They need to be `x, y, z` respectively.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,132 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                      all_pixels_mapped_to_halo_cells = .false.
+                  end if
+
+               ! For outermost boundary cells, additional work is needed to verify that the pixel
+               ! actually lies within the nearest cell
+               else
+                  ! Model cell coordinates assume a "full" sphere radius
+                  ! at this point, so we need to use the same radius
+                  ! for source pixel coordinates
+                  call mpas_latlon_to_xyz(x, y, z, sphere_radius, lat_pt, lon_pt)
```
MiCurry, line 407: I also think this can be removed as well.

MiCurry:
@mgduda I just pushed an update that fixes the problems we were having and I think this is ready for a review!

Although we had a problem with using `xPixel, yPixel` and `zPixel` in the above comment, instead of using `x, y, z`, In the last update I changed using `x, y, z` to `xPixel, yPixel` and `zPixel` as I thought it was a bit more clearer as the what the `x, y, z` values are representing!

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,130 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                  if (res % id <= nCellsSolve) then
+                      all_pixels_mapped_to_halo_cells = .false.
+                  end if
+
+               ! For outermost boundary cells, additional work is needed to verify that the pixel
+               ! actually lies within the nearest cell
+               else
+                  ! Model cell coordinates assume a "full" sphere radius
+                  ! at this point, so we need to use the same radius
+                  ! for source pixel coordinates
```
MiCurry, line 406: I think we can remove this comment above

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -135,6 +140,20 @@ subroutine init_atm_static(mesh, dims, configs)
...
+ type (mpas_kd_type), pointer :: res
+
+ type (mpas_geotile_mgr_type) :: mgr
+ type (mpas_geotile_type), pointer :: tile
+
+ real (kind=RKIND) :: tval
+ integer, pointer :: tile_bdr
+ integer, pointer :: tile_nx, tile_ny
+
+ logical :: all_pixels_mapped_to_halo_cells = .true.
```
mgduda, line 154: It probably doesn't matter, but just as a matter of convention, it might be good to remove the initialization of this variable so that it doesn't take on the `SAVE` attribute implicitly.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -225,6 +245,22 @@ subroutine init_atm_static(mesh, dims, configs)
...
+do i = 1, nCells
+    allocate(kd_points(i) % point(3))
+    kd_points(i) % point = (/xCell(i), yCell(i), zCell(i)/)
+    kd_points(i) % id = i ! Cell ID
+enddo
+tree => null()
+tree => mpas_kd_construct(kd_points, 3)
+if (.not. associated(tree)) then
+    call mpas_log_write('Error creating the KD-Tree', messageType=MPAS_LOG_CRIT)
+    return
```
mgduda, line 261: Can we indent this code block by one column to match the surrounding code?

I think the `return` statement is also unnecessary, since a critical error message will stop execution of the program.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,127 @@ subroutine init_atm_static(mesh, dims, configs)
...
-          ! For outermost boundary cells, additional work is needed to verify that the pixel
-          ! actually lies within the nearest cell
-          else
-             zPixel = sphere_radius * sin(lat_pt)                  ! Model cell coordinates assume a "full" sphere radius
-             xPixel = sphere_radius * cos(lon_pt) * cos(lat_pt)    ! at this point, so we need to ues the same radius
-             yPixel = sphere_radius * sin(lon_pt) * cos(lat_pt)    ! for source pixel coordinates
+ ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+ if (ierr /= 0) then
+    call mpas_log_write('Error initializing ter Geotile Manager', messageType=MPAS_LOG_CRIT)
+    return
```
mgduda, line 329: The `return` statement here will never be reached.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,127 @@ subroutine init_atm_static(mesh, dims, configs)
...
-
-          ! For outermost boundary cells, additional work is needed to verify that the pixel
-          ! actually lies within the nearest cell
-          else
-             zPixel = sphere_radius * sin(lat_pt)                  ! Model cell coordinates assume a "full" sphere radius
-             xPixel = sphere_radius * cos(lon_pt) * cos(lat_pt)    ! at this point, so we need to ues the same radius
-             yPixel = sphere_radius * sin(lon_pt) * cos(lat_pt)    ! for source pixel coordinates
+ ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+ if (ierr /= 0) then
+    call mpas_log_write('Error initializing ter Geotile Manager', messageType=MPAS_LOG_CRIT)
```
mgduda, line 328: I don't have any suggestions at the moment, but are there ways that this error message could be made to be more helpful or understandable by a user? I suspect most users would have no idea what it means to not be able to initialize the "ter Geotile Manager", nor why that might have occurred.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,127 @@ subroutine init_atm_static(mesh, dims, configs)
...
+ allocate(ter_integer(nCells))
+ allocate(nhs(nCells))
 
-             end if
-          end if
-       end do
-       end do
+ !
+ ! Store tile values as a I8KIND integer temporarily to avoid floating
+ ! point round off differences
```
mgduda, line 338: Here we could provide a little more insight into why we are accumulating in a 64-bit integer array, and what implications this has for resolution and precision of input datasets in relation to the MPAS mesh.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,127 @@ subroutine init_atm_static(mesh, dims, configs)
...
+ ter_integer(:) = 0_I8KIND
+ ter(:) = 0.0_RKIND
+ nhs(:) = 0
 
-    end do
+ call mpas_pool_get_config(mgr % pool, 'tile_bdr', tile_bdr)
+ call mpas_pool_get_config(mgr % pool, 'tile_x', tile_nx)
+ call mpas_pool_get_config(mgr % pool, 'tile_y', tile_ny)
+
+ do i = 1, nCells
```
mgduda, line 348: As we discussed, it might follow existing conventions better to use `iCell` as the loop variable here.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,127 @@ subroutine init_atm_static(mesh, dims, configs)
...
+         if (tile % is_processed) then
+             cycle
+         end if
+
+         call mpas_log_write('Processing tile: '//trim(tile%fname))
+
+         all_pixels_mapped_to_halo_cells = .true.
+
+         do j = 1 + tile_bdr, tile_ny + tile_bdr, 1
+            do k = 1 + tile_bdr, tile_nx + tile_bdr, 1
```
mgduda, line 381: As we discussed, let's switch to `i` as the loop variable for the x-direction, and `j` as the loop variable for the y-direction. I think the `, 1` loop increments could also be removed, unless we plan to implement sub-sampling.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,127 @@ subroutine init_atm_static(mesh, dims, configs)
...
  end do
- deallocate(rarray)
+
+ deallocate(ter_integer)
  deallocate(nhs)
+
+ ierr = mgr % finalize()
+ if (ierr /= 0) then
+    call mpas_log_write('Error finalizing the ter Geotile Manager', messageType=MPAS_LOG_CRIT)
+    return
```
mgduda, line 444: The `return` statement here will never be reached.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -306,66 +323,127 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                call mpas_log_write("Error pushing the tile neighbors of: "//trim(tile%fname), messageType=MPAS_LOG_CRIT)
+            end if
+        end if
+     end do
  end do
 
- do iCell = 1,nCells
-    ter(iCell) = ter(iCell) / real(nhs(iCell))
+ do i = 1, nCells
+    ter(i) = real(ter_integer(i), kind=RKIND) / real(nhs(i), kind=RKIND)
```
mgduda, line 435: For current datasets I don't think it would make much difference, but because its easy enough to do so, we could type-convert the integer values to `R8KIND` rather than `RKIND` before performing the division.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: There is also a `return` statement in the`finalize` routine, which I will also remove.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: The `init` function of the `geotile_manager` can only produce a non-zero return value in two circumstances: [inability to find an `index` file in the given directory](https://github.com/MPAS-Dev/MPAS-Model/blob/a667cd3108cae728a977daa8bb7eef7811d31aa5/src/core_init_atmosphere/mpas_geotile_manager.F#L129-L133) and [if `parse_geoindex` returns an error](https://github.com/MPAS-Dev/MPAS-Model/blob/a667cd3108cae728a977daa8bb7eef7811d31aa5/src/core_init_atmosphere/mpas_geotile_manager.F#L139-L143). Both of which produce their own log messages. As well, the `parse_geoindex` module also produces its own log message. So this error is a catch-all for all of those possible errors and really only serves as a call kill the model (via `MPAS_LOG_CRIT`). 

So it would hard to get more specific with the error, without repeated whats already been logged, but I agree that the error message might be a bit confusing and could be better. How does the following sound?

`'Error occurred when initializing the interpolation of terrain height (ter)'`

Or,

`'Error occurred when initializing the interpolation of terrain height (ter) from '//trim(directory)`

Where `directory` would be the complete path to the data-set.

Let me know what your thoughts are or if you have any other ideas!

MiCurry:
Okay @mgduda I have updated this PR with changes that address your comments and what we discussed.

mgduda:
I just rebased this PR on 515afd7 .

MiCurry:
@mgduda thanks! That helped with the problems when rebasing the `init_atmosphere/par_cat_interp` with this branch!

mgduda (approved):

PR #656, opened 2020-08-20
Correct sign in FCT 3rd order advection coefficients

mark-petersen:
Two third-order advection coefficients incorrectly have a plus sign, should be minus. Thanks to @ambrad for his work on this. See discussion at this issue: https://github.com/MPAS-Dev/MPAS-Model/pull/583#issuecomment-675274063

Note that this is a framework merge, but the ocean is the only core that calls the subroutine, `mpas_tracer_advection_coefficients`, that initializes the FCT cell weights. The atmosphere has nearly identical lines in `src/core_atmosphere/mpas_atm_core.F`, subroutine `atm_adv_coef_compression`, which is clearly where this code was originally copied from in 2013.

Also copied comments from the atmospheric core to better explain each term, and cleaned up other comments.

### DIFF
--- a/src/operators/mpas_tracer_advection_helpers.F
+++ b/src/operators/mpas_tracer_advection_helpers.F
@@ -229,14 +229,14 @@ module mpas_tracer_advection_helpers
            k = mpas_binary_search(sorted_cell_indices, 2, 1, nAdvCellsForEdge(iEdge), indexToCellID(cell2))
            if(k <= nAdvCellsForEdge(iEdge)) then
              adv_coefs(k, iEdge) = adv_coefs(k, iEdge) + deriv_two(1,2,iEdge)
-             adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(1,2,iEdge)
+             adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(1,2,iEdge)
            end if
 
            do iCell = 1, nEdgesOnCell(cell2)
              k = mpas_binary_search(sorted_cell_indices, 2, 1, nAdvCellsForEdge(iEdge), indexToCellID(cellsOnCell(iCell,cell2)))
              if(k <= nAdvCellsForEdge(iEdge)) then
                adv_coefs(k, iEdge) = adv_coefs(k, iEdge) + deriv_two(iCell+1, 2, iEdge)
-               adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(iCell+1, 2, iEdge)
+               adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(iCell+1, 2, iEdge)
              end if
            end do ! loop over iCell
 
--- a/src/operators/mpas_tracer_advection_helpers.F
+++ b/src/operators/mpas_tracer_advection_helpers.F
@@ -209,6 +209,21 @@ module mpas_tracer_advection_helpers
              advCellsForEdge(iCell, iEdge) = sorted_cell_indices(2, iCell)
            end do
 
+           ! equation 7 in Skamarock, W. C., & Gassmann, A. (2011):
+           ! F(u,psi)_{i+1/2} = u_{i+1/2} *
+           !  [   1/2 (psi_{i+1} + psi_i)                       term 1
+           !    - 1/12(dx^2psi_{i+1} + dx^2psi_i)               term 2
+           !    + sign(u) beta/12 (dx^2psi_{i+1} - dx^2psi_i)]  term 3 (note minus sign)
+           !
+           ! adv_coefs accounts for terms 1 and 2 in SG11 equation 7. Term 1 is
+           ! the 2nd-order flux-function term. adv_coefs accounts for this with
+           ! the "+ 0.5" lines below. In the advection routines that use these
+           ! coefficients, the 2nd-order flux loop is then skipped. Term 2 is
+           ! the 4th-order flux-function term. adv_coefs_3rd accounts for term
+           ! 3, the beta term. beta > 0 corresponds to the third-order flux
+           ! function. The - sign in the deriv_two accumulation is for the i+1
+           ! part of term 3, while the + sign is for the i part.
+
            adv_coefs(:,iEdge) = 0.
            adv_coefs_3rd(:,iEdge) = 0.
 
--- a/src/operators/mpas_tracer_advection_helpers.F
+++ b/src/operators/mpas_tracer_advection_helpers.F
@@ -188,7 +188,7 @@ module mpas_tracer_advection_helpers
                sorted_cell_indices(2, n) = cellsOnCell(i, cell1)
                call mpas_hash_insert(cell_hash, cellsOnCell(i, cell1))
              end if
-           end do ! loop over i
+           end do
 
            do i = 1, nEdgesOnCell(cell2)
              if(.not. mpas_hash_search(cell_hash, cellsOnCell(i, cell2))) then
@@ -198,7 +198,7 @@ module mpas_tracer_advection_helpers
                sorted_cell_indices(2, n) = cellsOnCell(i, cell2)
                call mpas_hash_insert(cell_hash, cellsOnCell(i, cell2))
              end if
-           end do ! loop over i
+           end do
 
            call mpas_hash_destroy(cell_hash)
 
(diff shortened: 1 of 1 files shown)


src/operators/mpas_tracer_advection_helpers.F
```diff
@@ -224,27 +226,30 @@ subroutine mpas_tracer_advection_coefficients( meshPool, horiz_adv_order, deriv_
...
-           end do ! loop over iCell
+           end do
 
+           ! pull together third and fourth order contributions to the flux
+           ! now from cell2
            k = mpas_binary_search(sorted_cell_indices, 2, 1, nAdvCellsForEdge(iEdge), indexToCellID(cell2))
            if(k <= nAdvCellsForEdge(iEdge)) then
              adv_coefs(k, iEdge) = adv_coefs(k, iEdge) + deriv_two(1,2,iEdge)
-             adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(1,2,iEdge)
+             adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(1,2,iEdge)
```
mark-petersen, line 236: Compare to corresponding line for atmosphere, which has a minus sign:
```
adv_coefs_3rd(j_in,iEdge) = adv_coefs_3rd(j_in,iEdge) - deriv_two(1,2,iEdge)
```
https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/core_atmosphere/mpas_atm_core.F#L1225

src/operators/mpas_tracer_advection_helpers.F
```diff
@@ -224,27 +226,30 @@ subroutine mpas_tracer_advection_coefficients( meshPool, horiz_adv_order, deriv_
...
            end if
 
            do iCell = 1, nEdgesOnCell(cell2)
              k = mpas_binary_search(sorted_cell_indices, 2, 1, nAdvCellsForEdge(iEdge), indexToCellID(cellsOnCell(iCell,cell2)))
              if(k <= nAdvCellsForEdge(iEdge)) then
                adv_coefs(k, iEdge) = adv_coefs(k, iEdge) + deriv_two(iCell+1, 2, iEdge)
-               adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) + deriv_two(iCell+1, 2, iEdge)
+               adv_coefs_3rd(k, iEdge) = adv_coefs_3rd(k, iEdge) - deriv_two(iCell+1, 2, iEdge)
```
mark-petersen, line 243: Compare to corresponding line for atmosphere, which has a minus sign:
```
adv_coefs_3rd(j_in,iEdge) = adv_coefs_3rd(j_in,iEdge) - deriv_two(iCell+1,2,iEdge)
```
https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/core_atmosphere/mpas_atm_core.F#L1233

mark-petersen:
@vanroekel and @ambrad, I am testing in various configurations of MPAS-Ocean and will report here. Please post reports of your testing here as well. Also note, @ambrad already conducted convergence tests here https://github.com/MPAS-Dev/MPAS-Model/pull/583#issuecomment-675274063 that show improvements with this bug fix.

mark-petersen:
@mgduda @matthewhoffman @akturner please add yourselves as reviewers if you like. This is in the operators subdirectory, but the ocean is the only core that calls this subroutine.

ambrad:
It might be useful to future code readers to add one further comment to the helpers.F code, something along the lines of this:

adv_coefs accounts for terms 1 and 2 in SG11 equation 7. Term 1 is the 2nd-order term. adv_coefs accounts for this with the "+ 0.5" lines. In the advection routines that use these coefficients, the 2nd-order flux loop is then skipped. Term 2 is the 4th-order term. adv_coefs_3rd accounts for term 3, the beta term. beta > 0 corresponds to the third-order flux function. The - sign in the deriv_two accumulation is for the i+1 part of term 3, while the + sign is for the i part.

mark-petersen:
Done. Thanks for the text @ambrad. That was perfect. I even added a text version of eqn 7 for reference to help the reader.

ambrad:
Testing was done in two ways. First, I developed a standalone code (part of other ongoing work) that implements these Eulerian flux-form methods. This code revealed some interesting patterns, described [in this comment](https://github.com/MPAS-Dev/MPAS-Model/pull/583#issuecomment-675274063) and repeated below. Second, once I knew what I was looking for, I ran the new MPAS-Ocean test of solid-body rotation of a cosine bell. This is described in [this comment](https://github.com/MPAS-Dev/MPAS-Model/pull/583#issuecomment-675862111) and again for convenience transcribed below.

Observations from standalone testing:
1. On a smooth quasiuniform grid (standard icosahedral grid, i.e., 10 pentagons at icosahedron vertices), all three flux functions -- 2nd-, 3rd-, 4th-order -- are empirically 2nd-order accurate on this test.
2. On a nonsmooth quasiuniform grid, e.g. one produced by Jigsaw, the 2nd- and 4th-order EFF methods empirically degrade to OOA ~1, while the 3rd-order EFF degrades only a small amount below OOA 2.

MPAS-Ocean testing:
1. I ran one full rotation of the cosine bell for 240, 120, 60km Jigsaw-created grids, with 720, 1440, 2880 time steps, respectively. I used RK4 as the time integration scheme.
2. I ran 2nd-, 3rd-, 4th-order flux functions, with the 3rd-order one both before and after the fix. I also ran 3rd-order after the fix with monotone advection rather than standard.
3. Results are as follows, where errors are RMSE as reported by compute_rmse.py in the linked PR:
```
order-2 flux               [0.04306733 0.0151201  0.00483401] OOA ~ 1.6
order-3 flux beta=1        [0.01030493 0.00226823 0.00050179] OOA > 2
order-3 flux beta=1/4      [0.00471197 0.00106989 0.00028895] OOA 1.9-2.1
order-3 flux beta=1/4 mono [0.00513074 0.00109597 0.00027202] OOA > 2
order-4 flux               [0.0071832  0.00447862 0.00253661] OOA < 1
order-3 flux beta=1/4, wrong sign in helper.F
                           [0.01326037 0.01270969 0.03431614] nonconvergent
```

ambrad:
@mark-petersen, thanks for adding the comment. Just to be clear, the minus sign in term 3 does not correspond to the - in the deriv_two accumulation. Rather, there are layers of negatives such that it's the i+1 part of term 3 that leads to the `- deriv_two`.

vanroekel:
@ambrad and @mark-petersen this looks great.  here are two snapshots from a G-case without the fix first then with.  These are salinity at 250m from high frequency
Screen Shot 2020-08-20 at 3 51 59 PM

Screen Shot 2020-08-20 at 3 51 59 PM

vanroekel:
this is 2.5 years into a G-case by the way

stephenprice:
Are these two images supposed to be identical?

ambrad:
@stephenprice they are not. The top image shows small-scale features due to the sign bug. In the bottom, these are gone, dissipated by the now-correct 4th-order hyperdiffusion (SG11 eq 9) produced by the third-order flux function.

@vanroekel you might try a simulation pair with the most extreme, Jigsaw-produced, RRM-like grid you have to look at beta=0.25 vs beta=1.0 with the sign fix. It's possible that beta=1.0 could be best for extremely nonuniform, nonsmooth grids.

stephenprice:
Thanks  @ambrad. I see the differences now that you point them out.

mark-petersen:
@vanroekel and @jamilgafur, would it be possible to rerun your advection convergence tests with these two changes? Getting the correct behavior and order of convergence on those tests would be a great confirmation of this change. I know @jamilgafur might be back in school next week.

vanroekel:
@mark-petersen sure, I can run the case in #583 and post convergence here

ambrad:
@vanroekel, I'll collect a few notes here re: the #583 convergence test:
- split_explicit should be switched to RK4.
- Currently, all three grids use the same 5min time step. For convergence analysis, I recommend a sequence of 30min, 15min, 7.5min.
- In your OOA script, I believe you should multiply the "Order of Convergence" numbers by d=2, the dimension of the grid. Each mesh refinement step increases the number of cells by 4, which means a refinement of sqrt(4)=2 in each direction.
- I was not able to get a good reading of OOA 2 until I lengthened the test to run one full rotation. Part of the challenge of convergence testing is being sure the test shows results in the asymptotic regime.
- That aside, a plot showing before-fix vs after-fix results should have a clear signal independent of any of the above details.

vanroekel:
Here is the OOA for the current code
convergence

Here is with this PR
convergenceFix

mattdturner:
@mgduda I would like to merge this today.  I wasn't brought into the loop on this PR until recently, so I'm not sure if you've weighed in or not.  Can you confirm whether or not you want to review this.  @mark-petersen asked if you wanted to review in  https://github.com/MPAS-Dev/MPAS-Model/pull/656#issuecomment-677891553

mgduda:
@mattdturner Thanks very much for checking. I haven't thoroughly reviewed the code changes with context, but the convergence results are convincing enough for me. I won't add myself as a reviewer, but I have no objections to merging this today.

PR #657, opened 2020-08-21
Change block tests to partition tests for QU240 RK4 and SE

xylar:
As of #496, MPAS-Ocean no longer supports multiple blocks.


mark-petersen:
@xylar, the previous block tests also served as a test for different domain partitions. So I simply changed these two tests to function again with partitions (4 vs 8).

mark-petersen (approved):
Tested nightly regression suite with gnu on grizzly. Partition test with SE and RK4 work fine.

mark-petersen:
@xylar If you agree with this, I'll merge.

xylar (commented):
I can't approve my own PR but I ran with the changes you made, @mark-petersen, and it works fine.

Probably a cleaner procedure in the future would be to just open a new PR, since this branch contains only changes that you made, is now incorrectly named, and I am not able to review it.

xylar:
I ran the two modified tests on my laptop with Gnu and they work as expected.

mark-petersen:
Yeah, looking back at this now, it would have been cleaner with two PRs. I'll keep that in mind on future ones.

PR #660, opened 2020-08-22
Fix validation for QU240 test step

xylar:
Folder is "test", not "forward".


mark-petersen (approved):
Yes, that's correct, thanks.

PR #666, opened 2020-08-25
Fix distribute calving method

matthewhoffman:
This merge rewrites the method for converting a calvingVelocity to a calvingThickness.

This requirement of this part of the code is to take a calving velocity and convert it to a
thickness to be removed.  The basic idea is (and was) to multiply the calving velocity
by the edge length and height to get a volume flux.  However, the details
are complicated by the unstructured Voronoi mesh and the use of partially-filled
non-dynamic floating grid cells.

The new algorithm rewrites this procedure.  It calculates the required calving volume flux based on edges, 
and then applies the volume removal to cells, converting that to a thickness removal using areaCell.
An important detail of the new scheme is that it assumes the direction of calvingVelocity is parallel to 
the ice surface velocity rather than perpendicular to every edge.

This ice removal process occurs in two phases:

The first phase handles non-dynamic floating cells. This is done by calculating the required calving flux
at the edges between those non-dynamic floating cells and the open ocean.
The calving velocity and ice thickness are copied from the upstream dynamic
floating cells, because these quantities do not have valid values on the non-dynamic cells.
After calculating the calving flux determined at each of these edges, the non-dyamic floating cells
are looped over, and volume is marked for removal based on the calving flux on the edges of each cell.

The final phase handles dynamic floating cells that are eligible for calving.
There are two ways this can happen - either they are adjacent to the open ocean (no buffer of
non-dynamic cells present), or they are adjacent to non-dynamic floating cells that had their
entire volume removed by calving, but there still is calving demand remaining.
The first case is handled easily by calculating calving flux on edges between dynamic floating cells
and open ocean using the calving velocity and ice thickness of the dynamic cell itself.
In the second case, "leftover" required calving flux is distributed from a non-dynamic cell that
has been completely "drained" to any edges between that non-dynamic cell and dynamic cells upstream.
This is done by weighting the flux to be distributed by the length of each such edge in the direction
perpendicular to the calving flux.

Further details of the new scheme can be seen in the commit messages.  The commit history contains a number of temporary and outdated commits to retain the history of what didn't work in addition to what did.

### DIFF
--- a/src/core_landice/mode_forward/mpas_li_calving.F
+++ b/src/core_landice/mode_forward/mpas_li_calving.F
@@ -1101,27 +1101,14 @@ module li_calving
       real(kind=RKIND), pointer :: config_calving_eigencalving_parameter_scalar_value
       character (len=StrKIND), pointer :: config_calving_eigencalving_parameter_source
       logical, pointer :: config_print_calving_info
-      real (kind=RKIND), pointer :: config_sea_level
-      real(kind=RKIND), pointer :: config_calving_thickness
-      real (kind=RKIND), dimension(:), pointer :: thickness
-      real (kind=RKIND), dimension(:), pointer :: bedTopography
       real (kind=RKIND), dimension(:), pointer :: eigencalvingParameter
-      real (kind=RKIND), dimension(:), pointer :: calvingThickness
-      real (kind=RKIND), dimension(:), pointer :: requiredCalvingVolumeRate
       real (kind=RKIND), dimension(:), pointer :: calvingVelocity
       real (kind=RKIND), dimension(:), pointer :: eMax, eMin
       integer, dimension(:), pointer :: cellMask
       type (field1dInteger), pointer :: calvingFrontMaskField
-      integer, dimension(:), pointer :: calvingFrontMask
       real (kind=RKIND), pointer :: deltat  !< time step (s)
-      integer, dimension(:), pointer :: nEdgesOnCell ! number of cells that border each cell
-      integer, dimension(:,:), pointer :: cellsOnCell ! list of cells that neighbor each cell
-      integer, dimension(:,:), pointer :: edgesOnCell
       real (kind=RKIND), dimension(:), pointer :: dvEdge
-      integer, pointer :: nCells
       integer :: iCell, jCell, iNeighbor
-      real(kind=RKIND) :: cellCalvingFrontLength, cellCalvingFrontHeight
-      logical :: dynamicNeighbor
       integer :: err_tmp
 
       err = 0
@@ -1131,8 +1118,6 @@ module li_calving
                config_calving_eigencalving_parameter_scalar_value)
       call mpas_pool_get_config(liConfigs, 'config_calving_eigencalving_parameter_source', &
               config_calving_eigencalving_parameter_source)
-      call mpas_pool_get_config(liConfigs, 'config_sea_level', config_sea_level)
-      call mpas_pool_get_config(liConfigs, 'config_calving_thickness', config_calving_thickness)
 
       ! block loop
       block => domain % blocklist
@@ -1145,25 +1130,13 @@ module li_calving
          call mpas_pool_get_subpool(block % structs, 'scratch', scratchPool)
 
          ! get fields
-         call mpas_pool_get_dimension(meshPool, 'nCells', nCells)
          call mpas_pool_get_array(meshPool, 'deltat', deltat)
-         call mpas_pool_get_array(meshPool, 'nEdgesOnCell', nEdgesOnCell)
-         call mpas_pool_get_array(meshPool, 'edgesOnCell', edgesOnCell)
-         call mpas_pool_get_array(meshPool, 'cellsOnCell', cellsOnCell)
-         call mpas_pool_get_array(meshPool, 'dvEdge', dvEdge)
-         call mpas_pool_get_array(geometryPool, 'calvingThickness', calvingThickness)
          call mpas_pool_get_array(geometryPool, 'cellMask', cellMask)
-         call mpas_pool_get_array(geometryPool, 'thickness', thickness)
-         call mpas_pool_get_array(geometryPool, 'bedTopography', bedTopography)
          call mpas_pool_get_array(geometryPool, 'eigencalvingParameter', eigencalvingParameter)
-         call mpas_pool_get_array(geometryPool, 'requiredCalvingVolumeRate', requiredCalvingVolumeRate)
          call mpas_pool_get_array(geometryPool, 'calvingVelocity', calvingVelocity)
          call mpas_pool_get_array(velocityPool, 'eMax', eMax)
          call mpas_pool_get_array(velocityPool, 'eMin', eMin)
 
-         call mpas_pool_get_field(scratchPool, 'iceCellMask2',  calvingFrontMaskField)
-         call mpas_allocate_scratch_field(calvingFrontMaskField, .true.)
-         calvingFrontMask => calvingFrontMaskField % array
 
          ! get parameter value
          if (trim(config_calving_eigencalving_parameter_source) == 'scalar') then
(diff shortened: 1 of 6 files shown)


xylar:
The PR of the Devil!!!

matthewhoffman:
@xylar , that PR number is very apt for this branch...

sbrus89:
@xylar, we were just wondering the other day who would get this PR#. Looks like @matthewhoffman takes the cake!

xylar:
image

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1276,170 +1249,411 @@ end subroutine eigencalving
...
+      call mpas_log_write("== Uncalved volumes: Non-dynamic cells=$r; Dynamic cells=$r", realArgs=(/globalInfo(5),globalInfo(6)/))
+      if (((globalInfo(5) + globalInfo(6)) / (globalInfo(4) + 1.0e-30_RKIND) > 0.001_RKIND) .and. &
+          (globalInfo(4) > 1000.0_RKIND**2)) then ! Include some small amount of total calving for comparison
+         call mpas_log_write("Calving scheme failed to calve $r m^3. ($r% of total calved)", MPAS_LOG_WARN, &
+                 realArgs=(/globalInfo(5) + globalInfo(6), &
+                 100.0_RKIND * (globalInfo(5) + globalInfo(6)) / (globalInfo(4)+1.0e-30_RKIND)/))
+      endif
+      if (((globalInfo(5) + globalInfo(6)) / (globalInfo(4) + 1.0e-30_RKIND) > 0.1_RKIND) .and. &
+          (globalInfo(4) > 1000.0_RKIND**2)) then ! Include some small amount of total calving for comparison
+         call mpas_log_write("Calving scheme failed to calve an amount greater than 10% of the ice calved.  " // &
```
tongzhangice, line 1607: I was wondering what the case is for calved amount is < 10% here. Do we have accumulated uncalved volume over time?

matthewhoffman: This is checking what fraction of volume was uncalved on this timestep.  10% seemed like a reasonable accuracy threshold to tolerate.

matthewhoffman:
@tongzhangice , @trhille , I'd like to try to get this merged soon (next week, ideally).  I've assigned you both as reviewers.  Now that you both have had a chance to try these changes with your respective calving schemes, please complete your review when you are ready.  We've tested this pretty thoroughly now, so your reviews can be as thorough or simple as you want.  Email or slack me if you want more info about the review process.

trhille:
@matthewhoffman, sorry for the long wait. I'll prioritize this on Monday.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1276,170 +1249,411 @@ end subroutine eigencalving
...
+                  calvLengthCell = calvLengthCell + calvLengthEdge
+               endif
+            enddo
+
+            ! Now pass along the required calving volume relative to the interface length
+            do iNeighbor = 1, nEdgesOnCell(iCell)
+               iEdge = edgesOnCell(iNeighbor, iCell)
+               jCell = cellsOnCell(iNeighbor, iCell)
+               if (li_mask_is_dynamic_ice(edgeMask(iEdge))) then
+                  edgeLengthScaling = scale_edge_length(angleEdge(iEdge), uReconstructX(1,iCell), uReconstructY(1,iCell))
```
trhille, line 1528: This seems redundant. Isn't it exactly the same as the do-loop in lines 1052–1509?

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1276,170 +1249,411 @@ end subroutine eigencalving
...
+      do iCell = 1, nCellsSolve
+         if ( (li_mask_is_floating_ice(cellMask(iCell)) .and. li_mask_is_dynamic_ice(cellMask(iCell)) .and. &
+                 li_mask_is_dynamic_margin(cellMask(iCell)) ) .or. (groundedMarineMarginMask(iCell) == 1) ) then
+            do iNeighbor = 1, nEdgesOnCell(iCell)
+               iEdge = edgesOnCell(iNeighbor, iCell)
+               ! No need to check what type of edge this is - only required edges are nonzero
+               requiredCalvingVolumeDynCell(iCell) = requiredCalvingVolumeDynCell(iCell) + &
+                          requiredCalvingVolumeDynEdge(iEdge) ! Keep running total
+            enddo
+            ! c. Apply calving here
```
trhille, line 1558: Same comment as above (line 1456). Need to be clear that we are updating accounting, but not actually removing the thickness yet.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Would it make sense to add a namelist option to control this threshold?

matthewhoffman: I think the 10% threshold is a good guardrail for now.  As we use the calving routine more, we can adjust it as needed.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1237,10 +1206,13 @@ subroutine eigencalving(domain, err)
          ! meaning just one 'row' of cells per timestep.  This could be expanded to continue
          ! removing ice backward until all connected too-thin ice has been removed.
          ! Tests of the current implementation show reasonable behavior.
-         where (calvingFrontMask == 1 .and. thickness<config_calving_thickness)
-             calvingThickness = calvingThickness + thickness
-             thickness = 0.0_RKIND
-         end where
+         do iCell = 1, nCells
```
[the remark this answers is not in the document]
matthewhoffman: It's not necessary, but the performance team has told us that where-constructs do not optimize as well as do-loops, and they have asked us to eliminate them.  I try to remove them as I come across them.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1276,170 +1249,411 @@ end subroutine eigencalving
...
+      real(kind=RKIND) :: thkSum
+      integer :: thkCount
       real(kind=RKIND) :: removeVolumeHere
-      type (field1dReal), pointer :: cellVolumeField
-      real(kind=RKIND), dimension(:), pointer :: cellVolume
-      real(kind=RKIND), dimension(:), pointer :: uncalvedVolume
-      integer :: inwardNeighbors
-      integer :: uncalvedCount
-      real(kind=RKIND) :: uncalvedTotal
+      real(kind=RKIND) :: volumeAvailableToPass
```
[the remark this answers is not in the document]
matthewhoffman: new comment added

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1276,170 +1249,411 @@ end subroutine eigencalving
...
+      call mpas_pool_get_array(geometryPool, 'uncalvedVolumeDynCell', uncalvedVolumeDynCell)
+      call mpas_pool_get_array(geometryPool, 'groundedMarineMarginMask', groundedMarineMarginMask)
       call mpas_pool_get_array(meshPool, 'nEdgesOnCell', nEdgesOnCell)
-      call mpas_pool_get_array(meshPool, 'cellsOnCell', cellsOnCell)
       call mpas_pool_get_array(meshPool, 'areaCell', areaCell)
+      call mpas_pool_get_array(meshPool, 'deltat', deltat)
+
+      groundedMarineMarginMask(:) = 0
+      do iCell = 1, nCells
+         ! Define marine m      arginal cells as those that are (1) at the ice
```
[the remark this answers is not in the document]
matthewhoffman: Fixed

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1276,170 +1249,411 @@ end subroutine eigencalving
...
+                   .not. li_mask_is_ice(cellMask(jCell)) .and. bedTopography(jCell) < config_sea_level & ! ensure margin is w/ocn
+                   ) then
+                  edgeLengthScaling = scale_edge_length(angleEdge(iEdge), uReconstructX(1,iCell), uReconstructY(1,iCell))
+                  requiredCalvingVolumeNonDynEdge(iEdge) = calvingVelocity(iCell) * &
+                          edgeLengthScaling * dvEdge(iEdge) * thicknessForCalving(iCell) * deltat
+                  requiredCalvingVolumeNonDynCell(iCell) = requiredCalvingVolumeNonDynCell(iCell) + &
+                          requiredCalvingVolumeNonDynEdge(iEdge) ! Keep running total
+               endif
+            enddo
+            ! c. Apply calving here
```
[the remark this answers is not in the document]
matthewhoffman: Fixed

src/core_landice/mode_forward/mpas_li_calving.F
matthewhoffman: The first loop is to calculate calvLengthCell, which is needed before the second loop can be evaluated.

src/core_landice/mode_forward/mpas_li_calving.F
matthewhoffman: done

tongzhangice:
@matthewhoffman sorry for the delayed reply. Based on what I have been testing, I don't have any revisions or comments. I am still a bit concerned about the 10% threshold and the potential limitation with the timestep we have to choose to avoid the 10% error. Except for that, I think the current calving scheme is perfect. Thanks for your fix!

PR #670, opened 2020-08-26
Openmp pgi compiler bug workaround

mattdturner:
This PR creates a workaround for a PGI compiler issue that causes a failure when running the model w/ PGI + OpenMP.  The workaround is wrapping the OpenMP directives for a few loops (loops that declare an allocatable array as `firstprivate`) in a preprocessor flag (`CPRPGI`).  This flag is already used in other parts of MPAS.

This resolves issue https://github.com/E3SM-Project/E3SM/issues/3793

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
@@ -144,9 +144,11 @@ contains
 
       nCells = nCellsArray( 3 )
 
+#ifndef CPRPGI
       !$omp parallel
       !$omp do schedule(runtime) private(depth, k, depLev) &
       !$omp firstprivate(weights)
+#endif
       do iCell = 1, nCells
          depth = 0.0_RKIND
          do k = 1, maxLevelCell(iCell)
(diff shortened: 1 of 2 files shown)


mattdturner:
This PR passes all tests in the MPAS-O nightly regression suite on Grizzly with intel v19, and also passes the `SMS_P12x2.ne4_oQU240.A_WCYCL1850.compy_pgi.allactive-mach_mods`, `PET_Ln9_PS.ne30_oECv3_ICG.A_WCYCL1850S.compy_pgi.allactive-mach-pet`, and `SMS_P256x2.T62_oEC60to30v3.CMPASO-NYF.summit_pgi` E3SM tests (which FAIL without this bugfix).

mark-petersen:
@mattdturner just to confirm, MPAS-Ocean now runs a single thread throughout until it hits an `omp parallel` region. So on PGI, this protected loop just runs with one thread?

mattdturner:
@mark-petersen That is correct.  Unfortunately, that was the only way that I could get the simulations to run successfully w/ PGI.  

If it becomes a performance problem (which I doubt, since it is only on 2 somewhat small loops), then I can do a bit of a refactoring for those loops to see if I can get rid of the need for a `firstprivate`.  

Ideally, future versions of PGI will have the bug fixed.

PR #679, opened 2020-08-31
Made seaice_mesh_pool use more explicit

akturner:
Made seaice_mesh_pool use more explicit


akturner:
@amametjanov Reviewed by email and requested several additional explicit use statements.

PR #683, opened 2020-09-04
Changes to CVMix forcing and interface

vanroekel:
This changes the amount of shortwave radiation in the boundary layer
which decreases vertical resolution sensitivity.

It also removes the shear instability driven mixing scheme from
operating in the boundary layer to reduce grid scale noise.


vanroekel:
@mark-petersen  this is a modification for CVMix.  In E3SM testing it improves boundary layer depths and also reduces (but doesn't fix) sensitivity to moving from 60-64 layers.

I have not run the nightly yet, will try to do so soon.

mark-petersen:
Rebased, tested on grizzly with gnu and intel 19, debug and optimized. Passes nightly regression suite.

vanroekel:
@mark-petersen thanks for rebasing and testing!

One note, I just found a cvmix related bug that I'd like to add here.  Please wait on merging this one until I can test that fix.

mark-petersen:
No problem. I still need to carefully read through the code on all these, and will double-check with you before merging.

mark-petersen:
@vanroekel, I'm pinging everyone on ocean PRs, as the queue is starting to move. What is the status of this one? We can still hold it as a draft if you want to postpone.

vanroekel:
@mark-petersen I'm testing these changes again, but with an updated version.  So let's leave this in draft for a bit longer.

mgduda:

PR #684, opened 2020-09-04
Fixes limiting on redi k33

vanroekel:
on the vertical redi term, the surface scaling is currently being
applied twice.  This PR removes that.  It also reorganizes the code to
increase the strength of slope limiting on k33.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -415,13 +415,25 @@ contains
                slopeTriadDown(k, iCellSelf, iEdge) = &
                   slopeTaperDown*sfcTaperDown*slopeTriadDown(k, iCellSelf, iEdge)
 
+               ! Griffies 1998 eqn 34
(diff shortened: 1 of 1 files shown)


vanroekel:
@mark-petersen we had originally put k33 before limiting slopes to have stronger diffusion for stability, but testing in E3SM has shown it creates too much mixing and overly deep boundary layers.  This PR moves the k33 calculation post limiting to reduce the associated vertical mixing.  I have not run into any stability issues with this change (30 years of G-case and 40 years of B-case).

vanroekel:
I will run the nightly soon and post results.

mark-petersen:
Rebased, tested on grizzly with gnu and intel 19, debug and optimized. Passes nightly regression suite.

mark-petersen (approved):
@vanroekel please check this last commit. Otherwise, this is ready to merge.

vanroekel:
@mark-petersen looks good.  Thanks for taking out the extra calculation.

PR #633, opened 2020-07-15
Improves buoyancy gradient calculation in GM

vanroekel:
Currently the GM horizontal buoyancy gradient is computed at a fixed vertical index.  At a number of places in the ocean this is problematic where layer thickness changes quickly between cells (especially near the bottom for partial bottom cells).

PR #687, opened 2020-09-08
Revert GM buoyancy gradient calculation to v1 form

vanroekel:
E3SM testing suggests that the new formulation of buoyancy gradient calculation that leverages infrastructure from Redi is at least partly responsible for excessive heat uptake in v2 alpha simulations.  This PR will revert those changes back to the v1 formulation.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -133,9 +133,9 @@ contains
       real(kind=RKIND)                 :: h1, h2, areaEdge, c, BruntVaisalaFreqTopEdge, rtmp
       real(kind=RKIND)                 :: kappaGMEdge, sumN2, countN2, maxN, kappaSum, ltSum
       real(kind=RKIND)                 :: sumRi, RiTopOfEdge, zEdge, zMLD, sshEdge, sfcTaper
-      real(kind=RKIND) :: gradDensityTopOfEdge, dcEdgeInv, drhoDx, drhoDT, drhoDS, dTdx, dSdx, BVFcent
+      real(kind=RKIND) :: dcEdgeInv, drhoDx, drhoDT, drhoDS, dTdx, dSdx, BVFcent
       real(kind=RKIND) :: slopeTaperUp, slopeTaperDown, sfcTaperUp, sfcTaperDown, invAreaCell
-      real(kind=RKIND) :: lt1, lt2, dzdxTopOfEdge, drhodz1, drhodz2, dTdz, dSdz, gradDensityConstZTopOfEdge
+      real(kind=RKIND) :: lt1, lt2, dzdxTopOfEdge, drhodz1, drhodz2, dTdz, dSdz!, gradDensityConstZTopOfEdge
       real(kind=RKIND), dimension(:), allocatable :: dzTop, dTdzTop, dSdzTop, k33Norm
       ! Dimensions
       integer :: nsmooth, nCells, nEdges
@@ -144,7 +144,9 @@ contains
 
       ! gradDensityEdge: Normal gradient of density
       !           units: none
-      real(kind=RKIND), dimension(:, :), allocatable :: gradDensityEdge
+      real(kind=RKIND), dimension(:, :), allocatable :: gradDensityEdge, gradDensityTopOfEdge, &
+        gradDensityConstZTopOfEdge, dDensityDzTopOfCell, dDensityDzTopOfEdge, &
+        gradZMidEdge, gradZMidTopOfEdge
       ! drhoDzTopOfEdge: Gradient of density in vertical on edges
       !           units: None
       real(kind=RKIND), dimension(:, :), allocatable :: drhoDzTopOfEdge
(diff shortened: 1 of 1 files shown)


xylar:
@darincomeau is seeing too little sea ice in the northern hemisphere in his SORRM runs.  Could this be related?

milenaveneziani:
@darincomeau: what hash are you using for the SORRM runs?

jonbob:
test merge into E3SM passes:
PET_Ln9_PS.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel
PEM_Ln9_PS.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel

waiting on:
ERS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel

milenaveneziani:
@vanroekel: is this the specific MPAS-Model PR that is involved in the new GM buoyancy gradient calculation?:
https://github.com/MPAS-Dev/MPAS-Model/pull/633
so, anything before that uses the v1 formulation?

vanroekel:
@xylar and @darincomeau it is not clear if the v2 buoyancy gradient would result in too little sea ice.  It is possible for sure as upper ocean stratification is pretty strongly affected by the choice.  

@milenaveneziani it is before that where it was introduced - #445 , but it was fixed in the PR you mention

darincomeau:
> @darincomeau: what hash are you using for the SORRM runs?
@milenaveneziani the hash is `b9f3341f2`, and the branch is here: https://github.com/E3SM-Project/E3SM/tree/darincomeau/SORRMr2-clubbv2. The branch was split off master on May 29th I believe.

milenaveneziani:
@darincomeau: so #445 (merged in Apr) was there but not #633.

jonbob:
test merge into E3SM also passed:
ERS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel

milenaveneziani:
so, @darincomeau, @xylar, I agree with @vanroekel that it is hard to say. I think best will be to restart the SORRM runs after v2 ocean is finalized, since so many changes are going in (and out) of v2 beta ocean. But maybe by then we'll have a final SORRM mesh to focus on?

darincomeau:
Yeah these runs used GM_closure=N2_dependent, and redi was off, so SORRM reruns are definitely warranted.

xylar:
Okay, happy to discuss next steps for SORRM but this PR isn’t the place.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -219,7 +221,13 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
 
       allocate(gradDensityEdge(nVertLevels, nEdges), &
                dzdxEdge(nVertLevels, nEdges), &
-               drhoDzTopOfEdge(nVertLevels, nEdges))
+               drhoDzTopOfEdge(nVertLevels, nEdges), &
+              dDensityDzTopOfCell(nVertLevels, nCells), &
```
mark-petersen, line 225: Some of these new arrays need nVertLevels+1. I can fix that.

vanroekel:
an update, I've tested this for 20+ years in a B-case and the results are positive.  Biases are reduced pretty substantially in terms of heat uptake and MLD bias.  @mark-petersen if you could give take a look I'd appreciate it.  This one is high priority for WC.

src/core_ocean/shared/mpas_ocn_gm.F
vanroekel: okay, thanks!  I could also take a look and fix as well if you'd like

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -219,7 +221,13 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
 
       allocate(gradDensityEdge(nVertLevels, nEdges), &
                dzdxEdge(nVertLevels, nEdges), &
-               drhoDzTopOfEdge(nVertLevels, nEdges))
+               drhoDzTopOfEdge(nVertLevels, nEdges), &
```
mark-petersen, line 224: @vanroekel in the current code, `dzdxEdge` and `drhoDzTopOfEdge` are allocated and set to zero but never used. I can take them out, but can you double check that not using them was intentional, rather than a copy and paste error?

vanroekel: @mark-petersen this was a copy-paste error   Thanks for catching that!

mark-petersen:
rebased, fixed index out of bounds errors. Now passes nightly regression with gnu debug.

mark-petersen:
@mattdturner This PR changes several loops to an older formulation from last year. Could you look at the OMP pragmas around the revised loops for this PR? We need to make sure it retains the newer format since #513.

mattdturner:
There were some missing `!$omp parallel` regions around some of the loops.  I added them, and the thread test PASSes on Cori w/ intel + debug.

Approved based on this test, and the other testing by @mark-petersen

mattdturner (approved):

mark-petersen:
Thanks @mattdturner. I also ran a test with gnu 6 debug, passes all.

mark-petersen (approved):
Reviewed code, passes stand-alone tests. Testing in E3SM now, but @vanroekel has tested as well.

PR #680, opened 2020-09-01
Update Earth radius to the CIME value

xylar:
Currently, COMPASS uses 3 different Earth radii.  With this merge and a corresponding update to MPAS-Tools, all Earth radii are the CIME value.

This PR addresses MPAS-Dev/MPAS-Model#549 along with https://github.com/MPAS-Dev/MPAS-Tools/pull/341.

PR #688, opened 2020-09-09
Update to version 0.1.11 of the compass environment

xylar:
This brings in several changes from MPAS-Tools:
* Changes to make sure the Earth radius is consistent throughout COMPASS and `mpas_tools.ocean` by using the CIME value (https://github.com/MPAS-Dev/MPAS-Tools/pull/347 and https://github.com/MPAS-Dev/MPAS-Tools/pull/341)
* Improved performance for interpolation of meshDensity and bathymetry (https://github.com/MPAS-Dev/MPAS-Tools/pull/344)
* Reorganization of `mpas_tools.mesh.creation` (https://github.com/MPAS-Dev/MPAS-Tools/pull/314)

### DIFF
--- a/testing_and_setup/compass/README_ocean.md
+++ b/testing_and_setup/compass/README_ocean.md
@@ -6,11 +6,11 @@ To set up and run ocean test cases from COMPASS, you will need a conda
 environment.  First, install Miniconda3 (if miniconda is not already
 installed), then create a new conda environment as follows:
 ``` bash
-conda create -n compass_0.1.9 -c conda-forge -c e3sm python=3.7 compass=0.1.9
+conda create -n compass_0.1.11 -c conda-forge -c e3sm python=3.7 compass=0.1.11
 ```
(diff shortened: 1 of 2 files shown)


xylar:
This update should have been part of https://github.com/MPAS-Dev/MPAS-Model/pull/680 but I missed adding it.

@mark-petersen, could you include this with the current round of COMPASS changes?  It is needed to document the correct environment to use for the current state of COMPASS on `ocean/develop`, which will not work correctly with v. 0.1.9 as stated in the documentation.

I'll try not to forget this important update in the future!

mark-petersen (approved):
I ran a quick test with 0.1.11 with gnu on grizzly. Nightly regression suite runs fine.

xylar:
Thanks @mark-petersen.  Just to note, we wouldn't expect the `global_ocean` test cases to be bit-for-bit with this change because of the changes in Earth radius.

PR #689, opened 2020-09-09
Update jigsaw_to_netcdf to use version from mpas_tools package

xylar:
This merge updates calls to jigsaw_to_netcdf to use the version from `mpas_tools` rather than from the outdated `compass/ocean/jigsaw_to_MPAS` location.

This merge also updates `README_landice.md` to describe the latest version of the `compass` environment and associated metapackage (v0.1.11).

This merge also adds `landice/load_compass_env/sh` as a potentially convenient shortcut for users to source on supported machines to get the compatible version of the compass environment.  This script gets symlinked into each test case if you use the `--link_load_compass` flag.

### DIFF
--- a/testing_and_setup/compass/README_landice.md
+++ b/testing_and_setup/compass/README_landice.md
@@ -2,16 +2,28 @@
 
 ## COMPASS conda environment
 
-To set up and run landice test cases from COMPASS, you will need a conda
+To set up and run ocean test cases from COMPASS, you will need a conda
 environment.  First, install Miniconda3 (if miniconda is not already
 installed), then create a new conda environment as follows:
 ``` bash
-conda create -n compass_py3.7 -c conda-forge -c xylar python=3.7 geometric_features mpas_tools jigsaw jigsawpy metis pyflann scikit-image basemap pyamg ffmpeg pyqt
+conda create -n compass_0.1.11 -c conda-forge -c e3sm python=3.8 compass=0.1.11
 ```
 Each time you want to work with COMPASS, you will need to run:
 ```
-conda activate compass_py3.7
+conda activate compass_0.1.11
 ```
+
+An appropriate conda environment is already available on Los Alamos National
+Laboratory's Institutional Computing (LANL IC) machines as well as Anvil, Compy
+and Cori.  In each case, you will run:
+```
+source <base_path>/load_latest_compass.sh
+```
+Values of `<base_path>` are:
+* grizzly and badger - `/usr/projects/climate/SHARED_CLIMATE/anaconda_envs`
+* anvil (blues) - `/lcrc/soft/climate/e3sm-unified/`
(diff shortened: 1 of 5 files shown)


xylar:
@matthewhoffman, the main change here is that I moved the old `triangle_jigsaw_to_netcdf.py` over to MPAS-Tools as part of getting rid of `ocean/jigsaw_to_MPAS`.  It has been split into `triange_to_netcdf` and `jigsaw_to_netcdf`, and i'm quite sure you only need the latter but still worth checking.  There won't be a `triangle_jigsaw_to_netcdf.py` availabe to you once `landice/develop` and `ocean/develop` get synced up with one another so switching over to the `mpas_tools` version of the tool from within the `compass` conda environment is the safer option in the long run.

xylar:
As a side note, the ocean code has moved entirely over to https://github.com/dengwirda/jigsaw-python and doesn't require any Matlab anymore.  The land-ice team may want to do the same so we can continue to share tools.  Take a look at 
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/mpas_tools/mesh/creation/build_mesh.py#L104
and
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/mpas_tools/mesh/creation/jigsaw_driver.py
as possible placed to get you started.  The ocean uses these in:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/mpas_tools/ocean/build_mesh.py#L74

xylar:
Since I don't have Matlab and I'm not very familiar with the 3 test cases I altered, I wasn't able to test them.  If you want to give me some instructions, I'm happy to learn, though.  Or I can leave it up to you to do a couple of (hopefully) quick tests.

matthewhoffman:
@xylar , I will take a look at this soon.  Thanks for looking out for us.  The Matlab code you see is legacy code that has not been updated (and I think will remain that way until we find a need to rerun them).  The more recent meshes we have worked on use the python interface, so I will focus on making sure those work.

matthewhoffman:
@xylar , actually, I have twisted the arm of @trhille , to test this out with the new GIS meshes he has been making.  I told him he could wait until next week to look at this if he is busy, but let us know if you are wanting to resolve this sooner.  I know I've let this sit a few weeks already.

xylar:
No rush.  It would be great to have @trhille use this for a new mesh.  It would be great if you can start transitioning to python instead of Matlab for future meshes but I totally understand that it doesn't make sense to redo Matlab code.

Please do let me know if there's a test or 2 I can run in the future before I hand PRs like this off to you all.

matthewhoffman:
@xylar , yes we have been using python exclusively for the past year or so.  It's only the older meshes that were still using Matlab and there was no foreseeable need to use those meshes again in the future.

xylar:
Moved to https://github.com/MPAS-Dev/compass/pull/14

PR #690, opened 2020-09-10
Fixes surface buoyancy forcing calculation

vanroekel:
Currently the KPP buoyancy flux is assuming the ```seaIceFreshWaterFlux``` and ```iceBergFreshwaterFlux``` enter the ocean at 0C, but the surface_bulk_forcing code assumes it enters at the freezing temperature.  This PR unifies the treatment.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -2898,11 +2898,9 @@
 		/>
 		<var name="seaIceFreshWaterFlux" type="real" dimensions="nCells Time" units="kg m^{-2} s^{-1}"
 			 description="Fresh water flux from sea ice at cell centers from coupler. Positive into the ocean."
-			 packages="thicknessBulkPKG"
 		/>
 		<var name="icebergFreshWaterFlux" type="real" dimensions="nCells Time" units="kg m^{-2} s^{-1}"
 			 description="Fresh water flux from iceberg melt at cell centers from coupler. Positive into the ocean."
-			 packages="thicknessBulkPKG"
 		/>
 		<var name="riverRunoffFlux" type="real" dimensions="nCells Time" units="kg m^{-2} s^{-1}"
 			 description="Fresh water flux from river runoff at cell centers from coupler. Positive into the ocean."
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -394,7 +394,7 @@
 					description="If true, use the Community Vertical MIXing routines to compute vertical diffusivity and viscosity"
 					possible_values="True or False"
 		/>
-		<nml_option name="config_cvmix_use_kpp_buoyancyForcing_fix" type="logical" default_value=".false." units="NA"
+		<nml_option name="config_cvmix_use_kpp_buoyancyForcing_fix" type="logical" default_value=".true." units="NA"
 					description="If true CVMix KPP buoyancy forcing fix for sea ice is enabled"
 					possible_values="True or False"
 		/>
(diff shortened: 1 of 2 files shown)


vanroekel:
compare code to https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_surface_bulk_forcing.F#L486-L490

vanroekel:
I've started a G-case to test this and will report back when it's finished.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1531,9 +1533,15 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, diagno
 
          nonLocalSurfaceTracerFlux(indexTempFlux, iCell) = activeTracersSurfaceFlux(indexTempFlux,iCell) &
                  + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell)  &
-                 - fracAbsorbed * (rainFlux(iCell) + evaporationFlux(iCell)) * activeTracers(indexTempFlux,1,iCell)/rho_sw  &
-                 - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell) &
-                  * min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)/rho_sw
+                 - fracAbsorbed * (rainFlux(iCell) + evaporationFlux(iCell)) * activeTracers(indexTempFlux,1,iCell)/rho_sw &
+                 - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell)* min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)/rho_sw
+
+         if(config_cvmix_use_kpp_buoyancyForcing_fix) then
```
mark-petersen, line 1539: @philipwjones @mattdturner or @amametjanov:
We put this if flag here for backwards compatibility. It is within a cell loop. 

Is this still a performance sin, or do the compilers handle this now for flags that never change during the run?

philipwjones: It's a sin my son and will require at least 3 Hail Marys. Particularly in this case, where the loop is complex enough that the compiler can't break it up, it's not optimal. I don't know that it'll be a huge performance hit on CPU architectures but will probably need to be broken out for the GPU version.

maltrud: this will hopefully be a relatively short term hack.  we just need to remember to remove it once we are satisfied with the new (correct) code.

vanroekel: agreed.  Once we test this in WC and are happy with results, I'll make a PR to remove the if statement.  I can't imagine this is in for more than a week.

mark-petersen: Perfect, thanks for the perspective. I made an issue to help us remember: https://github.com/MPAS-Dev/MPAS-Model/issues/692

src/core_ocean/Registry.xml
```diff
@@ -394,6 +394,10 @@
 					description="If true, use the Community Vertical MIXing routines to compute vertical diffusivity and viscosity"
 					possible_values="True or False"
 		/>
+		<nml_option name="config_cvmix_use_kpp_buoyancyForcing_fix" type="logical" default_value=".false." units="NA"
```
mark-petersen, line 397: @vanroekel do you want this default true in E3SM? If so, I'll make it true here as well.

mark-petersen (approved):
Reviewed code, passes stand-alone tests. Testing in E3SM now, but @vanroekel has tested as well.

PR #694, opened 2020-09-15
New Mesh: ECwISC30to60kmL60E3SMv2r01 and add 1900 state estimate initial condition

xylar:
This merge updates the ECwISC30to60 test case to allow for use of the Met Office EN4 1900 state estimate as the ocean initial condition.  It attempts to create a mesh with ice-shelf cavities that is as similar as possible to #669 (EC30to60kmL60E3SMv2r03) with the new IC from #671.


xylar:
I am currently spinning up this mesh here:
```
/lustre/scratch4/turquoise/xylar/spin_up_ECwISC30to60kmL60E3SMv2r01_1900_EN4/ocean/global_ocean/EC60to30wISC/
```
I will run a 1-hour spin-up following @vanroekel's suggestions for the EC30to60km mesh.

xylar:
@jonbob, @darincomeau, @vanroekel and @mark-petersen, I think the review process here is just to try to verify in whatever ways we can that the initial condition produced here is, indeed, similar to the EC30to60kmL60E3SMv2r03 one we're currently testing.  I think the mesh without ice-shelf cavities (land-ice mask is zero) should be the same as in EC30to60kmL60E3SMv2r03 and I'll try to verify that later today.  The vertical grid near Antarctica will be different because we have to prevent Haney-number blow-up when the ocean surface dips under ice shelves.  Mostly, I just wanted to make sure you know that this PR would be the place to note issues you see with this mesh or questions you might have.

@milenaveneziani and @proteanplanet, I initially thought we might want your review but I think we're at a point where the only priority is making this mesh as similar to EC30to60kmL60E3SMv2r03 as possible.  You're welcome to weigh in but you shouldn't feel like you need to.

xylar:
I looked at the culled mesh from the EC30to60kmL60E3SMv2r03 and compared that to the `no_ISC_culled_mesh.nc` for this setup.  They have very slightly different numbers of cells: 236853 vs 236855.  That's not good news but maybe we can live with it.  I looked at the setup I used to create the EC30to60E2r2 and EC30to60E2r3 meshes, and I can't see any difference.  I ran the cell culling step on the login node on Grizzly rather than a compute node.  But I just cant see how that could make the difference.

If folks are concerned, I can keep investigating.  Images from Paraview look identical to my eyes:
EC30to60_culled
no_IsC_culled

xylar:
A note to self: I need to bring over `create_files_for_e3sm.py` from by branch of the `EC30to60E2r3` `EN4_1900` work.  Currently, there's a complaint about an unrecognized flag.

xylar:
Moved to https://github.com/MPAS-Dev/compass/pull/15

PR #697, opened 2020-09-16
Hydro model updates for Humboldt and MISMIP+

matthewhoffman:
Updates to subglacial hydrology model to support the Humboldt Glacier and MISMIP+ subglacial hydrology simulations.

* Add ocean_connection_N hydrology parameterization: calculates effective pressure assuming perfect connection with the ocean (used for MISMIP+)
* Adds option to limit max N used by velocity solver
* A number of fixes to the pressure boundary conditions at the edge of hydro domain
* Improvements to sheet flow and channel flow at low flow situations
* Other small fixes/improvements to hydro model


matthewhoffman:
Note: minor answer change for hydro tests due to change in use of regularization in commit 1bab0538e .  I checked the hydro-radial case run to steady state against the near-exact solution, and there is negligible change (old top, new bottom):
image

Issue #685, opened 2020-09-04
runtime error in ALE routine with gnu 6.4.0 debug during EC60to30 spinup


mark-petersen:
This error occurs only with `gcc/6.4.0` with debug, not intel or `gcc/5.3.0`, and during EC60to30 `prep_spinup1`. It does not occur in the nightly regression suite.

This is an error in the first timestep.  Stack trace:
```
#1  0x8ec46d in __ocn_thick_ale_MOD_ocn_ale_thickness
	at /usr/projects/climate/lvanroekel/meshprs/src/core_ocean/shared/mpas_ocn_thick_ale.F:162
#2  0x91834f in __ocn_diagnostics_MOD_ocn_vert_transport_velocity_top
	at /usr/projects/climate/lvanroekel/meshprs/src/core_ocean/shared/mpas_ocn_diagnostics.F:1060
#3  0xa5c300 in __ocn_time_integration_split_MOD_ocn_time_integrator_split
	at /usr/projects/climate/lvanroekel/meshprs/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F:1519
#4  0xa4b88e in __ocn_time_integration_MOD_ocn_timestep
	at /usr/projects/climate/lvanroekel/meshprs/src/core_ocean/mode_forward/mpas_ocn_time_integration.F:111
#5  0x8dac07 in __ocn_forward_mode_MOD_ocn_forward_mode_run
	at /usr/projects/climate/lvanroekel/meshprs/src/core_ocean/mode_forward/mpas_ocn_forward_mode.F:588
#6  0x9fdcb6 in __ocn_core_MOD_ocn_core_run
	at /usr/projects/climate/lvanroekel/meshprs/src/core_ocean/driver/mpas_ocn_core.F:107
#7  0x4094a9 in __mpas_subdriver_MOD_mpas_run
	at /usr/projects/climate/lvanroekel/meshprs/src/driver/mpas_subdriver.F:347
#8  0x4082b3 in mpas
	at /usr/projects/climate/lvanroekel/meshprs/src/driver/mpas.F:16
#9  0x40830f in main
	at /usr/projects/climate/lvanroekel/meshprs/src/driver/mpas.F:10
```

vanroekel:
Thanks for tracking this down @mark-petersen!  In the meantime, I'll update my modules on grizzly to avoid this.

mark-petersen:
closed by #676 .

mark-petersen:
[...]

PR #698, opened 2020-09-17
Fix OMP Directives and Loop Bounds in Ocean

mattdturner:
This PR is a bugfix that corrects errors in a few OpenMP private clauses, and fixes loop bounds in an ALE thickness routine.  

For additional details of the errors that this PR resolves, see https://github.com/E3SM-Project/E3SM/issues/3826

This also potentially fixes https://github.com/MPAS-Dev/MPAS-Model/issues/685 , but I will need to test to confirm.

### DIFF
--- a/src/core_ocean/analysis_members/mpas_ocn_eliassen_palm.F
+++ b/src/core_ocean/analysis_members/mpas_ocn_eliassen_palm.F
@@ -482,7 +482,6 @@ contains
          !--------------------------------------------------
          ! allocate scratch field variables
          !--------------------------------------------------
-         !$omp master
          allocate(firstLayerBuoyCoor(ncells), &
                   lastLayerBuoyCoor(ncells), &
                   heightMidBuoyCoor(nBuoyancyLayers, nCells), &
@@ -520,8 +519,6 @@ contains
                   uPVMidBuoyCoorEA(nBuoyancyLayers, nCells), &
                   vPVMidBuoyCoorEA(nBuoyancyLayers, nCells), &
                   PVFluxTest(2,nBuoyancyLayers, nCells))
-         !$omp end master
-         !$omp barrier
 
          !--------------------------------------------------
          ! assign pointers for ensemble average (EA) and thickness-weighted averaged state
@@ -1063,7 +1060,6 @@ contains
          !-------------------------------------------------------------
          ! deallocate scratch space and test space variables
          !-------------------------------------------------------------
-         !$omp master
          deallocate(firstLayerBuoyCoor, &
                     lastLayerBuoyCoor, &
                     heightMidBuoyCoor, &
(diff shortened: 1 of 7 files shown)


mattdturner:
This branch PASSes the nightly regression suite, as well as the following E3SM tests:
```
```

Prior to this PR, the `SMS_D` tests listed above all failed.

This is non-bfb in the `analysis_test` when comparing to a baseline.  This is because the baseline data that is generated is incorrect due to the baseline being generated with a variable in a `private` clause that should not be private (`interp_local`).  Even when running single-threaded (as the baseline is), if the executable is compiled w/ OpenMP then the `private` clause still applies.  This resulted in `interp_local` being uninitialized in the loop, resulting in all interpolation coefficients having a value of `0.0`.  This also means that comparing a single-threaded run that is compiled with OpenMP against one without OpenMP should also be non-bfb.

@philipwjones Continuing on our previous discussion, Valgrind likely would have pointed out these issues even though all of our tests were PASSing.

mattdturner:
This PR does indeed appear to fix #685 when running on Badger (I first confirmed that the code fails in the EC60to30 prep_spin_up1 test without these changes).

mark-petersen (approved):
@mattdturner, this is perfect. I was able to test the PR in the way that we discussed. These are all nightly regression suites with gnu 6 on badger:
1. current ocean/develop, OPENMP=false versus multi-threaded: Global ocean cases **FAIL** to match
1. This PR, OPENMP=false versus multithreaded: **PASS** match bfb
1. This PR OPENMP=false versus ocean/develop OPENMP=false: **PASS** match bfb

This is excellent! Thanks for testing the EC60to30 spinup as well.

mark-petersen:
Also just did this same comparison but with Intel19:
1. current ocean/develop, OPENMP=false versus multi-threaded: PASS match bfb
2. This PR versus ocean/develop, both OPENMP=false: PASS match bfb

That is good to check, but it wasn't a problem to begin with.

mark-petersen:
@ambrad, @ndkeen, @rljacob, @vanroekel, @golaz
Just a note, per the discussion at yesterday's water cycle meeting. The [comment](https://github.com/MPAS-Dev/MPAS-Model/pull/698#pullrequestreview-491583538) just above, we had a non-bfb match between compiling without openmp, and compiling with openmp. In this case, the compiling without was correct, and the compiling with had some arrays that were being filled incorrectly, but just for certain compilers (gnu). We now test MPAS-Ocean regularly between with/without openmp compile flags.

@ambrad if your E3SM tests were run from master after https://github.com/E3SM-Project/E3SM/pull/3844 was merged (Sept 23) then this is not the specific source of your bfb mismatch. Still, I think the current mismatch indicates that there is a bug, and the no-threading is much more likely to be correct. Until the mismatch is found, I would recommend compiling with OpenMP off.

ambrad:
Thanks @mark-petersen; this is good information. The branch @ndkeen ran for his PE layouts analysis does _not_ include PR 3844. For clarity, the branch we are advertising to others (https://github.com/E3SM-Project/E3SM/pull/3879) _does_, because it's even with master as of Tues 6 Oct night. Thus, the 4 out of 24 threaded runs that diff may be explained by not having PR 3844 included in the build.

PR #699, opened 2020-09-18
Alter vertical transport in stage 1 of split explicit

mark-petersen:
Add back computation of vertical transport, removed in 44e5a6c6

The original change was done to make the first stage of split explicit Lagrangian. This PR changes it back, to see if that aids stability and heat transport problems.

### DIFF
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
@@ -312,8 +312,6 @@ module ocn_time_integration_split
          call mpas_pool_get_array(statePool, 'lowFreqDivergence', lowFreqDivergenceCur, 1)
          call mpas_pool_get_array(statePool, 'lowFreqDivergence', lowFreqDivergenceNew, 2)
 
-         call mpas_pool_get_array(diagnosticsPool, 'vertAleTransportTop', vertAleTransportTop)
-
          call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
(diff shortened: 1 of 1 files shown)


mark-petersen:
Passes nightly regression suite in debug for gnu 6 and intel 19.

vanroekel:
Tested in E3SM, changes are mostly small, but I think this PR is important.  As we think about further alterations to GM, we don't want to inadvertently add the bolus velocity.  I think we should do one of two things
```
1 - Merge this PR as is
2 - do stage 1 as "Lagrangian" and in stage 2 have momentum advection be due to normal velocity only
```
since 1 seems cleaner, I'd advocate for merging this PR

vanroekel (approved):
approved based on visual inspection and E3SM testing in G and B-cases

mark-petersen:
I agree with (1). I think that on a time stepping scheme with two iterations, it is better to have a uniform method for all iterations, so that the 'predictor' solution is as close as possible to the 'corrector' solution. 

@jonbob I want to do a little more testing, but will plan to get this in today.

jonbob:
@mark-petersen - sounds good. I'll make sure there's time at the end of the day to get this in when it's ready

mark-petersen:
Tested in E3SM. passes:
```
PEM_Ld5_P512.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_intel
PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-haswell_intel
```

PR #703, opened 2020-09-23
Fixes and Updates to CVMix

vanroekel:
This commit introduces 3 developments
1) Fixes indexing issue in normalVelocity Averaging
2) removes shear instability mixing from the boundary layer
3) forces convection to be one full layer below the boundary layer

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
@@ -530,13 +530,13 @@ contains
                  normalVelocitySum(1) = normalVelocity(1, iEdge)*layerThicknessEdge(1,iEdge)
                  layerThicknessEdgeSum(1) = layerThicknessEdge(1,iEdge)
 
-                 do kIndexOBL = 2, maxLevelEdgeTop(iEdge)
+                 do kIndexOBL = 2, maxLevelCell(iCell)
                     normalVelocitySum(kIndexOBL) = normalVelocitySum(kIndexOBL-1) + &
                                       layerThicknessEdge(kIndexOBL, iEdge)*normalVelocity(kIndexOBL, iEdge)
                     layerThicknessEdgeSum(kIndexOBL) = layerThicknessEdgeSum(kIndexOBL-1) + layerThicknessEdge(kIndexOBL, iEdge)
                  end do
 
-                 do kIndexOBL = 1, maxLevelEdgeTop(iEdge)
+                 do kIndexOBL = 1, maxLevelCell(iCell)
                     normalVelocityAv = normalVelocitySum(surfaceAverageIndex(kIndexOBL )) / &
                         layerThicknessEdgeSum(surfaceAverageIndex(kIndexOBL))
                     delU2 = ( normalVelocityAv - normalVelocity(kIndexOBL, iEdge) )**2
@@ -674,23 +674,10 @@ contains
             ! intent out of BoundaryLayerDepth is boundary layer depth measured in meters and vertical index
             indexBoundaryLayerDepth(iCell) = cvmix_variables % kOBL_depth
 
-
-
-            if(config_cvmix_kpp_matching .eq. 'SimpleShapes') then
-                do k = 1, int(indexBoundaryLayerDepth(iCell))
-                   vertViscTopOfCell(k,iCell) = vertViscTopOfCell(k,iCell) + cvmix_variables % Mdiff_iface(k)
-                   vertDiffTopOfCell(k,iCell) = vertDiffTopOfCell(k,iCell) + cvmix_variables % Tdiff_iface(k)
-                end do
-                do k = int(indexBoundaryLayerDepth(iCell))+1, maxLevelCell(iCell)+1
-                   vertViscTopOfCell(k,iCell) = cvmix_variables % Mdiff_iface(k)
-                   vertDiffTopOfCell(k,iCell) = cvmix_variables % Tdiff_iface(k)
-                enddo
-            else
-                do k = 1, maxLevelCell(iCell) + 1
-                   vertViscTopOfCell(k, iCell) = cvmix_variables % Mdiff_iface(k)
-                   vertDiffTopOfCell(k, iCell) = cvmix_variables % Tdiff_iface(k)
-                end do
-            endif
+            do k = 1, maxLevelCell(iCell) + 1
+                vertViscTopOfCell(k, iCell) = cvmix_variables % Mdiff_iface(k)
+                vertDiffTopOfCell(k, iCell) = cvmix_variables % Tdiff_iface(k)
+            end do
 
             ! store non-local flux terms
             ! these flux terms must be multiplied by the surfaceTracerFlux field
@@ -713,7 +700,7 @@ contains
             ! add convective mixing to vertical viscosity/diffusivity
             ! if using KPP, then do not apply convective mixing within the ocean boundary layer
             if(config_use_cvmix_kpp) then
-               do k = int(indexBoundaryLayerDepth(iCell)) + 1, maxLevelCell(iCell)
+               do k = int(ceiling(indexBoundaryLayerDepth(iCell))) + 1, maxLevelCell(iCell)
                   vertViscTopOfCell(k,iCell) = vertViscTopOfCell(k,iCell) + cvmix_variables % Mdiff_iface(k)
                   vertDiffTopOfCell(k,iCell) = vertDiffTopOfCell(k,iCell) + cvmix_variables % Tdiff_iface(k)
                enddo
--- a/src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
@@ -404,10 +404,10 @@ contains
             ! The standard atan formulation is approximated following equation (13) of Girones et al (2012)
             ! https://doi.org/10.1109/MSP.2012.2219677
             do k = 1,maxLevelCell(iCell)
-               x = (abs(cvmix_variables % zw_iface(k)) - config_cvmix_bryanLewis_transitionDepth) / &
-                   config_cvmix_bryanLewis_transitionWidth
-               vertDiffTopOfCell(k, iCell) = vertDiffTopOfCell(k, iCell) + config_cvmix_bryanLewis_bl1 + &
-                        (config_cvmix_bryanLewis_bl2/2.0_RKIND)*(sign(1.0_RKIND, x)*(0.596227_RKIND*     &
+               x = (abs(cvmix_variables % zw_iface(k)) - config_cvmix_BryanLewis_transitionDepth) / &
+                   config_cvmix_BryanLewis_transitionWidth
+               vertDiffTopOfCell(k, iCell) = vertDiffTopOfCell(k, iCell) + config_cvmix_BryanLewis_bl1 + &
+                        (config_cvmix_BryanLewis_bl2/2.0_RKIND)*(sign(1.0_RKIND, x)*(0.596227_RKIND*     &
                         abs(x)+x**2)/(1.0_RKIND+1.192454_RKIND*abs(x)+x**2))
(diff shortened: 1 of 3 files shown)


vanroekel:
This is being tested in E3SM now.  Will post results once testing is complete.

vanroekel:
@mark-petersen this PR is ready for your review.  It has run stably in G- and B-cases.  Mixed layer depth biases are improved with these changes.

vanroekel:
@mark-petersen I'm going to add one more cvmix feature, please hold on this PR.  I'll have this ready for you tomorrow morning.

vanroekel:
@mark-petersen this one is ready now.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -283,7 +283,7 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, diagnost
       !
       ! start by adding the mininum background values to the viscocity/diffusivity arrays
       !
-      if (cvmixBackgroundOn) then
+      if (config_cvmix_background_scheme == 'constant') then
```
mark-petersen, line 286: @philipwjones and @mattdturner as we write new code, should we avoid string comparisons when possible? They are easier to read, but we can use logicals if it is better for performance.

mattdturner: Per @philipwjones in this comment (https://github.com/MPAS-Dev/MPAS-Model/issues/394#issuecomment-553016388):

> Either way, please convert strings to other forms. Simple binary options should be implemented as logical flags. For conditions where we have multiple options, the preferred way would be to use the Fortran equivalent of enums - basically named integer parameters.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -384,8 +384,17 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, diagnost
          cvmix_variables % WaterDensity_cntr => density(1:maxLevelCell(iCell), iCell)
          cvmix_variables % AdiabWaterDensity_cntr => displacedDensity(1:maxLevelCell(iCell), iCell)
 
-         ! fill Ri
          if (kpp_stage == 2) then
+         if (config_cvmix_background_scheme == 'bryanLewis') then
+            do k = 1,maxLevelCell(iCell)
+               vertDiffTopOfCell(k, iCell) = vertDiffTopOfCell(k, iCell) + config_cvmix_bryanLewis_bl1 + &
+                        (config_cvmix_bryanLewis_bl2/pii)*atan((abs(cvmix_variables % zw_iface(k)) -  &
+                        config_cvmix_bryanLewis_transitionDepth) / config_cvmix_bryanLewis_transitionWidth)
```
mark-petersen, line 392: @vanroekel I'm just being thorough - this is equation 9 from the paper, right?
image
with 
```
  95     config_cvmix_bryanLewis_bl1 = 8.0e-5
  96     config_cvmix_bryanLewis_bl2 = 1.05E-4
  97     config_cvmix_bryanLewis_transitionDepth = 2500
  98     config_cvmix_bryanLewis_transitionWidth = 222.   (1/222. = 0.0045)
```
It all looks right, except he has a 10^4 out front (also in mks units) and you have 10^{-4}.  The paper must have a typo, because he talks about values in the range of 10^{-4}, which makes sense for horizontal mixing.

Also, `abs(cvmix_variables % zw_iface(k))` must be some depth offset.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
mark-petersen: Thanks @mattdturner, that is helpful. Can you point me to an example of enums? Is that when we use caps variables at the top of the module?

philipwjones: @mark-petersen see for example how we treat equation of state choices

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
philipwjones: atan is an expensive function. how variable is zw_iface? any chance we could mitigate by precomputing? Understand if we can't. Looks like we could precompute a number of other factors in this expression too.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
mattdturner: I think the idea is to use strings in the namelist, then in `init` convert to integers and create new variables for the different values.  For example, keep `config_cvmix_background_scheme` as it is in the namelist.  Then in init
```
! Define the named integer parameters
integer :: config_cvmix_background_scheme_type
integer, parameter :: &
    backgroundSchemeNone = 0, &
    backgroundSchemeConstant = 1, &
    backgroundSchemeBryanLewis = 1

if config_cvmix_background_scheme == 'constant':
    config_cvmix_background_scheme_type = backgroundSchemeConstant
else if config_cvmix_background_scheme == 'bryanLewis':
    config_cvmix_background_scheme_type = backgroundSchemeBryanLewis
else if config_cvmix_background_scheme == 'none':
    config_cvmix_background_scheme_type == backgroundSchemeNone
```

Then use a `select case (convig_cvmix_background_scheme_type)` in the code

mattdturner: Or reference equation of state like @philipwjones said (I wasn't aware of where else in the model this was done)

mattdturner:
Code passes nightly regression suite / intelv19 + DEBUG on grizzly.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
vanroekel: @mark-petersen yes that's the one.

@philipwjones I don't imagine z changing too much, less sure near ice cavities.  I could move that to the init phase of KPP.

mark-petersen: Looks like `cvmix_variables % zw_iface(k)` is simply depth (z in eqn 9) - @vanroekel is that right?

Over z=0 to 5000, tanh() would have inputs of -10 to 10. It is that same function all the time. In straight z-level we could pre-compute once. We could do some approximation here, like a linear fit at each mid-depth for z-star. It might not apply to more general vertical coordinates, but we don't use those in practice.

@vanroekel is this potentially a default option in E3SM, or just for testing?

xylar: We're hoping to move to simulations on a mesh with ice-shelf cavities and I don't think it's at all safe to assume z-level anymore.

vanroekel: @mark-petersen yes this is depth of interfaces (zTop but referenced to SSH not z=0).  Instead of precomputing, I'd much rather prefer search out a similar, less expensive functional approximation.  Let me look into this

Yes I'd like to target this for default.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -713,7 +709,7 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, diagnost
             ! add convective mixing to vertical viscosity/diffusivity
             ! if using KPP, then do not apply convective mixing within the ocean boundary layer
             if(config_use_cvmix_kpp) then
-               do k = int(indexBoundaryLayerDepth(iCell)) + 1, maxLevelCell(iCell)
+               do k = int(ceiling(indexBoundaryLayerDepth(iCell))) + 1, maxLevelCell(iCell)
```
mattdturner, line 712: `ceiling` returns an integer by default ([GCC documentation](https://gcc.gnu.org/onlinedocs/gfortran/CEILING.html), [Intel documentation](https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-developer-guide-and-reference/top/language-reference/a-to-z-reference/c-to-d/ceiling.html)), so `int(ceiling(` is redundant.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
philipwjones: @vanroekel I think we did a similar substitution for atan in the original KPP. Wish I could remember what we used...

vanroekel: @philipwjones exactly what I was thinking.  I found a possible approximation from POP, checking how close it is now.  Will push up a change soon.

mattdturner: If you need another option...

From [“Efficient approximations for the arctangent function”, Rajan, S. Sichun Wang Inkol, R. Joyal, A., May 2006](http://www-labs.iro.umontreal.ca/~mignotte/IFT2425/Documents/EfficientApproximationArctgFunction.pdf), `atan` can be approximated via the following C code:
```
#define M_PI_4 (3.1415926535897932384626433832795/4.0)

double FastArcTan(double x) {
  return M_PI_4*x - x*(fabs(x) - 1)*(0.2447 + 0.0663*fabs(x));
}
```

It is valid for `x` values of -1 to 1, a bit over 3x faster than `atan`,  and has a max error of approx 0.0015 radians.

vanroekel: @mattdturner this is what I was looking at too, but our range goes from -10, 10, and the error gets very large as soon as you go outside -1 to 1 with this formula

vanroekel: @mattdturner and @philipwjones what about something like this?  In fortran of course, this is my python testing
```
for i in range(len(z)):
    x = (abs(z[i])-2500.)/222.
    if abs(x) < 1:
        zmix2[i] = 8e-5 + 1.05E-4/np.pi*(np.pi*x-x*(abs(x)-1)*(0.2447+0.0663*abs(x)))
    else:
        zmix2[i] = 8e-5 + 1.05E-4/np.pi*(np.pi/2*np.sign(x) - 4*x/(4*x**2+1))
```
this gives a nice approximation through the full column

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
mattdturner: I replaced these string comparisons w/ named integer parameters

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
vanroekel: thanks to @mattdturner I have a viable alternative
```
zmix4 = 8e-5 + 1.05E-4/2*(np.sign(x)*(0.596227*abs(x)+x**2)/(1+2*0.596227*abs(x)+x**2))
```
This works well through the entire column, without the nasty embedded if.  I'll push this up soon.

philipwjones: Great, glad you guys found one. Otherwise I was about to suggest a spline fit where we precompute the spline coefficients and then it becomes just a spline (polynomial) evaluation.  This is better.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
vanroekel: Thanks, fixed.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
vanroekel: okay this is changed now.

mark-petersen (approved):
Ran test on last commit, with gnu 6 debug and intel 19 debug, passes nightly regression suite. Also compared with 2 threads versus no openmp, and get bfb match.

PR #615, opened 2020-06-29
Adds free surface gradient forcing for idealized 1D turbulence

pwolfram:
Adds free surface gradient forcing for
 * config_zonal_ssh_grad (x-component for planar)
 * config_meridional_ssh_grad (y-component for planar)

Case is applied in periodic flow in
`ocean/periodic_planar/2.5km/default_gotm` case, which can be
used to validate vertical momentum balances with constant forcing
due to double periodicity, which is useful to indicate potential
instabilities with constant forcing due to unstructured mesh.

Note, for full case comparison against literature (e.g., section 5.1 of https://www.sciencedirect.com/science/article/pii/S1463500319302082), needs to have Generaliized
Ocean Turbulence Model to be incorporated.

PR #704, opened 2020-09-24
GOTM in MPAS-Ocean

qingli411:
This PR adds the General Ocean Turbulence Model ([GOTM](https://gotm.net/portfolio/)) in MPAS-Ocean as an additional option for the vertical turbulence closure, especially for coastal simulations in which the present implementation of KPP in CVMix does not work well. GOTM is a turbulence closure library that includes a set of two equation models such as k-epsilon and the generic length scale model (GLS; [Umlauf and Burchard, 2003](https://doi.org/10.1357/002224003322005087)).


qingli411:
It has been tested in a single column test case following Section 5.1 of [Kärnä, 2020](https://doi.org/10.1016/j.ocemod.2020.101619), which is an updated version of the surface gradient forcing test case in https://github.com/MPAS-Dev/MPAS-Model/pull/615.  The test case is in compass test number 138. 
image

The simulated velocity and viscosity profiles are compared with the analytical solution of Equations (67) and (69), and a simulation using constant viscosity of 0.01 m^2/s in CVMix.

- The velocity profile
image

- The viscosity profile
image

The results depends on the bottom drag coefficient, here set to 0.0173, corresponding to the value inferred from the analytical solution with a layer thickness of 0.06 m.

qingli411:
Currently the GOTM module reads in its own namelist file ('gotmturb.nml', included in the compass test) at the beginning of the simulation. The path of the GOTM namelist file can be set in the MPAS-Ocean namelist. This is coded in the GOTM library so we will need to customize GOTM or have a way to generate the GOTM namelist if we want to control the behavior of GOTM from the MPAS-Ocean namelist. 

GOTM requires input of bottom drag coefficient (to compute the bottom friction velocity) and the bottom roughness length, both set to constant in the namelist right now. But later may use the values from the bottom drag parameters in other part of the MPAS-Ocean code once they are ready.

qingli411:
@sbrus89 Could you please take a look at this when you have time? Looks like I don't have the access to assign a reviewer...

qingli411:
The implementation of GOTM in MPAS-Ocean has also been tested in a JRA55do forced G-case before rebasing to the `ocean/develop` branch. I only ran it for two years to make sure it does not break anything in a global case. It gives shallower mixed layer depth than CVMix in most of the regions and is quite sensitive to the time step.

Here are some MPAS-Analysis results:
- [GOTM with 1 min time step versus CVMix](https://portal.nersc.gov/project/e3sm/qingli/mpas_analysis/20200605.GOTM1mvsCVMix.GMPAS-JRA.TL319_oEC60to30v3.anvil/0002-0002/)
- [GOTM with 10 min time step versus CVMix](https://portal.nersc.gov/project/e3sm/qingli/mpas_analysis/20200605.GOTM10mvsCVMix.GMPAS-JRA.TL319_oEC60to30v3.anvil/0002-0002/)
- [GOTM with 30 min time step versus CVMix](https://portal.nersc.gov/project/e3sm/qingli/mpas_analysis/20200605.GOTM30mvsCVMix.GMPAS-JRA.TL319_oEC60to30v3.anvil/0002-0002/)

mark-petersen:
@qingli411 please post a `gotmturb.nml` file so we can test this PR. Also, tell us what flags need to be changed to exercise all parts of the code you wrote. Is it simply `config_use_gotm = .true.`, or do we need to test with various other settings as well?

qingli411:
@mark-petersen The `gotmturb.nml` file is in the `default_gotm` test case (`testing_and_setup/compass/ocean/periodic_planar/2.5km/default_gotm/gotmturb.nml`). The `default_gotm` test case is the single column test case with GOTM I mentioned earlier. To activate GOTM, use `config_use_gotm = .true.`. Since GOTM replaces CVMix here, we also need to turn off CVMIx by `config_use_cvmix = .false.`.

mark-petersen:
@qingli411 I rebased on the head of `ocean/develop` and then tested with gnu 5 debug and intel 19 debug on the nightly regression suite, with cvmix off and gotm on, and `gotmturb.nml` in every run directory. All the tests run, which is a good sign, as debug catches any out of bounds array index.

<details>
<summary> This is the output from the nightly regression suite: </summary> <p>

```
 ** Running case Baroclinic Channel 10km - Default Test
      PASS
      PASS
      PASS
   ** FAIL (See case_outputs/Global_Ocean_240km_-_Restart_Test for more information)
 ** Running case Global Ocean 240km - Partition Test
      PASS
      PASS
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - BGC Ecosys Test
      PASS
      PASS
      PASS
      PASS
      PASS
   ** FAIL (See case_outputs/Baroclinic_Channel_10km_-_Restart_Test for more information)
   ** FAIL (See case_outputs/sub-ice-shelf_2D_-_restart_test for more information)
```
</details>

for both gnu and intel, the restart tests all fail. This means one of these:
1. upon restart, the proper fields are not computed on init, or are in the wrong order.
2. there is a field in local memory that is needed from step to step, but not known at the beginning of a restart. 

either way, the 'state' of the system is not identical just after beginning a restart, as it would be if the model proceeded without restarting. Try to fix (1) first. The solution to (2) is to add variables to the restart file. But we only do that if they are absolutely needed. If they can be computed on init instead, that is better.

You can copy my restart test here, for your convenience:
/lustre/scratch3/turquoise/mpeterse/runs/nightly/pr704_qingli_gotm/ocean/global_ocean/QU240/restart_test
or the whole regression suite:
/lustre/scratch3/turquoise/mpeterse/runs/nightly/pr704_qingli_gotm

you can tar up that whole directory, copy it to your own space, and run `nightly_ocean_test_suite.py`. You will need to add the link to your own executable in the forward subdirectories.

qingli411:
@mark-petersen OK. I think it's probably due to the second reason. GOTM solves prognostic equations for TKE and a length scale and requires the state of some variables from the previous time step. I added a list of variables that GOTM needs in the `statePool` for that reason (See [this line](https://github.com/MPAS-Dev/MPAS-Model/pull/704/files#diff-04eb0970c6f8ad0aa0624d056d718f00L1951) in `src/core_ocean/Registry.xml`). These arrays are used in `src/core_ocean/shared/mpas_ocn_vmix_gotm.F` to assign values from the previous time step to GOTM. So maybe we should save these state variables in the restart file? If so is there a way to save these variables to the restart file only when GOTM is used?

Related to this I have a question on the way to initialize these arrays in MPAS-O. [Here](https://github.com/qingli411/MPAS-Model/blob/a8d321754019bcddca7edf18fe0388fd2bbe5625/src/core_ocean/shared/mpas_ocn_vmix_gotm.F#L255) I initialize the arrays only once using a saved variable `first`. Now I realized that I should probably do this in the init step and check `config_do_restart` so that these arrays do not get initialized (to zero) again when doing restart. What do you think?

src/core_ocean/Registry.xml
```diff
@@ -1949,6 +1979,35 @@
 			description="The effective ocean density within ice shelves based on Archimedes' principle."
 			packages="landIceCouplingPKG"
 		/>
+        <!-- FIELDS FOR GOTM -->
+		<var name="gotmVertViscTopOfCell" type="real" dimensions="nVertLevelsP1 nCells Time" units="m^2 s^{-1}"
+			 description="GOTM: vertical viscosity defined at the cell center (horizontally) and top (vertically)"
+			 packages="forwardMode;analysisMode"
```
mark-petersen, line 1985: Here you are adding 3D fields that are allocated whether gotm is in use or not. To fix that, add a new package in Registry.xml called `gotmPKG` and put all of your new variables in that package (not `forwardMode;analysisMode`).  You will need to add code to turn on the package if the first gotm flag is true. Follow the example here:
```
src/core_ocean/driver/mpas_ocn_core_interface.F
238       !
239       ! test for variable bottom drag of momentum, variableBottomDragPKG
240       call mpas_pool_get_package(packagePool, 'variableBottomDragPKGActive', variableBottomDragPKGActive)
241       call mpas_pool_get_config(configPool, 'config_use_variable_drag', config_use_variable_drag)
242       if ( config_use_variable_drag ) then
243         variableBottomDragPKGActive = .true.
244       end if
```
For bfb restarts, add the prognostic variables that are needed for restart to the restart stream in the Registry.xml file here:
```
1550         <stream name="restart"
```
Also, think carefully about which variables go into the state var_struct. Those should really only be the variables we have a prognostic equation for (i.e. use a time-stepping method to advance). All the other variables that are secondary should go into the diagnostic var_struct. 

The difference is important because the state variables have 2 time levels for the timestepping, and the diagnostics only have one timelevel. From Registry:
```
1883     <var_struct name="state" time_levs="2">
...
2220     <var_struct name="diagnostics" time_levs="1">
```
so think carefully about what what variables belong in state. Those get double the memory, and that is needed for the time-stepping routines. The diagnostics are computed from the state variables, but only need a single instance in memory, for one time slice.

mark-petersen:
> Related to this I have a question on the way to initialize these arrays in MPAS-O. Here I initialize the arrays only once using a saved variable first. Now I realized that I should probably do this in the init step and check config_do_restart so that these arrays do not get initialized (to zero) again when doing restart. What do you think?

Correct, you do not want to use a `first` logical that you flip. All modules have at init routine at the very bottom, and all the init routines are called on start-up. That is where you conduct any process that you do on start-up or restart, but not on every time step. This includes initialization of variables.  

You can use the flag `config_do_restart` which is false at the very beginning of a simulation, and true for restart runs. So for prognostic variables you may initialize if `config_do_restart=.false.` and then depend on the restart read of a prognostic variable after that.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
```diff
@@ -0,0 +1,520 @@
...
+      do iCell = 1, nCells
+         vertViscTopOfCell(:, iCell) = 0.0_RKIND
+         vertDiffTopOfCell(:, iCell) = 0.0_RKIND
+      end do
+      !$omp end do
+
+      ! initialize GOTM arrays at the first step
+      ! MPAS-O (1:nlev+1) <-> GOTM (nlev:0)
+      if (first) then
+         !$omp do schedule(runtime)
```
mark-petersen, line 256: @qingli411 all our omp calls have changed format in https://github.com/MPAS-Dev/MPAS-Model/pull/513. Please search for all !$omp lines in your PR and change to this:
```
      !$omp parallel
      !$omp do schedule(runtime) &
      !$omp private(k, ki)
      do iCell = 1, nCells
...
```
for all cell, edge, and vertex loops. The private variables are any index or scalar value used within the loop.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
```diff
@@ -0,0 +1,520 @@
...
+               gotmEpsbTopOfCellCur(k, iCell) = gotm_epsb(ki)
+               gotmVertViscTopOfCellNew(k, iCell) = gotm_vertVisc(ki)
+               gotmVertDiffTopOfCellNew(k, iCell) = gotm_vertDiffT(ki)
+               gotmDissTopOfCellNew(k, iCell) = gotm_eps(ki)
+               gotmLengthTopOfCellNew(k, iCell) = gotm_L(ki)
+               gotmKbTopOfCellNew(k, iCell) = gotm_kb(ki)
+               gotmEpsbTopOfCellNew(k, iCell) = gotm_epsb(ki)
+            end do
+         end do
+         !$omp end do
```
mark-petersen, line 275: At the end of a cell, edge, or vertex loop, change all omp lines to:
```
      !$omp end do
      !$omp end parallel
```

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
mark-petersen: Note that the first index (here iCell) is not in the private line on purpose. That is because openMP knows to run thread-parallel on that first loop.

qingli411:
@mark-petersen I have made the requested changes and run `nightly_ocean_test_suite.py` with gnu debug on badger. Looks like the tests all pass. For some reason the script stopped at the end when reporting the runtime (see the output below)...

<details>
<summary>
Output from the nightly regression suite
</summary>
<p>

```
 ** Running case Baroclinic Channel 10km - Default Test
      PASS
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - Partition Test
      PASS
      PASS
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - BGC Ecosys Test
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
      PASS
TEST RUNTIMES:
  File "./nightly_ocean_test_suite.py", line 240, in <module>
    runtime = np.ceil(float(subprocess.check_output(
ValueError: could not convert string to float: 'file'
```

</p>
</details>

src/core_ocean/Registry.xml
qingli411: This is done in 2f63289 and 6693f2b. I think all the GOTM variables here need to be state variables rather than diagnostics variables as values from the previous time step are required to step GOTM forward.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
qingli411: Initialization of the GOTM state variables are now in the init subroutine (see 7b00c93).

I also changed the format for the OpenMP directives. But since my knowledge of OpenMP is very limited I'm not sure whether it work with OpenMP... Here GOTM has its own initialization step (called in the subroutine ` ocn_vmix_gotm_init`), which allocates memory for a list of 1D arrays (vertical profiles of TKE, length scales, viscosity etc). Then I loop over all the MPAS-O cells to call GOTM to update these variables for each cell (the GOTM state variables allocated in MPAS-O).

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
qingli411: Done

mark-petersen:
@qingli411 please compile with openmp on, i.e.
```
make $COMPILER CORE=ocean USE_PIO2=true OPENMP=true DEBUG=true GEN_F90=true
```
and you will see some mismatching `!$omp` statements.

qingli411:
@mark-petersen That should be fixed now.

mark-petersen:
@qingli411 try to run the nightly regression suite with openmp on and debug. I get a floating invalid error:
```
206 ocean_model        000000000168B22C  ocn_vmix_gotm_mp_         313  mpas_ocn_vmix_gotm.f90
```
note the line number is for *.f90, after the C-pre processor (use `GEN_F90=true` on compile to see it).

<details>
<summary>All this is printing to the screen.  </summary> <p>

```
  init_turbulence
    init_turbulence
        reading turbulence namelists..
        done.
        allocation memory..
        allocation memory..
        allocation memory..
        done.
        reading turbulence namelists..
        done.
        allocation memory..
        allocation memory..
        allocation memory..
        done.

        --------------------------------------------------------
        You are using the generic two-equation model
        with the following properties:

            exponent of k in psi-equation,    m  =   1.50000000000000
            exponent of l in psi-equation,    n  =  -1.00000000000000
            exponent of cm0 in psi-equation,  p  =   3.00000000000000
            cpsi1                                =   1.44000000000000
            cpsi2                                =   1.92000000000000
            cpsi3minus                           = -0.630000000000000
            cpsi3plus                            =   1.00000000000000
            sig_k                                =   1.00000000000000
            sig_psi                              =   1.30000000000000

            Value of the stability function
            in the log-law,                   cm0 =  0.526464710835324
            in shear-free turbulence,        cmsf =  0.731006013490010

            von Karman constant,           kappa =  0.415873799673737
            homogeneous decay rate,            d =  -1.08695652173913
            spatial decay rate (no shear), alpha =  -4.97449646698836
            length-scale slope (no shear),     L =  7.333286748097885E-002
            steady-state Richardson-number, Ri_st=  0.248302382206140
        --------------------------------------------------------

    init_tridiagonal
```
</details>

You should remove all printing to the screen (print statements).  If you want to keep verbose debug printing, use `call mpas_log_write(` so it goes into the output files. In addition, you should have a local CPP flag for verbose output, and have it default off for the PR. For example, see
```
src/framework/
mpas_forcing.F:#define FORCING_DEBUG_WRITE(M) !call mpas_log_write( M )
```
where you can remove the `!` when you want verbose output.

mark-petersen:
You may need to set 
```
export OMP_NUM_THREADS=1
```
in your window before running to test the omp lines properly.

qingli411:
@mark-petersen OK I will do that. The printing to the screen is in the GOTM library... We will need to modify the GOTM repository to write that information to the log file.

qingli411:
@mark-petersen With the above fix (7c8fc6e) the nightly regression suite with openmp on and debug fails on the Thread and Decomp tests with
```
SIGFPE: Floating-point exception - erroneous arithmetic operation.
```

<details>
<summary>
See the output of the nightly regression suite here.
</summary>

```
 ** Running case Baroclinic Channel 10km - Default Test
      PASS
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - Partition Test
      PASS
      PASS
      PASS
      PASS
 ** Running case Global Ocean 240km - BGC Ecosys Test
      PASS
      PASS
      PASS
   ** FAIL (See case_outputs/Baroclinic_Channel_10km_-_Thread_Test for more information)
   ** FAIL (See case_outputs/Baroclinic_Channel_10km_-_Decomp_Test for more information)
      PASS
      PASS
      PASS
      PASS
TEST RUNTIMES:
  File "./nightly_ocean_test_suite.py", line 240, in <module>
    runtime = np.ceil(float(subprocess.check_output(
ValueError: could not convert string to float: 'file'
```

</details>

I guess some variables were not correctly initialized in the Thread tests. I have set `export OMP_NUM_THREADS=1` before running the test suite. Any suggestions?

mark-petersen:
@qingli411 thanks for your work on this. We don't want excessive output written, even to log files. Lines like `allocation memory.. done.` should really just be on when a verbose setting is on.

qingli411:
@mark-petersen  I understand that. But I don't think we have an easy way to control how GOTM writes out these messages as it is an external library. Same is for the way it reads in its own namelist. We will need to have a customized version of GOTM in MPAS-O if we want to have better control.

mark-petersen:
Yes, I understand now. It is best to not customize GOTM, just use it straight. Are there actual Fortran print statements in GOTM? They don't have an internal verbose flag to shut them off? If every processor uses print, and we are using 1k to 10k processors, then GOTM would be completely unusable in E3SM. If so, please try to get GOTM to put the print statements within a CPP verbose flag. Even if it takes several months to go through, it's worth making an issue to them now. Also, better to work with them than split off your own version of the code. It's a really reasonable request.

qingli411:
@mark-petersen I'm still in contact with the GOTM developers on possible better solutions. But, with their help, here (a603b7e) is a workaround to suppress the GOTM messages which only involves changing one line of code in GOTM. So I added the change in the Makefile. What do you think?

mark-petersen:
@qingli411 thanks for your work on this, and sorry for the delay. Now that V2 is done, we are working through the backlog of PRs. We can work on this the next week, and merge after that. I'm rebasing and testing now. I'll also move the compass cases to the new compass repo instead. 

For your part, please think about whether any additions or changes are needed since last September. Please confirm that this would not change simulations with the default setting 
`config_use_gotm = .false.`.

@vanroekel and @sbrus89 FYI on this, if you have any comments while we review this next week.

qingli411:
@mark-petersen This sounds good. Thanks!

Yes, with `config_use_gotm = .false.` the GOTM subroutines will be skipped. It shouldn't change the simulations.

mark-petersen:
@qingli411 I rebased this onto the current head of `ocean/develop`. I squashed all your commits into one so the rebase was easier. You can still see the granular commits above. There were significant updates performance and modules.  Right now, I get a compile error with gnu on grizzly:
```
mpas_ocn_vmix_gotm.f90:41:7:

    use turbulence, only: init_turbulence, do_turbulence, clean_turbulence
       1
Fatal Error: Can't open module file ‘turbulence.mod’ for reading at (1): No such file or directory
```
I checked that gotm is compiled, and see
```
ls src/core_ocean/gotm/build/*a
  src/core_ocean/gotm/build/liboutput_manager.a
  src/core_ocean/gotm/build/libturbulence.a
  src/core_ocean/gotm/build/libutil.a
```
So it did compile. It looks like the Makefile lines are all included in my revised code changes.

@qingli411 since you added gotm the first time, can you do a test compile and try to find the problem?

src/core_ocean/Makefile
```diff
@@ -6,12 +6,12 @@ $(info "EXCLUDE_INIT_MODE found")
 endif
 
 OCEAN_SHARED_INCLUDES = -I$(PWD)/../framework -I$(PWD)/../external/esmf_time_f90 -I$(PWD)/../operators
-OCEAN_SHARED_INCLUDES += -I$(PWD)/BGC -I$(PWD)/shared -I$(PWD)/analysis_members -I$(PWD)/cvmix/src/shared -I$(PWD)/mode_forward -I$(PWD)/mode_analysis
+OCEAN_SHARED_INCLUDES += -I$(PWD)/BGC -I$(PWD)/shared -I$(PWD)/analysis_members -I$(PWD)/cvmix/src/shared -I$(PWD)/gotm/build -I$(PWD)/mode_forward -I$(PWD)/mode_analysis
```
mark-petersen, line 9: I can see that `gotm/build` is in the include here, so it should find it...

mark-petersen: Never mind. I found it. I didn't get
`-I$(PWD)/gotm/build/modules`

qingli411:
@mark-petersen Glad that you figured it out. Thanks! So it is working now, right?

mark-petersen:
Compiles, passes nightly regression suite with gnu 5.3 on badger and `config_use_gotm = .false.`. When I switch to true, with gotmturb.nml copied from 
```
./testing_and_setup/compass/ocean/periodic_planar/2.5km/default_gotm/gotmturb.nml
```
I still see an error. Working on it...

mark-petersen:
Runs now. Using the QU240 performance test with gnu 5.3 debug and gotm on

<details>
<summary> output with four cores is: </summary> <p>

```
    init_turbulence
    init_turbulence
    init_turbulence
    init_turbulence
        reading turbulence namelists..
        reading turbulence namelists..
        reading turbulence namelists..
        done.
        allocation memory..
        allocation memory..
        allocation memory..
        done.
        reading turbulence namelists..
        done.
        allocation memory..
        allocation memory..
        allocation memory..
        done.
        done.
        allocation memory..
        allocation memory..
        allocation memory..
        done.
        done.
        allocation memory..
        allocation memory..
        allocation memory..
        done.

        --------------------------------------------------------
        You are using the generic two-equation model
        with the following properties:

            exponent of k in psi-equation,    m  =   1.5000000000000000
            exponent of l in psi-equation,    n  =  -1.0000000000000000
            exponent of cm0 in psi-equation,  p  =   3.0000000000000000
            cpsi1                                =   1.4399999999999999
            cpsi2                                =   1.9199999999999999
            cpsi3minus                           = -0.63000000000000000
            cpsi3plus                            =   1.0000000000000000
            sig_k                                =   1.0000000000000000
            sig_psi                              =   1.3000000000000000

            Value of the stability function
            in the log-law,                   cm0 =  0.52646471083532442
            in shear-free turbulence,        cmsf =  0.73100601349000971

            von Karman constant,           kappa =  0.41587379967373655
            homogeneous decay rate,            d =  -1.0869565217391306
            spatial decay rate (no shear), alpha =  -4.9744964669883514

        --------------------------------------------------------
        You are using the generic two-equation model
        with the following properties:

            exponent of k in psi-equation,    m  =   1.5000000000000000
            exponent of l in psi-equation,    n  =  -1.0000000000000000
            exponent of cm0 in psi-equation,  p  =   3.0000000000000000
            cpsi1                                =   1.4399999999999999
            cpsi2                                =   1.9199999999999999
            cpsi3minus                           = -0.63000000000000000
            cpsi3plus                            =   1.0000000000000000
            sig_k                                =   1.0000000000000000
            sig_psi                              =   1.3000000000000000

            Value of the stability function
            in the log-law,                   cm0 =  0.52646471083532442
            in shear-free turbulence,        cmsf =  0.73100601349000971

            von Karman constant,           kappa =  0.41587379967373655
            homogeneous decay rate,            d =  -1.0869565217391306
            spatial decay rate (no shear), alpha =  -4.9744964669883514
            length-scale slope (no shear),     L =   7.3332867480978059E-002
            steady-state Richardson-number, Ri_st=  0.24830238220613976
        --------------------------------------------------------

    init_tridiagonal
            length-scale slope (no shear),     L =   7.3332867480978059E-002
            steady-state Richardson-number, Ri_st=  0.24830238220613976
        --------------------------------------------------------

    init_tridiagonal

        --------------------------------------------------------
        You are using the generic two-equation model
        with the following properties:

            exponent of k in psi-equation,    m  =   1.5000000000000000
            exponent of l in psi-equation,    n  =  -1.0000000000000000
            exponent of cm0 in psi-equation,  p  =   3.0000000000000000
            cpsi1                                =   1.4399999999999999
            cpsi2                                =   1.9199999999999999
            cpsi3minus                           = -0.63000000000000000
            cpsi3plus                            =   1.0000000000000000
            sig_k                                =   1.0000000000000000
            sig_psi                              =   1.3000000000000000

            Value of the stability function
            in the log-law,                   cm0 =  0.52646471083532442
            in shear-free turbulence,        cmsf =  0.73100601349000971

            von Karman constant,           kappa =  0.41587379967373655
            homogeneous decay rate,            d =  -1.0869565217391306
            spatial decay rate (no shear), alpha =  -4.9744964669883514
            length-scale slope (no shear),     L =   7.3332867480978059E-002
            steady-state Richardson-number, Ri_st=  0.24830238220613976
        --------------------------------------------------------

    init_tridiagonal

        --------------------------------------------------------
        You are using the generic two-equation model
        with the following properties:

            exponent of k in psi-equation,    m  =   1.5000000000000000
            exponent of l in psi-equation,    n  =  -1.0000000000000000
            exponent of cm0 in psi-equation,  p  =   3.0000000000000000
            cpsi1                                =   1.4399999999999999
            cpsi2                                =   1.9199999999999999
            cpsi3minus                           = -0.63000000000000000
            cpsi3plus                            =   1.0000000000000000
            sig_k                                =   1.0000000000000000
            sig_psi                              =   1.3000000000000000

            Value of the stability function
            in the log-law,                   cm0 =  0.52646471083532442
            in shear-free turbulence,        cmsf =  0.73100601349000971

            von Karman constant,           kappa =  0.41587379967373655
            homogeneous decay rate,            d =  -1.0869565217391306
            spatial decay rate (no shear), alpha =  -4.9744964669883514
            length-scale slope (no shear),     L =   7.3332867480978059E-002
            steady-state Richardson-number, Ri_st=  0.24830238220613976
        --------------------------------------------------------

    init_tridiagonal
 pionfput_mod.F90         128           1           1          64           1 0001-01-01_00:00:00
 pionfput_mod.F90         128           1           1          64           1 0001-01-01_06:00:00
    clean_turbulence
        de-allocating turbulence memory ...
        done.
    clean_turbulence
        de-allocating turbulence memory ...
        done.
    clean_turbulence
        de-allocating turbulence memory ...
        done.
    clean_turbulence
        de-allocating turbulence memory ...
        done.


     *****************************
     ** Finished model run step **
     *****************************

```
</details>

so everything appears to run successfully. @qingli411 please test in with your standard test to verify it is the same as before my re-base.

We moved compass to it's own repo here: https://github.com/MPAS-Dev/compass. Sorry for the inconvenience. I need to move your compass files in this PR and make a new compass PR, and you can use that for testing. Will need to do that later today...

src/core_ocean/shared/mpas_ocn_vel_pressure_grad.F
```diff
@@ -537,6 +540,30 @@ subroutine ocn_vel_pressure_grad_tend(ssh, pressure, &
...
+
+#ifdef MPAS_OPENACC
+         !$acc parallel loop &
+         !$acc    present(angleEdge,maxLevelEdgeTop, tend, edgeMask)
+         !$acc    private(pGrad)
+#else
+         !$omp parallel
+         !$omp do schedule(runtime) &
+         !$omp    private(pGrad)
+#endif
```
mark-petersen, line 553: @mattdturner I followed the examples above here. Please check.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
```diff
@@ -0,0 +1,531 @@
...
+   use mpas_constants
+   use mpas_log
+   use mpas_timekeeping
+
+   use ocn_constants
+   use ocn_config
+   use ocn_diagnostics_variables
+
+   ! TODO: read in GOTM namelist for testing <20200512, Qing Li> !
+   use mpas_io_units
```
mark-petersen, line 28: @qingli411 is there something to do here, or should this comment be removed?

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
```diff
@@ -0,0 +1,531 @@
...
+!> \brief   Computes mixing coefficients using GOTM
+!> \author  Qing Li
+!> \date    11 May 2020
+!> \details
+!>  This routine computes the vertical mixing coefficients for momentum
+!>  and tracers by calling GOTM routines.
+!
+!-----------------------------------------------------------------------
+
+   subroutine ocn_vmix_coefs_gotm_build(meshPool, statePool, forcingPool, err, timeLevelIn)!{{{
```
mark-petersen, line 86: @mattdturner, I rebased this from last September, when we passed the meshPool in. We should add
```
use ocn_mesh
```
at the top and remove all meshPool variable declarations and lines with
```
mpas_pool_get_dimension(meshPool
```
correct? @qingli411 let's wait on that step until after you have successfully verified your test case with the current code.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
```diff
@@ -0,0 +1,531 @@
...
+      !
+      if (.not. gotmOn) return
+
+      !
+      ! When GOTM is turned on, all other vertical mixing schemes should be off
+      ! Test to make sure this is the case.
+      !
+      ! test here, err=1 if a problem
+
+      ! TODO: use MPAS-O dt for testing. Sub-stepping is possible, though it may not help too much without updating the shear and stratification in the substeps <20200604, Qing Li> !
```
mark-petersen, line 410: @qingli411 please search for all lines with TODO. We want to either take care of it, or add it as a comment that says `comment, possible future work`. Committing TODO lines into the code base makes it look like we forgot something.

src/core_ocean/shared/mpas_ocn_vel_pressure_grad.F
mattdturner: This should be the following
```
#ifdef MPAS_OPENACC
         !$acc parallel loop &
         !$acc    present(angleEdge,maxLevelEdgeTop, tend, edgeMask) &
         !$acc    private(pGrad, k)
#else
         !$omp parallel
         !$omp do schedule(runtime) &
         !$omp    private(pGrad, k)
#endif
```

I can't **bold** the changes in a codeblock, but the `acc    present` line needs an `&` at the end, and both the OpenMP and OpenACC `private` clauses should have `k` added.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
mattdturner: Correct, you would need to add `use ocn_mesh`, remove all meshPool variable declarations (or you'll get unassociated pointer errors), and all calls to `mpas_pool_get_dimension(meshPool)`.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
qingli411: @mark-petersen I think this comment can be removed. I will check and clean up all the lines with TODO.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
qingli411: @mark-petersen OK. I will do that.

qingli411:
> Runs now. Using the QU240 performance test with gnu 5.3 debug and gotm on
> 
>  output with four cores is:
> so everything appears to run successfully. @qingli411 please test in with your standard test to verify it is the same as before my re-base.
> 
> We moved compass to it's own repo here: https://github.com/MPAS-Dev/compass. Sorry for the inconvenience. I need to move your compass files in this PR and make a new compass PR, and you can use that for testing. Will need to do that later today...

@mark-petersen Here (a603b7e) is a way to suppress all the GOTM messages, which doesn't seem to be included here.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
philipwjones: And also remember the changes to account for the changes in nCells, nCellsArray, nCellsSolve (to nCellsAll, nCellsHalo(n-1), nCellsOwned) and equivalent changes to the nEdges, nVertex analogues.

src/core_ocean/shared/mpas_ocn_vel_pressure_grad.F
mark-petersen: done, thanks.

mark-petersen:
Currently failing `Global Ocean 240km - Thread Test`. I need to update openMP lines in `mpas_ocn_vmix_gotm.F`.

mattdturner:
@mark-petersen I just tested the HEAD of the branch, and it passed the `Global Ocean 240km - Thread Test` for Intel + Debug.  What compiler were you using when you got the failure?

mark-petersen:
@mattdturner when I said thread test failed on head (182b2127a) I didn't look close enough. It's an actual error, not a mismatch. Occurs with grizzly gnu 5.3:
```
/lustre/scratch3/turquoise/mpeterse/runs/nightly/210301_pr704_qingli_gotm/ocean/global_ocean/QU240/thread_test
.12) gr0774:thread_test$ ./run.py
 * Running 1thread_run
  - Complete
 * Running 2thread_run

Backtrace for this error:
#0  0x2B61BA567C57
#1  0x2B61BA566E50
#2  0x2B61BB4383FF
#3  0xB80F95 in __mtridiagonal_MOD_tridiagonal
#4  0xB82015 in diff_face_
#5  0xB805B9 in tkeeq_
#6  0xB731B0 in __turbulence_MOD_do_turbulence
#7  0x99B1F2 in __ocn_vmix_gotm_MOD_ocn_vmix_coefs_gotm_build._omp_fn.2 at mpas_ocn_vmix_gotm.f90:330 (discriminator 12)
```
but I didn't spend any time fixing the openMP lines in `mpas_ocn_vmix_gotm.F`. If you have a fixed up version, you can push your commits, I'll try again too.

qingli411:
@mark-petersen @mattdturner Just pushed two commits to clean up the comments in `mpas_ocn_vmix_gotm.F` and to fix a bug in `mpas_ocn_vel_pressure_grad.F` that the pressure gradient type `constant_forced` was not initialized.

While I was testing the code in the single column case with constant pressure gradient and explicit bottom drag as described above, I got the following error:
```
At line 144 of file mpas_ocn_vel_forcing_explicit_bottom_drag.F
Fortran runtime error: Index '0' of dimension 1 of array 'tend' below lower bound of 1
``` 
Any suggestions?

philipwjones:
@qingli411  Maybe maxLevelEdgeTop is zero? If so, you can put an if (k > 0) around that statement. I'll leave it to others to say whether that particular event is problematic or not.

mattdturner:
@mark-petersen @qingli411 Regarding the thread test failure...

I think the problem is that the subroutine `do_turbulence` within GOTM is not thread-safe.  If you look in `src/core_ocean/gotm/src/turbulence/turbulence.F90` at line 2131 (within `subroutine do_turbulence`), there is a call to the `production` subroutine:
```
      if ( PRESENT(xP) ) then                                                                                
!        with seagrass turbulence                                                                            
         call production(nlev,NN,SS,xP)                                                                      
      else                                                                                                   
!        without                                                                                             
         call production(nlev,NN,SS)                                                                         
      end if
```

Within `src/core_ocean/gotm/src/turbulence/production.F90`, the subroutine sets the values for `P`, `B`, and `Pb`:
```
   if ( PRESENT(xP) ) then                                                                                   
      do i=0,nlev                                                                                            
         P(i)    =  num(i)*( SS(i)+alpha_eff*NN(i) ) + xP(i)                                                 
         B(i)    = -nuh(i)*NN(i)                                                                             
         Pb(i)   = -  B(i)*NN(i)                                                                             
      enddo                                                                                                  
   else                                                                                                      
      do i=0,nlev                                                                                            
         P(i)    =  num(i)*( SS(i)+alpha_eff*NN(i) )                                                         
         B(i)    = -nuh(i)*NN(i)                                                                             
         Pb(i)   = -  B(i)*NN(i)                                                                             
      enddo                                                                                                  
   endif
```

The variables `P`, `B`, and `Pb` are module-level variables in `turbulence.F90`.  When running `do_turbulence` with more than 1 thread, each thread will be writing to those 3 variables at the same time, creating a race condition.  There are likely other instances of this type of race condition in GOTM as well.

So, unless I'm mistaken, I don't think we can call `do_turbulence` within a threaded region.  We either need to do some refactoring so that `do_turbulence` is only called on 1 thread at a time, or omit the OpenMP pragmas around lines 235-240.

Alternatively, we could add an `!$omp critical` region around the `call do_turbulence` (I've confirmed that this allows the thread test to pass).  However, I've generally seen that `!$omp critical` regions negate any performance gain from OpenMP regions in earth system component models.

mark-petersen:
@mattdturner, gotm is an external code, so it's best for us not to change it. We could easily add the `!$omp critical` region. But when you say

> However, I've generally seen that `!$omp critical` regions negate any performance gain from OpenMP regions in earth system component models.

do you mean it negates threading performance gain just of that region, or of the OpenMP performance gain of the whole model? I expect it would be just of that region, and that would be fine.

mattdturner:
@mark-petersen I mean just of that particular region.  I didn't actually test to see how threading the region with `!$omp critical` compares to just removing the `!$omp` directives completely.  I can quickly test the `!$omp critical` implementation and the no-threading implementation to see what the performance difference is.

mark-petersen:
I'm sure it will be just a small percentage of the code, so doesn't matter much. I think using the `!$omp critical` region for gotm is fine.

qingli411:
Thanks, @philipwjones for the suggestion! Putting an `if (k > 0)` around that statement does fix the error. Then I realized that I was using __implicit__ bottom drag in my previous test case instead of __explicit__ bottom drag. Now I can reproduce the figures at the top of this PR with the new code. So I'm leaving `mpas_ocn_vel_forcing_explicit_bottom_drag.F` as it is now here -- I'm not sure if we need `if (k > 0)` for other cases.

For the thread test issue, I'm not very familiar with multi-thread with OpenMP, but here is my understanding when I implement GOTM in MPAS-Ocean. GOTM is a single column turbulence library so everything is done column by column. It allocates its own variables to solve prognostic equations for the TKE and the length scale in a single column and it requires variables at the previous and current time step. The values of these variables are save in MPAS-Ocean as `state` variables for each grid cell but only one set of 1D GOTM variables is allocated for each MPAS-Ocean `block` (done by `init_turbulence`), and I'm looping over the grid cells in each MPAS-Ocean `block` which share one set of 1D GOTM variables. This may prevent parallelization within one `block` of MPAS-Ocean grid cells. Hope this makes sense...

mattdturner:
@mark-petersen I have no real objection to using the `omp critical` region here since it is such a small percentage of the runtime.

However, for reference here are runtime (average of 3 simulations) differences for the `2thread_run` simulation within QU240 `thread_test`.  Timer (gotm cell loop) is specific to the threaded loop:

**GNU Optimized** (gnu v5.3.0)
```
OMP Critical: .2333 seconds
No Openmp: .06605 seconds
```

**Intel Optimized** (intel v19.0.4)
```
OMP Critical: 0.09616 seconds
No Openmp: 0.0856 seconds
```

mark-petersen:
That is a big change for gnu. But if we don't change the gotm code, that is our only option, right?

mattdturner:
We can either use OMP critical, or just remove all OpenMP from that loop.  Either will work just fine.

philipwjones:
@qingli411 Since your case is an existence proof that that condition can occur, probably best to leave the k>0 conditional in.

Regarding the threading, as @mattdturner noted, because you only have one copy of the 1-d work arrays as module variables, they are by default shared by all the threads so all of the threads are trying to share the same column array rather than having their own local copy and are competing with each other to update values in those arrays. If you want thread parallelism across all the columns, these 1-d arrays need to be local and declared private or threadprivate within the parallel regions. Mark and Matt have a short-term fix.

Unfortunately, in the longer term, if we want this (and other single-column models like cvmix) to work on GPUs, we're going to need to push horizontal loops down into these libraries so that there are loops over columns and not operating as single-column models. We won't be able to run this part of the code on GPU until that is done.

mark-petersen:
OK. @mattdturner when you are happy with your performance changes, please push the commit and approve this PR. Then I will retest, and we can merge next week.

qingli411:
@philipwjones OK. That `k>0` conditional is added. Thanks.

mattdturner:
@mark-petersen I'll finish implementing `ocn_mesh` in the GOTM code on Monday, and push / approve.

mattdturner:
I implemented `ocn_mesh` on this branch.  I don't seem to have write permissions to @qingli411 repo, so here is a patch file:

<details><summary> 0001-Use-ocn_mesh-in-GOTM.patch </summary>

```
From 3b648cb3953908b8cbb4547c39f2db65bc1ec464 Mon Sep 17 00:00:00 2001
From: Matt Turner <mturner@lanl.gov>
Date: Mon, 8 Mar 2021 11:09:07 -0700
Subject: [PATCH 1/1] Use ocn_mesh in GOTM

Replace the meshPool interface with the ocn_mesh module within the GOTM
driver file in MPAS-O
---
 src/core_ocean/shared/mpas_ocn_vmix.F      |  2 +-
 src/core_ocean/shared/mpas_ocn_vmix_gotm.F | 55 ++++------------------
 2 files changed, 9 insertions(+), 48 deletions(-)

diff --git a/src/core_ocean/shared/mpas_ocn_vmix.F b/src/core_ocean/shared/mpas_ocn_vmix.F
index 7b999f9d0..1c8514854 100644
--- a/src/core_ocean/shared/mpas_ocn_vmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix.F
@@ -174,7 +174,7 @@ contains
 
       call ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, err1, timeLevel)
       call ocn_vmix_coefs_redi_build(meshPool, statePool, err2, timeLevel)
-      call ocn_vmix_coefs_gotm_build(meshPool, statePool, forcingPool, err3, timeLevel)
+      call ocn_vmix_coefs_gotm_build(statePool, forcingPool, err3, timeLevel)
 
       err = ior(ior(err1, err2), err3)
 
diff --git a/src/core_ocean/shared/mpas_ocn_vmix_gotm.F b/src/core_ocean/shared/mpas_ocn_vmix_gotm.F
index e425671e9..97fa892a3 100644
--- a/src/core_ocean/shared/mpas_ocn_vmix_gotm.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix_gotm.F
@@ -22,6 +22,7 @@ module ocn_vmix_gotm
 
    use ocn_constants
    use ocn_config
+   use ocn_mesh
    use ocn_diagnostics_variables
 
    use mpas_io_units
@@ -82,7 +83,7 @@ contains
 !
 
-   subroutine ocn_vmix_coefs_gotm_build(meshPool, statePool, forcingPool, err, timeLevelIn)!{{{
+   subroutine ocn_vmix_coefs_gotm_build(statePool, forcingPool, err, timeLevelIn)!{{{
 
       !-----------------------------------------------------------------
       !
@@ -90,9 +91,6 @@ contains
       !
       !-----------------------------------------------------------------
 
-      type (mpas_pool_type), intent(in) :: &
-         meshPool          !< Input: mesh information
-
       integer, intent(in), optional ::     &
          timeLevelIn !< Input: time level for state pool
 
@@ -122,17 +120,6 @@ contains
       !
       !-----------------------------------------------------------------
 
-      integer, pointer :: nVertLevels, nVertLevelsP1
-
-      integer, dimension(:), pointer :: &
-        maxLevelCell, nEdgesOnCell, nCellsArray
-
-      integer, dimension(:,:), pointer :: edgesOnCell
-
-      real (kind=RKIND), dimension(:), pointer :: &
-        bottomDepth, &
-        dcEdge, dvEdge, areaCell
-
       real (kind=RKIND), dimension(:,:), pointer :: &
         normalVelocity, layerThickness, &
         gotmVertViscTopOfCellCur, gotmVertDiffTopOfCellCur, &
@@ -175,22 +162,7 @@ contains
       !
       ! set parameters
       !
-      call mpas_pool_get_dimension(meshPool, 'nVertLevels', nVertLevels)
-      call mpas_pool_get_dimension(meshPool, 'nVertLevelsP1', nVertLevelsP1)
-      call mpas_pool_get_dimension(meshPool, 'nCellsArray', nCellsArray)
-      call mpas_pool_get_array(meshPool, 'dcEdge', dcEdge)
-      call mpas_pool_get_array(meshPool, 'dvEdge', dvEdge)
-      call mpas_pool_get_array(meshPool, 'nEdgesOnCell', nEdgesOnCell)
-      call mpas_pool_get_array(meshPool, 'areaCell', areaCell)
       call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocity, timeLevel)
-      call mpas_pool_get_array(meshPool, 'edgesOnCell', edgesOnCell)
-
-      !
-      ! set pointers for fields related to vertical mesh
-      !
-      call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
-      call mpas_pool_get_array(meshPool, 'bottomDepth', bottomDepth)
-
       call mpas_pool_get_array(statePool, 'layerThickness', layerThickness, timeLevel)
 
       !
@@ -212,7 +184,7 @@ contains
       call mpas_pool_get_array(statePool, 'gotmEpsbTopOfCell', gotmEpsbTopOfCellCur, 1)
       call mpas_pool_get_array(statePool, 'gotmEpsbTopOfCell', gotmEpsbTopOfCellNew, 2)
 
-      nCells = nCellsArray( size(nCellsArray) )
+      nCells = nCellsAll
 
       !$omp parallel
       !$omp do schedule(runtime)
@@ -223,7 +195,7 @@ contains
       !$omp end do
       !$omp end parallel
 
-      nCells = nCellsArray( 3 )
+      nCells = nCellsHalo( 2 )
 
       ! allocate squared shear and buoyancy frequencies
       allocate(shearSquaredTop(nVertLevelsP1))
@@ -232,12 +204,8 @@ contains
 
       call mpas_timer_start('gotm cell loop', .false.)
 
-      !$omp parallel
-      !$omp do schedule(runtime) &
-      !$omp private(i, k, ki, shearSquaredTop, NSquaredTop, hLayer, &
-      !$omp         invAreaCell, iEdge, factor, delU2, bottomFrictionVelocity, &
-      !$omp         gotm_vertVisc, gotm_vertDiffT, gotm_vertDiffS, gotm_TKE, gotm_eps, &
-      !$omp         gotm_L, gotm_kb, gotm_epsb)
+      ! MDT, Mar 2021: Removed OpenMP directives because the subroutine
+      !                do_turbulence is not thread safe
       do iCell = 1, nCells
 
          ! initialize arrays
@@ -333,8 +301,6 @@ contains
          end do
 
       end do
-      !$omp end do
-      !$omp end parallel
 
       call mpas_timer_stop('gotm cell loop')
 
@@ -371,11 +337,9 @@ contains
       integer, intent(out) :: err !< Output: error flag
 
       type (block_type), pointer :: block
-      type (mpas_pool_type), pointer :: meshPool
       type (mpas_pool_type), pointer :: statePool
 
-      integer, pointer :: nVertLevels, nVertLevelsP1
-      integer, dimension(:), pointer :: nCellsArray
+      integer, pointer :: nVertLevels
 
       real (kind=RKIND), dimension(:,:), pointer :: &
         gotmVertViscTopOfCellCur, gotmVertDiffTopOfCellCur, &
@@ -437,10 +401,7 @@ contains
       ! initialize GOTM arrays
       block => domain%blocklist
       do while (associated(block))
-         call mpas_pool_get_subpool(block % structs, 'mesh', meshPool)
          call mpas_pool_get_subpool(block % structs, 'state', statePool)  !
-         call mpas_pool_get_dimension(meshPool, 'nVertLevelsP1', nVertLevelsP1)
-         call mpas_pool_get_dimension(meshPool, 'nCellsArray', nCellsArray)
 
          ! set pointers for GOTM viscosity/diffusivity, TKE, rate of dissipation and turbulent length scale
          ! (current and new state)
@@ -452,7 +413,7 @@ contains
          call mpas_pool_get_array(statePool, 'gotmKbTopOfCell', gotmKbTopOfCellCur, 1)
          call mpas_pool_get_array(statePool, 'gotmEpsbTopOfCell', gotmEpsbTopOfCellCur, 1)
 
-         nCells = nCellsArray( size(nCellsArray) )
+         nCells = nCellsAll
          !$omp parallel
          !$omp do schedule(runtime) &
          !$omp private(k, ki)
-- 
2.21.0


```

</details>

This passed the nightly suite on grizzly w/ Intel + Debug (after setting `config_use_cvmix` to false and `config_use_gotm` to true).

mattdturner:
I just pushed the `ocn_mesh` in GOTM commit.  This should be good to go.  I'll start a few tests in E3SM.

mattdturner:
Tests w/ Intel and GNU seem to be good
```
PASS:
SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel
PEM_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_intel
PEM_Ln9.T62_oQU240.GMPAS-IAF.anvil_intel
PET_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_intel
PET_Ln9.T62_oQU240.GMPAS-IAF.anvil_intel
PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel
PEM_Ln9_D.T62_oQU240.GMPAS-IAF.anvil_intel
PET_Ln9_P8192.T62_oRRS18to6v3.GMPAS-IAF.cori-knl_gnu
```

However there is a compilation failure whenever using PGI (at least the versions of PGI I have tested)
```
FAIL:
FAIL SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi MODEL_BUILD 
FAIL PEM_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi MODEL_BUILD 
FAIL PET_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi MODEL_BUILD 
```

The failures are due to PGI complaining about an intrinsic used in gotm
```
PGF90-S-0155-Intrinsic not supported in initialization: spacing (/qfs/people/turn923/E3SM_test/components/mpas-source/src/core_ocean/gotm/src/util/field_manager.F90: 43)               
PGF90-S-0457-Illegal expression in initialization (/qfs/people/turn923/E3SM_test/components/mpas-source/src/core_ocean/gotm/src/util/field_manager.F90: 85)
```

mark-petersen:
Rebased. Testing again in stand-alone.

mattdturner:
> However there is a compilation failure whenever using PGI (at least the versions of PGI I have tested)
> 
> ```
> FAIL:
> FAIL SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi MODEL_BUILD 
> FAIL PEM_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi MODEL_BUILD 
> FAIL PET_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi MODEL_BUILD 
> ```
> 
> The failures are due to PGI complaining about an intrinsic used in gotm
> 
> ```
> PGF90-S-0155-Intrinsic not supported in initialization: spacing (/qfs/people/turn923/E3SM_test/components/mpas-source/src/core_ocean/gotm/src/util/field_manager.F90: 43)               
> PGF90-S-0457-Illegal expression in initialization (/qfs/people/turn923/E3SM_test/components/mpas-source/src/core_ocean/gotm/src/util/field_manager.F90: 85)
> ```

The `spacing` Fortran Intrinsic appears to be supported by pgfortran (according to the Fortran Reference page of the documentation, https://docs.nvidia.com/hpc-sdk/pgi-compilers/19.10/x86/fortran-ref-guide/index.htm#fort-intrin-func-tbl-arithmetic).  It just isn't supported in initialization, according to the error message

philipwjones:
Indeed, using intrinsics for compile-time evaluation can be risky for anything but the simplest intrinsics. In this case, it's probably better to find a hard-wired special value to use for these. I would not rely on PGI supplying a fix.

mark-petersen:
@mattdturner I set up a [simple test case](https://gist.github.com/mark-petersen/a5c53f32a8e8b47389d63146cce522d8) to confirm a potential solution.

With intel and gnu, both produce this output:
```
 default_fill_value             -1.7976931348623157E+308
 spacing(default_fill_value)     1.9958403095347198E+292
 digits(default_fill_value)             53
 precision(default_fill_value)          15
 default_minimum                -1.7976931348623155E+308
 default_maximum                 1.7976931348623157E+308
```
See [descriptions of these intrinsic functions](https://docs.nvidia.com/hpc-sdk/pgi-compilers/19.10/x86/fortran-ref-guide/index.htm#fort-intrin-func-tbl-90-95-numeric-inquiry).

Could you test both with PGI? I know `test_spacing_on_initialization.F90` will fail on compile. The question is if `test_spacing_during_run.F90` works for PGI. If so, the solution is to have GOTM change this line:
```
<    real*8, parameter :: default_minimum = default_fill_value + spacing(default_fill_value)
---
>    real*8 :: default_minimum
...
>    default_minimum = default_fill_value + spacing(default_fill_value)
```
where `default_minimum` is set when the code is first run. This is a little tricky, because it is declared within an introduced data type at the top of the module. But let's just see first if it is a possibility for PGI.

mark-petersen:
@qingli411 and @mattdturner, I thought about this for a while, and this seems like a reasonable solution in the gotm library:
```
--- a/src/util/field_manager.F90
+++ b/src/util/field_manager.F90
@@ -40,7 +40,7 @@ module field_manager
    integer, parameter, public :: output_level_debug    = 32

    real(rk),parameter :: default_fill_value = -huge(_ONE_)
-   real(rk),parameter :: default_minimum = default_fill_value + spacing(default_fill_value)
+   real(rk),parameter :: default_minimum  = default_fill_value*0.9999999_rk
```
That is not exactly the same, but it serves the same purpose. It looks to me as if default_minimum is only used as a default to provide a minimum value to the netcdf file. With the new line, this is what I get:
```
Single precision:
 default_fill_value             -3.40282347E+38
 spacing(default_fill_value)     2.02824096E+31
 default_minimum (old)          -3.40282326E+38
 default_minimum (new)          -3.40282306E+38
Double precision:
 default_fill_value             -1.7976931348623157E+308
 spacing(default_fill_value)     1.9958403095347198E+292
 default_minimum (old)          -1.7976931348623155E+308
 default_minimum (new)          -1.7976929550930023E+308
```
Note I keep seven digits of accuracy in both single and double precision because I have seven digits in `0.9999999_rk`.  But I hate to make a mountain out of a molehill. For our purposes, `default_minimum` is still greater than, but close to the minimum number representation.

We could propose this to the developers of GOTM but first could you both check that it works? @mattdturner can you give this a test compile with that one line changed? @qingli411 could you rerun your tests and make sure they produce the identical result as before?

qingli411:
@mark-petersen Running the simple single column test with GNU in the debug mode, I got identical results with and without the change of the `default_minimum` line in `field_manager.F90`. 

For the issue of `default_minimum`, I think we are not using that part of the code. So I think your change shouldn't affect the results. GOTM has a 'library' mode (only including the turbulent mixing subroutines) and a 'standalone' mode (a 1D column model with time stepping and input/output). We are using GOTM in the 'library' mode. It looks like that the GOTM `turbulence` library uses part of the `util` library, of which `field_manager.F90` is also a part. So `field_manager.F90` gets compiled even if it is not used in the 'library' mode.

I also tried to use PGI (NVIDIA HPC SDK) on my Linux workstation with the change in `field_manager.F90`, it builds, but gives me the following error when I run it. This is probably an issue of the PGI environment on my Linux workstation as I got the same error for other MPAS-Ocean branches without GOTM. Still trying to figure out why as PGI was working fine previously on my workstation...
```
./ocean_model: error while loading shared libraries: libomptarget.so: cannot open shared object file: No such file or directory
```

mattdturner:
@mark-petersen Your proposed fix does allow E3SM to be compiled and run on Compy w/ PGI

mark-petersen:
@qingli411 would it make sense to make a pull request into GOTM with that fix? At the same time, could we have the MPAS GOTM submodule point to a branch on your fork of GOTM with that fix, so we can proceed with this PR?

qingli411:
@mark-petersen Looks like that in the new stable release of GOTM (recently released about 10 days ago), `field_manager.F90` is reorganized into a submodule of GOTM. I guess it's probably better for us to update to the latest GOTM release in this PR. I think the GOTM `turbulence` module that we are using doesn't change too much so I hope we don't need to change too much on the MPAS-Ocean side. I will test MPAS-Ocean with the latest release of GOTM to see if this issue is still there. Since we are only using GOTM in the 'library' mode, I guess the latest GOTM release might just work. 

I also notice that the line using `spacing` is still there (now in the new submodule 'flexout' of GOTM), so I think we could perhaps make a PR into 'flexout' with your fix.

mark-petersen:
@qingli411 that sounds like a good plan. See if you can update and test the new GOTM in this PR. We will still need to change the line with `spacing` for the PGI compiler, so would still need to make the PR and point to a branch on your fork for right now.

qingli411:
@mark-petersen I just tested the new GOTM release and we are having the same issue with the use of `spacing` for PGI. So I will point to a branch on my GOTM fork with the fix for now and make the PR to GOTM. Looks like that there is another issue when compiling the new GOTM release in the 'library' mode. So I will still use the previous stable GOTM release in my branch for this PR.

qingli411:
@mark-petersen Just pointed GOTM to a branch on my GOTM folk with your fix for PGI.

mattdturner:
I just rebased to the current HEAD of `ocean/develop` and will work on getting the E3SM PR ready now.

mark-petersen:
@qingli411 I moved the compass files here: https://github.com/MPAS-Dev/compass/pull/66. Can you run your test case from there with the head of this PR, and make sure it all works?

qingli411:
@mark-petersen The head of this PR now crashes with the following error without much useful backtrace information at the first time step. @mattdturner What was the change in the rebase? Do we need to change the GOTM interface `src/core_ocean/shared/mpas_ocn_vmix_gotm.F`?

```
```

mark-petersen:
@qingli411 compile with `DEBUG=true` and it will give you a backtrace. We rebased on #825, where we replaced the top index with an array.

qingli411:
So we need to use `minLevel*` in all the k loops, right? I'm not sure why this can break the single column test though as there are no inactive top cells...

I did compile with `DEBUG=true` but for some reason the backtrace information is all something like `#0  0x102b619cc`. I was compiling MPAS-O with `gfortran` on a MacBook Pro, and a Linux workstation as well.

mark-petersen:
We are moving to `minLevel*` module by module. You do not need to change the gotm module to use it at this time. The intel compiler is available on IC and produces better error messages.

mark-petersen:
I also have problems running, and don't see any error messages or traceback with debug on using either intel or gnu. I added `git revert 86ffe2eac` so that GOTM writes verbose output, but still don't see anything.  I am running `ocean/global_ocean/QU240/performance_test/forward` with `config_use_gotm = .true.` My log file ends with
```
Pressure type is: Jacobian_from_TS
```
and the direct output to screen ends with
```
STOP init_turbulence
```
Does STOP here mean that it completed `init_turbulence` successfully, or that it stopped with an error? 

I've never seen the model stop without a trace-back in debug mode. We should check that the debug options go down to the GOTM compile.

Are we supposed to always use `config_use_cvmix = .false.` when gotm is on, or does it matter? Error occurs either way.

mattdturner:
I thought I had tested with Intel on Grizzly after rebasing (prior to pushing), but perhaps I'm remembering incorrectly.  I can start looking into the crash sometime this afternoon.

xylar:
> STOP init_turbulence

Doesn't `stop` quit a Fortran code entirely?

xylar:
> So we need to use minLevel* in all the k loops, right? I'm not sure why this can break the single column test though as there are no inactive top cells...

> We are moving to minLevel* module by module. You do not need to change the gotm module to use it at this time. The intel compiler is available on IC and produces better error messages.

Regarding adding `minLevel*`, my feeling is that it would be *safer* if you were to do that work as part of this PR, even if it takes a little longer to go in as a result.  If @cbegeman and I add the loops, we're going to have to make our best judgment calls about what the right behavior is with inactive top cells.  If you make the changes, you might be better at avoiding edge cases we might miss.

cbegeman:
> So we need to use minLevel* in all the k loops, right? I'm not sure why this can break the single column test though as there are no inactive top cells...

@qingli411 If this comment means that you tried to add `minLevel*` and it didn't pass the test case, I'm happy to help work through the issue. I agree with @xylar that it would be great to have your involvement in this change to `minLevel*`. We're not far from the full `minLevel*` implementation (@mark-petersen I anticipate opening the PR within a week).

mark-petersen:
> Doesn't stop quit a Fortran code entirely?

Here `STOP init_turbulence` is just a text output from gotm, so could mean anything.

To clarify on the `minLevel*` comments: for all our standard tests, the current code is expected to work after the rebase without any changes. That is because we are changing a hard-coded `for k=1,...` to  `for k=minLevel*,...` etc. But for the nightly regression suite tests, those 'minLevel*` arrays are all ones so it shouldn't matter.

In other words, changing the top index from 1 to `minLevel*` in the gotm module is not the first step in debugging the current problem. It must be something else that was introduced by the rebase.

xylar:
@mark-petersen, thanks for clarifying.  I agree, `minLevel*` is very unlikely to be the problem on its own.

qingli411:
Looks like with the code (https://github.com/MPAS-Dev/MPAS-Model/commit/5717df33b33b78b099d754c1c3322d264e7880be) before the latest rebase, this test case still crashes, which is strange as it was working fine... Still trying to figure out what caused the issue. 

Thanks for the clarifying the `minLevel*` changes, @xylar @cbegeman and @mark-petersen. I agree it makes sense to include the necessary changes in this PR. But as @mark-petersen suggests that should be the next step.

qingli411:
I have traced the error to these lines when initializing GOTM arrays
https://github.com/qingli411/MPAS-Model/blob/1bf3482737ae47b35d96075d967a3a98f317aa42/src/core_ocean/shared/mpas_ocn_vmix_gotm.F#L416-L432

Looks like for some reason both `nCellsAll` and `nVertLevelsP1` are zero here, so that the GOTM arrays did not get properly initialized and were set to the default zero values. So we were getting a division by zero error when running GOTM. Any suggestions?

cbegeman:
@qingli411 My guess is that it is because `ocn_vmix_init` is called before `ocn_meshCreate` so you can't just `use ocn_mesh` in this gotm submodule.

mattdturner:
@qingli411 @cbegeman Is correct that the issue is `ocn_vmix_init` being called before `ocn_meshCreate`.  I've run into this before, where tests pass for some unknown reason if using `ocn_mesh` before it is initialized.  We'll have to revert the use of `ocn_mesh` during initialization, and instead use the subpool retrievals again.

qingli411:
@cbegeman @mattdturner OK I see. Thanks.

mattdturner:
@qingli411 Sorry about this being an issue.  I should have double-checked initialization when I implemented `ocn_mesh` in this PR.  Let me know if you need help reverting the use of `ocn_mesh` in initialization.

qingli411:
@mattdturner No problem! I will try to revert it in the initialization subroutine and let you know if any problem arises. Thanks.

qingli411:
Just made the `minLevel*` changes. The same single column test for GOTM works fine with these changes. But I don't have a way to test if inactive top cells work with GOTM. @cbegeman @xylar, could either of you please take a look and see if these changes make sense to you? 

Note that, different from MPAS-O, GOTM uses indices `(0:nlev)` to represent the vertical layers from the bottom to the top, so that the index `0` represents the bottom of the bottom layer `1`.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -1093,6 +1095,24 @@ subroutine ocn_vmix_implicit(dt, meshPool, statePool, forcingPool, scratchPool,
...
+      if ( config_use_gotm ) then
+
+         nEdges = nEdgesArray( 1 )
+         call mpas_timer_start('GOTM avg', .false.)
+         !$omp do schedule(runtime) private(cell1, cell2, k)
+         do iEdge=1,nEdges
+            vertViscTopOfEdge(:, iEdge) = 0.0_RKIND
+            cell1 = cellsOnEdge(1,iEdge)
+            cell2 = cellsOnEdge(2,iEdge)
+            do k=1,maxLevelEdgeTop(iEdge)
```
cbegeman, line 1108:
```suggestion
            do k=minLevelEdgeBot(iEdge),maxLevelEdgeTop(iEdge)
```

cbegeman (commented):
@qingli411 Thanks for making those changes! I think for now we don't need to test the case where `minLevelCell != 1`. Do you mind running your test case again with the `minLevelEdgeBot` changes I suggested? If that works out then I'm satisfied.

src/core_ocean/shared/mpas_ocn_vmix.F
qingli411: @cbegeman Thanks for these suggestions! Just want to make sure should we use `minLevelEdgeBot` or `minLevelEdgeTop` here? I notice that a few lines above this when CVMix is used, the `k` loop starts from `minLevelEdgeTop`. Although it probably does not matter as I think we will not use the value of `vertViscTopOfEdge` at the top of the top active layer, right? So I guess we should use `minLevelEdgeBot` for CVMix as well?

cbegeman: @qingli411 That's a great question. I think you're right that we should use `minLevelEdgeBot` for CVMix above. I'll test/include that change in my next `minLevel*` PR so you only need to make the change in the GOTM section.

xylar: @qingli411, `minLevelEdgeBot` is the deeper of the two `minLevelCell` on the edge, and `minLevelEdgeTop` is the shallower of the 2.  So they are both 1 for `minLevelCell=1` everywhere.

qingli411: Thanks, @xylar! OK. I misunderstood it... Now it makes sense to me.

cbegeman: @qingli411 Glad you asked. `minLevelEdgeBot` is actually 1. It's defined as `max(minLevelCell(cell1),minLevelCell(cell2))` so in the case where there are no inactive cells `minLevelCell(cell1)=minLevelCell(cell2)=1`. You can think of `minLevelEdgeBot` as the uppermost level where cells on both sides are active. 

Conceptually, when there is topography under ice shelves and thus inactive top cells, do you think that these vertical mixing variables should be computed at the top level of a cell when it is **not** horizontally connected to its neighbor? If the answer is yes, then we should use `minLevelEdgeTop`, if no, `minLevelEdgeBot`.

qingli411: Thanks, @cbegeman! Yes, it makes sense to me now to use `minLevelEdgeBot` here as the vertical mixing should only be applied away from the boundaries.

qingli411:
With these changes @cbegeman suggested the single column test works.

cbegeman:
Thanks, @qingli411!

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
```diff
@@ -0,0 +1,494 @@
...
+         call do_turbulence(maxLevelCell(iCell) - minLevelCell(iCell) + 1,                &
+                            gotm_dt,                                                      &
+                            waterDepth,                                                   &
+                            surfaceFrictionVelocity(iCell),                               &
+                            bottomFrictionVelocity,                                       &
+                            gotmSurfaceRoughnessLength,                                   &
+                            gotmBottomRoughnessLength,                                    &
+                            hLayer(maxLevelCell(iCell)+1:minLevelCell(iCell):-1),         &
+                            NSquaredTop(maxLevelCell(iCell)+1:minLevelCell(iCell):-1),    &
+                            shearSquaredTop(maxLevelCell(iCell)+1:minLevelCell(iCell):-1))
```
mark-petersen, line 284: @qingli411 when the order is inverted here, the compiler is forced to create temporary arrays in memory within the `iCell` loop. This is bad for performance and unnecessary, since these three arrays are created above. Please change these three variables to be ordered from bottom to top above, and then pass them in without the `-1` index at the end. You can test it by compiling with intel debug. Temporary array creation produces a warning like this:
```
forrtl: warning (406): fort: (1): In call to DO_TURBULENCE, an array temporary was created for argument #10

ocean_model        0000000002E0D2C6  Unknown               Unknown  Unknown
ocean_model        00000000016EDAD2  ocn_vmix_gotm_mp_         286  mpas_ocn_vmix_gotm.f90
ocean_model        00000000016F701F  ocn_vmix_mp_ocn_v         188  mpas_ocn_vmix.f90
```
or you can just let me test again.

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
```diff
@@ -0,0 +1,494 @@
...
+
+      ! initialize GOTM parameters
+      gotmSurfaceRoughnessLength = config_gotm_constant_surface_roughness_length
+      gotmBottomRoughnessLength = config_gotm_constant_bottom_roughness_length
+      gotmBottomDragCoef = config_gotm_constant_bottom_drag_coeff
+
+      !
+      ! initialize GOTM turbulence module
+      ! read in GOTM namelist, required by init_turbulence
+      call mpas_new_unit(namlstUnit)
```
mark-petersen, line 389: When I tested gotm I forgot to include the gotm namelist file in the run directory. But the error messages don't indicate what is wrong. Please check here if the file exists. If it does not, call 
```
call mpas_log_write(' fatal error: gotm namelist file does not exist: config_gotm_namelist_file = '//config_gotm_namelist_file, MPAS_LOG_ERR)
```
so the user knows. Thanks!

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
qingli411: @mark-petersen I redefined these arrays with indices in the GOTM order, which starts from 0. Could you please test it again and see if there are any problems?

src/core_ocean/shared/mpas_ocn_vmix_gotm.F
qingli411: @mark-petersen This should be checked in GOTM. But since we are suppressing all the messages from GOTM we won't see the error message. I added a check here as suggested.

mark-petersen:
rebased. Passes nightly regression suite with gnu and intel, both debug, gotm off. Also tested QU240 with gotm on with both.

PR #705, opened 2020-09-24
adds new GM option -- eden greatbatch

vanroekel:
This adds a simplified version of the Eden and Greatbatch (2008) EKE budget to compute the GM bolus kappa.  It is a steady state budget but is fully three-dimensional.

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -2536,7 +2536,7 @@
 		<var name="eddyTime" type="real" dimensions="nEdges Time" units="s^{-1}"
 			description="inverse eddy time scale in Visbeck 1997 closure"
 		/>
-		<var name="gmKappaScaling" type="real" dimensions="nVertLevelsP1 nEdges Time" units="NA"
+		<var name="gmKappaScaling" type="real" dimensions="nVertLevelsP1 nEdges Time" units="NA" default_value="1.0"
 			description="spatially and depth varying GM kappa.  The scaling is based on the Brunt Vaisala Frequency relative to a maximum value below the mixed layer, follows from Danabasoglu and Marshall 2007.  If config_GM_closure is not set to N2_dependent the scaling value is set to 1 everywhere."
 			packages="gm"
 		/>
@@ -2645,11 +2645,11 @@
 		<var name="indMLD" type="integer" dimensions="nCells Time" units="unitless"
 			description="index of model where mixed layer depth occurs (always one past)"
 		/>
-		<var name="slopeTriadUp" type="real" dimensions="nVertLevels TWO nEdges Time" units="non-dimensional"
+		<var name="slopeTriadUp" type="real" dimensions="nVertLevels TWO nEdges Time" units="non-dimensional" default_value="0.0"
 			 description="Magnitude of slope of isopycnal surface, using triad through this cell and edge, angled up. Uses expansion of equation of state."
 			 packages="forwardMode;analysisMode"
 		/>
-		<var name="slopeTriadDown" type="real" dimensions="nVertLevels TWO nEdges Time" units="non-dimensional"
+		<var name="slopeTriadDown" type="real" dimensions="nVertLevels TWO nEdges Time" units="non-dimensional" default_value="0.0"
 			 description="Magnitude of slope of isopycnal surface, using triad through this cell and edge, angled up. Uses expansion of equation of state."
 			 packages="forwardMode;analysisMode"
 		/>
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -320,28 +320,28 @@
 		<nml_option name="config_GM_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$. Used when config_GM_kappa_function is set to constant."
 		/>
-		<nml_option name="config_GM_closure" type="character" default_value="edenGreatbatch" units="unitless"
-					description="Control what method used to compute GM $\kappa$. Both 'constant' and 'N2_dependent' use the method in Ferrari et al. 2010 (https://doi.org/10.1016/j.ocemod.2010.01.004). 'constant' uses a constant kappa in eqn 16a, while 'N2_dependent' varies kappa in the vertical according to Danabasoglu and Marshall 2007 (https://doi.org/10.1016/j.ocemod.2007.03.006). 'visbeck' implements a horizontally varying diffusivity of Visbeck et al 1997. edenGreatbatch implements a simplified form of the EKE scheme in Eden and Greatbatch (2008) Ocean modeling"
-					possible_values="'N2_dependent', 'constant', 'visbeck', 'edenGreatbatch'"
+		<nml_option name="config_GM_closure" type="character" default_value="EdenGreatbatch" units="unitless"
+					description="Control what method used to compute GM $\kappa$. Both 'constant' and 'N2_dependent' use the method in Ferrari et al. 2010 (https://doi.org/10.1016/j.ocemod.2010.01.004). 'constant' uses a constant kappa in eqn 16a, while 'N2_dependent' varies kappa in the vertical according to Danabasoglu and Marshall 2007 (https://doi.org/10.1016/j.ocemod.2007.03.006). 'Visbeck' implements a horizontally varying diffusivity of Visbeck et al 1997. EdenGreatbatch implements a simplified form of the EKE scheme in Eden and Greatbatch (2008) Ocean modeling"
+					possible_values="'N2_dependent', 'constant', 'Visbeck', 'EdenGreatbatch'"
 		/>
 		<nml_option name="config_GM_baroclinic_mode" type="real" default_value="1.0" units="NA"
 					description="baroclinic wave mode chosen for the ferrari et al 2010 calculation"
 					possible_values="small positive numbers"
 		/>
-		<nml_option name="config_GM_visbeck_alpha" type="real" default_value="0.005" units="NA"
-					description="scaling factor on the visbeck diffusivity parameterization"
+		<nml_option name="config_GM_Visbeck_alpha" type="real" default_value="0.005" units="NA"
+					description="scaling factor on the Visbeck diffusivity parameterization"
 					possible_values="small positive numbers"
 		/>
-		<nml_option name="config_GM_visbeck_max_depth" type="real" default_value="1000.0" units="m"
+		<nml_option name="config_GM_Visbeck_max_depth" type="real" default_value="1000.0" units="m"
 					description="minimum depth for calculation of vertical average"
 					possible_values="values between zero and bottom depth"
 		/>
-		<nml_option name="config_GM_visbeck_min_kappa" type="real" default_value="300.0" units="m^2 s^{-1}"
-					description="minimum value of bolus diffusivity for visbeck scheme"
+		<nml_option name="config_GM_Visbeck_min_kappa" type="real" default_value="300.0" units="m^2 s^{-1}"
+					description="minimum value of bolus diffusivity for Visbeck scheme"
 					possible_values="values around 100s"
 		/>
-		<nml_option name="config_GM_visbeck_max_kappa" type="real" default_value="1800.0" units="m^2 s^{-1}"
-					description="minimum value of bolus diffusivity for visbeck scheme"
+		<nml_option name="config_GM_Visbeck_max_kappa" type="real" default_value="1800.0" units="m^2 s^{-1}"
+					description="minimum value of bolus diffusivity for Visbeck scheme"
 					possible_values="values around 100s"
 		/>
 		<nml_option name="config_gm_EG_riMin" type="real" default_value="200.0" units="NA"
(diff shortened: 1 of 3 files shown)


vanroekel:
@mark-petersen this has been tested in G- and B-cases in E3SM, but I have cleaned up and reorganized, so will run another E3SM G-case before merging. 

@mattdturner could you check openmp declarations?  I'm working on a few fixes of this I just noticed.

vanroekel:
@mark-petersen this is the last PR for the ocean in WC v2

src/core_ocean/Registry.xml
```diff
@@ -321,7 +321,7 @@
...
 		/>
 		<nml_option name="config_GM_closure" type="character" default_value="N2_dependent" units="unitless"
-					description="Control what method used to compute GM $\kappa$. Both 'constant' and 'N2_dependent' use the method in Ferrari et al. 2010 (https://doi.org/10.1016/j.ocemod.2010.01.004). 'constant' uses a constant kappa in eqn 16a, while 'N2_dependent' varies kappa in the vertical according to Danabasoglu and Marshall 2007 (https://doi.org/10.1016/j.ocemod.2007.03.006). 'visbeck' implements a horizontally varying diffusivity of Visbeck et al 1997"
+					description="Control what method used to compute GM $\kappa$. Both 'constant' and 'N2_dependent' use the method in Ferrari et al. 2010 (https://doi.org/10.1016/j.ocemod.2010.01.004). 'constant' uses a constant kappa in eqn 16a, while 'N2_dependent' varies kappa in the vertical according to Danabasoglu and Marshall 2007 (https://doi.org/10.1016/j.ocemod.2007.03.006). 'visbeck' implements a horizontally varying diffusivity of Visbeck et al 1997. edenGreatbatch implements a simplified form of the EKE scheme in Eden and Greatbatch (2008) Ocean modeling"
```
milenaveneziani, line 324: shouldn't `edenGreatbatch` be included as an option below?

src/core_ocean/Registry.xml
```diff
@@ -321,7 +321,7 @@
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$. Used when config_GM_kappa_function is set to constant."
 		/>
 		<nml_option name="config_GM_closure" type="character" default_value="N2_dependent" units="unitless"
```
milenaveneziani, line 323: shouldn't we have `constant` as the default?

src/core_ocean/Registry.xml
vanroekel: it's there, just way at the end of the text

src/core_ocean/Registry.xml
vanroekel: I wasn't sure what to put here.  This was what's there.  In E3SM we set it to visbeck and when we make the e3sm PR we'll set to edenGreatbatch in E3SM.  I don't think it matters too much what we set it to here.  What do you think @mark-petersen ?

src/core_ocean/Registry.xml
milenaveneziani: no, I mean the row below, sorry, under `possible_values`

vanroekel: I see, yes, thanks for catching that!

vanroekel:
Tested in G-case for 1 year, results look consistent with previous implementation.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -486,86 +488,93 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
-            do iCell=1,nCells
+            do iEdge=1,nEdges
                k=1
-               maxN = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
+               cell1 = cellsOnEdge(1,iEdge)
+               cell2 = cellsOnEdge(2,iEdge)
+               BruntVaisalaFreqTopEdge = 0.5_RKIND*(BruntVaisalaFreqTop(k,cell1) + &
+                                                    BruntVaisalaFreqTop(k,cell2))
+               maxN = max(BruntVaisalaFreqTopEdge, 0.0_RKIND)
                do k = 2, maxLevelCell(iCell)
```
mattdturner, line 556: This loop has a limit of `maxLevellCell(iCell)`, but the parent loop is now over `iEdge`, not `iCell`.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -486,86 +488,93 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
-                 maxN = max(maxN,max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND))
+                 BruntVaisalaFreqTopEdge = 0.5_RKIND*(BruntVaisalaFreqTop(k,cell1) + &
+                                                      BruntVaisalaFreqTop(k,cell2))
+                 maxN = max(maxN,max(BruntVaisalaFreqTopEdge, 0.0_RKIND))
                enddo
 
                do k = 1,maxLevelCell(iCell)
```
mattdturner, line 562: This is `maxLevelCell(iCell)` as well.  For my own reference, would `maxLevelCell(cell1)` equal `maxLevelCell(cell2)` in all cases?

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -486,86 +488,93 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
+         end do
+         !$omp end do
+         !$omp end parallel
 
+         nEdges = nEdgesArray(3)
          ! For config_GM_closure = 'N2_dependent' use a scaling to taper gmBolusKappa
          ! based on stratification relative to the maximum in the column
          if (local_config_GM_kappa_lat_depth_variable) then
             !$omp parallel
             !$omp do schedule(runtime) private(k, BruntVaisalaFreqTopEdge, maxN)
```
mattdturner, line 548: This OpenMP directive will need `cell1` and `cell2` added to the `private()` clause.

src/core_ocean/shared/mpas_ocn_gm.F
mark-petersen: I think they should all be `maxLevelEdgeTop(iEdge)`.  cell1 and cell2 would, in general, have different values of `maxLevelCell`.  That is why there are two arrays for the edge: `maxLevelEdgeTop` and `maxLevelEdgeBottom`.

src/core_ocean/shared/mpas_ocn_gm.F
mark-petersen: fixed.

mattdturner (approved):
Thanks for the quick updates @mark-petersen .  After those fixes, this now passes the nightly regression suite w/ intelv19 + DEBUG.  Approved

src/core_ocean/Registry.xml
mark-petersen: These defaults set how the code is exercised for our testing suite. Let's use the same defaults as E3SM for that reason, so set it to edenGreatbatch.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -666,6 +675,40 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
+                 sigma = max(abs(fEdge(iEdge)),sqrt(2.0*betaEdge(iEdge)*cGMphaseSpeed(iEdge))) / &
+                         sqrt(sumRi + config_gm_EG_riMin)
+                 L_rhines = sigma / (1.0E-11_RKIND + betaEdge(iEdge))
+                 Lr = min(cGMphaseSpeed(iEdge) / abs(fEdge(iEdge)), sqrt(cGMphaseSpeed(iEdge) / &
+                         (2.0_RKIND*betaEdge(iEdge))))
+                 Length = min(Lr,L_rhines)
+                 !For compatibility with other schemes, we normalize by the 2-D kappa field, which will
+                 !be removed below in the computation of the stream function
+                 gmKappaScaling(k,iEdge) = max(config_GM_visbeck_min_kappa, min(config_gm_EG_kappa_factor* &
+                         Length**2.0*sigma, config_GM_visbeck_max_kappa)) / gmBolusKappa(iEdge)
```
mark-petersen, line 705: @vanroekel this line had gmBolusKappa(iCell) rather than iEdge. I just fixed it. I think this could alter your E3SM cases with Eden & Greatbatch that you already ran.

BTW, if you are working in E3SM, you can do a quick test with _D for debug, like
```
./create_test SMS_Ln9_D.T62_oQU120.GMPAS-IAF
```
and it will catch those.

vanroekel:
@mark-petersen I found a couple bugs in the physical formulation.  Working on them right now.  Hope to have them for you soon.

src/core_ocean/shared/mpas_ocn_gm.F
vanroekel: thanks for the catch and the tip @mark-petersen!

vanroekel:
@mark-petersen this should be ready now.  I've made all fixes.

vanroekel:
results from a 5-year G-case look good.

vanroekel:
Passes ```SMS_Ln9_D.T62_oQU120.GMPAS-IAF.anvil_intel``` sanity test

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -410,8 +410,7 @@ subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnostic
 
       deallocate(div_hu,div_huTransport,div_huGMBolus)
 
-      nEdges = nEdgesArray( 2 )
-
+      nEdges = nEdgesArray( 3 )
```
mark-petersen, line 413: @vanroekel FYI, once we used `tangentialVelocity` in the Richardson number, we had to expand this halo to pass the partition test.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -561,7 +561,7 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
                  maxN = max(maxN,max(BruntVaisalaFreqTopEdge, 0.0_RKIND))
                enddo
 
-               do k = 1, maxLevelEdgeTop(iEdge)
+               do k = 2, maxLevelEdgeTop(iEdge)
```
mark-petersen, line 553: Index 1 on `gmKappaScaling` is never used, and 1 caused an out-of bounds error.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -235,16 +235,6 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
       end do
       !$omp end do
 
-      !$omp do schedule(runtime) private(k)
-      do iEdge = 1, nEdges
-         do k = 1, nVertLevels
-            slopeTriadUp(k, :, iEdge) = 0.0_RKIND
-            slopeTriadDown(k, :, iEdge) = 0.0_RKIND
-            gmKappaScaling(k,iEdge) = 1.0_RKIND
```
mark-petersen: These variables didn't need to be reset every time step. They are computed below. I added lines to initialize them the first time in the Registry. I did a check, results of QU240 are bfb before/after this change.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -305,8 +305,7 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
       !$omp do schedule(runtime)  &
       !$omp private(invAreaCell, k33Norm, k, dzTop, dTdzTop, dSdzTop, i, iEdge, cell1, cell2, &
       !$omp         iCellSelf, dcEdgeInv, areaEdge, drhoDT, drhoDS, dTdx, dSdx, drhoDx, &
-      !$omp         slopeTaperUp, slopeTaperDown, sfcTaperUp, sfcTaperDown, sfcTaper) &
-      !$omp firstprivate(slopeTriadUp, slopeTriadDown)
```
mark-petersen, line 315: Array variables are not supposed to be in a private clause. I confirmed this change with @mattdturner. I'm surprised this had passed a threading bfb test.

philipwjones: be a little careful with the language here - arrays can certainly be in private clauses if they are truly private to each loop iteration (eg the dzTop, etc arrays).  And in the old threading model, all local arrays ended up being thread private which was part of the problem.  Having not looked carefully at slopeTriadUp/Down, it's not completely obvious - these look like they are large-ish local copies replicated across icell iterations. (k,2,nEdges) for each iCell iteration? Could be wrong

mark-petersen: These are registry arrays that have dimensions (k, 2, nEdges) - it wouldn't make sense for these to be thread private. They are used later, and are not just a local copy.

vanroekel:
@mark-petersen just reran an E3SM G-case with the head of this PR and the results look good.  I think this is ready to go

mark-petersen (approved):
Ran test on last commit, with gnu 6 debug and intel 19 debug, passes nightly regression suite. Also compared with 2 threads versus no openmp, and get bfb match.

PR #706, opened 2020-09-25
Changed default namelist and streams to E3SM default

akturner:
Changed Registry.xml configs and streams to E3SM default
Added forcing directory to forcing filenames in Registry
Updated SnowAgingPropertiesInput file from snicar_drdt_bst_fit_60_c04182019.nc to snicar_drdt_bst_fit_60_c04262019.nc
Modified testing to allow exceptions in individual tests without then stopping and allowing remaining tests to run
Added e3sm default configuration to configurations
Added e3sm default configuration to tests

### DIFF
--- a/src/core_seaice/Registry.xml
+++ b/src/core_seaice/Registry.xml
@@ -438,7 +438,7 @@
 			description="The number of ice layers in the vertical direction to use."
 			possible_values="Any positive integer."
 		/>
-		<nml_option name="config_nSnowLayers" type="integer" default_value="1" units="unitless"
+		<nml_option name="config_nSnowLayers" type="integer" default_value="5" units="unitless"
 			description="The number of snow layers in the vertical direction to use."
 			possible_values="Any positive integer."
 		/>
@@ -671,7 +671,7 @@
 			description="Run the ridging column physics calculation."
 			possible_values="true or false"
 		/>
-		<nml_option name="config_use_column_snow_tracers" type="logical" default_value="false" units="unitless"
+		<nml_option name="config_use_column_snow_tracers" type="logical" default_value="true" units="unitless"
 			description="Allow snow modifications plus additional snow tracers"
 			possible_values="true or false"
 		/>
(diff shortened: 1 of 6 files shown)


akturner:
Note file change for SnowAgingPropertiesInput

njeffery:
@akturner, the namelist for e3sm_default still has the old default values.   The streams file looks okay.

akturner:
@njeffery: Thanks for noticing that - I will fix ASAP.

akturner:
@njeffery, @proteanplanet: With the e3sm default options the restartability test failed, so maybe a variable is missing from restart? Any ideas what that would be?

proteanplanet:
> @njeffery, @proteanplanet: With the e3sm default options the restartability test failed, so maybe a variable is missing from restart? Any ideas what that would be?

@akturner Thanks for trying to get this through.  I'll take a look at this next week when V2 is out of the way.  For the record, I've done an ERS test within E3SM on G-cases with these default settings, and it passed, so a quick comparison between the two configurations should reveal the problem.

njeffery:
@akturner :  there are two config options that need to be turned on during restarts: config_do_restart_snow_density = true and  config_do_restart_snow_grain_radius = true.    The testsuite.snow_tracer_physics.xml has them included with the option "snow_tracer_physics" value="True".

akturner:
@njeffery , @proteanplanet : I've fixed the restartability testing issue

proteanplanet (approved):
Since this passes all tests, and all namelist options for the default E3SM V2 configuration are switched on, I approve this request.   Thanks @akturner!

njeffery:
@akturner : While testing on anvil,  I found a couple of things that you may want to roll into this PR.  The get_data.py script has old python 2 syntax that no longer works with python 3.  string.strp should be changed to str.strp.   Another really minor thing, the testing  readme file describing set-up of the mpas_seaice conda environment should have  conda install -n mpas_seaice -c anaconda six

njeffery (approved):
Looks good. Passes parallelism and restartability.  I'll add a consistency check in MPAS for snow restart flags in a later PR.

akturner:
@njeffery I removed python 2 support from the script, removing the string and six module usage

mgduda:

PR #707, opened 2020-09-26
Update test for libpio to skip libpio.settings

### LINKED ISSUE #695: An error occurred when building the MPAS.
when i use ‘make gfortran CORE=ocean’ this command to build mpas，but the program occurred an error，as follows.
the directory of pio installation is '/usr/local/lib,/usr/local/include',How should I specify the environment variable of pio?
thanks ！

Checking for a usable PIO library...
************ ERROR ************
Failed to compile a PIO test program
Please ensure the PIO environment variable is set to the PIO installation directory
************ ERROR ************
make[1]: *** [pio_test] Error 1
make[1]: Leaving directory `/opt/mpas/MPAS-Model-master'
make: *** [gfortran] Error 2

xylar:
The latest PIO2 (2.5.2) includes a file `libpio.settings` that
was giving our Makefile trouble because it matches the wildcard
`libpio.*`, `-lpio` will fail.

Closes #695

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-MODEL_FORMULATION = 
+MODEL_FORMULATION =
 
 ifneq "${MPAS_SHELL}" ""
         SHELL = ${MPAS_SHELL}
@@ -507,9 +507,9 @@ llvm:
 	"OPENMP = $(OPENMP)" \
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI" )
 
-CPPINCLUDES = 
-FCINCLUDES = 
-LIBS = 
+CPPINCLUDES =
+FCINCLUDES =
+LIBS =
 
 #
 # If user has indicated a PIO2 library, define USE_PIO2 pre-processor macro
@@ -544,9 +544,15 @@ endif
 # Depending on PIO version, libraries may be libpio.a, or libpiof.a and libpioc.a
 # Keep open the possibility of shared libraries in future with, e.g., .so suffix
 #
+# Check if libpio.* exists and link -lpio if so, but we make an exception for
+# libpio.settings (a file added in PIO2), which is not a library to link
 ifneq ($(wildcard $(PIO_LIB)/libpio\.*), )
-	LIBS += -lpio
+	# Makefiles don't support "and" operators so we have nested "if" instead
+	ifneq "$(wildcard $(PIO_LIB)/libpio\.*)" "$(PIO_LIB)/libpio.settings"
+		LIBS += -lpio
+	endif
 endif
+
 ifneq ($(wildcard $(PIO_LIB)/libpiof\.*), )
 	LIBS += -lpiof
 endif


xylar:
@philipwjones, @mgduda, @mark-petersen, @matthewhoffman and @akturner, we got reports in #695 that MPAS isn't building with the latest PIO2 because of the new file `libpio.settings`.  This is my suggestion for how to fix that issue.  Please let me know if you have questions, concerns or suggestions.

I have done a test build of the current `ocean/develop` with a cherry-pick of ca51467 and it builds with both the latest PIO2 (2.5.2) and `scorpio` 1.1.1 under Ubuntu 18.04 with gnu.

Makefile
```diff
@@ -545,8 +545,11 @@ endif
 # Keep open the possibility of shared libraries in future with, e.g., .so suffix
 #
 ifneq ($(wildcard $(PIO_LIB)/libpio\.*), )
-	LIBS += -lpio
+	ifneq "$(wildcard $(PIO_LIB)/libpio\.*)" "$(PIO_LIB)/libpio.settings"
```
mark-petersen, line 548: My understanding here is:
case 1: only $(PIO_LIB)/libpio.settings exists, so they are equal, and -lpio is not added.
case 2: only $(PIO_LIB)/libpio.a exists, they are not equal, and -lpio is added.
case 3: both files exist, so two items are not equal, so -lpio is added.

why don't we just check for the existence of libpio.a? That seems much more straightforward:
```
ifneq ("$(wildcard $(PIO_LIB)/libpio\.a)","")
```
I got the format from http://www.humbug.in/2012/makefile-check-if-a-file-exists-using-wildcard-function/

Would it ever be named something other than `libpio.a`? I think that could replace the first ifneq on line 547 as well.

mark-petersen: I see. We are leaving it open to libpio.a or libpio.so. So we could do this:
```
 544 # Depending on PIO version, libraries may be libpio.a, or libpiof.a and libpioc.a
 545 # Keep open the possibility of shared libraries in future with, e.g., .so suffix
 546 #
 547 ifneq ($(wildcard $(PIO_LIB)/libpio\.a), )
 548     LIBS += -lpio
 549 else ifneq ($(wildcard $(PIO_LIB)/libpio\.so), )
 550     LIBS += -lpio
 551 endif
```
That is more readable, at least. It looks like make files don't support booleans, which is why I used two ifneq lines. You could read it as "if libpio.a exists" and "if libpio.so exists".

xylar: > why don't we just check for the existence of libpio.a?

It could be called `libpio.a`, `libpio.so`, or `libpio.dylib`, depending on the system and whether these are static or dynamic libraries.  I don't think we want to lock ourselves into static libraries.

Your second suggestion only covers linux, not OSX, which we also want to support.  I think it's just easier to treat `libpio.settings` as a special case.  I'm open to another suggestion but the nested if was the way I had seen suggested for "and" in a makefile.

mark-petersen: I see, I understand now. It is just Makefile clumsiness because of three possible suffixes and lack of boolean operators. Yes, considering that, your formulation is fine.

mark-petersen (approved):

matthewhoffman (approved):
@xylar , based on the discussion with @mark-petersen , I was convinced this was a good solution.  However, I wouldn't have been able to follow what the changes are doing without that discussion.  Can you add to the comment above this bit of code to elaborate on the logic being employed?  I think that could save some time if this needs to be adjusted again in the distant future.

xylar:
Thanks, @matthewhoffman.  I added additional comments.  Let me know if they address your concerns

matthewhoffman:
@xylar , looks good - thanks.

mgduda (changes requested):
The changes to the `Makefile` look good to me, and I've tested that they don't have any impact when building with PIO1 or PIO2 prior to the introduction of the `libpio.settings` file. I'd recommend two changes, though:
1. If we base this PR off of the `v7.0` tag, then anyone using the v7.0 release of MPAS with PIO 2.5.2 can simply merge the head of the branch in this PR; for example, we can tell users to simply run `git merge abcd123` in order to pick up this fix. In fact, I could even include the changes in the MPAS v7.1 release (which I've been meaning to make for quite a long time).
2. We could squash the two commits in this PR into a single commit for simplicity.

xylar:
@mgduda, those are good suggestions.  I have rebased onto v7.0 and squashed, as you requested.

mgduda (approved):
@xylar Thanks very much for the updates, and apologies for taking so long to approve. I've tested that builds still work with PIO1 and PIO2 (<2.5.2), so all looks good from my perspective.

xylar:
Thanks @mgduda.  

@akturner and @philipwjones, do you want to review this PR?

philipwjones (approved):
Approve based on visual inspection and testing reported by others.

mark-petersen:
Rebased, as there was a very small merge conflict. Did a local merge into ocean/develop and tested with gnu 6. All passed. Merging in...

xylar:
Thanks, @mark-petersen

PR #708, opened 2020-09-27
Add Piecewise Polynomial Reconstruction (PPR) submodule

xylar:
This module will be used in the future for vertical remapping as an alternative to vertical advection.  Advantages of vertical remapping are:
* higher order accuracy
* time step is not limited by vertical CFL
* efficiency (since remapping need not occur every time step)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "src/core_ocean/BGC"]
 	path = src/core_ocean/BGC
 	url = https://github.com/E3SM-Project/Ocean-BGC.git
+[submodule "src/core_ocean/ppr"]
+	path = src/core_ocean/ppr
+	url = https://github.com/dengwirda/PPR.git
--- a/src/core_ocean/Makefile
+++ b/src/core_ocean/Makefile
@@ -11,7 +11,7 @@ ifneq "$(EXCLUDE_INIT_MODE)" "true"
 	OCEAN_SHARED_INCLUDES += -I$(PWD)/mode_init
 endif
 
-all: shared libcvmix analysis_members libBGC
+all: shared libcvmix analysis_members libBGC libppr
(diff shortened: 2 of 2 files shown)


xylar:
@dengwirda, so far, I'm only adding PPR to the build but not using it yet anywhere in MPAS-Ocean.  There are a couple of options for how to use this PR.  If there is somewhere you could imagine adding PPR calls as a starting point that wouldn't require too much design work, we could go ahead and include that here.  If not, it might make sense to keep this as a starting point but I still don't think we want to merge until we have some code in MPAS-O that actually uses PPR.

xylar:
@mark-petersen and @cbegeman, just keeping you in the loop for now but please let me know if you have thoughts or concerns.

dengwirda:
Thanks @xylar, this looks good --- I'm not sure how best to check that all is/will work when called from within MPAS-O yet, but as you say, I think this is a good starting point. 

One addition that may be useful is a link to the `PPR` examples: [https://github.com/dengwirda/PPR/tree/master/example](https://github.com/dengwirda/PPR/tree/master/example). These describe how to call the routines, etc, as well as being simple unit tests.

I think moving to `PPR`-style ("remap") advection may require broader dycore changes, so let's approach that via a design doc, etc.

@vanroekel, just checking whether your use of `PPR` might be affected by this (I suspect not)?

vanroekel:
@dengwirda thanks for pointing this PR out to me. I've included PPR in a different part of the code in a different way in my super parameterization branch, but will plan to utilize this PR instead.

xylar:
It seems like we might need a special version of PPR for MPAS-Ocean to be performant.  I'm going to close this PR.  We can explore this or other approaches within E3SM.

xylar:
To be clear, we haven't explored the performance of PPR in MPAS-Ocean.  So we don't know whether it will be performant or not as it is.  But there is reason to think the long-term strategy will be to have a version that is aware of MPAS-Ocean data structures.

It still might make sense to have a submodule for PPR to begin with, but this will be added to E3SM, rather than here.

PR #711, opened 2020-10-01
PR from VIMS-UM fork, which contains work done to 2-way couple SCHISM and MPAS-O

josephzhang8:
The new addition is relatively simple to integrate. The only new file created is 

core_ocean/analysis_members/mpas_ocn_schism.F

which incorporates SCHISM as a new analysis member. The top level Makefile now has schism inside (e.g. LIBS). In addition, schism is also registered in: Registry_analysis_members.xml, Registry_schism.xml, mpas_ocn_analysis_driver.F

Corresponding inputs are also augmented with schism sections:
namelist.ocean, namelist.ocean.analysis, namelist.ocean.forward, streams.ocean

Note that the exchange of some barotropic variables is not completed as MPAS-O crashes when forced by SSH, velocity etc. This will be fixed later. Also exchange of BGC variables can be expanded later.

Joseph, Zhengui and Fei


mgduda:

PR #712, opened 2020-10-01
Added registry file defining namelist parameters used by e3sm driver

akturner:
Added registry file defining namelist parameters used only by the e3sm driver. These do not appear anywhere in MPAS-Seaice. They include a new character array type since the driver anmelist requires this for e3sm streams.


mgduda:

PR #714, opened 2020-10-06
Adds modifications to the E&G length scale

vanroekel:
equation (28) of eden and greatbatch (2008) shows that implementation
requires a constant factor on the Rossby and Rhines length.  This is
implemented in this PR.


vanroekel:
@mark-petersen this is a very small PR that based on testing in WC.  The paper I pulled the new GM from has this modification deeper in the paper.  It has been fully tested in E3SM (G + B cases).  If you could review as soon as possible, I'd greatly appreciate it.

mark-petersen (approved):
Approved. This matches eqn 28 in E&G. Compiled and ran a quick regression suite for a sanity test.

PR #715, opened 2020-10-06
Fix memory corruption and leak issues in gen_inc for parse executable

markjolah:
The `parse` executable exhibits memory corruption under certain compiler/libc combinations.   The source of the problems are `sprintf()` calls with overlapping pointers which is not permitted by the C standard.   In particular, the memory corruption leads to malformed `.inc` files on MacOS when linking to Apple `libc` and on Linux linking to `glibc` under various compilers whenever any C-compiler optimizations are enabled.

From [`man sprintf`](https://linux.die.net/man/3/sprintf):
> **Notes**
> Some programs imprudently rely on code such as the following
> `sprintf(buf, "%s some further text", buf);`
> to append text to buf. However, the standards explicitly note that the results are undefined if source and destination buffers overlap when calling sprintf(), snprintf(), vsprintf(), and vsnprintf(). Depending on the version of gcc(1) used, and the compiler options employed, calls such as the above will not produce the expected results. 

This PR fixes those problematic locations in the source, relying on `strcat()` instead.

Additionally, in multiple locations variables named `tofree` are never `free()`'d.  While these memory leaks are arguably not a major correctness problem for this very simple short-running executable, they do present a myriad of false-positives for memory-checking and static-analysis tools, making it difficult to find any actual memory corruption errors.

With these changes, the parser executable corruption bugs appear to be fixed.

Tested with platforms/compilers:
* Linux
  * `GCC-10.2.0`
  * `Clang-10.0.1`
  * `Intel-20`
* MacOS
  * `GCC-10.2.0`
  * `Clang-11.0.3`

I am submitting this PR against `master` on advice from @mgduda to merge to the oldest branch exhibiting the bug, but I think the same patch will also apply cleanly to `develop` or `v7.0`.

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -153,16 +153,17 @@ int add_package_to_list(const char * package, const char * package_list){/*{{{*/
 	token = strsep(&string, ";");
 
 	if(strcmp(package, token) == 0){
+		if(tofree) free(tofree);
 		return 0;
 	}
 
 	while( (token = strsep(&string, ";")) != NULL){
 		if(strcmp(package, token) == 0){
-
+			if(tofree) free(tofree);
 			return 0;
 		}
 	}
-
+	if(tofree) free(tofree);
 	return 1;
 }/*}}}*/
 
@@ -225,12 +226,13 @@ int build_struct_package_lists(ezxml_t currentPosition, char * out_packages){/*{
 				if(out_packages[0] == '\0'){
 					sprintf(out_packages, "%s", token);
 				} else if(add_package_to_list(token, out_packages)){
-					sprintf(out_packages, "%s;%s", out_packages, token);
+					strcat(out_packages, ";");
+					strcat(out_packages, token);
 				}
 
 				while( (token = strsep(&string, ";")) != NULL){
 					if(add_package_to_list(token, out_packages)){
-						sprintf(out_packages, "%s;%s", out_packages, token);
+						strcat(out_packages, token-1);
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


mgduda:
@markjolah Apologies for the long delay in getting around to working on this PR. I've just changed the base to the `hotfix-v7.1` branch, as I'm hoping we can release v7.1 very soon, merging the fixes from the release onto `develop` immediately afterward.

matthewhoffman (approved):
Approved by inspection.  Thanks for the work in identifying this and sorting it out.

mark-petersen (approved):
Merged locally, passes all ocean tests with gnu 6 and intel 19 in debug. I must admit I don't understand this correction. But if it corrects your problem, then I approve.

mgduda (changes requested):
With the changes in this PR, we're getting differences in the generated code in `structs_and_variables.inc` for the atmosphere core:
```diff
8812,8813c8812
<       if (limited_areaActive .or. bl_mynn_inActive .or. cu_tiedtke_inActive .or. mp_kessler_inActive .or. mp_thompson_inActive &
< .or. mp_wsm6_inActive) then
---
>       if (limited_areaActive .or. bl_mynn_inActive .or. mp_kessler_inActive .or. mp_thompson_inActive) then
16657,16658c16656
<       if (iauActive .or. bl_mynn_inActive .or. bl_ysu_inActive .or. cu_tiedtke_inActive .or. mp_kessler_inActive .or. &
< mp_thompson_inActive .or. mp_wsm6_inActive) then
---
>       if (iauActive .or. bl_mynn_inActive .or. mp_kessler_inActive .or. mp_thompson_inActive) then
```
My initial thought is that the use of `token-1` isn't correct, as the `;` delimiter is replaced with a `\0` as described in the man page for `strsep`:
```
This token is terminated by overwriting the delimiter with a null byte ('\0')
and *stringp is updated to point past the token.
```

MiCurry (approved):
I'm seeing the same differences in `structs_and_variables.inc` as @mgduda, but otherwise this looks good.

snyderdad:
@markjolah, I know you're departing soon.  It seems like there is a real bug here and that you are most of the way to fixing it.  Could you have a look and see if this can be resolved before you leave?

PR #717, opened 2020-10-07
changes GM config option names for clarity

vanroekel:
This PR only changes names of a few config flags for clarity.  right now, multiple GM options use config flags with visbeck in the name and this is not appropriate.  This PR should be bfb.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -260,7 +260,7 @@ contains
       ! It is only needed when redi mixing is enabled.
       if ( config_use_redi ) then
 
-      if ( config_Redi_use_N2_based_taper .and. .not. local_config_GM_kappa_lat_depth_variable ) then
+      if ( config_Redi_N2_based_taper_enable .and. .not. local_config_GM_kappa_lat_depth_variable ) then
 
          !$omp parallel
          !$omp do schedule(runtime) private(maxLocation, k, BruntVaisalaFreqTopEdge, maxN)
@@ -275,10 +275,9 @@ contains
             maxLocation = k
             do k = maxLocation, maxLevelCell(iCell)
                BruntVaisalaFreqTopEdge = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
-!!!!!!!!! LUKE WILL CHANGE THIS LOGIC, OCT 8 2020
-!               RediKappaScaling(k, iCell) = min(max(config_GM_min_stratification_ratio, &
-!                                               BruntVaisalaFreqTopEdge/(maxN + 1.0E-10_RKIND)), &
-!                                           1.0_RKIND)
+               RediKappaScaling(k, iCell) = min(max(config_Redi_N2_based_taper_min,                  &
+                                                    BruntVaisalaFreqTopEdge/(maxN + 1.0E-10_RKIND)), &
+                                                    1.0_RKIND)
             end do
          end do
          !$omp end do
@@ -555,10 +554,10 @@ contains
                                                        BruntVaisalaFreqTop(k,cell2))
                   BruntVaisalaFreqTopEdge = max(BruntVaisalaFreqTopEdge, 0.0_RKIND)
 
-!!!!!!!!! LUKE WILL CHANGE THIS LOGIC, OCT 8 2020
-!                  gmKappaScaling(k, iEdge) = min(max(config_GM_min_stratification_ratio, &
-!                                               BruntVaisalaFreqTopEdge/(maxN + 1.0E-10_RKIND)), &
-!                                               1.0_RKIND)
+                  gmKappaScaling(k, iEdge) = min(max(config_GM_spatially_variable_min_kappa /         &
+                                                     config_GM_spatially_variable_max_kappa,          &
+                                                     BruntVaisalaFreqTopEdge/(maxN + 1.0E-10_RKIND)), &
+                                                     1.0_RKIND)
                end do
            enddo
            !$omp end do
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -251,180 +251,180 @@ contains
       !$omp end do
       !$omp end parallel
 
-      allocate (dzTop(nVertLevels + 1))
-      allocate (dTdzTop(nVertLevels + 1))
-      allocate (dSdzTop(nVertLevels + 1))
-      allocate (k33Norm(nVertLevels + 1))
-
       ! The following code computes scaling for Redi mixing terms and the slope triads
-      ! It is only needed when redi mixing is enabled.
-      if ( config_use_redi ) then
+      ! It is only needed when Redi mixing is enabled.
+      if ( config_use_Redi ) then
 
-      if ( config_Redi_N2_based_taper_enable ) then
+         allocate(dzTop(nVertLevels + 1))
+         allocate(dTdzTop(nVertLevels + 1))
+         allocate(dSdzTop(nVertLevels + 1))
+         allocate(k33Norm(nVertLevels + 1))
 
-         !$omp parallel
-         !$omp do schedule(runtime) private(maxLocation, k, BruntVaisalaFreqTopEdge, maxN)
-         do iCell = 1, nCells
-            k = min(maxLevelCell(iCell) - 1, max(1, indMLD(iCell)))
-            maxN = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
-            do while (BruntVaisalaFreqTop(k + 1, iCell) > maxN .and. k < maxLevelCell(iCell) - 1)
-              k = k + 1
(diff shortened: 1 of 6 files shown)


vanroekel:
@mark-petersen as we start to add more GM options, the config names are getting confusing.  This PR cleans that up.  I think this can be combined into E3SM with #714

mark-petersen:
@vanroekel thanks. Longer config names are better. In my mind there are two different types of spatial variations:
1. Horizontal variations based on grid cell size (or data field in the future)
2. Vertical variations that start with (1) as a base value but then modify it in the column based on stratification or N^2 using Visbeck or E&G or Ghokan's paper.

We always have (1) on, and the testing the past few months is on (2). When we say "3D GM" we mean (2). Is that a fair summary?

Since we are changing names, would it make sense to use "hor", "3D" or "vert" in any of these flags? Let me look at them again...

vanroekel:
@mark-petersen it was tough to figure what a good renaming is.  my changes were all related to (2), I think we used ```gmResolutionTaper``` for the horizontal (1) above.  The tough part about (2) though is that it actually does do horizontal and vertical variation.  So maybe we should do ```gmHorizontalResolutionTaper``` for (1) and add ```3D``` for (2). so like ```config_GM_3D_spatially_variable_```?

mark-petersen:
Rebased. Based on discussion with @vanroekel, I reordered and changed flag names for GM and Redi.  Test compile and passes regression suite.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -275,10 +275,10 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
             maxLocation = k
             do k = maxLocation, maxLevelCell(iCell)
                BruntVaisalaFreqTopEdge = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
-
-               RediKappaScaling(k, iCell) = min(max(config_GM_min_stratification_ratio, &
-                                               BruntVaisalaFreqTopEdge/(maxN + 1.0E-10_RKIND)), &
-                                           1.0_RKIND)
+!!!!!!!!! LUKE WILL CHANGE THIS LOGIC, OCT 8 2020
+!               RediKappaScaling(k, iCell) = min(max(config_GM_min_stratification_ratio, &
```
mark-petersen, line 279: @vanroekel please change this line and also 559, where `config_GM_min_stratification_ratio` was removed.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -681,29 +684,19 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
...
-                         Length**2.0*sigma, config_GM_Visbeck_max_kappa)) / gmBolusKappa(iEdge)
+                 gmKappaScaling(k,iEdge) = max(config_GM_spatially_variable_min_kappa, min(config_GM_EG_kappa_factor* &
+                         Length**2.0*sigma, config_GM_spatially_variable_max_kappa)) / gmBolusKappa(iEdge)
               enddo
            enddo
            !$omp end do
            !$omp end parallel
          endif
 
-         if(config_Redi_set_RediKappa_to_GMKappa) then
```
mark-petersen, line 697: We decided to remove this flag, so I removed this block completely. @vanroekel please double check this deletion.

mark-petersen:
Here is our naming scheme:
```
 54 &mesoscale_eddy_parameterizations
 55     config_eddying_resolution_taper = 'ramp'
 56     config_eddying_resolution_ramp_min = 20e3
 57     config_eddying_resolution_ramp_max = 30e3
 58 /
 59 &Redi_isopycnal_mixing
 60     config_use_Redi = .false.
 61     config_Redi_closure = 'constant'
 62     config_Redi_constant_kappa = 600.0
 63     config_Redi_maximum_slope = 0.3
 64     config_Redi_use_slope_taper = .true.
 65     config_Redi_use_surface_taper = .true.
 66     config_Redi_use_N2_based_taper = .true.
 67     config_Redi_N2_limit_term1 = .true.
 68 /
 69 &GM_eddy_parameterization
 70     config_use_GM = .false.
 71     config_GM_closure = 'EdenGreatbatch'
 72     config_GM_constant_kappa = 600.0
 73     config_GM_constant_gravWaveSpeed = 0.3
 74     config_GM_spatially_variable_min_kappa = 300.0
 75     config_GM_spatially_variable_max_kappa = 1800.0
 76     config_GM_spatially_variable_baroclinic_mode = 1.0
 77     config_GM_Visbeck_alpha = 0.005
 78     config_GM_Visbeck_max_depth = 1000.0
 79     config_GM_EG_riMin = 200.0
 80     config_GM_EG_kappa_factor = 3.0
 81     config_GM_EG_Rossby_factor = 2.0
 82     config_GM_EG_Rhines_factor = 0.3
 83 /
```
On all the GM ones, we tried to unify the naming:
```
config_GM_closure = 'constant', 'N2_dependent', 'Visbeck', 'EdenGreatbatch'

so all the rest of the flags say what closure they apply to:
  config_GM_constant_*
  config_GM_spatially_variable_* means any except constant
  config_GM_Visbeck_*
  config_GM_EG_*
```

src/core_ocean/shared/mpas_ocn_gm.F
vanroekel: looks good, thanks!

vanroekel:
okay @mark-petersen I think I'm done now.  For the Redi_N2_based_taper I've changed to the following
```
config_Redi_N2_based_taper_enable -- Logical to turn on or off
config_Redi_N2_based_taper_limit_term1 -- logical to taper horizontal component of redi with N2
config_Redi_N2_based_taper_min -- minimum floor for the taper
```

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -260,7 +260,7 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
       ! It is only needed when redi mixing is enabled.
       if ( config_use_redi ) then
 
-      if ( config_Redi_use_N2_based_taper .and. .not. local_config_GM_kappa_lat_depth_variable ) then
+      if ( config_Redi_N2_based_taper_enable .and. .not. local_config_GM_kappa_lat_depth_variable ) then
```
mark-petersen, line 263: @vanroekel can you check this logic? We were trying to separate the Redi and GM flags. From below, we have
```
if (config_GM_closure == 'N2_dependent') then
...
 960             local_config_GM_kappa_lat_depth_variable = .true.
```
otherwise it is false. So if `config_GM_closure == 'N2_dependent'` we _do not_ do this section. Does that still make sense? I suspect it is now incorrect, since we only compute RediKappaScaling below, not anything for GM.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -275,10 +275,9 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
             maxLocation = k
             do k = maxLocation, maxLevelCell(iCell)
                BruntVaisalaFreqTopEdge = max(BruntVaisalaFreqTop(k, iCell), 0.0_RKIND)
-
-               RediKappaScaling(k, iCell) = min(max(config_GM_min_stratification_ratio, &
-                                               BruntVaisalaFreqTopEdge/(maxN + 1.0E-10_RKIND)), &
-                                           1.0_RKIND)
+               RediKappaScaling(k, iCell) = min(max(config_Redi_N2_based_taper_min,                  &
```
mark-petersen, line 278: @vanroekel I was expecting to replace ratio with a min/max. Is the max assumed to be one here, so we just have the min? If so, then this is correct.

vanroekel: yes, the max is just 1 and shouldn't be changed

src/core_ocean/shared/mpas_ocn_gm.F
vanroekel: Thanks for the catch!  you are exactly right.  I just pushed a fix, the later condition should not be there.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -260,7 +260,7 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, diagnosticsPool, &
       ! It is only needed when redi mixing is enabled.
       if ( config_use_redi ) then
```
mark-petersen, line 261: Almost done... On this logical, we need to make sure that nothing in GM below depends on this Redi section. @vanroekel do you agree?

It looks like this section only computes `slopeTriadUp, slopeTriadDown, k33, k33Norm, RediKappaScaling` and those are not used in the sections below, so this if statement is correct.

mark-petersen: Confirmed with @vanroekel verbally, this is all correct.

mark-petersen (approved):
This looks good. Recompiled and retested nightly suite with both intel and gnu. Passes all.

PR #718, opened 2020-10-09
Fix typos in Registry descriptions of rthratenlw and rthratensw

MiCurry:
This commit fixes the description of rthratenlw which previously stated that it
was due to 'short wave' radiation instead of 'long wave' radiation.

The descriptions for rthratenlw and rthratensw reference 'long wave' and
'short wave', respectfully, as two words. Other descriptions in the registry
reference these as 'longwave' and 'shortwave', i.e. as a single word, so this
commit changes 'long wave' and 'short wave' to a single word to match the rest
of the registry. This helps when finding fields that are related to shortwave
and longwave radiation.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -2955,10 +2955,10 @@
                 <!--  rthratenlw:uncoupled theta tendency due to longwave radiation                             [K s-1] -->
 
                 <var name="rthratensw" type="real" dimensions="nVertLevels nCells Time" units="K s^{-1}"
-                     description="tendency of potential temperature due to short wave radiation"/>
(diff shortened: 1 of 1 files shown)


mgduda:
Since PR #347 has already changed "short wave" to "long wave" in the description of `rthratenlw`, and since we have other variants of spelling ("long-wave", "longwave"), removing whitespace from two occurrences doesn't seem critical. Ideally, at some point before the next release we can review the `Registry.xml` file and make all spelling, hyphenation, etc. consistent.

@MiCurry I'll close this PR, but if you would like to repurpose it to make all "shortwave"/"short-wave"/"short wave" occurrences consistent, please feel free to re-open it.

PR #719, opened 2020-10-14
Breakup the diagnostic_solve routine

mattdturner:
This PR breaks up the diagnostic_solve subroutine, and modifies the split-explicit timestepping scheme to have 1 of the calls to diagnostic solve only calculate a few of the variables.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -516,6 +516,7 @@ contains
       ! Gradient Richardson number
       !
 
+      if ( full_compute ) then
       nCells = nCellsArray(3)
       !$omp parallel
       !$omp do schedule(runtime) private(invAreaCell1, k, shearSquared, i, iEdge, delU2, shearMean)
@@ -536,6 +537,9 @@ contains
           RiTopOfCell(1,iCell) = RiTopOfCell(2,iCell)
       end do
       !$omp end do
+      !$omp end parallel
+
+      end if    ! full_compute
 
       !
       ! extrapolate tracer values to ocean surface
@@ -544,6 +548,7 @@ contains
       ! field will be updated below is better approximations are available
 
 !TDR need to consider how to handle tracersSurfaceValues
+      !$omp parallel
       !$omp do schedule(runtime)
       do iCell = 1, nCells
          tracersSurfaceValue(:, iCell) = activeTracers(:,1, iCell)
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -92,7 +92,7 @@ contains
 !-----------------------------------------------------------------------
 
    subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnosticsPool, scratchPool, tracersPool, &!{{{
-                                   timeLevelIn)
+                                   timeLevelIn, full)
 
       real (kind=RKIND), intent(in) :: dt !< Input: Time step
       type (mpas_pool_type), intent(in) :: statePool !< Input: State information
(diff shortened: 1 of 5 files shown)


mattdturner:
This PR passes the nightly test suite for Intel v19 w/ debug on Grizzly.  It also passes the following E3SM tests:

```
PASS:
SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu
```

For those E3SM tests (not including the ones w/ Debug enabled), the resulting speedup is
**Maximum**: 15.13% (although the total runtime for this case was 2.2 seconds)
**Average**: 3.34%

The speedup for the higher resolution test (`PET_Ln9_P8192.T62_oRRS18to6v3.GMPAS-IAF.cori-knl_intel`) is 6.9%.

philipwjones:
@mattdturner - this looks good. One quick question - was there a reason why the meshPool hasn't been completely removed? Was it causing problems or were there other dependencies preventing it?

mattdturner:
This PR will probably go in after https://github.com/MPAS-Dev/MPAS-Model/pull/611 , which does completely remove meshPool.  Although, the actual reason that this PR doesn't completely remove meshPool is because I forgot to remove all instances of it.

Once #611 goes in, I imagine there will be a rather time-consuming rebase for this PR.  But I wanted to get `ocn_diagnostic_solve` broken up to see what kind of performance improvement could be achieved.

mark-petersen:
@vanroekel we would like this PR to proceed to the special E3SM PR list for V2. It is bit-for-bit, and has a measured ocean speed-up of 3.4%. I wanted to give you a heads-up, in case we need to ask during the Wed water cycle meeting. I still need to review, but it should be ready later this week.

@mattdturner is bringing over the needed pieces from #611 right now. @philipwjones, could you also finish your review after he adds that (probably tomorrow).

mattdturner:
@mark-petersen and @philipwjones I just pushed some changes to this PR.  It passes the nightly regression suite, and is bit-for-bit compared to the current ocean/develop.  

The overall runtime reduction for this PR will be a little less than previously reported, since I had to add back some computation to run every time `ocn_diagnostic_solve` is called.

philipwjones (approved):
Approve based on code inspection and prior testing.

mark-petersen:
Tested on badger with gnu 6. Matches bfb against `ocean/develop` with optimized. Matches between compile with and without OpenMP, 2 threads versus none.

mattdturner:
The following E3SM tests pass:
PASS:
```
```

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -604,43 +735,523 @@ subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnostic
...
+      !
+      ! Brunt-Vaisala frequency (this has units of s^{-2})
+      !
+      coef = -gravity / rho_sw
+      !$omp parallel
+      !$omp do schedule(runtime) private(k)
+      do iCell = 1, nCellsAll
+         BruntVaisalaFreqTop(1,iCell) = 0.0_RKIND
+         do k = 2, maxLevelCell(iCell)
+            BruntVaisalaFreqTop(k,iCell) = coef * (displacedDensity(k-1,iCell) - density(k,iCell)) &
```
mark-petersen, line 816: @mattdturner and @philipwjones here is a question: Would it be beneficial to merge loops where possible? Here we have a loop over iCell and k, and there is another one below at line 830. The loops share the `BruntVaisalaFreqTop` variable. As a rule of thumb, should we avoid separate loops when possible? I'm guessing the benefit is reduced memory access, so there is only a speed-up if the loops share variables, so the single loop reduces memory retrievals.

On the flip side, separate loops can be more clear, since each variable, like Brunt-Vaisala here, get their own block of code.

mattdturner: This loop is over `nCellsAll`, whereas the loop at line 830 is over `nCellsHalo( 2 )`.  So these 2 loops probably couldn't be merged without some conditionals or other trickery.  

Generally speaking, I try to fuse loops when possible to reduce the overhead associated with more loops.  Loop fission, though, can be beneficial if the number of computations in a loop becomes excessive, leading to register spills that degrade performance.  So unfortunately there isn't a 1 size fits all solution.

mark-petersen (approved):
@mattdturner looks great. Thanks for your work on this.

Do you have a new number for speed-up on this? You said 3.3% before, but that was before you added the KE and vorticity computation back in.

mattdturner:
I didn't re-test the performance within E3SM, just tested that all tests PASSed.  I can re-run a handful of tests to look at the performance.

mark-petersen:
It's always nice to document the speed-up, if you could run a few. Thanks.

mattdturner:
After having to add back a few portions of code, the average runtime reduction was just below 2%.  The high res run had a reduction of 2.3%.

PR #720, opened 2020-10-15
Parallel interpolation of the categorical fields soilcat, landuse

MiCurry:
This merge allows the init_atmosphere model to safely interpolate the categorical fields, x1 and x1 in parallel. However, because other fields have not yet been updated to run in parallel, the multiprocessor check in init_atm_cases is still present.

This merge is in connection with the merge at 91d8539. Before this merge, parallel interpolation of the x1 and x2 fields could not be done unless one was using a convex partition file, which are difficult to generate. This merge, along with 91d8539 (and future merges), make it possible to process static fields in parallel using a decomposition file created from metis.

Because other fields have not yet been updated to run in parallel, this merge still only allows for single processing of static fields.


MiCurry:
I'd like to clean up a few of the commit messages. I'll change this from a draft PR to a normal PR once I've done so.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -282,13 +283,17 @@ subroutine init_atm_static(mesh, dims, configs)
...
+       ! Should we grab these values from the index file?
        isice_lu = 24
        iswater_lu = 16
        ismax_lu = 24
        write(mminlu,'(a)') 'USGS'
     case('MODIFIED_IGBP_MODIS_NOAH')
+       ! Should we grab these values from the index file?
        isice_lu = 15
        iswater_lu = 17
        ismax_lu = 20
```
MiCurry, line 299: The index file for the MODIS dataset contains the above values (`isice`, `iswater` and `category_max`).

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -282,13 +283,17 @@ subroutine init_atm_static(mesh, dims, configs)
...
 !
+! TODO: Should this be moved to be above the interpolation of land use and soil at? Or
+! is it fine here?
  surface_input_select0: select case(trim(config_landuse_data))
     case('USGS')
+       ! Should we grab these values from the index file?
        isice_lu = 24
        iswater_lu = 16
        ismax_lu = 24
        write(mminlu,'(a)') 'USGS'
```
MiCurry, line 294: The index file for the USGS dataset does not contain the above values (`isice` and `iswater`), but does contain `category_max`.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
- scalefactor = 1.0
- allocate(rarray(nx,ny,nz))
- allocate(ncat(ismax_lu,nCells))
+
+ ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+ if (ierr /= 0) then
+    call mpas_log_write('Error initalizing lu_index Geotile Manager', messageType=MPAS_LOG_CRIT)
+ endif
+
+ allocate(ncat(ismax_lu, nCells))
```
MiCurry, line 477: Instead of using `ismax_lu` as the upper bound for `ncat`, should we use `category_min:category_max` so we have a lower and an upper bound for categories as specified in the index files?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
 
-             if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(iPoint), yCell(iPoint), zCell(iPoint), &
-                         nEdgesOnCell(iPoint), verticesOnCell(:,iPoint), xVertex, yVertex, zVertex)) then
+         do j = 1 + tile_bdr, tile_ny + tile_bdr, 1
+             do k = 1 + tile_bdr, tile_nx + tile_bdr, 1
 
-                ncat(int(rarray(i,j,1)),iPoint) = ncat(int(rarray(i,j,1)),iPoint) + 1
+                !
+                ! The MODIS dataset has zeros at the north pole, and possibly other places, so
+                ! skip any value that is below the category_min or category_max as reported in
```
MiCurry, line 519: This line should read:
                `! skip any value that is less than category_min or greater than category_max as reported in`

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                call mpas_kd_search(tree, (/xPixel, yPixel, zPixel/), res)
 
-    end do
+                !
+                ! For all but the outermost boundary cells, we can safely assume that the nearest
+                ! model grid cell contains the pixel (else, a different cell would be nearest)
+                !
+                if (bdyMaskCell(res % id) < nBdyLayers) then
+
+                   !write(0,*) tile % tile(k,j,1)
```
MiCurry, line 536: This line needs to be removed

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
-          else
-             zPixel = sphere_radius * sin(lat_pt)                  ! Model cell coordinates assume a "full" sphere radius
-             xPixel = sphere_radius * cos(lon_pt) * cos(lat_pt)    ! at this point, so we need to ues the same radius
-             yPixel = sphere_radius * sin(lon_pt) * cos(lat_pt)    ! for source pixel coordinates
+         all_pixels_mapped_to_halo_cells = .true.
 
-             if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(iPoint), yCell(iPoint), zCell(iPoint), &
-                         nEdgesOnCell(iPoint), verticesOnCell(:,iPoint), xVertex, yVertex, zVertex)) then
+         do j = 1 + tile_bdr, tile_ny + tile_bdr, 1
+             do k = 1 + tile_bdr, tile_nx + tile_bdr, 1
```
MiCurry, line 515: Replace `k` with `i`

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
- rarray_ptr = c_loc(rarray)
+ call mpas_pool_get_config(mgr % pool, 'tile_bdr', tile_bdr)
+ call mpas_pool_get_config(mgr % pool, 'tile_x', tile_nx)
+ call mpas_pool_get_config(mgr % pool, 'tile_y', tile_ny)
+ call mpas_pool_get_config(mgr % pool, 'category_min', category_min)
+ call mpas_pool_get_config(mgr % pool, 'category_max', category_max)
 
- do jTileStart = 1,20401,ny
-    jTileEnd = jTileStart + ny - 1
+ do i = 1, nCells
```
MiCurry, line 487: Replace `i` with `iCells`

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
- isigned  = 1
- endian   = 0
- wordsize = 1
- scalefactor = 1.0
- allocate(rarray(nx,ny,nz))
- allocate(ncat(ismax_lu,nCells))
+
+ ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+ if (ierr /= 0) then
+    call mpas_log_write('Error initalizing lu_index Geotile Manager', messageType=MPAS_LOG_CRIT)
```
MiCurry, line 474: Update this error message.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
+         end if
+     end do
  end do
- deallocate(rarray)
  deallocate(ncat)
- call mpas_log_write('--- end interpolate LU_INDEX')
 
+ ierr = mgr % finalize()
+ if (ierr /= 0) then
+    call mpas_log_write('Error finalizing the lu_index Geotile Manager', messageType=MPAS_LOG_CRIT)
```
MiCurry, line 583: Also update this error message.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
+            do k = 1 + tile_bdr, tile_nx + tile_bdr, 1
+
+                ! TODO: Should we have a check if a tile value is greater or less than
+                ! the category_max or category_min, respectfully?
+
+                call mgr % tile_to_latlon(tile, j, k, lat_pt, lon_pt)
+                call mpas_latlon_to_xyz(x, y, z, sphere_radius, lat_pt, lon_pt)
+
+                res => null()
+                call mpas_kd_search(tree, (/x, y, z/), res)
```
MiCurry, line 643: Instead of using `x, y, z` here, use `xPixel, yPixel, zPixel`.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                       all_pixels_mapped_to_halo_cells = .false.
+                   end if
+
+                ! For outermost boundary cells, additional work is needed to verify that the pixel
+                ! actually lies within the nearest cell
+                else
+                   ! Model cell coordinates assume a "full" sphere radius
+                   ! at this point, so we need to use the same radius
+                   ! for source pixel coordinates
+                   call mpas_latlon_to_xyz(xPixel, yPixel, zPixel, sphere_radius, lat_pt, lon_pt)
```
MiCurry, line 663: If we change mpas_kd_search to use `xPixel, yPixel and zPixel` then the above function call and comment can be removed.

src/core_init_atmosphere/mpas_init_atm_static.F
mgduda: Agreed.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -282,13 +283,17 @@ subroutine init_atm_static(mesh, dims, configs)
 !
 ! Set land use and soil category parameters for water and ice
 !
+! TODO: Should this be moved to be above the interpolation of land use and soil at? Or
+! is it fine here?
```
MiCurry, line 287: This should be left here (so we can throw an error as soon as possible if the user specified a bad value for `config_landuse_data`), but we should move setting `isice_lu`, `iswater_lu` and `ismax_lu` into the `geotile_manager`.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +468,240 @@ subroutine init_atm_static(mesh, dims, configs)
...
- do iCell = 1,nCells
-    lu_index(iCell) = 1
-    do i = 2,ismax_lu
-       if(ncat(i,iCell) > ncat(lu_index(iCell),iCell)) then
-          lu_index(iCell) = i
-       end if
-    end do
+ do iCell = 1, nCells
+     lu_index(iCell) = 1
+     do i = 2, ismax_lu
```
MiCurry, line 573: @mgduda was there a reason for looping between 2 and `ismax_lu` here instead of 1? Both of the landuse datasets and soiltop dataset have `category_min` set to 1 in their index files.

mgduda: If we looped from 1 to `ismax_lu`, the first iteration would check whether `ncat(1,iCell) > ncat(1,iCell)`, which is always false and therefore unnecessary.

Actually, though, we could probably use `maxloc` here rather than a loop to find the index of the maximum.

MiCurry:
Okay @mgduda I have just pushed an update to this branch with the changes that address the comments above and what we discussed over hangouts, so its now ready for your review!

You'll notice that I added two new commits. One that sets the default values for the `iswater`, `isice` and `isurban` parameters (and uses them or the values from the index file) and a commit to fix a GCC warning within `mpas_geotile_manager`.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
+ call mpas_pool_get_config(mgr % pool, 'isice', isice_lu_ptr)
+ call mpas_pool_get_config(mgr % pool, 'tile_bdr', tile_bdr)
+ call mpas_pool_get_config(mgr % pool, 'tile_x', tile_nx)
+ call mpas_pool_get_config(mgr % pool, 'tile_y', tile_ny)
+ call mpas_pool_get_config(mgr % pool, 'category_min', category_min)
+ call mpas_pool_get_config(mgr % pool, 'category_max', category_max)
+
+ ! Transfer iswater_lu_ptr, isice_lu_ptr to iswater_lu and isice_lu because
+ ! isawter_lu_ptr and isice_lu_ptr will become unassociated when calling
+ ! mgr % finalize. iswater_lu and isice_lu will be outputed to the output
```
mgduda, line 477: I think "outputed" should be "output" here. Alternatively, we could say that these will be "written" to the output stream.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
-             xPixel = sphere_radius * cos(lon_pt) * cos(lat_pt)    ! at this point, so we need to ues the same radius
-             yPixel = sphere_radius * sin(lon_pt) * cos(lat_pt)    ! for source pixel coordinates
+         do j = 1 + tile_bdr, tile_ny + tile_bdr, 1
+             do i = 1 + tile_bdr, tile_nx + tile_bdr, 1
 
-             if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(iPoint), yCell(iPoint), zCell(iPoint), &
-                         nEdgesOnCell(iPoint), verticesOnCell(:,iPoint), xVertex, yVertex, zVertex)) then
+                !
+                ! The MODIS dataset has zeros at the north pole, and possibly other places, so
+                ! skip any value that is less tha category_min or greater than category_max as
```
mgduda, line 518: "tha" -> "than"

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                ! reported in the index file.
+                !
+                if (tile % tile(i,j,1) <= category_min .or. tile % tile(i,j,1) >= category_max) then
```
mgduda, line 521: The comment above says "less than" or "greater than", but here we're using `<=` and `>=`. Should the comparisons here use `<` and `>`?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
+ do iCell = 1, nCells
+    lu_index(iCell) = maxloc(ncat(:,iCell), dim=1)
  end do
- deallocate(rarray)
  deallocate(ncat)
- call mpas_log_write('--- end interpolate LU_INDEX')
 
+ ierr = mgr % finalize()
+ if (ierr /= 0) then
+    call mpas_log_write('Error occured when finalizing the interpolation of landuse category (lu_index)', &
```
mgduda, line 576: "occured" -> "occurred"

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
- endian      = 0
- wordsize    = 1
- scalefactor = 1.0
- allocate(rarray(nx,ny,nz))
- allocate(ncat(16,nCells))
+ geog_sub_path = 'soiltype_top_30s/'
+
+ ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+ if (ierr /= 0) then
+    call mpas_log_write('Error occured when initalizing the interpolation of soil top category (soilcat_top)', &
```
mgduda, line 589: "occured" -> "occurred"

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
- isigned  = 1
- endian   = 0
- wordsize = 1
- scalefactor = 1.0
- allocate(rarray(nx,ny,nz))
- allocate(ncat(ismax_lu,nCells))
+
+ ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+ if (ierr /= 0) then
+    call mpas_log_write('Error occured when initializing the interpolation of landuse category (lu_index)', &
```
mgduda, line 463: "occured" -> "occurred"

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
+
+        call mpas_log_write('Processing tile: '//trim(tile%fname))
+
+        all_pixels_mapped_to_halo_cells = .true.
+
+        do j = 1 + tile_bdr, tile_ny + tile_bdr, 1
+            do i = 1 + tile_bdr, tile_nx + tile_bdr, 1
+
+                ! TODO: Should we have a check if a tile value is greater or less than
+                ! the category_max or category_min, respectfully?
```
mgduda, line 633: "respectfully" -> "respectively". Should we add this check as was done for land use category?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +457,238 @@ subroutine init_atm_static(mesh, dims, configs)
...
+ do iCell = 1, nCells
+    soilcat_top(iCell) = maxloc(ncat(:,iCell), dim=1)
  end do
- deallocate(rarray)
+
  deallocate(ncat)
+
+ ierr = mgr % finalize()
+ if (ierr /= 0) then
+    call mpas_log_write('Error occured when finalizing the interpolation of soil top category (soilcat_top)', &
```
mgduda, line 689: "occured" -> "occurred"

mgduda:
@MiCurry There are a few places in the static interpolation (e.g., in the [consistency check](https://github.com/MPAS-Dev/MPAS-Model/blob/v7.0/src/core_init_atmosphere/mpas_init_atm_static.F#L563-L578) between land use category and soil category) where we use the value `14`. Should we update this to `isoilwater`?

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Yes I think you're correct. Because categories can include `category_min` and `category_max`. I'll update it to be `<` and `>` instead of `<=` and `>=`.

MiCurry:
> `isoilwater`
Yes, that's a great idea. We can also use isoilice to replace where the soil type is fixed over Antarctica.

MiCurry:
> > @MiCurry There are a few places in the static interpolation (e.g., in the [consistency check](https://github.com/MPAS-Dev/MPAS-Model/blob/v7.0/src/core_init_atmosphere/mpas_init_atm_static.F#L563-L578) between land use category and soil category) where we use the value `14`. Should we update this to `isoilwater`?
> 
> Yes, that's a great idea. We can also use isoilice to replace where the soil type is fixed over Antarctica.

Actually, it looks like the value used to fix the soil type over Antartica (16) is not the same as the default value for isice, which is 24.

Oh actually, I'm wrong in this. I'm making an assumption that the `isice` and `iswater` index options are for both land use and soil type, but they are just for landuse. So I'll leave the fix as it is.

MiCurry:
Okay, @mgduda I have just pushed an update to this branch that address your review, so it should be ready to go now!

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -464,171 +460,246 @@ subroutine init_atm_static(mesh, dims, configs)
...
 
- do iCell = 1,nCells
-    lu_index(iCell) = 1
-    do i = 2,ismax_lu
-       if(ncat(i,iCell) > ncat(lu_index(iCell),iCell)) then
-          lu_index(iCell) = i
-       end if
-    end do
+ do iCell = 1, nCells
+    lu_index(iCell) = maxloc(ncat(:,iCell), dim=1)
```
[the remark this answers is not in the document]
MiCurry: Ah, you're right. What if we update line 573 (and the counterpart for `soilcat`) to:

```
lu_index(iCell) = (category_min - 1) + maxloc(ncat(:,iCell), dim=1)
```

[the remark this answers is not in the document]
mgduda: I think that should work. It may be worth adding a one-line comment to explain why we need to add `(category_min - 1)`, too. I don't think the parentheses are needed, and in my opinion the `- 1` is more closely associated with the `maxloc` call than with `category_min`.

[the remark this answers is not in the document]
MiCurry: Sounds good! I just pushed another update!

mgduda (approved):

PR #722, opened 2020-10-16
Add support for building with NAG compilers

mgduda:
This PR adds a build target for the NAG compilers to the top-level Makefile, and
it also fixes several compilation issues with the NAG compilers in shared code.

Although the 'test' core can now be compiled with the NAG compilers, other cores
contain issues in their core-specific code that prevent compilation, and this PR
does not address those issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -396,6 +396,31 @@ bluegene:
 	"OPENMP = $(OPENMP)" \
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI" )
 
+nag:
+	( $(MAKE) all \
+	"FC_PARALLEL = mpifort" \
+	"CC_PARALLEL = mpicc" \
+	"CXX_PARALLEL = mpic++" \
+	"FC_SERIAL = nagfor" \
+	"CC_SERIAL = gcc" \
+	"CXX_SERIAL = g++" \
+	"FFLAGS_PROMOTION = -r8" \
+	"FFLAGS_OPT = -free -mismatch -O3 -convert=big_ieee" \
+	"CFLAGS_OPT = -O3" \
+	"CXXFLAGS_OPT = -O3" \
+	"LDFLAGS_OPT = -O3" \
+	"FFLAGS_DEBUG = -free -mismatch -O0 -g -C -convert=big_ieee" \
+	"CFLAGS_DEBUG = -O0 -g -Wall -pedantic" \
+	"CXXFLAGS_DEBUG = -O0 -g -Wall -pedantic" \
+	"LDFLAGS_DEBUG = -O0 -g -C" \
+	"FFLAGS_OMP = -qsmp=omp" \
+	"CFLAGS_OMP = -qsmp=omp" \
+	"CORE = $(CORE)" \
+	"DEBUG = $(DEBUG)" \
+	"USE_PAPI = $(USE_PAPI)" \
+	"OPENMP = $(OPENMP)" \
+	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI -DUNDERSCORE -DNAG_COMPILER" )
+
 CPPINCLUDES = 
 FCINCLUDES = 
 LIBS = 
--- a/src/external/Makefile
+++ b/src/external/Makefile
@@ -3,7 +3,7 @@
 all: esmf_time ezxml-lib
 
 esmf_time:
-	( cd esmf_time_f90; $(MAKE) FC="$(FC) $(FFLAGS)" CPP="$(CPP)" CPPFLAGS="$(CPPFLAGS) -DHIDE_MPI" GEN_F90=$(GEN_F90) )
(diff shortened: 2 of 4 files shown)


mgduda:
Our work with MPAS-Atmosphere in CESM would benefit from the ability to build with the NAG compilers, and a first step in this seemed to be the addition of a build target that would allow us to compile and test stand-alone MPAS-Atmosphere with the NAG compilers.

As the description of this PR indicates, most cores probably won't build right now with the NAG compilers, but it seems reasonable to leave it up to core developers to create future PRs to address this as needed/desired. We'll need to create a future PR to address compilation and runtime issues in MPAS-Atmosphere.

mgduda:
I just force-pushed an update that adds `-gline` to the NAG debug flags.

mgduda:
@matthewhoffman I agree that the loss of the ability to place namelist groups in any order is a nasty side-effect of supporting NAG. I scanned through the Fortran 2003 standard, and it wasn't easy for me to discern whether the standard requires groups to be in order or whether this as a NAG idiosyncrasy. [This discussion](https://groups.google.com/g/comp.lang.fortran/c/3-dVbjGEzB8) suggests that requiring namelist groups to be in order isn't a violation of the standard, but neither is allowing them to be in any order.

One way to work around this might be to define a pre-processing macro, like `MPAS_NAG_FORTRAN` when using the `nag` build target, then we can protect the `REWIND` statement with `#ifndef MPAS_NAG_FORTRAN`, for example.

mgduda:
@matthewhoffman [This commit](https://github.com/mgduda/MPAS-Model/commit/0520ec2) illustrates what I think might allow us to only require in-order namelist groups with the NAG compiler. I'll do a little more testing, but if that commit seems reasonable to you, I can add a cleaned-up version of it to this PR.

philipwjones:
@mgduda , @matthewhoffman - while the standard is ambiguous on the order, it does specify a number of different error codes that can be used to query/skip/rewind to search for and read the correct namelist in a file. We used this in POP to good effect so that we could support any order and also trap some other common errors. So if you want this capability, it is pretty easy to implement.

matthewhoffman:
@mgduda , the commit you referenced looks like a good solution.  I'll leave it up to you if you want to handle it that way or look into what @philipwjones suggested.

mgduda:
@philipwjones We do use the `IOSTAT` values when trying to read namelists to some degree, though, for example, we don't try to check whether `IOSTAT` is equal to `IOSTAT_END` or `IOSTAT_EOR` as defined in the Fortran 2003 standard. I would be interested to see how POP handles namelist reading (if it's easy to give a link to the code somewhere online); but, I'm inclined to take the minimum necessary steps to get the code building and running (albeit, with the requirement that namelist groups are given in order) with NAG. I don't expect we'll actually be using the NAG builds for any real-world runs, but rather, just for testing and debugging.

@matthewhoffman I'll update this PR shortly with changes to detect whether the NAG compiler is used, and to only skip the rewinding of namelist files in that case. I.e., we'll get exactly the same behavior as before with all other compilers.

philipwjones:
@mgduda  You can look at the POP version in
https://github.com/ESCOMP/POP2-CESM/blob/master/source/POP_ConfigMod.F90 

The relevant code is in each of the ConfigRead routines (one for each data type but aliased to the generic read). In POP, there is a single namelist input file (pop_in) that contains all of the namelists associated with each module and each module reads its own config namelist from that combined file. So when the routine refers to module name or module found, that is the same as the namelist name and finding the proper namelist in the file.

mgduda:
Having a look through the POP configuration-reading code was informative. However, being rather lazy, and having fixed-up this PR so that it has no impact on the behavior of namelist reading with existing compiler options, I'd propose to keep this PR as-is.

matthewhoffman:
@mgduda , that's fine with me.

mark-petersen (approved):
I tested this merged into the ocean core with gnu 5 and intel 19, both debug and optimized, and this passes the nightly regression suite on all. This does not alter the ability to change flag order on a namelist, so it all looks good.

mark-petersen:
@philipwjones or @amametjanov do we use NAG for E3SM? If so, would you like to test this PR?

mgduda:
There's no particular rush, but I just wanted to see whether @philipwjones or @amametjanov wanted to test with NAG and E3SM?

@matthewhoffman I've amended the PR so that namelist groups can be in any order (as before) for any compiler except the NAG compiler. The inability to have out-of-order namelist groups with NAG isn't terribly detrimental for us, as we primarily intend to use the NAG compiler for debugging and for spotting non-standard code that other compilers accept.

philipwjones:
@mgduda - Sorry, but E3SM has actually been trying to prune the number of supported compilers, so we don't have any interest in testing/supporting NAG within E3SM builds unless there is a compelling need for any of our machines. If the others find it works in standalone mode, that's fine for this PR.

matthewhoffman (approved):
Approved based on inspection and discussion.

PR #724, opened 2020-10-20
1km Delaware Bay hurricane case with synthetic winds and vegetation

sbrus89:
This sets up a 1km resolution Delaware Bay hurricane run in COMPASS. It uses a parametric wind model based off of track information and the formula to calculate manning's n bottom drag from vegetation properties.

### DIFF
--- a/src/core_ocean/mode_init/Registry_hurricane.xml
+++ b/src/core_ocean/mode_init/Registry_hurricane.xml
@@ -59,4 +59,8 @@
 			description="Value of channel drag below config_marsh_z_limit for either Cd or Manning's n."
 			possible_values="Any real number."
 		/>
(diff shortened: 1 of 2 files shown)


sbrus89:
@mark-petersen, this is COMPASS only with the exception of a new config variable for the hurricane case init mode.

mark-petersen:
@sbrus89 should this be merged? If so I can separate out the compass changes and make a PR to the new compass repo. The init mode changes would remain in here.

mark-petersen:
I changed this to just the init mode piece. I'll make another PR to compass with those parts.

mark-petersen (approved):
Init mode only. Compiled, tested with regression suite (doesn't make this case). Approved by visual inspection.

PR #726, opened 2020-10-21
Adds saving of GM induced MOC streamfunction

vanroekel:
This PR adds the ability to compute and separately save the MOC stream function due to GM bolus velocity.  This is saved separately from the MOC from the resolved velocity.

### DIFF
--- a/src/core_ocean/analysis_members/mpas_ocn_moc_streamfunction.F
+++ b/src/core_ocean/analysis_members/mpas_ocn_moc_streamfunction.F
@@ -470,7 +470,7 @@ contains
 
          call mpas_pool_get_dimension(block % dimensions, 'nCellsSolve', nCellsSolve)
 
-         call mpas_pool_get_array(diagnosticsPool, config_AM_mocStreamfunction_vertical_velocity_value, vertVelocityTop)
+         call mpas_pool_get_array(diagnosticsPool, 'vertVelocityTop', vertVelocityTop)
          call mpas_pool_get_array(meshPool, 'latCell', latCell)
          call mpas_pool_get_array(meshPool, 'areaCell', areaCell)
 
@@ -482,11 +482,7 @@ contains
          call mpas_pool_get_array(transectPool,'transectEdgeMaskSigns',transectEdgeMaskSigns)
          call mpas_pool_get_array(transectPool,'transectEdgeMasks',transectEdgeMasks)
          call mpas_pool_get_subpool(block % structs, 'state', statePool)
-         if (trim(config_AM_mocStreamfunction_normal_velocity_value)=="normalVelocity") then
-            call mpas_pool_get_array(statePool, trim(config_AM_mocStreamfunction_normal_velocity_value), normalVelocity, timeLevel)
-         else
-            call mpas_pool_get_array(diagnosticsPool, trim(config_AM_mocStreamfunction_normal_velocity_value), normalVelocity, timeLevel)
-         end if
+         call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocity, timeLevel)
          call mpas_pool_get_array(meshPool, 'cellsOnEdge', cellsOnEdge)
          call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
          call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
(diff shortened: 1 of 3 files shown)


vanroekel:
Here are a couple figures from a 1 month G-case test

Here is the time monthly from a one month test of AMOC of resolved velocities
Screen Shot 2020-10-21 at 1 54 35 PM

Here is that from GM bolus velocity
Screen Shot 2020-10-21 at 1 54 40 PM

vanroekel:
@xylar @milenaveneziani @maltrud I'd like to remove all occurrences of 
```
config_AM_mocStreamfunction_normal_velocity_value
```
and
```
config_AM_mocStreamfunction_vertical_velocity_value
```
These are never altered and if this capability seems useful to you all it doesn't seem necessary to specify the velocity field.  What do you think?

xylar:
I agree, remove those config options.

vanroekel:
@maltrud @milenaveneziani @mark-petersen I think this is ready for your review.

milenaveneziani:
Thanks @vanroekel for doing this.
I guess the only part that I can't figure out from the code is at what point the GM streamfunction component is added to the previously computed non-GM part: what am I missing?

vanroekel:
@milenaveneziani it's actually not added anywhere.  I'm outputting two stream function fields that can be added offline, but then allow flexibility of plotting both.

milenaveneziani:
>  I'm outputting two stream function fields that can be added offline, but then allow flexibility of plotting both.

Couldn't that be a bit misleading though? From the variable name, at least, one would think that the first streamfunction is the total value.
I would vote for adding them together at this point, to avoid confusion, while of course still outputting the GM part separately. @maltrud, @xylar: what is your opinion on this?

vanroekel:
@milenaveneziani I see what you are saying.  I'm more than happy to add them together, it would be a very easy change.  @maltrud or @xylar any objections to that?

maltrud:
i think we need 2 streamfunctions, but i don't think it matters too much which ones.  either eulerian mean and EM+bolus, or EM and bolus separately.  does that answer the question?

src/core_ocean/analysis_members/mpas_ocn_moc_streamfunction.F
```diff
@@ -559,6 +559,113 @@ subroutine ocn_compute_moc_streamfunction(domain, timeLevel, err)!{{{
...
+       call mpas_pool_get_subpool(domain % blocklist % structs, 'mocStreamfunctionAM', mocStreamfunctionAMPool)
+       call mpas_pool_get_array(mocStreamfunctionAMPool, 'mocStreamvalLatAndDepthGM', mocStreamvalLatAndDepthGM)
+       mocStreamvalLatAndDepthGM = mocStreamvalLatAndDepthTotal * m3ps_to_Sv
+
+       call mpas_pool_get_array(mocStreamfunctionAMPool, 'mocStreamvalLatAndDepthRegionGM', mocStreamvalLatAndDepthRegionGM)
+       mocStreamvalLatAndDepthRegionGM = mocStreamvalLatAndDepthRegionTotal * m3ps_to_Sv
+
+       !Add GM bolus contribution to resolved streamfunction to create total streamfunction
+       mocStreamvalLatAndDepthRegion = mocStreamvalLatAndDepthRegion + mocStreamvalLatAndDepthRegionGM
+       mocStreamvalLatAndDepth = mocStreamvalLatAndDepth + mocStreamvalLatAndDepthGM
```
vanroekel, line 665: @milenaveneziani I've added these two lines to create a total overturning stream function

xylar:
@vanroekel, I agree that @milenaveneziani's suggestion is more intuitive.  Then, you can always subtract off the bolus MOC to get the Eulerian MOC if you want that.  @milenaveneziani, thanks for being on top of this!

vanroekel:
Sounds good.  I've added this in the last commit.

src/core_ocean/analysis_members/mpas_ocn_moc_streamfunction.F
milenaveneziani: great, thanks @vanroekel

milenaveneziani (approved):

maltrud (approved):

mark-petersen (approved):
This looks good. I ran a quick test with analysis on using debug with intel and gnu.

@vanroekel this should be merged today and a PR created for E3SM V2, correct? If so, I can do that.

vanroekel:
yes, that would be great.  For E3SM, please request adding
```
<var name="mocStreamvalLatAndDepthGM"/>
<var name="mocStreamvalLatAndDepthRegionGM"/>
```
to default monthly average output for all meshes except RRS.  It would also be good to fully switch on the mocAM for E3SM runs.  It should work now.

src/core_ocean/analysis_members/mpas_ocn_moc_streamfunction.F
```diff
@@ -125,7 +125,7 @@ subroutine ocn_init_moc_streamfunction(domain, err)!{{{
       integer :: currentRegion, i, iCell
 
       !! region arrays/variables
-      character (len=STRKIND), dimension(:), pointer :: regionGroupNames
+      character (len=ShortStrKIND), dimension(:), pointer :: regionGroupNames
```
mark-petersen, line 128: @vanroekel why is this changed to `ShortStrKIND`?  It makes compile fail at
```
mpas_ocn_moc_streamfunction.F:150:80:
       call mpas_pool_get_array(regionPool, 'regionGroupNames', regionGroupNames)
```
because this subroutine only supports STRKIND. When I change it back it compiles, but I wonder if you changed it for another reason.

vanroekel: I thought this was needed as the regionGroupName variable in the MOC transect file was ShortStrKINd (64) and the routine wasn't reading that field.  Can you try a test with the AM on and reverted?  

Related, what do you mean it only supports STRKND?  I was able to run this in e3sm with that setting with no issue.  Does it only come up in debug or something?

mark-petersen: This PR compiles and runs with intel 18 (debug and opt) but produces that compile error with gnu 5.3 on grizzly (debug and opt).

I don't see anywhere that shortStrKind is used in the ocean core. On `ocean/develop`:
```
cd src/core_ocean
git grep -in shortstrkind
```
produces nothing, and 
```
git grep -in regionGroupName|grep character
analysis_members/mpas_ocn_meridional_heat_transport.F:132:      character (len=STRKIND), dimension(:), pointer :: regionGroupNames
analysis_members/mpas_ocn_meridional_heat_transport.F:344:      character (len=STRKIND), dimension(:), pointer :: regionNames, regionGroupNames
analysis_members/mpas_ocn_moc_streamfunction.F:128:      character (len=STRKIND), dimension(:), pointer :: regionGroupNames
analysis_members/mpas_ocn_moc_streamfunction.F:358:      character (len=STRKIND), dimension(:), pointer :: regionNames, regionGroupNames
analysis_members/mpas_ocn_water_mass_census.F:208:      character (len=STRKIND), dimension(:), pointer :: regionNames, regionGroupNames
```
shows everything is `STRKIND`. I'll go ahead and change that one back. 

BTW, I compiled E3SM on cori, and both gnu and intel compiled fine, so it must be version dependent. 

What I meant above is that the general interface for `mpas_pool_get_array` only supports two character versions:
```
cd src/framework
git grep -in mpas_pool_get_array|grep subrou|grep '{{{'|grep char
mpas_pool_routines.F:4480:   subroutine mpas_pool_get_array_0d_char(inPool, key, string, timeLevel)!{{{
mpas_pool_routines.F:4510:   subroutine mpas_pool_get_array_1d_char(inPool, key, array, timeLevel)!{{{
```
and in both of those, the string is defined as a `strKind`:
```
4486       character (len=StrKIND), pointer :: string
4516       character (len=StrKIND), dimension(:), pointer :: array
```
So gnu 5.3 is correct to complain about it.

vanroekel:
@mark-petersen thanks for explaining.  I see now.  Can you confirm that the model still reads RegionGroupName with use of StrKND?  I'm assuming it does, but would be good to verify the log file isn't full of this errors about not being able to read regionGroupNames.

mark-petersen:
Yes, confirmed that moc reads RegionGroupName with use of StrKND. Merging...

PR #728, opened 2020-10-23
Ocean/split explicit corrections

mark-petersen:
This PR corrects the following issues in the Split Explicit time-stepping routine:

1. Re-split baroclinic/barotropic velocity at the beginning of every time step
2. Set momentum velocity (`normalVelocity`) to:
    u = F/H + u' - btrAvg( u' )
3. Set transport velocity (`normalTransportVelocity`) to:
    uTr = F/H + u' - btrAvg( u' ) + uBolus - btrAvg( uBolus )

where:
F is the accumulated barotropic flux
H is the total column depth, including SSH, from the last time known.
u' is the baroclinic velocity from stage 1

Previously, `normalVelocity` could deviate from `normalTransportVelocity` and become noisy. Since `normalVelocity` does not feed back into the pressure gradient, there was no control on this noise. The previous version also used the `velocityCorrection` variable, which was extremely noisy, but also misleading, because it was really just the difference between `normalVelocity` and `normalTransportVelocity`. There is no need for `velocityCorrection` when the velocities are built directly, as shown above.


mark-petersen:
This work was originally done in 2017 in the old private repo, with a PR here: https://github.com/MPAS-Dev/MPAS/pull/1388. It currently compiles and runs, but is not bfb across partitions or threading.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -323,18 +325,20 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
          !$omp parallel
          !$omp do schedule(runtime) private(k)
```
mattdturner, line 326: `thicknessSum` needs to be private here

mark-petersen: Thanks, that solved the threading problem.

src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -1404,72 +1398,48 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
...
-               !{\bf u}^{corr} = \left( {\overline {\bf F}}
-               !  - \sum_{k=1}^{N^{edge}} h_{k,*}^{edge}  {\bf u}_k^{avg} \right)
-               ! \left/ \sum_{k=1}^{N^{edge}} h_{k,*}^{edge}   \right.
-
-               if (config_vel_correction) then
-                  useVelocityCorrection = 1
-               else
-                  useVelocityCorrection = 0
-               endif
+               nEdges = min(nEdgesPtr, nEdgesArray( config_num_halos + 1 ))
```
mark-petersen, line 1407: This solved the bfb partition problem. It needed one wider halo layer.

mark-petersen:
@vanroekel this now passes all MPAS-O tests. I suspect it won't make a big difference, but it is a much cleaner method of splitting. If you could test it this week, that would be great.

vanroekel:
@mark-petersen yes of course, will do.  Is it okay if I just drop in the files?  Or can you make an e3sm branch?  I'm happy to just pull in the diffs from this PR into my e3sm branch if that's okay with you.

mark-petersen:
@vanroekel I just rebased on the current `ocean/develop` so you can merge or copy the two files within E3SM. Thanks a lot.

mattdturner (approved):
Approved based on visual inspection and @mark-petersen testing

vanroekel:
@mark-petersen I'm running this in e3sm now, I've run 6 months so far.  These changes do make a difference, especially in the shallower ocean.  It's too soon to fully quantify the impact, but it is looking like it is going in the right direction with these changes.

xylar:
@mark-petersen, I just saw the commit about bottom drag.  Is equivalent code needed for top drag under ice shelves?

vanroekel:
@mark-petersen please hold on this PR. I was looking at more depth at my G-case with these changes.  It makes a bigger difference than I saw after a few years.  Below are plots of AMOC from this PR.  You can see the deep overturning is very strong

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.vanroekel/20201025_JRA_splitExplicit_EC30to60r2_yrs/ocean/index.html#&gid=29&pid=2

I'd like to run this longer to see if it is just an initial condition issue.  I'll post again after it runs 30 years.

mark-petersen:
@xylar I added commit 761177b719 here by mistake. That added bottom drag to the Stage 2 barotropic subcycling, which we should consider (along with wind forcing) but this PR is not the right place to do it. 

@vanroekel can you make sure your test is running with the current PR head 10b3d3f, and not with 761177b? Yes, I will hold on this until I hear further.

mark-petersen:
@vanroekel I think this is the correct way to do the splitting. In this PR we re-compute the barotropic at the beginning of every time step. In E3SM we currently carry the barotropic mode from one time step to the next, which is a poor design.

On your [comment above](https://github.com/MPAS-Dev/MPAS-Model/pull/728#issuecomment-720524981) you said the AMOC's bottom branch is too strong. Now that we have more time, I wonder if we should reevaluate this change. Maybe it would help with the low AMOC problems. Do you remember if your results above were due to the initial condition?

If you agree that we should consider this PR again, it would be after https://github.com/MPAS-Dev/MPAS-Model/pull/781, which will merge in a week or two.

vanroekel:
@mark-petersen yes lets test this again.  I'd like to double check and do another test after all the other PRs are in.  Revisiting this in a week or two sounds good.

mark-petersen:
Agreed. There have been major performance revisions since I made this PR, and formatting in #781, so let's look at again in a few weeks when I can rebase this PR.

PR #729, opened 2020-10-23
Add support for building with NAG compilers (atmosphere/cam)

mgduda:
This PR adds a build target for the NAG compilers to the top-level Makefile, and
it also fixes several compilation issues with the NAG compilers in shared code.

Although the 'test' core can now be compiled with the NAG compilers, other cores
contain issues in their core-specific code that prevent compilation, and this PR
does not address those issues.

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -645,7 +645,11 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 		}
 
 		fortprintf(fd, "      if (dminfo %% my_proc_id == IO_NODE) then\n");
+		fortprintf(fd, "! Rewinding before each read leads to errors when the code is built with\n");
+		fortprintf(fd, "! the NAG Fortran compiler. If building with NAG, be kind and don't rewind.\n");
+		fortprintf(fd, "#ifndef NAG_COMPILER\n");
 		fortprintf(fd, "         rewind(unitNumber)\n");
+		fortprintf(fd, "#endif\n");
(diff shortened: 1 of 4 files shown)


mgduda:
This PR is identical to PR #722 , except that it targets the `atmosphere/cam` branch.

src/tools/registry/gen_inc.c
```diff
@@ -645,7 +645,11 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 		}
 
 		fortprintf(fd, "      if (dminfo %% my_proc_id == IO_NODE) then\n");
+		fortprintf(fd, "! Rewinding before each read leads to errors when the code is built with\n");
+		fortprintf(fd, "! the NAG Fortran compiler. If building with NAG, be kind and don't rewind.\n");
```
MiCurry, line 649: 😒 ...

I'm guessing the NAG compiler rewinds automatically once read automatically reaches the end of a record?

MiCurry:
I don't believe I have access to a NAG compiler, so I wont be able to test to see if this branch compiles or not.

MiCurry (approved):

src/tools/registry/gen_inc.c
mgduda: No, it appears that the NAG compiler expects the file to be positioned at the beginning of the next namelist group (possibly preceded by comments); if we rewind the file, this isn't true. For example, if we have the following `namelist.atmosphere` file:
```
&foo
    a = 2.71828
/
&blah
    i = 42
/
```
rewinding before trying to read the `&blah` group results in an error.

MiCurry:
I was able to successfully compile the `test_core` of this branch with the NAG compiler.

PR #734, opened 2020-10-28
Use an explicit format when converting dt to a string in mpas_set_timeInterval

mgduda:
This merge switches to the use of an explicit format string when converting `dt`
to a string in the `mpas_set_timeInterval` routine.

If the real-valued `dt` optional argument is given to the `mpas_set_timeInterval`
routine, the value of this variable is written into a string using an internal
write statement. Before this commit, list-directed output was used in this write
(i.e., the format was `*`), which permits the compiler to choose either the 'f'
or 'e' edit descriptors. If the 'e' edit descriptor was chosen, the value of `dt`
would be incorrectly converted to a fractional seconds value.

For example, if dt = 600.0, the NAG compiler generates the `timeString_`

00:00:   6.0000000000000000E+02

which lead to an interval of 6 seconds rather than 600 seconds.

The solution adopted in this commit is to use an explicit 'f' edit descriptor
in a format string when writing the `timeString_` string.

Although the Fortran 2003 standard permits an 'f' edit descriptor with a zero
width, e.g., `f0.9`, the use of a zero width leaves the choice of whether to print
a leading zero up to the compiler. In the case of the GNU (and possibly other)
compilers, a `dt` value of, say, 0.1 results in a `timeString_` of

00:00:.100000000

Parsing of the whole seconds part of `timeString_` then leads to errors.

To avoid leaving choices of how to format the `dt` value up to the compiler, this
commit dynamically builds a format string that uses the 'f' edit descriptor.

### DIFF
--- a/src/framework/mpas_timekeeping.F
+++ b/src/framework/mpas_timekeeping.F
@@ -1334,6 +1334,9 @@ module mpas_timekeeping
       character (len=StrKIND) :: timeSubString
       character (len=StrKIND) :: secDecSubString
       character(len=StrKIND), pointer, dimension(:) :: subStrings
+      character(len=16) :: fmtString
+      integer :: iwidth
+      integer :: idecimals
 
 !      if (present(DD)) then
 !         days = DD
@@ -1403,8 +1406,29 @@ module mpas_timekeeping
       if (present(timeString) .or. present(dt)) then
 
 
-         if(present(dt)) then
-            write (timeString_,*) "00:00:", dt         
+         if (present(dt)) then
+            !
+            ! Before writing dt into a timeString, first construct an appropriate format string
+            !
+
+            ! Number of decimal places of precision (9 = nanosecond precision)
+            idecimals = 9
+
+            ! Scale total width of representation based on max(log10(dt),0.0)
+            ! (+2 for at least a leading zero and a '.')
+            if (dt /= 0.0_RKIND) then
+               iwidth = int(max(log10(abs(dt)),0.0_RKIND)) + idecimals + 2
+            else
+               iwidth = idecimals + 2
+            end if
+
+            ! Add an extra character for a minus sign if needed
+            if (dt < 0.0_RKIND) then
+               iwidth = iwidth + 1
+            end if
+
+            write(fmtString, '(a,i2.2,a,i2.2,a)') '(a,f', iwidth, '.', idecimals, ')'
+            write(timeString_,trim(fmtString)) '00:00:', dt
          else
             timeString_ = timeString
          end if


mgduda:
If you'd like a test program to experiment with, the following mirrors the code change in this PR:
```fortran
program foobar

   implicit none

   integer, parameter :: RKIND = selected_real_kind(12)

   real(kind=RKIND) :: dt
   integer :: idecimals
   integer :: iwidth
   character(len=16) :: fmtString
   character(len=512) :: timeString

   read(5,*) dt

   !
   ! Before writing dt into a timeString, first construct an appropriate format string
   !

   ! Number of decimal places of precision (9 = nanosecond precision)
   idecimals = 9

   ! Scale total width of representation based on max(log10(dt),0.0)
   ! (+2 for at least a leading zero and a '.')
   if (dt /= 0.0_RKIND) then
      iwidth = int(max(log10(abs(dt)),0.0_RKIND)) + idecimals + 2
   else
      iwidth = idecimals + 2
   end if

   ! Add an extra character for a minus sign if needed
   if (dt < 0.0_RKIND) then
      iwidth = iwidth + 1
   end if

   write(fmtString, '(a,i2.2,a,i2.2,a)') '(a,f', iwidth, '.', idecimals, ')'
   write(timeString,trim(fmtString)) '00:00:', dt
   write(timeString,'(a,f0.9)') '00:00:', dt

   write(0,*) timeString

end program foobar
```

matthewhoffman:
@mgduda , thank you for the test program!  That greatly lowered the bar to me actually looking carefully at this.  I did discover some behavior that I'm not sure if is correct or not.  Some examples:

Do we want a leading 0 (or actually two 0s here)?
```
$ ./a.out
5.6456464565465465e-2
 00:00:.056456465
```

Same as previous, but I'm also not sure what a negative in the seconds place of a timestamp even means.
```
$ ./a.out
-5.4324234e-2
 00:00:-.054324234
```

Like first example, do we want a leading 0 in the 10s place of the seconds?  (like we have for minutes and hours)
```
$ ./a.out
5
 00:00:5.000000000
```

I confess to not being particularly familiar with what this routine does.  Do we want the decimal part to be dropped if the value is a whole number?  Has this routine always returned decimal seconds?  (I was under the impression the MPAS timekeeping did not support decimal seconds.)

mark-petersen (approved):
This all works fine for me in the ocean core, and we call this routine with dt requested. I tested with both intel and gnu, passes the nightly regression suite.

mgduda:
@matthewhoffman Sorry about that -- I apparently forgot to remove a line from the test program:
```fortran
write(timeString,'(a,f0.9)') '00:00:', dt
```
If you remove this line (the third to last line), the results of the test program should match what we get in the code change. I was trying to see whether letting the compiler choose the width (e.g., with `f0.9` format) would simplify the code, but I found that some compilers omit the leading zero.

I could update the PR to include a tens place, but I don't think this is important. Ultimately, the time string is parsed into hours, minutes, and seconds and fractional seconds before creating an ESMF time interval instance. For example, the integer part of the interval's seconds are read into an integer variable using list-directed input in one of several locations: [L.1485](https://github.com/MPAS-Dev/MPAS-Model/blob/v7.0/src/framework/mpas_timekeeping.F#L1485), [L.1490](https://github.com/MPAS-Dev/MPAS-Model/blob/v7.0/src/framework/mpas_timekeeping.F#L1490), or [L.1495](https://github.com/MPAS-Dev/MPAS-Model/blob/v7.0/src/framework/mpas_timekeeping.F#L1495).

mgduda:
@matthewhoffman To answer your question about what a negative in the seconds place means, since the string that is built describes a time interval rather than a timestamp, it simply means that the seconds contribute negatively to the interval, I think. It's probably best not to place too much weight on the internally built string, since it is later parsed into hours, minutes, and seconds; in essence, the string is just an intermediate representation on the way to an ESMF time interval.

mgduda:
@philipwjones I added you as a reviewer in case you had any thoughts on better ways to get to the required formatting of a real value more cleanly, but please feel free to remove yourself if you'd prefer.

matthewhoffman (approved):

MiCurry (approved):

PR #736, opened 2020-10-29
Fix non-standard BOZ-literal integer constant in module_ra_cam_support

mgduda:
This merge corrects a non-standard BOZ-literal constant in module_ra_cam_support.

The GNU 10.1.0 compiler identified a non-standard BOZ-literal constant in
module_ra_cam_support.F, failing to compile this file with the following error:
```
  module_ra_cam_support.F:11:35:

     11 |       integer, parameter:: bigint = O'17777777777'           ! largest possible 32-bit integer
  Error: BOZ literal constant at (1) is neither a data-stmt-constant nor an actual argument to INT, REAL, DBLE, or CMPLX intrinsic function [see ‘-fno-allow-invalid-boz’]
```
As per the Fortran 2003 standard, this constant should be typed with the INT
intrinsic (or be used in a data statement):
```
  4.4.1 Integer type

  C410 (R411) A boz-literal-constant shall appear only as a data-stmt-constant in
  a DATA statement, as the actual argument associated with the dummy argument A of
  the numeric intrinsic functions DBLE, REAL or INT, or as the actual argument
  associated with the X or Y dummy argument of the intrinsic function CMPLX.
```
This commit uses the INT intrinsic with an argument of O'17777777777' to
initialize the bigint parameter.

### DIFF
--- a/src/core_atmosphere/physics/physics_wrf/module_ra_cam_support.F
+++ b/src/core_atmosphere/physics/physics_wrf/module_ra_cam_support.F
@@ -8,7 +8,7 @@ MODULE module_ra_cam_support
   implicit none
       integer, parameter :: r8 = 8
       real(r8), parameter:: inf = 1.e20 ! CAM sets this differently in infnan.F90
-      integer, parameter:: bigint = O'17777777777'           ! largest possible 32-bit integer 
+      integer, parameter:: bigint = int(O'17777777777')      ! largest possible 32-bit integer 
 
       integer :: ixcldliq 
       integer :: ixcldice


src/core_atmosphere/physics/physics_wrf/module_ra_cam_support.F
```diff
@@ -8,7 +8,7 @@ MODULE module_ra_cam_support
   implicit none
       integer, parameter :: r8 = 8
       real(r8), parameter:: inf = 1.e20 ! CAM sets this differently in infnan.F90
-      integer, parameter:: bigint = O'17777777777'           ! largest possible 32-bit integer 
+      integer, parameter:: bigint = int(O'17777777777')      ! largest possible 32-bit integer
```
MiCurry, line 11: Would using the `huge` intrinsic be better here? GNU, Intel, PGI and Nag are cool with the following:
```
integer, parameter :: bigint = huge(bigint)
```

ldfowler58:
Hi Michael and Miles:
I do not think that I can really help you as it is a compiler related issue. Better leave it to software engineers. Let me check where bigint is used in MPAS-Model.

src/core_atmosphere/physics/physics_wrf/module_ra_cam_support.F
mgduda: That is a good suggestion and something that I had considered. However, I don't know that `huge(bigint)` will be equivalent to `int(O'17777777777')` on every machine (it depends on the machine's representation of integers, and I would agree that what the code is doing now is completely not portable), and I was concerned about inadvertently changing the value of `bigint`. In the end, I considered it better to modify the code as minimally as possible while getting it to comply with the Fortran 2003 standard.

MiCurry: That's what I figured your rational was, sounds good to me!

MiCurry (approved):

mgduda:
@ldfowler58 Since this change modifies code in a physics scheme, I had added you as a reviewer. The code change does not alter the results, but simply expresses a constant in a way that is standards-conformant. If you have no objections, that would be good enough to count as an approval, I think. But, please don't hesitate to let me know if you have any questions or concerns about the code change.

PR #737, opened 2020-10-29
Fix cmake for tool exes

jgfouca:
Regardless of MPAS_TOOL_DIR, the tool exes need to be targets
and targets that depend of these exes need to always depend on the
needed exe. This will ensure the exes get built regardless of which
targets the user is building.

I tested this change with the E3SM case ne4pg2_r05_oQU480 A_WCYCL1850. I tested both `case.build` and `case.build --separate-builds` with both MPAS_TOOL_DIR enabled and disabled.

### DIFF
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -1,9 +1,18 @@
 
 if (DEFINED ENV{MPAS_TOOL_DIR})
   message(STATUS "*** Using MPAS tools from $ENV{MPAS_TOOL_DIR} ***")
-  file(COPY $ENV{MPAS_TOOL_DIR}/input_gen/namelist_gen DESTINATION .)
-  file(COPY $ENV{MPAS_TOOL_DIR}/input_gen/streams_gen  DESTINATION .)
-  file(COPY $ENV{MPAS_TOOL_DIR}/registry/parse         DESTINATION .)
+  add_custom_target(namelist_gen)
+  add_custom_command(
+    TARGET namelist_gen PRE_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy $ENV{MPAS_TOOL_DIR}/input_gen/namelist_gen ${CMAKE_CURRENT_BINARY_DIR}/namelist_gen)
+  add_custom_target(streams_gen)
+  add_custom_command(
+    TARGET streams_gen PRE_BUILD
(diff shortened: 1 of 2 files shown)


jgfouca:
@amametjanov @mark-petersen

rljacob:
Can a test be added for the cmake build within MPAS-Dev ?

src/tools/CMakeLists.txt
```diff
@@ -1,9 +1,18 @@
...
+  add_custom_target(streams_gen)
+  add_custom_command(
+    TARGET streams_gen PRE_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy $ENV{MPAS_TOOL_DIR}/input_gen/streams_gen ${CMAKE_CURRENT_BINARY_DIR}/streams_gen)
+  add_custom_target(parse)
+  add_custom_command(
+    TARGET parse PRE_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy $ENV{MPAS_TOOL_DIR}/input_gen/parse ${CMAKE_CURRENT_BINARY_DIR}/parse)
 else()
   message(STATUS "*** Building MPAS tools from source ***")
```
amametjanov, line 17: This change is needed by `./case.build --separate-builds` (aggregated`./case.build` [without separate builds] works fine).

It's not clear why `src/tools/CMakeLists.txt` is not getting picked up while building `ice`: messages are getting logged into e3sm.bldlog, instead of ice.bldlog. But the calls to cmake's `add_custom_target` generate explicit dependencies. Okay.

amametjanov (approved):

mark-petersen:
@mgduda @matthewhoffman @akturner FYI, this changes cmake files used by E3SM only.

mark-petersen:
@jgfouca I just started a few tests on cori. Should I go ahead and merge after that?

mark-petersen:
Passes E3SM tests:
```
```

jgfouca:
@mark-petersen , yes, this is ready to merge.

mark-petersen (approved):

jgfouca:
@mark-petersen , thanks for merging this. What's the process for requesting an MPAS submodule update in E3SM?

mark-petersen:
Done. See https://github.com/E3SM-Project/E3SM/pull/3927

Issue #732, opened 2020-10-27
Bugs in tracer advection code


vanroekel:
There are a few issues in the tracer advection code, both _mono and _std.

(1) If we look at vertOrder3 here

https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F#L663-L695

and compare this with Skamarock and Gassmann (2011), their equation 11
![Screen Shot 2020-10-27 at 12 55 12 PM](https://user-images.githubusercontent.com/13662545/97348382-adabc600-1853-11eb-8e7f-54fb7c186399.png)

there is a sign error on the term multiplying coef3rdOrder.  This occurs here and in vflux3 in the framework

(2) the third order and fourth order fluxes are not weighted by thickness.  This could lead to lower than expected accuracy in vertical advection

mark-petersen:
On (2), see [comment on PR](https://github.com/MPAS-Dev/MPAS-Model/pull/738#discussion_r514760374). It is a linear interpolation, not a weighted average, so original is correct.

vanroekel:
For both the original formulation is correct.  Closing this issue.  See discussion in #738 for further information.

PR #738, opened 2020-10-29
Fixes to tracer advection mono

vanroekel:
This PR fixes in part #732, there is a sign error in vertOrder3 advection and the second order vertical weights are multiplying the wrong level of temperature.

These fixes show positive impacts (shallowing of MLD, especially in BCE regions, Kuroshio, Gulf Stream) in G-cases.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
@@ -682,7 +682,7 @@ module ocn_tracer_advection_mono
             do k=3,kmax-1
               highOrderFlx(k, iCell) = (w(k,iCell)* &
                    (7.0_RKIND*(tracerCur(k  ,iCell) + tracerCur(k-1,iCell)) - &
-                              (tracerCur(k+1,iCell) + tracerCur(k-2,iCell))) - &
+                              (tracerCur(k+1,iCell) + tracerCur(k-2,iCell))) + &
                         coef3rdOrder*abs(w(k,iCell))* &
                              ((tracerCur(k+1,iCell) - tracerCur(k-2,iCell)) - &
                     3.0_RKIND*(tracerCur(k  ,iCell) - tracerCur(k-1,iCell))))/ &
--- a/src/operators/mpas_tracer_advection_helpers.F
+++ b/src/operators/mpas_tracer_advection_helpers.F
@@ -69,7 +69,7 @@ module mpas_tracer_advection_helpers
         real (kind=RKIND), intent(in) :: w !< Input: vertical veloicity
         real (kind=RKIND), intent(in) :: coef !< Input: Advection coefficient
 
-        mpas_tracer_advection_vflux3 = (w * (7.0_RKIND * (q_i + q_im1) - (q_ip1 + q_im2)) - coef * abs(w) * ((q_ip1 - q_im2) - 3.0_RKIND*(q_i-q_im1)))/12.0_RKIND
+        mpas_tracer_advection_vflux3 = (w * (7.0_RKIND * (q_i + q_im1) - (q_ip1 + q_im2)) + coef * abs(w) * ((q_ip1 - q_im2) - 3.0_RKIND*(q_i-q_im1)))/12.0_RKIND
    end function!}}}
 
 !|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||


vanroekel:
@mark-petersen There is also a bug in tracer_vflux3 in the framework, which affects tracer_advection_std.  Should I fix that here too, I did not as that is in framework, but looking again it affects tracer_advection_mono too.  It is an easy fix.

vanroekel:
@ambrad I can't seem to add you as a reviewer, but please look over and make sure I made corrections right.

ambrad:
@vanroekel yes, I will review by tonight.

As long as we discussed this in some detail, re: the sign fix, here is some documentation relating the code to SG11:
```
equation 7 in SG11:
    F(u,p)[i+1/2] = u[i+1/2] (1/2 (p[i+1] + p[i]) - 1/12 (dx2 p[i+1] + dx2 p[i]) +
                              sign(u) beta/12 (dx2 p[i+1] - dx2 p[i])),
where
    dx2 p[i] = p[i+1] - 2 p[i] + p[i-1]
expand the first using the second:
    F(u,p)[i+1/2] = u[i+1/2] (1/2 (p[i+1] + p[i]) - 1/12 ((p[i+2] - 2 p[i+1] + p[i]) + (p[i+1] - 2 p[i] + p[i-1])) +
                              sign(u) beta/12 ((p[i+2] - 2 p[i+1] + p[i]) - (p[i+1] - 2 p[i] + p[i-1])))
    = u[i+1/2] (6/12 (p[i+1] + p[i]) - 1/12 ((p[i+2] - 2 p[i+1] + p[i]) + (p[i+1] - 2 p[i] + p[i-1])) +
                sign(u) beta/12 ((p[i+2] - 2 p[i+1] + p[i]) - (p[i+1] - 2 p[i] + p[i-1])))
    = u[i+1/2] (7/12 (p[i+1] + p[i]) - 1/12 (p[i+2] + p[i-1]) +
                sign(u) beta/12 ((p[i+2] - p[i-1]) - 3 (p[i+1] - p[i])))
then map i -> k-1 to match the code:
    F(u,p)[k-1/2] = u[k-1/2] (7/12 (p[k] + p[k-1]) - 1/12 (p[k+1] + p[k-2]) +
                              sign(u) beta/12 ((p[k+1] - p[k-2]) - 3 (p[k] - p[k-1])))
finally, deal with u, abs(u), sign(u):
    F(u,p)[k-1/2] = u[k-1/2] (7/12 (p[k] + p[k-1]) - 1/12 (p[k+1] + p[k-2])) +
                    sign(u) u[k-1/2] beta/12 ((p[k+1] - p[k-2]) - 3 (p[k] - p[k-1]))
                  = u[k-1/2] (7/12 (p[k] + p[k-1]) - 1/12 (p[k+1] + p[k-2])) +
                    abs(u[k-1/2]) beta/12 ((p[k+1] - p[k-2]) - 3 (p[k] - p[k-1]))
```

Re: the vflux3 function, just so this to-do fix is documented, we need to put the sign fix into src/operators/mpas_tracer_advection_helpers.F, subroutine mpas_tracer_advection_vflux3.

vanroekel:
@ambrad thanks for pasting in that part of our slack discussion.  Talking with @mark-petersen I'm going to make the vflux3 change in this PR and unify the ocn_advection routines to call that routine instead of a remade version in the ocean core.  I'll let you know when it is ready, likely not till later tonight.

philipwjones:
@vanroekel  I purposely in-lined those vertical advection routines recently for performance reasons. I know that duplicates code, but it does make a difference.

vanroekel:
@philipwjones okay thanks that makes a ton of sense  I realize I was mistaken I thought there were calls to vflux3 in advection_mono, but was mistaken (likely looking at develop).  I'll fix advection-std to be like this one

@ambrad okay, but I will say the changes are very positive in a G-case right now, but I will be very interested to see what you find.

ambrad:
A. @vanroekel are you testing vertOrder2 or just vertOrder3?

B. The proposed verticalWeightK(m1) modifications occur in two spots: the interior if vertOrder2 is used, and the boundaries regardless of order. Thus, if you are testing just vertOrder3, I believe the sign fix for vertOrder3 will dominate the signal in the G-case; it won't tell us if the other modifications are right. On the other hand, if you're testing vertOrder2 in a G-case, then we _do_ have some signal for those changes.

However, my testing suggests these modifications are not correct. (Of course, I may be wrong.) If I use the current settings, I see this convergence plot in my standalone 1D code for vertOrder2:
vertOrder2-current
If I flip the weights as in the PR, I get this instead:
vertOrder2-proposed

C. One way to think of the current implementation is to view it as a linear interpolation of the mixing ratio field. From this perspective, it's clear that a flux built on this linear interpolation would have the verticalWeightK(m1) values that are already in the code, not the proposed modified values.

D. In summary, I believe the sign flip for the vertOrder3 beta coefficient, both in _mono.F and in the vflux3 function, are the only two lines that should be modified in this PR.

E. As future work, whether or not we ever switch to vertical remap, it would be good to get the hProv-weighting used in vertOrder2 into the higher-order flux functions. Right now, the vertOrder3 vertical discretization I think is _worse_ than the vertOrder2 one. Here again I can demonstrate this with my 1D standalone code. Currently, with a nonuniform discretization, vertOrder3 gives this:
vertOrder3-current
But if we add hProv-weighting to the flux functions, we can get this:
vertOrder3-hProv

vanroekel:
@ambrad thanks for the testing.  I’m still really confused by the vertOrder2 weights still.  Shouldn’t the flux here be a weighted average of tracerCur(k-1) and tracerCur(k)?  If so, it seems the flux should be
```
W(k-1/2)*(hProv(k)/(hProv(k-1)+hProv(k))*tracerCur(k) + hProv(k-1)/(hProv(k) + hProv(k-1))*tracerCur(k-1))
```
?

But instead the averaging is flipped
```
W(k-1/2)*(hProv(k)/(hProv(k-1)+hProv(k))*tracerCur(k-1) + hProv(k-1)/(hProv(k) + hProv(k-1))*tracerCur(k))
```

This second flux (which is what the code has I think) seems wrong to me.  I’m obviously missing something as your testing clear shows my modification (first block in this comment) is wrong.  Could you point out what I’m missing?

(2) Interesting results with vertOrder2 vs vertOrder3.  It does suggest we should add hPrv weighting.  I’ve previously worked out what that should look like.  I can add it to this PR.  I don’t think we want vertOrder2 given the increase in dispersion in advection, which could be problematic.

vanroekel:
About your later two points
(F) I can try that but I think we may need to add vertical diffusivity, but on second thought I guess not since the vertical is FCT as well.  I can give this a try

(G) Whole heartedly agree!  It would be extremely great to have vertical advection convergence tests too

src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
```diff
@@ -701,9 +701,9 @@ subroutine ocn_tracer_advection_mono_tend(tend, tracers, layerThickness,    &
           do iCell = 1, nCells
             kmax = maxLevelCell(iCell)
             do k=3,kmax-1
-              verticalWeightK   = hProv(k-1,iCell) / &
+              verticalWeightK   = hProv(k  ,iCell) / &
```
mark-petersen, line 704: @vanroekel the original version was correct on all these. It is not a weighted average, it is an interpolation.
- weighted average: ( h_k T_k + h_{k-1} T_{k-1} ) / ( h_k + h_{k-1})
- interpolation: ( h_{k-1} T_k + h_k T_{k-1} ) / ( h_k + h_{k-1})

image
It seems opposite. I saw this in 2017 and thought it was an error too. But for interpolation, as a layer thickens (h_k) the value at the interface gets closer and closer to the value in the _opposite_ layer (h_{k-1}). For the equation of a line, you would actually have factors of 1/2 on all the thicknesses, but those all divide out.

vanroekel: ahh, thanks @mark-petersen!  That makes sense now.  I'll take those out.

src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
```diff
@@ -682,7 +682,7 @@ subroutine ocn_tracer_advection_mono_tend(tend, tracers, layerThickness,    &
             do k=3,kmax-1
               highOrderFlx(k, iCell) = (w(k,iCell)* &
                    (7.0_RKIND*(tracerCur(k  ,iCell) + tracerCur(k-1,iCell)) - &
-                              (tracerCur(k+1,iCell) + tracerCur(k-2,iCell))) - &
+                              (tracerCur(k+1,iCell) + tracerCur(k-2,iCell))) + &
```
mark-petersen, line 685: I agree this looks suspicious. I'll look more carefully tomorrow. One consideration is that k index is positive down and w is positive up. 

Does this equation appear in the sw or atm core for comparison? I didn't see it, with a quick check.

vanroekel: @mark-petersen it is here in the atm core -- src/core_atmosphere/dynamics/mpas_atm_time_integration.F L3720 with the + instead of -

ambrad (approved):

src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
mark-petersen: Thanks for the reference to the atmospheric core. Those lines are here:
```
3717       flux4(q_im2, q_im1, q_i, q_ip1, ua) =                     &
3718           ua*( 7.*(q_i + q_im1) - (q_ip1 + q_im2) )/12.0
3719
3720       flux3(q_im2, q_im1, q_i, q_ip1, ua, coef3) =              &
3721                 flux4(q_im2, q_im1, q_i, q_ip1, ua) +           &
3722                 coef3*abs(ua)*((q_ip1 - q_im2)-3.*(q_i-q_im1))/12.0
```
Where the atmosphere `+` is at the end of 3721. I think the negative is correct for the ocean, because:
1. In the atmosphere, positive ua points with increasing k
2. In the ocean, positive w points with decreasing k

so `+abs(ua)` in the atmospheric stencil translates to `-abs(w)` in the ocean stencil. In other words, upwinding is in index space, but the direction of upwinding comes from the velocity. In the ocean the sign of the velocity is in physical space, which is opposite.

I can see the advantage of an ocean model that counts from the bottom up for stuff like this, but an upside-down k index seems to be a well-accepted standard in ocean models.

Here is a thought experiment: Make new variable `wDown=-w` for w pointed down, aligned with index space. If we do that everywhere, do we get the correct stencil relative to the S&G2011 paper?  Fluxes would also change sign, so `highOrderFlxDown = -highOrderFlx` etc is what we expect. I think the negative sign propagates through all the formulas, with the exception of `abs(w)` and `abs(wDown)`, which is why we see that negative sign:
```
abs(wDown) = - abs(w)
```
because in this case the absolute value clears the sign, and we have to put it back in manually.

mark-petersen:
After we settle this, let's make sure to put comments in the code for the next time someone is confused.

src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
vanroekel: @mark-petersen I don't understand the logic here.  Both @ambrad and I worked from SG11 and defined derivatives appropriately relative to height, which seems correct.  It's not clear to me at all why it matters which way index space points.  If we define our derivatives appropriately that seems like the correct direction.  I have to think about this more.

Also note that looking at the copy pasted code, I think the atmospheric formulation is wrong.  The way it is written there should be a sign(ua) not abs(ua).  Ours is fine because we have ua*abs(ua) which is sign(ua).

vanroekel: thinking about this further, I think you may be right.  I see how the index space direction matters here now.  But I'd like to write this out explicitly to fully convince myself.

mark-petersen: Also, note that `w = sign(w)*abs(w)` is the relation for comparing the paper to the code. The paper uses `sign(w)` and the code uses `abs(w)`. The lines are here:
```
 683               highOrderFlx(k, iCell) = (w(k,iCell)* &
 684                    (7.0_RKIND*(tracerCur(k  ,iCell) + tracerCur(k-1,iCell)) - &
 685                               (tracerCur(k+1,iCell) + tracerCur(k-2,iCell))) - &
 686                         coef3rdOrder*abs(w(k,iCell))* &
 687                              ((tracerCur(k+1,iCell) - tracerCur(k-2,iCell)) - &
 688                     3.0_RKIND*(tracerCur(k  ,iCell) - tracerCur(k-1,iCell))))/ &
 689                    12.0_RKIND
```
but I'll write it simpler with T as tracer and ignore iCell:
```
F(k) = w(k)*(      7*(T(k) + T(k-1)) - (T(k+1) + T(k-2))   ) &
       - coef3rdOrder*abs(w(k)) * ( T ...  etc)
```
The point here is that in SG2011, the first `u` is completely outside the square brackets, and multiplies all the way through:
image
In the code, we multiply through by the u, and use `u*sign(u) = abs(u)` to get: 
```
F = u* (...) + abs(u)*(...)
```

ambrad: Mark, that is a good point concerning the sign w.r.t. positive w. In SG11, "sign(u) is positive for u indicating flow from cell i to cell i+1." In MPAS-O, I think you're saying positive w points up and k=1 is the surface, so the sign is already correct.

ambrad (commented):
I'm removing my approval because of Mark's comment about sign w.r.t. increasing k.

Once we get convergence tests for each discretization operator, we'll be able to get clear evidence from the code itself regarding changes of this sort.

ambrad (changes requested):
Trying to change my review again. It looks like this PR can be closed without merging.

ambrad:
@vanroekel there is still the possibility of getting better vertical advection by using the vertOrder2 formula in the vertical since the vertOrder3 one does not use hProv weighting. But, again, since we don't have a convergence test for the vertical right now, it's hard to decide on something like that.

vanroekel:
@ambrad and @mark-petersen I agree with everything here, the sign is correct as is.  It is very clear if you take SG11 eq. 7 and set beta = 1.  I will close this PR now and make a new PR that increases commenting to reduce confusion in the advection code.  I will also try address the hProv weighting in the third and 4th order reconstructions.  Hopefully soon.

Thanks for the great discussion on this PR, it was very helpful.

PR #739, opened 2020-11-02
Modifications to tracer advection mono

vanroekel:
This PR does two things

1) adds hProv weighting to the high order vertical fluxes (3rd/4th)
2) adds commenting of fluxes to reduce confusion in interpretation

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
@@ -47,9 +47,9 @@ module ocn_tracer_advection_mono
       vertOrder           !< choice of order for vertical advection
    integer, parameter :: &! enum for supported vertical algorithm order
       vertOrder2 = 2,    &!< 2nd order scheme
-      vertOrder3 = 3,    &!< 3rd order scheme
-      vertOrder4 = 4      !< 4th order scheme
-
+      vertOrder3 = 3      !< 3rd order scheme
+                          !< there is no enum for 4th order, it uses the same
+                          !< code as vertOrder3, but with coef3rdOrder=0.0
    ! These temporary allocatable arrays will eventually be moved back into
    ! the tendency routine, but must be declared here for now so that they
    ! retain the shared attribute. Once the threading model has been changed,
@@ -175,7 +175,10 @@ module ocn_tracer_advection_mono
          invAreaCell2,      &! inverse cell area
          verticalWeightK,   &! vertical weighting
          verticalWeightKm1, &! vertical weighting
-         coef1, coef3        ! temporary coefficients
+         coef1, coef3,      &! temporary coefficients
+         dzEdge,            &! distance between cell centers in the vertical
+         deriv2K, deriv2Km1  ! second derivative of vertical tracer gradient for 
+                             ! 3rd and 4th order reconstructions
 
       real (kind=RKIND), dimension(size(tracers, dim=2)) :: &
          wgtTmp,            &! vertical temporaries for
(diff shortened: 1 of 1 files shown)


vanroekel:
@mark-petersen and @ambrad I've pulled together a draft PR to add thickness weighting to the high order fluxes.  I've also included more commenting of 3rd/4th order fluxes to avoid confusion.  I have not tested this beyond building it, but wanted to post for your comments.  In particular can you verify my implementation of SG equation (11) with the full derivatives?  I've included my derivation.

File_001

I will report back after testing.

src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
```diff
@@ -656,46 +659,62 @@ subroutine ocn_tracer_advection_mono_tend(tend, tracers, layerThickness,    &
...
-
+              deriv2K = (tracerCur(k-1,iCell) - tracerCur(k,iCell))/(0.5_RKIND*hProv(k,iCell)* &
+                        (hProv(k-1,iCell) + hProv(k,iCell))) - (tracerCur(k,iCell) - tracerCur(k+1,iCell))/ &
+                        (0.5_RKIND*hProv(k,iCell)*(hProv(k,iCell) + hProv(k+1,iCell)))
+              deriv2Km1 = (tracerCur(k-2,iCell) - tracerCur(k-1,iCell))/(0.5_RKIND*hProv(k-1,iCell)* &
+                        (hProv(k-2,iCell) + hProv(k-1,iCell))) - (tracerCur(k-1,iCell) - tracerCur(k,iCell)) / &
+                        (0.5_RKIND*hProv(k-1,iCell)*(hProv(k-1,iCell) + hProv(k,iCell)))
+              dzEdge = 0.5*(hProv(k-1,iCell) + hProv(k,iCell))
+
+              highOrderFlx(k, iCell) = w(k,iCell)*(0.5_RKIND*(tracerCur(k,iCell) + tracerCur(k-1,iCell)) - &
```
ambrad, line 699: I think the primary spot where using hProv weighting will benefit the flux is in the 2nd-order part. Here, you've kept this as `1/2 (q(k) + q(k-1))` and used hProv only in the second-derivative estimates. The vertOrder2 code uses the hProv-weighted version based on a linear interpolant; I think you should bring this into this expression.

That aside, I think we should be cautious about making changes to this code without a vertical-direction convergence test.

vanroekel: Thanks @ambrad you're right, I didn't even think about that.  It is of course obvious that part is fixed grid too.

I also agree we should have a vertical advection convergence test.  I started implementing that test with a postbac but didn't get it fully cleaned up.

src/core_ocean/shared/mpas_ocn_tracer_advection_mono.F
```diff
@@ -656,46 +659,65 @@ subroutine ocn_tracer_advection_mono_tend(tend, tracers, layerThickness,    &
...
-                   12.0_RKIND
-
-
+              deriv2K = (tracerCur(k-1,iCell) - tracerCur(k,iCell))/(0.5_RKIND*hProv(k,iCell)* &
+                        (hProv(k-1,iCell) + hProv(k,iCell))) - (tracerCur(k,iCell) - tracerCur(k+1,iCell))/ &
+                        (0.5_RKIND*hProv(k,iCell)*(hProv(k,iCell) + hProv(k+1,iCell)))
+              deriv2Km1 = (tracerCur(k-2,iCell) - tracerCur(k-1,iCell))/(0.5_RKIND*hProv(k-1,iCell)* &
+                        (hProv(k-2,iCell) + hProv(k-1,iCell))) - (tracerCur(k-1,iCell) - tracerCur(k,iCell)) / &
+                        (0.5_RKIND*hProv(k-1,iCell)*(hProv(k-1,iCell) + hProv(k,iCell)))
+              dzEdge = 0.5*(hProv(k-1,iCell) + hProv(k,iCell))
```
ambrad, line 697: `0.5_RKIND` rather than `0.5`.

These formulas look fine, but again, it's best to verify them with a convergence test.

The arithmetic could be decreased a bit in the deriv2(m1)* and highOrderFlx lines with some algebra.

mgduda:

PR #743, opened 2020-11-07
Use appropriate units for core_atmosphere file I/O

jjguerrette:
This PR supports interfacing MPAS with the JEDI data assimilation framework.

The atmospheric core is updated to use the MPAS framework subroutines mpas_new_unit and mpas_release_unit for defining appropriate file units for I/O that do not conflict with other units already used throughout MPAS or in a surrounding program when MPAS is used as a library.


src/core_atmosphere/physics/mpas_atmphys_camrad_init.F
```diff
@@ -745,13 +736,14 @@ subroutine aer_optics_initialize(dminfo)
...
 !=================================================================================================================
- subroutine oznini(mesh,atm_input)
+ subroutine oznini(dminfo,mesh,atm_input)
 !=================================================================================================================
 
 !This subroutine assumes a uniform distribution of ozone concentration. It should be replaced
 !with monthly climatology varying ozone distribution.
 
 !input arguments:
+ type(dm_info):: dminfo
```
mgduda, line 746: I think this can be safely declared as `intent(in)`.

src/core_atmosphere/physics/mpas_atmphys_camrad_init.F
```diff
@@ -789,30 +779,40 @@ subroutine oznini(mesh,atm_input)
...
- open(pin_unit,file='OZONE_PLEV.TBL',action='READ',status='OLD',iostat=istat)
+ if(dminfo % my_proc_id == IO_NODE) then
+    call mpas_new_unit(read_unit)
+    if(read_unit < 0) &
+       call physics_error_fatal('module_ra_cam: oznini: Cannot find unused '//&
+                                'fortran unit to read in lookup table.')
+ end if
+
+!distribute unit to other processors:
+ call mpas_dmpar_bcast_int(dminfo,read_unit)
```
mgduda, line 790: I don't think this is guaranteed to be correct. It's possible that on MPI rank 0, unit *N* is available, but that unit *N* is not available on one or more other ranks. In other words, we cannot assume that the usage of I/O units is identical across all MPI ranks.

Regardless, I think we should stick to the existing pattern of having just MPI rank 0 read data and broadcast that data to all other ranks.

src/core_atmosphere/physics/mpas_atmphys_camrad_init.F
```diff
@@ -822,12 +822,16 @@ subroutine oznini(mesh,atm_input)
...
+    read(read_unit,*) ozmixin(i,k,j,m)
  enddo
  enddo
  enddo
  enddo
- close(oz_unit)
+ close(read_unit)
+
+ if(dminfo % my_proc_id == IO_NODE) then
+    call mpas_release_unit(read_unit)
```
mgduda, line 833: If we let all MPI ranks read the file (which I would argue is not best practice, but it's what has apparently been working without issue so far), then I think `mpas_release_unit` should be called by all ranks.

jjguerrette: Ok, I will remove the usage of dminfo within `ozini`.  All ranks will get their own file unit (new behavior) and read the same file (same as existing behavior).  I think I was trying to follow `radaeini` with the rank checks, but I never added the communications.  Adding new comms is not my intention in this PR.

jjguerrette: I was going to add another commit that accomplishes my description above, but it seems like I could alternatively remove commit 22a14a9fd8efeadba9e7ca463389159c from history.  Doing that requires yet another rebase, which seems very frowned upon in all the reading I am doing about git histories.  Maybe it doesn't matter for a feature branch.  Is rebasing the preferred option from an MPAS-Atmosphere development standpoint?  The conflating factor is that this branch was originally stemmed from an old `develop` commit several years ago.  So removing old commits requires going through all the new develop commits that has to be merged into this branch.  If you would rather that I rebase, what is the easiest way to do it?  Create another branch and PR?

jjguerrette: I gave it a try by checking out the commit before the one that needed to be removed, then doing a force push with `git push -f origin HEAD:atmosphere/unformatted_io`

mgduda: In general, rebasing feature branches is not frowned upon. The problem with rebasing a branch that many people might have in their forks is that your rebased branch then becomes inconsistent (history-wise) with everyone else's. But, if the intent is to soon merge that branch into 'develop', this is probably not a problem.

mgduda:
@ldfowler58 The changes in this PR modify code in the `physics` directory, so if you'd like to offer any comments, please feel free to add yourself as a reviewer and post your review comments. If everything looks fine to you, though, there's no need to do anything. I'll try to get this merged by the end of this week, but if you need more time, just let me know.

mgduda (approved):

PR #744, opened 2020-11-07
Move domain_ptr and corelist from module variables to arguments in subdriver

mgduda:
Towards the goal of supporting multiple instances of domains within a single
simulation, this merge moves domain_ptr and corelist from module variables
in the mpas_subdriver module to arguments to the mpas_init, mpas_run, and
mpas_finalize routines.

The main program (in mpas.F) now passes instances of a core_type and
a domain_type to the subriver's mpas_init, mpas_run, and mpas_finalize routines.

### DIFF
--- a/src/driver/mpas.F
+++ b/src/driver/mpas.F
@@ -8,14 +8,18 @@
 program mpas
 
    use mpas_subdriver
+   use mpas_derived_types, only : core_type, domain_type
 
    implicit none
 
-   call mpas_init()
+   type (core_type), pointer :: corelist => null()
+   type (domain_type), pointer :: domain => null()
 
-   call mpas_run() 
+   call mpas_init(corelist, domain)
 
-   call mpas_finalize()
+   call mpas_run(domain) 
+
+   call mpas_finalize(corelist, domain)
 
    stop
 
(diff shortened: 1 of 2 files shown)


mgduda:
This PR handles the first commit in PR #308. By creating separate PRs for each piece of PR #308, it will hopefully be easier to review each of the changes in isolation.

mark-petersen (approved):
Thanks @mgduda! I merged locally, this runs fine on mpas-ocean with gnu 5 and intel 19, debug and optimized.

I guess there will be a later phase where each core will need to specify the particular domain in their core code?

mgduda:
@mark-petersen For the stand-alone MPAS models, I don't think any further changes are needed to use a specific domain, except if cores decide in future to work with multiple domain instances in the same executable. The specific use case for these changes is a data assimilation system that works with multiple instances of the atmosphere, and that also re-uses the sub-driver code, necessitating the removal of module variables from the sub-driver.

If I recall correctly, E3SM doesn't use the MPAS sub-driver module, right? If this is correct, then I don't foresee any impact on E3SM.

matthewhoffman (approved):
@mgduda , that is correct - E3SM does not use the subdriver module.

MiCurry (approved):

PR #745, opened 2020-11-09
New module for ocean diagnostics variables

mattdturner:
This PR add a new module to house the variables contained in `diagnosticsPool` in the ocean model.  This also implements the use of the new module in the `mpas_ocn_diagnostics.F` file.

This is the first of a handful of PRs that will replace #611 by breaking it up into smaller PRs.


mattdturner:
This passes the nightly regression suite, and matches `ocean/develop` bit-for-bit for each test in the MPAS-O nightly regression suite.

I am about to start testing in E3SM.

mark-petersen (approved):
Thanks @mattdturner. I also ran the nightly regression suite with gnu and intel, and looked through the code. Since this is really moving declarations into a new module, any errors would be detected by these tests.

PR #750, opened 2020-11-10
Parallel interpolation of annual max snow albedo

MiCurry:
This merge allows the init_atmosphere model to safely interpolate maximum snow albedo in parallel on non-cvt partitions. However, because other fields have not yet been updated to run in parallel, the multiprocessor check in init_atm_cases is still present.

This merge is in connection with merges 91d8539, and 679cf1b. Parallel interpolation of monthly climatological monthly vegetation fraction and monthly-mean climatological surface albedo still need to be interpolated and will be implemented in a future commit.

### DIFF
--- a/src/core_init_atmosphere/mpas_geotile_manager.F
+++ b/src/core_init_atmosphere/mpas_geotile_manager.F
@@ -682,16 +682,22 @@ module mpas_geotile_manager
     !
     !  public subroutine mpas_geotile_mgr_tile_to_latlon => tile_to_latlon
     !
-    !> \brief   Translate a tile indices to latitude and longitude,
+    !> \brief   Find the latitude, longitude location of a tile's pixels
     !> \author  Miles A. Curry
     !> \date    02/20/2020
     !> \details
-    !> Given a tile, translate the relative tile coordinates, i, j, of that
-    !> tile to a latitude and longitude coordinate. Upon success, lat, lon
-    !> will be in the range of -1/2 * pi to 1/2 * pi and 0 to 2.0 * pi, respectively.
+    !> Given a tile, translate the pixel coordinates, i, j to a corresponding latitude
+    !> and longitude coordinate.
+    !>
+    !> If supersample_fac is present each pixel will be subdivided into supersample_fac ^ 2
+    !> sub-pixels. If supersample_fac is greater than 1, then the calling code will need
+    !> to pass in supersampled i and j coordinates.
+    !>
+    !> Upon success, lat, lon will be in the range  of -1/2 * pi to 1/2 * pi and 0 to
+    !> 2.0 * pi, respectively.
     !
     !-----------------------------------------------------------------------
-    subroutine mpas_geotile_mgr_tile_to_latlon(mgr, tile, j, i, lat, lon)
+    subroutine mpas_geotile_mgr_tile_to_latlon(mgr, tile, j, i, lat, lon, supersample_fac)
 
         implicit none
 
@@ -701,15 +707,23 @@ module mpas_geotile_manager
         integer, value :: i
         real (kind=RKIND), intent(out) :: lat
         real (kind=RKIND), intent(out) :: lon
+        integer, optional, intent(in) :: supersample_fac
 
         integer, pointer :: tile_bdr
         real (kind=RKIND), pointer :: known_lon
         real (kind=RKIND), pointer :: known_lat
         real (kind=RKIND), pointer :: dx
         real (kind=RKIND), pointer :: dy
+        integer :: supersample_lcl
         integer :: ierr
 
         ierr = 0
+
+        if (present(supersample_fac)) then
+            supersample_lcl = supersample_fac
+        else
+            supersample_lcl = 1
+        end if
 
         call mpas_pool_get_config(mgr % pool, 'tile_bdr', tile_bdr)
         call mpas_pool_get_config(mgr % pool, 'known_lat', known_lat)
@@ -717,8 +731,10 @@ module mpas_geotile_manager
         call mpas_pool_get_config(mgr % pool, 'dx', dx)
         call mpas_pool_get_config(mgr % pool, 'dy', dy)
 
-        lat = real((j - (tile_bdr + 1) + tile % y - 1), kind=RKIND) * dy + known_lat
-        lon = real((i - (tile_bdr + 1) + tile % x - 1), kind=RKIND) * dx + known_lon
+        lat = known_lat + real(j - (supersample_lcl * tile_bdr + 1) + (supersample_lcl * (tile % y - 1)), kind=RKIND) * dy &
+                               / supersample_lcl
+        lon = known_lon + real(i - (supersample_lcl * tile_bdr + 1) + (supersample_lcl * (tile % x - 1)), kind=RKIND) * dx &
+                              / supersample_lcl
 
         call deg2Rad(lat)
         call deg2Rad(lon)
--- a/src/core_init_atmosphere/mpas_geotile_manager.F
+++ b/src/core_init_atmosphere/mpas_geotile_manager.F
@@ -274,8 +274,8 @@ module mpas_geotile_manager
         ! Calculate the total number of pixels in x dir
         ! NOTE: This calculation assumes that a dataset is a global dataset and may
         ! not work correctly for non-global datasets
-        mgr % pixel_nx = nint(360.0_RKIND / dx)
-        mgr % pixel_ny = nint(180.0_RKIND / dy)
+        mgr % pixel_nx = nint(360.0_RKIND / abs(dx))
+        mgr % pixel_ny = nint(180.0_RKIND / abs(dy))
 
         ! Calculate the number of tiles in the x, y directions
         ! NOTE: This calculation assumes that a dataset is a global dataset and may
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -829,107 +829,14 @@
 !
  if (trim(config_maxsnowalbedo_data) == 'MODIS') then
 
+    geog_sub_path = 'maxsnowalb_modis/'
+
     call mpas_log_write('Using MODIS 0.05-deg data for maximum snow albedo')
     if (supersample_fac > 1) then
        call mpas_log_write('   Dataset will be supersampled by a factor of $i', intArgs=(/supersample_fac/))
     end if
 
-    nx = 1206
-    ny = 1206
-    nz = 1
-    isigned  = 1
-    endian   = 0
-    wordsize = 2
-    scalefactor = 0.01
-    msgval = real(-999.0,kind=R4KIND)*real(0.01,kind=R4KIND)
-    fillval = 0.0
-    allocate(rarray(nx,ny,nz))
-    allocate(nhs(nCells))
-    nhs(:) = 0
     snoalb(:) = 0.0
-
-    rarray_ptr = c_loc(rarray)
-
-    start_lat = 90.0 - 0.05 * 0.5 / supersample_fac
-    start_lon = -180.0 + 0.05 * 0.5 / supersample_fac
-    geog_sub_path = 'maxsnowalb_modis/'
-
-    do jTileStart = 1,02401,ny-6
-       jTileEnd = jTileStart + ny - 1 - 6
-
-       do iTileStart=1,06001,nx-6
-          iTileEnd = iTileStart + nx - 1 - 6
-          write(fname,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(geog_data_path)//trim(geog_sub_path), &
-                       iTileStart,'-',iTileEnd,'.',jTileStart,'-',jTileEnd
-          call mpas_log_write(trim(fname))
-          call mpas_f_to_c_string(fname, c_fname)
-
-          call read_geogrid(c_fname,rarray_ptr,nx,ny,nz,isigned,endian, &
-                            wordsize,istatus)
-          call init_atm_check_read_error(istatus, fname)
-          rarray(:,:,:) = rarray(:,:,:) * scalefactor
-
-          iPoint = 1
-          do j=supersample_fac * 3 + 1, supersample_fac * (ny-3)
-          do i=supersample_fac * 3 + 1, supersample_fac * (nx-3)
-             ii = (i - 1) / supersample_fac + 1
-             jj = (j - 1) / supersample_fac + 1
-
-             lat_pt = start_lat - (supersample_fac*(jTileStart-1) + j - (supersample_fac*3+1)) * 0.05 / supersample_fac
-             lon_pt = start_lon + (supersample_fac*(iTileStart-1) + i - (supersample_fac*3+1)) * 0.05 / supersample_fac
-             lat_pt = lat_pt * PI / 180.0
-             lon_pt = lon_pt * PI / 180.0
-
-             iPoint = nearest_cell(lat_pt,lon_pt,iPoint,nCells,maxEdges, &
-                                   nEdgesOnCell,cellsOnCell, &
-                                   latCell,lonCell)
-             if (rarray(ii,jj,1) /= msgval) then
-
-                !
-                ! This field only matters for land cells, and for all but the outermost boundary cells,
-                ! we can safely assume that the nearest model grid cell contains the pixel (else, a different
-                ! cell would be nearest)
-                !
-                if (landmask(iPoint) == 1 .and. bdyMaskCell(iPoint) < nBdyLayers) then
-                   snoalb(iPoint) = snoalb(iPoint) + rarray(ii,jj,1)
-                   nhs(iPoint) = nhs(iPoint) + 1
-
-                ! For outermost land cells, additional work is needed to verify that the pixel
-                ! actually lies within the nearest cell
-                else if (landmask(iPoint) == 1) then
-                   zPixel = sphere_radius * sin(lat_pt)                  ! Model cell coordinates assume a "full" sphere radius
-                   xPixel = sphere_radius * cos(lon_pt) * cos(lat_pt)    ! at this point, so we need to ues the same radius
-                   yPixel = sphere_radius * sin(lon_pt) * cos(lat_pt)    ! for source pixel coordinates
-
-                   if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(iPoint), yCell(iPoint), zCell(iPoint), &
-                               nEdgesOnCell(iPoint), verticesOnCell(:,iPoint), xVertex, yVertex, zVertex)) then
-                      snoalb(iPoint) = snoalb(iPoint) + rarray(ii,jj,1)
-                      nhs(iPoint) = nhs(iPoint) + 1
-                   end if
-                end if
-             end if
-          end do
-          end do
-
-       end do
-    end do
-
-    do iCell = 1,nCells
-       !
-       ! Mismatches in land mask can lead to MPAS land points with no maximum snow albedo.
-       ! Ideally, we would perform a search for nearby valid albedos, but for now using
-       ! the fill value will at least allow the model to run. In general, the number of cells
-       ! to be treated in this way tends to be a very small fraction of the total number of cells.
-       !
-       if (nhs(iCell) == 0) then
-          snoalb(iCell) = fillval
-       else
-          snoalb(iCell) = snoalb(iCell) / real(nhs(iCell))
-       end if
-       snoalb(iCell) = 0.01_RKIND * snoalb(iCell)        ! Convert from percent to fraction
-    end do
-    deallocate(rarray)
-    deallocate(nhs)
 
  else if (trim(config_maxsnowalbedo_data) == 'NCEP') then
 
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -88,7 +88,7 @@
  integer,dimension(:,:),allocatable:: ncat
       
  real(kind=RKIND), pointer :: scalefactor_ptr
- real(kind=c_float):: scalefactor
+ real(kind=RKIND) :: scalefactor
  real(kind=c_float),dimension(:,:,:),pointer,contiguous :: rarray
  type(c_ptr) :: rarray_ptr
 
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -836,7 +836,44 @@
        call mpas_log_write('   Dataset will be supersampled by a factor of $i', intArgs=(/supersample_fac/))
     end if
 
+    ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+    if (ierr /= 0) then
+       call mpas_log_write('Error occurred when initializing the interpolation of snow albedo (snoalb)', &
+                            messageType=MPAS_LOG_CRIT)
+    endif
+
     snoalb(:) = 0.0
+
+    do iCell = 1, nCells
+
+       tile => null()
+       ierr = mgr % get_tile(latCell(iCell), lonCell(iCell), tile)
+       if (ierr /= 0 .or. .not. associated(tile)) then
+           call mpas_log_write('Could not get the tile that contained cell $i', intArgs=(/iCell/), messageType=MPAS_LOG_CRIT)
+       end if
+
+       if (tile % is_processed) then
+           cycle
+       end if
+
+       tile % is_processed = .True.
+
+       ierr = mgr % push_tile(tile)
+       if (ierr /= 0) then
+           call mpas_log_write('Error pushing this tile onto the stack: '//trim(tile%fname), messageType=MPAS_LOG_CRIT)
+       end if
+    end do
+
+    do while(.not. mgr % is_stack_empty())
+       tile => mgr % pop_tile()
+       call mpas_log_write('Processing tile: '//trim(tile % fname))
+    end do
+
+    ierr = mgr % finalize()
+    if (ierr /= 0) then
+       call mpas_log_write('Error occurred when finalizing the interpolation of snow albedo (snoalb)', &
+                            messageType=MPAS_LOG_CRIT)
+    endif
 
  else if (trim(config_maxsnowalbedo_data) == 'NCEP') then
 
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -129,10 +129,12 @@
  integer (kind=I8KIND), dimension(:), pointer :: ter_integer
  real (kind=RKIND), dimension(:), pointer :: soiltemp
  real (kind=RKIND), dimension(:), pointer :: snoalb
+ integer (kind=I8KIND), dimension(:), pointer :: snoalb_integer
  real (kind=RKIND), dimension(:), pointer :: shdmin, shdmax
  real (kind=RKIND), dimension(:,:), pointer :: greenfrac
  real (kind=RKIND), dimension(:,:), pointer :: albedo12m
  real (kind=RKIND) :: msgval, fillval
+ real (kind=RKIND), pointer :: missing_value
  integer, pointer :: category_min, category_max
  integer, dimension(:), pointer :: lu_index
  integer, dimension(:), pointer :: soilcat_top
@@ -150,6 +152,7 @@
  type (mpas_geotile_type), pointer :: tile
 
  real (kind=RKIND) :: tval
+ integer (kind=I8KIND) :: i8val
  integer, pointer :: tile_bdr
  integer, pointer :: tile_nx, tile_ny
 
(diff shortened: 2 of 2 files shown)


src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -837,33 +838,110 @@ subroutine init_atm_static(mesh, dims, configs)
...
-       ierr = mgr % push_tile(tile)
-       if (ierr /= 0) then
-           call mpas_log_write('Error pushing this tile onto the stack: '//trim(tile%fname), messageType=MPAS_LOG_CRIT)
-       end if
+            call mpas_log_write('Processing tile: '//trim(tile % fname))
+
+            all_pixels_mapped_to_halo_cells = .true.
+
+            do i = supersample_fac * tile_bdr + 1, supersample_fac * (tile_ny + tile_bdr), 1
+                do j = supersample_fac * tile_bdr + 1, supersample_fac * (tile_nx + tile_bdr), 1
```
MiCurry, line 878: I think these loops need to be flipped i.e.:
```
 do j = supersample_fac * tile_bdr + 1, supersample_fac * (tile_ny + tile_bdr), 1
                do i = supersample_fac * tile_bdr + 1, supersample_fac * (tile_nx + tile_bdr), 1
```

mgduda: I agree.

MiCurry: This has been updated.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -274,8 +274,8 @@ function mpas_geotile_mgr_init(mgr, path) result(ierr)
         ! Calculate the total number of pixels in x dir
         ! NOTE: This calculation assumes that a dataset is a global dataset and may
         ! not work correctly for non-global datasets
-        mgr % pixel_nx = nint(360.0_RKIND / dx)
-        mgr % pixel_ny = nint(180.0_RKIND / dy)
+        mgr % pixel_nx = abs(nint(360.0_RKIND / dx))
+        mgr % pixel_ny = abs(nint(180.0_RKIND / dy))
```
mgduda, line 278: Just for consideration, an alternative here would be to call the `abs` intrinsic with only `dx` and `dy` as arguments, since the literal constants `360.0_RKIND` and `180.0_RKIND` can obviously never be negative. The result is the same, but it seems like it better expresses to the reader the possibility that `dx` and `dy` might be negative if these variables alone are the arguments to the `abs` call.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -824,107 +825,129 @@ subroutine init_atm_static(mesh, dims, configs)
...
-    snoalb(:) = 0.0
-
-    rarray_ptr = c_loc(rarray)
-
-    start_lat = 90.0 - 0.05 * 0.5 / supersample_fac
-    start_lon = -180.0 + 0.05 * 0.5 / supersample_fac
-    geog_sub_path = 'maxsnowalb_modis/'
+    ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+    if (ierr /= 0) then
+       call mpas_log_write('Error occurred when initalizing the interpolation of snow albedo (snoalb)', &
```
mgduda, line 837: "initalizing" -> "initializing"

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -824,107 +825,129 @@ subroutine init_atm_static(mesh, dims, configs)
...
-          iTileEnd = iTileStart + nx - 1 - 6
-          write(fname,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(geog_data_path)//trim(geog_sub_path), &
-                       iTileStart,'-',iTileEnd,'.',jTileStart,'-',jTileEnd
-          call mpas_log_write(trim(fname))
-          call mpas_f_to_c_string(fname, c_fname)
+    allocate(nhs(nCells))
+    snoalb(:) = 0.0
+    nhs(:) = 0
+    fillval = 0.0
+    missing_value = real(missing_value, kind=R4KIND) * real(0.01, kind=R4KIND)
```
mgduda, line 850: Here's an alternative that might be more compact:
```fortran
    missing_value = real(0.01_RKIND * missing_value, kind=R4KIND)
```
This could arguably be more precise, too, when `RKIND` specifies 64-bit reals, since the multiplication of 0.01 and `missing_value` will occur at 64-bit precision.

src/core_init_atmosphere/mpas_geotile_manager.F
MiCurry: I like this a lot better. The more succinct and clearer the better. I'll update it.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Fixed.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: I like this better as well. It has been updated.

src/core_init_atmosphere/mpas_geotile_manager.F
MiCurry: Fixed.

MiCurry:
This PR will now need to updated after #757 is merged in.

MiCurry:
@mgduda this PR has been rebased with the latest changes from PR #750.

MiCurry:
@mgduda this probably should have been addressed in PR #750, but should we now remove the `c_float` `scalefactor` declared at line 91. As well, should we rename `scalefactor_ptr` to just `scalefactor`? Since now there is no reason to have two versions of scalefactor?

 real(kind=RKIND), pointer :: scalefactor_ptr
 real(kind=c_float):: scalefactor

https://github.com/MPAS-Dev/MPAS-Model/blob/bcfc98633de307699eb8cb6b09b64282795fa25d/src/core_init_atmosphere/mpas_init_atm_static.F#L90-L91

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,108 +830,132 @@ subroutine init_atm_static(mesh, dims, configs)
...
-          iTileEnd = iTileStart + nx - 1 - 6
-          write(fname,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(geog_data_path)//trim(geog_sub_path), &
-                       iTileStart,'-',iTileEnd,'.',jTileStart,'-',jTileEnd
-          call mpas_log_write(trim(fname))
-          call mpas_f_to_c_string(fname, c_fname)
+    allocate(nhs(nCells))
+    snoalb(:) = 0.0
+    nhs(:) = 0
+    fillval = 0.0
+    missing_value = real(0.01 * missing_value, kind=R4KIND) * real(0.01, kind=R4KIND)
```
mgduda, line 857: Is this correct? I think this will scale `missing_value` by 0.01 (which is the scaling factor?) twice.

MiCurry: No, it should only be multiplied by the 0.01 once. I'll update it to be multiplied by `scalefactor` and to be multiplied only once.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,108 +830,132 @@ subroutine init_atm_static(mesh, dims, configs)
...
+
+            all_pixels_mapped_to_halo_cells = .true.
+
+            do j = supersample_fac * tile_bdr + 1, supersample_fac * (tile_ny + tile_bdr), 1
+                do i = supersample_fac * tile_bdr + 1, supersample_fac * (tile_nx + tile_bdr), 1
+
+                    ii = (i - 1) / supersample_fac + 1
+                    jj = (j - 1) / supersample_fac + 1
+
+                    call mgr % tile_to_latlon(tile, j, i, lat_pt, lon_pt, supersample_fac)
```
mgduda, line 890: When `supersample_fac` != 1, is calling `tile_to_latlon` with `i`, `j`, and `supersample_fac` equivalent to calling the routine with `ii`, `jj`, and `1`? In other words, if we "un-super-sample" the pixel coordinates at this level in the code, is there any need to handle a supersampling factor in the `tile_to_latlon` routine?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,108 +830,132 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                        ! actually lies within the nearest cell
+                        else if (landmask(res % id) == 1) then
+                             if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(res % id), yCell(res % id), zCell(res % id), &
+                                              nEdgesOnCell(res % id), verticesOnCell(:,res % id), xVertex, yVertex, zVertex)) then
+
+                                    snoalb(res % id) = snoalb(res % id) + tile % tile(ii, jj, 1)
+                                    nhs(res % id) = nhs(res % id) + 1
+
+                                if (res % id <= nCellsSolve) then
+                                    all_pixels_mapped_to_halo_cells = .false.
```
mgduda, line 916: I think the logic to set `all_pixels_mapped_to_halo_cells` to `.false.` can only happen if `landmask(res % id) == 1`. Would this cause erroneous results if some pixel from a tile did map to a non-halo cell, but that non-halo cell was a water cell?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,108 +830,132 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                            if (res % id <= nCellsSolve) then
+                                all_pixels_mapped_to_halo_cells = .false.
+                            end if
+                        ! For outermost land cells, additional work is needed to verify that the pixel
+                        ! actually lies within the nearest cell
+                        else if (landmask(res % id) == 1) then
+                             if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(res % id), yCell(res % id), zCell(res % id), &
+                                              nEdgesOnCell(res % id), verticesOnCell(:,res % id), xVertex, yVertex, zVertex)) then
+
+                                    snoalb(res % id) = snoalb(res % id) + tile % tile(ii, jj, 1)
```
mgduda, line 912: For the interpolation of terrain, we performed the accumulation in a 64-bit integer array to avoid order-of-summation differences when running with different MPI task counts. Should we adopt a similar approach here? In single-precision, we can represent integral values up to about 2^24 (23 mantissa bits, plus an implicit 1 bit, I think), or 16777216, without having any gaps in the range of values. With the albedo values having up to four significant digits (up to 100 percent, with two decimal places of precision), it seems like we could run into issues when we accumulate around ~1700 pixels in each grid cell.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: I don't think the two are equivalent. `supersample_fac` needs to be passed into `tile_to_latlon` so that we can divide the pixel into subpixels: `dx / supersample`. 

``` Fortran
lat = known_lat + real(j - (supersample_lcl * tile_bdr + 1) + (supersample_lcl * (tile % y - 1)), kind=RKIND) * dy &
                               / supersample_lcl
lon = known_lon + real(i - (supersample_lcl * tile_bdr + 1) + (supersample_lcl * (tile % x - 1)), kind=RKIND) * dx &
                              / supersample_lcl
```

Here is a toy program which I made that demonstrates that the two aren't equivalent:

https://gist.github.com/MiCurry/a237abff4ec5596ec4c754e38e0f4a99

```
 not equal
 First:           11          10   89.9749985      -179.958344               3
 Second:            4           4   89.9749985      -179.975006               1
```

Does that seem correct/right to you?

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: I think code at lines [915 and 916](https://github.com/MiCurry/MPAS-Model/blob/92bbe60116ab64d7cfe06aec51304f80fdb6d203/src/core_init_atmosphere/mpas_init_atm_static.F#L915-L917) can only occur when landmask(res % id) == 1, because of the `else if` statement at line [908](https://github.com/MiCurry/MPAS-Model/blob/92bbe60116ab64d7cfe06aec51304f80fdb6d203/src/core_init_atmosphere/mpas_init_atm_static.F#L908). Does that seem correct to you? Or, perhaps I am missing something?

src/core_init_atmosphere/mpas_init_atm_static.F
mgduda: I think the result printed in the `Second:` line is correct. With a supersampling ratio of 3, the first 3x3 block of sub-pixels should all map to the same latitude and longitude, i.e., the latitude and longitude of the first full pixel.

src/core_init_atmosphere/mpas_init_atm_static.F
mgduda: Let me try to ask the question a different way: what if we did have pixels from a tile that mapped only to non-halo _water_ cells? Would `all_pixels_mapped_to_halo_cells` be correctly set in that case?

MiCurry: Ah, yes I see you what you mean and you are right. I'll rearrange the logic to make sure `all_pixels_mapped_to_halo_cells` is correctly set.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: I thought the idea behind supersample was so that the sub-pixels did not map to the same location as the full pixel. So for instance: With a supersample of 2, if an edge of two cells, c1 and c2, divided a full pixel equally in half in the x-direction, then the first half of the sub-pixels would be mapped to c1 and the second half of sub-pixels would be mapped to c2. Is that assumption incorrect?

mgduda: Sorry for the confusion -- you're correct: when computing the latitude and longitude of sub-pixels, the computed location should vary! I somehow got stuck thinking about the reverse problem: that different latitudes and longitudes should map to the same (j,i) location.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Yes that sounds like a good idea. I've just pushed a branch to accumulate values using an integer as you described.

MiCurry:
@mgduda I have just pushed an update to this branch that reworks the logic for setting `all_pixels_mapped_to_halo_cells` and to use an integer to accumulate values, similar to terrain!

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +831,146 @@ subroutine init_atm_static(mesh, dims, configs)
...
-
-       do iTileStart=1,06001,nx-6
-          iTileEnd = iTileStart + nx - 1 - 6
-          write(fname,'(a,i5.5,a1,i5.5,a1,i5.5,a1,i5.5)') trim(geog_data_path)//trim(geog_sub_path), &
-                       iTileStart,'-',iTileEnd,'.',jTileStart,'-',jTileEnd
-          call mpas_log_write(trim(fname))
-          call mpas_f_to_c_string(fname, c_fname)
+    allocate(nhs(nCells))
+    allocate(snoalb_integer(nCells))
+    snoalb_integer(:) = 0.0
```
mgduda, line 856: I think `0.0` should be `0` since `snoalb_integer` is integer-valued.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +831,146 @@ subroutine init_atm_static(mesh, dims, configs)
...
+
+            all_pixels_mapped_to_halo_cells = .true.
+
+            do j = supersample_fac * tile_bdr + 1, supersample_fac * (tile_ny + tile_bdr), 1
+                do i = supersample_fac * tile_bdr + 1, supersample_fac * (tile_nx + tile_bdr), 1
+
+                    ii = (i - 1) / supersample_fac + 1
+                    jj = (j - 1) / supersample_fac + 1
+
+                    sval = tile % tile(ii, jj, 1)
```
mgduda, line 893: In the only two places where `sval` is used, it's converted to an `I8KIND` integer. Would it be simpler to just declare `sval` as an `integer (kind=I8KIND)` (and perhaps name it something like `i8val`) and perform the conversion from `real` to `integer (kind=I8KIND)` here?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +831,146 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                        else
+                            if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(res % id), yCell(res % id), zCell(res % id), &
+                                              nEdgesOnCell(res % id), verticesOnCell(:,res % id), xVertex, yVertex, zVertex)) then
+
+                                if (landMask(res % id) == 1) then
+                                    snoalb_integer(res % id) = snoalb_integer(res % id) + int(sval, kind=I8KIND)
+                                    nhs(res % id) = nhs(res % id) + 1
+                                end if
+
+                                if (res % id <= nCellsSolve) then
```
mgduda, line 925: Would the code at line 923 obviate the need for the same code here?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +831,146 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                            ! This field only matters for land cells, and for all but the outermost boundary cells,
+                            ! we can safely assume that the nearest model grid cell contains the pixel (else, a different
+                            ! cell would be nearest)
+                            !
+                            if (landmask(res % id) == 1) then
+                                snoalb_integer(res % id) = snoalb_integer(res % id) + int(sval, kind=I8KIND)
+                                nhs(res % id) = nhs(res % id) + 1
+                            end if
+
+                            if (res % id <= nCellsSolve) then
```
mgduda, line 911: Would the code at line 932 obviate the need for the same code here?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +831,146 @@ subroutine init_atm_static(mesh, dims, configs)
...
+        ! Ideally, we would perform a search for nearby valid albedos, but for now using
+        ! the fill value will at least allow the model to run. In general, the number of cells
+        ! to be treated in this way tends to be a very small fraction of the total number of cells.
+        !
+        if (nhs(iCell) == 0) then
+            snoalb(iCell) = fillval
+        else
+            snoalb(iCell) = real(snoalb_integer(iCell), kind=R8KIND) / real(nhs(iCell), kind=R8KIND)
+            snoalb(iCell) = snoalb(iCell) * scalefactor
+            snoalb(iCell) = snoalb(iCell) / real(nhs(iCell))
```
mgduda, line 961: Does this line repeat what was done on line 959?

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: I like this idea and haven't like having `tval`, `sval`, and soon to be `gval` and `aval` with PR #764. Perhaps it would be easiest to add a commit to PR #764 that replaces all of the above with `i8val`? Would that work for you?

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Yes, looks like I forgot to remove the code at line 932.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Yes, you are correct. I will remove it.

MiCurry:
Okay, this branch has been updated with the requested changes as well. The outer if statement containing the comparison to `missing_value` and the pixel value has been removed and replaced with a compound if statement that only accumulates values if the cell is a land cell and if the pixel is not a missing_value. That way, if a pixel maps to an own cell that is not a land cell and/or is a missing value, its value will not be accumulated and `all_pixels_mapped_to_halo_cells` will be set correctly.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                            !
+                            ! When a pixel maps to a non-land cell or is not a missing value, the values are not accumulated
+                            ! aboved; however, these pixels may still reside in an own cell, in which case we will still need
+                            ! to push the tile's neighbors unto the stack for processing.
+                            !
+                            if (res % id <= nCellsSolve) then
+                                all_pixels_mapped_to_halo_cells = .false.
+                            end if
+                        end if
+                    end if
```
mgduda, line 942: Just for consideration, there's an alternative for the code between lines 899 and 942 that doesn't require any duplication of code and that doesn't require any nested if-tests:
```fortran
                    !
                    ! If the nearest cell center is that of an outer-boundary cell (i.e., bdyMaskCell == nBdyLayers),
                    ! we need to check whether the pixel is actually contained in that cell, since the cell center
                    ! might simply have been the closest point to a pixel that lies outside of a limited-area domain
                    !
                    if (bdyMaskCell(res % id) == nBdyLayers .and. &
                        .not. mpas_in_cell(xPixel, yPixel, zPixel, xCell(res % id), yCell(res % id), zCell(res % id), &
                                           nEdgesOnCell(res % id), verticesOnCell(:,res % id), xVertex, yVertex, zVertex)) then

                           ! No further processing for this pixel is needed...
                           cycle

                    end if

                    !
                    ! This field only matters for land cells. If the cell containing this pixel is a land cell,
                    ! and if the pixel isn't a missing value, accumulate the pixel value in the cell.
                    ! NB: The i8val value is not scaled by scale_factor and neither is missing_value, so no
                    !     scaling is required in the comparison, below.
                    !
                    if (landmask(res % id) == 1 .and. i8val /= int(missing_value, kind=I8KIND)) then
                        snoalb_integer(res % id) = snoalb_integer(res % id) + i8val
                        nhs(res % id) = nhs(res % id) + 1
                    end if

                    !
                    ! Regardless of whether we accumulate a pixel value, we need to consider whether we have pixels
                    ! from a tile that map to owned (i.e., non-halo) cells
                    !
                    if (res % id <= nCellsSolve) then
                        all_pixels_mapped_to_halo_cells = .false.
                    end if
```
@MiCurry What do you think of this alternative?

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -686,12 +686,15 @@ end function mpas_geotile_mgr_gen_tile_name
...
     !> \date    02/20/2020
     !> \details
-    !> Given a tile, translate the relative tile coordinates, i, j, of that
-    !> tile to a latitude and longitude coordinate. Upon success, lat, lon
-    !> will be in the range of -1/2 * pi to 1/2 * pi and 0 to 2.0 * pi, respectively.
+    !> Given a tile, translate the relative tile coordinates, i, j, of that tile to a
+    !> latitude and longitude coordinate. Upon success, lat, lon will be in the range
+    !> of -1/2 * pi to 1/2 * pi and 0 to 2.0 * pi, respectively. If supersample_fac is
+    !> present, then the tile will be subdivided into 2 * supersample_fac pixels
+    !> (supersample_fac for both the north-south and east-west directions).
```
mgduda, line 693: @MiCurry As we had discussed, I think, at a minimum, `2 * supersample_fac` should be `supersample_fac ^ 2`. But, the phrase "the tile will be subdivided into ..." might suggest to some readers that it is the `mpas_geotile_mgr_to_latlon` routine that is doing the subdividing, when in fact it is the calling code that has chosen to sub-divide a tile, and it's this routine that simply needs to take that into account when computing lat/lon coordinates.

MiCurry: @mgduda how does this sound:

> Given a tile, translate the relative tile coordinates, i, j, of that tile to a corresponding latitude and longitude coordinate.
> 
> If supersample_fac is present, the latitude, longitude coordinates returned will represent pixel centers of a tile that has been subdivided into 2 ^ supersample_fac pixels. Therefore, the calling code will need to pass in supersampled i, j coordinates as each i, j coordinate will represent a divided tile with supersample_fac * (tile_nx/tile_ny + tile_bdr) pixels.
> 
> Upon success, lat and lon will be in the range of -1/2 * pi to 1/2 pi and 0 to 2.0 * pi, respectively.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: @mgduda I like this a lot better. I'll implement it for this PR.

MiCurry: Actually, thinking a bit more about this change, I’d like to forgo it for now. I think I had mentioned earlier that I would like to remove the duplicated code, but I figured it would be best if we did it when we wrote a unified function. I’d also like to keep this code as more or less consistence with the other interpolated fields, knowing that it (and the other interpolations) will be updated in a future PR.

src/core_init_atmosphere/mpas_geotile_manager.F
MiCurry: @mgduda I have just pushed an update with changes to the above. Let me know if you think it still needs more work. Thanks for the feedback on this.

src/core_init_atmosphere/mpas_init_atm_static.F
mgduda: That's fine. In this case, there are a few minor mistakes in the comments (which would have been replaced by the proposed code and comments) that we should fix before merging this PR. I'll leave comments for these.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                    call mpas_kd_search(tree, (/xPixel, yPixel, zPixel/), res)
+
+                    if (bdyMaskCell(res % id) < nBdyLayers) then
+                        !
+                        ! This field only matters for land cells, and for all but the outermost boundary cells,
+                        ! we can safely assume that the nearest model grid cell contains the pixel (else, a different
+                        ! cell would be nearest)
+                        !
+                        ! Since values are currently not scaled, we can compare missing_value, which also
+                        ! isn't scaled, we can compare the two without scaling.
```
mgduda, line 906: This comment says "we can compare" twice in the same sentence.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                        if (landmask(res % id) == 1 .and. i8val /= int(missing_value, kind=I8KIND)) then
+                            snoalb_integer(res % id) = snoalb_integer(res % id) + i8val
+                            nhs(res % id) = nhs(res % id) + 1
+                        end if
+
+                        !
+                        ! When a pixel maps to a non-land cell or is not a missing value, the values are not accumulated
```
mgduda, line 913: Should "is not a missing value" be "is a missing value"?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                        ! aboved; however, these pixels may still reside in an own cell, in which case we will still need
```
mgduda, line 914: "aboved" -> "above"
"own" -> "owned"

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                        ! to push the tile's neighbors unto the stack for processing.
```
mgduda, line 915: "unto" -> "onto"

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                            all_pixels_mapped_to_halo_cells = .false.
+                        end if
+                    ! For outermost land cells, additional work is needed to verify that the pixel
+                    ! actually lies within the nearest cell
+                    else
+                        if (mpas_in_cell(xPixel, yPixel, zPixel, xCell(res % id), yCell(res % id), zCell(res % id), &
+                                          nEdgesOnCell(res % id), verticesOnCell(:,res % id), xVertex, yVertex, zVertex)) then
+
+                            ! Since values are currently not scaled, we can compare missing_value, which also
+                            ! isn't scaled, we can compare the two without scaling.
```
mgduda, line 927: The comment above applies here, too.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                            if (landmask(res % id) == 1 .and. i8val /= int(missing_value, kind=I8KIND)) then
+                                snoalb_integer(res % id) = snoalb_integer(res % id) + i8val
+                                nhs(res % id) = nhs(res % id) + 1
+                            end if
+
+                            !
+                            ! When a pixel maps to a non-land cell or is not a missing value, the values are not accumulated
+                            ! aboved; however, these pixels may still reside in an own cell, in which case we will still need
+                            ! to push the tile's neighbors unto the stack for processing.
```
mgduda, line 936: The comment above applies here, too.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+
+                        !
+                        ! When a pixel maps to a non-land cell or is not a missing value, the values are not accumulated
+                        ! aboved; however, these pixels may still reside in an own cell, in which case we will still need
+                        ! to push the tile's neighbors unto the stack for processing.
+                        !
+                        if (res % id <= nCellsSolve) then
+                            all_pixels_mapped_to_halo_cells = .false.
+                        end if
+                    ! For outermost land cells, additional work is needed to verify that the pixel
```
mgduda, line 920: The `else` block, below, all cells (not just land cells), so we could consider removing the word "land" in this comment.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Thanks for the catch

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Yes!

MiCurry:
Thanks for the additional review, @mgduda. I've pushed another change that addresses your latest comments.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -829,107 +832,152 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                    call mpas_latlon_to_xyz(xPixel, yPixel, zPixel, sphere_radius, lat_pt, lon_pt)
+                    call mpas_kd_search(tree, (/xPixel, yPixel, zPixel/), res)
+
+                    if (bdyMaskCell(res % id) < nBdyLayers) then
+                        !
+                        ! This field only matters for land cells, and for all but the outermost boundary cells,
+                        ! we can safely assume that the nearest model grid cell contains the pixel (else, a different
+                        ! cell would be nearest).
+                        !
+                        ! Since values in i8val are not yet scalled, we can compare them to missing_value, which
```
mgduda, line 905: "scalled" should be "scaled" (and the same in the duplicated code, below).

src/core_init_atmosphere/mpas_geotile_manager.F
mgduda: The updated version looks good.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: @mgduda good catch. I just pushed an update that corrects the above.

PR #751, opened 2020-11-10
Remove ocean core from COMPASS

xylar:
This merge removes the ocean core from COMPASS, because it has moved to https://github.com/MPAS-Dev/compass


xylar:
@mark-petersen and @sbrus89, I'm not sure when we will be ready to remove the ocean core from COMPASS on MPAS-Model but hopefully soon.  This is the PR to do that when we are ready.

sbrus89:
@xylar, thanks for the heads up.

mark-petersen (approved):
Yes, this looks right to me. Will try to merge in the coming week.

mark-petersen:
@matthewhoffman do you want to make a similar PR for land ice? We could merge both at the same time, for a single E3SM PR.

xylar:
@mark-petersen, I don't think `landice` is there yet. @matthewhoffman hasn't had a chance to review https://github.com/MPAS-Dev/compass/pull/3 yet.  While it would be tidy to have both branches remove COMPASS at the same time, I don't think it's necessary.

mark-petersen:
OK, thanks. It is not necessary to do remove them together.

xylar:
>  It is not necessary to do remove them together.

Right, as long as each branch only removes the code it "owns", meaning `develop` deletes the framework, we should be fine after a merge back to `develop`.

sbrus89 (approved):
This looks good to me as well.

vanroekel (approved):
Does not affect e3sm simulations -  approving for merging

PR #754, opened 2020-11-13
Align atmospheric coupling with CIME and Icepack for E3SM V2

proteanplanet:
This change aligns atmospheric coupling with CIME ocean flux calculations being used in E3SM Version 2 at https://github.com/ESMCI/cime/blob/master/src/share/util/shr_flux_mod.F90.  It also aligns the code with Icepack at https://github.com/CICE-Consortium/Icepack/blob/master/columnphysics/icepack_atmo.F90.   The code has been tested extensively in E3SM Version 2 candidiate configurations, and is required to help address polar biases. 

We request that this pull request is expedited to help finalized E3SM Version 2.


proteanplanet:
This was mistakenly submitted to master instead of seaice/develop.

PR #755, opened 2020-11-13
Align atmospheric coupling with CIME and Icepack for E3SM V2

proteanplanet:
This change aligns atmospheric coupling with CIME ocean flux calculations being used in E3SM Version 2 at https://github.com/ESMCI/cime/blob/master/src/share/util/shr_flux_mod.F90. It also aligns the code with Icepack at https://github.com/CICE-Consortium/Icepack/blob/master/columnphysics/icepack_atmo.F90. The code has been tested extensively in E3SM Version 2 candidate configurations and is required to help address polar biases.

We request that this pull request is expedited to help finalized E3SM Version 2.


njeffery (approved):
I tested these changes in stand a lone MPAS-SI.  It's nonBFB so regression failed as expected but restartability and parallelism passed for standard and bgc configurations.  Andrew already tested the fidelity of the simulation in fully coupled mode and found improvements.

mark-petersen:
@akturner and @proteanplanet, on this PR, Andrew merged `master` into `seaice/develop` as the first commit in this PR aeb2da0, on Aug 5. It’s not the normal order of our cascading merges, but it looks like it had no impact, since we merged `develop` into `seaice/develop` since Aug 5, unrelated to this PR. So let's just leave this one be.

@proteanplanet just for future reference, this would be the normal order:
1. Make new PRs from `seaice/develop`, and back to `seaice/develop`
2. For framework changes, new PRs from `develop`, and back to `develop`

The person in charge of the E3SM PR does:

3. Merge cores from `CORE/develop` to `e3sm/develop`
4. or merge `develop` to `e3sm/develop`

On the rare occasion of an MPAS release:

5. Merge `CORE/develop` to into `develop` and then `develop` to `master`.

proteanplanet:
@mark-petersen Thanks. Got it. You are right, it had no impact, but I won't do it again.

PR #756, opened 2020-11-13
Bugfix for PGI failure after diagnostic solve breakup

mattdturner:
Add back the retrieval of edgeSignOnCell from meshPool to the
ocn_vert_transport_velocity_top subroutine on mpas_ocn_diagnostics.F.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -1489,6 +1489,7 @@ contains
       integer :: iEdge, iCell, k, i, nCells
 
       real (kind=RKIND) :: flux, invAreaCell, div_hu_btr
+      integer, dimension(:, :), pointer :: edgeSignOnCell
 
       ! Scratch Arrays
       ! projectedSSH: projected SSH at a new time
@@ -1507,6 +1508,8 @@ contains
         vertAleTransportTop=0.0_RKIND
         return
       end if
+
+      call mpas_pool_get_array(meshPool, 'edgeSignOnCell', edgeSignOnCell)
 
       ncells = nCellsAll
 


mattdturner:
After adding back this pointer retrieval, 1 of the Compy tests that previously was FAILing is now passing…
```
    SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi.C.20201112_154934_b3ucd2.cpl.hi.0001-01-06-00000.nc matched cpl.hi.0001-01-06-00000.nc
PASS

  Most recent bless: sha:8697827cb20c3290b39f0236937c16c749752883 date:2020-10-12_10:19:36
```
Note that I haven’t re-run any of the MPAS-O standalone tests, or any of the E3SM tests on Cori after adding back the commit.

mark-petersen:
Is `edgeSignOnCell` special? I thought all variables could be retrieved from the mesh pool now, and don't need individual declarations and `mpas_pool_get_array` lines.

mattdturner:
There are a few select variables that are currently defined as `integer` in Registry.xml, but defined a `real` in `ocn_mesh`.  Those variables are:

- edgeMask
- vertexMask
- cellMask
- highOrderAdvectionMask
- edgeSignOnCell
- edgeSignOnVertex
- boundaryEdge
- boundaryVertex
- boundaryCell

The issue that arises here is that `ocn_diagnostic_solve` is called between the call to `ocn_meshCreate` and `ocn_meshUpdateFields`.  It appears that `edgeSignOnCell` is also modified somewhere between those 2 `ocn_mesh` routines.  So `ocn_diagnostic_solve` was using the `real` `edgeSignOnCell` array from the module, which wasn't yet kept up to date with the `integer` `edgeSignOnCell` from the pool.  

This will not be an issue once the rest of the model is updated to use `ocn_mesh`.  I decided that it would be quicker to just grab the array from `meshPool` right now, instead of digging to find where it was updated between the two `ocn_mesh` routines.

mark-petersen (approved):
Passes nightly regression suite with gnu 5 on grizzly. Approved based on @mattdturner's tests with PGI.

PR #757, opened 2020-11-13
Remove scaling from read_geogrid

MiCurry:
In preparation for a new parallel interpolation method, this merge removes
the scaling done within the read_geogrid C function and moves it into the
init_atm_static function.

For continuous datatypes, such as terrain height or snow albedo, the new
parallel interpolation method will accumulate values as integers and then
convert them to reals once all accumulation is complete. This method ensures
correctness across processes.

After this commit any new interpolation interpolation of a dataset will need to
scale any read tiles *after* calling read_geogrid.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -34,7 +34,7 @@
 
  interface
     subroutine read_geogrid(fname, rarray, nx, ny, nz, isigned, endian, &
-                            scalefactor, wordsize, status) bind(C)
+                            wordsize, status) bind(C)
        use iso_c_binding, only : c_char, c_int, c_float, c_ptr
        character (c_char), dimension(*), intent(in) :: fname
        type (c_ptr), value :: rarray
@@ -43,7 +43,6 @@
        integer (c_int), intent(in), value :: nz
        integer (c_int), intent(in), value :: isigned
        integer (c_int), intent(in), value :: endian
-       real (c_float), intent(in), value :: scalefactor
        integer (c_int), intent(in), value :: wordsize
        integer (c_int), intent(inout) :: status
     end subroutine read_geogrid
@@ -88,6 +87,7 @@
  integer,dimension(:),allocatable  :: nhs
  integer,dimension(:,:),allocatable:: ncat
       
+ real(kind=RKIND), pointer :: scalefactor_ptr
  real(kind=c_float):: scalefactor
  real(kind=c_float),dimension(:,:,:),pointer,contiguous :: rarray
  type(c_ptr) :: rarray_ptr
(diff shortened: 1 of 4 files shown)


src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1152,7 +1156,7 @@ subroutine init_atm_static(mesh, dims, configs)
     call mpas_f_to_c_string(fname, c_fname)
 
     call read_geogrid(c_fname,rarray_ptr,nx,ny,nz,isigned,endian, &
-                      scalefactor,wordsize,istatus)
+                      wordsize,istatus)
     call init_atm_check_read_error(istatus,fname)
     vegfra(-2:1250,-2:1253,1:12) = rarray(1:1253,1:1256,1:12)
```
mgduda, line 1161: Although the scaling factor is 1.0, should we scale `rarray` by `scalefactor` here (and a few other places, below) for consistency?

MiCurry: Yes, that is probably the best idea. Should we also update the terrain, landuse and soil category to also be scaled by their scale factor? If so I'll think we'll need to do so after computing the average height for each cell (or most dominate category). I.E.:
``` Fortran
 do iCell = 1, nCells
    ter(iCell) = real(ter_integer(iCell), kind=R8KIND) / real(nhs(iCell), kind=R8KIND)
    ter(iCell) = ter(iCell) * scalefactor
 end do
```

mgduda: I think we should definitely update the terrain field -- in principle, if the terrain dataset did have a non-unity scaling factor (e.g., a scale factor of 0.001, so that we had vertical resolution of 1 mm), we wouldn't get correct results without scaling.

I'm unsure that scaling would ever be used for categorical fields, so I'm fine with not performing any scaling for land use category and soil category. Categories are necessarily integer values, and it's hard to imagine a dataset in which, e.g., someone provided categories 100, 110, 120, 130, etc. but represented them in a file with values 10, 11, 12, 13 and a scaling factor of 10.0.

MiCurry: That makes sense! Okay, I just pushed an update with the above request.

mgduda: I just scanned through the diffs so I may be mistaken, but are we still missing code to scale `rarray` before lines 1177 and 1368?

MiCurry: You were correct. Apologies about missing that one. I have just pushed an update. We should now be scaling `rarray` after any call to `read_geogrid`.

mgduda: It looks like we're still missing code to scale `rarray` around L.1368.

MiCurry: Okay, scaling has been added to rarray around line 1368.

mgduda:
@MiCurry As we discussed, let's add a second commit to remove the scaling factor from the calls to `read_geogrid` in `mpas_init_atm_gwd.F`.

MiCurry:
> @MiCurry As we discussed, let's add a second commit to remove the scaling factor from the calls to `read_geogrid` in `mpas_init_atm_gwd.F`.

Thanks for the reminder, I just updated this branch with a second commit that removes the scaling factor from calls to `read_geogrid`.

mgduda (approved):

PR #759, opened 2020-11-19
Consolidate ocean diagnostics modules

mark-petersen:
There had been two diagnostics modules, but one only had one subroutine in it.  Removed `mpas_ocn_diagnostics_routines.F` and moved the relevant subroutine to mpas_ocn_diagnostics.F.


mark-petersen:
Compared to 30ca47000 on `ocean/develop` (before #745). Nightly regression suite is bfb using both intel19 and gnu6, both optimized. Also tested both with debug.

mattdturner (approved):
Approved based on visual inspection and above testing

mark-petersen:
Tested in E3SM by @mattdturner, see https://github.com/E3SM-Project/E3SM/pull/3937#issuecomment-730564590

PR #763, opened 2020-11-24
Vertical Lagrangian-remap design doc

cbegeman:
This design doc is for the implementation of the vertical Lagrangian-remap method. This development is motivated by the occurrence of instabilities in cases of evolving ice-shelf geometry, but should have broad advantages including higher-order accuracy and lifting the vertical CFL constraint.

### DIFF
--- a/docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
+++ b/docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
@@ -1,0 +1,124 @@
+
+Vertical Lagrangian-remap method
+======================
+
+date: 2020/11/24
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+
+
+Summary
+-------
+
+The vertical Lagrangian-remap method is to be implemented to support the choice 
+of hybrid vertical coordinates. It can be seen as an extension of the Arbitrary 
+Lagrangian-Eulerian method which is already implemented in MPAS-Ocean, a key 
+difference being that the solution of diasurface transport is coordinate-free. 
+This has the advantage of removing the vertical CFL constraint, which is of 
+particular concern for an evolving ice shelf geometry in terrain-following 
+coordinate schemes. Generalization of the vertical coordinate in the horizontal 
+equations of motion is not in the scope of this development. This development 
+will include some infrastructure for specifying the target grid for regridding. 
+The minimum criteria for success are not degrading the accuracy of the solution 
+or the performance by more than [an acceptable threshold] for several basic test 
+cases... .
+
+*The purpose of this section is to summarize what capability is to be added to
+MPAS-Model through this design process. It should be clear what new code will do
+that the current code does not. Summarizing the primary challenges with respect
+to software design and implementation is also appropriate for this section.
+Finally, this statement should contain a general statement with regard to what
+is "success."*
+
+
+Requirements
+------------
+
+Requirement: name-of-requirement-here
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/11/24
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Requirement: A grid that relaxes to an analytical expression for a target grid.
+
+Requirement: The ability to implement hybrid coordinate systems.
+
+Requirement: XX order accuracy of the solution.
+
+Requirement: Performance is not degraded below an acceptable level.
+
+Requirement: Layer thicknesses are non-negative (and above a minimum thickness).
+
+*Possible additional requirements:
+- re. development of instabilities
+- re. conservation/monotonicity of interpolants
+*
+
+
+*Each requirement is to be listed under a "section" heading, as there will be a
+one-to-one correspondence between requirements, design, proposed implementation
+and testing. Requirements should not discuss technical software issues, but
+rather focus on model capability. To the extent possible, requirements should
+be relatively independent of each other, thus allowing a clean design solution,
+implementation and testing plan.*
+
+
+Algorithmic Formulations (optional)
+-----------------------------------
+
+Design solution: short-description-of-proposed-solution-here
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/11/24
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+*For each requirement, there is a design solution that is intended to meet that
+requirement. Design solutions can include detailed technical discussions of
+PDEs, algorithms, solvers and similar, as well as technical discussion of
+performance issues. In general, this section should steer away from a detailed
+discussion of low-level software issues such as variable declarations,
+interfaces and sequencing.*
+
+
+Design and Implementation
+-------------------------
+
+Implementation: short-description-of-implementation-here
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: YYYY/MM/DD
+
+Contributors: (add your name to this list if it does not appear)
+
+*This section should detail the plan for implementing the design solution for
+requirement XXX. In general, this section is software-centric with a focus on
+software implementation. Pseudo code is appropriate in this section. Links to
+actual source code are appropriate. Project management items, such as git
+branches, timelines and staffing are also appropriate. Pseudo code can be
+included via blocks like*
+
+.. code-block:: python
+
+   def example_function(foo):
+       return foo**2.0
+
+
+Testing
+-------
+
+Testing and Validation: short-description-of-testing-here
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: YYYY/MM/DD
+
+Contributors: (add your name to this list if it does not appear)
+
+*How will XXX be tested, i.e., how will be we know when we have met requirement
+XXX? Which tests from the regression suites are appropriate?  How would they
+need to be configured or modified to test that the new software is working
+properly?  What additions or modifications to the nightly regression suite might
+be made to ensure that the new capability continues to work as expected?*


xylar:
Looks good, thanks for the new PR!

dengwirda:
Thanks for putting this together @cbegeman!

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
+
+Vertical Lagrangian-remap method
+======================
```
xylar, line 3:
```suggestion
Vertical Lagrangian-remap method
================================
```
RestructuredText is strict about these needing to match

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+Lagrangian-Eulerian method which is already implemented in MPAS-Ocean, a key 
+difference being that the solution of diasurface transport is coordinate-free. 
+This has the advantage of removing the vertical CFL constraint, which is of 
+particular concern for an evolving ice shelf geometry in terrain-following 
+coordinate schemes. Generalization of the vertical coordinate in the horizontal 
+equations of motion is not in the scope of this development. This development 
+will include some infrastructure for specifying the target grid for regridding. 
+The minimum criteria for success are not degrading the accuracy of the solution 
+or the performance by more than [an acceptable threshold] for several basic test 
+cases... .
```
xylar, line 25: I think we can leave the threshold ambiguous for now.  We can also fill in which test cases we will use for checking accuracy as the design progresses.  Probably ISOMIP+ Ocean0 would be one, maybe the baroclinic channel?  We can ask @mark-petersen which tests he would suggest.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+
+*The purpose of this section is to summarize what capability is to be added to
+MPAS-Model through this design process. It should be clear what new code will do
+that the current code does not. Summarizing the primary challenges with respect
+to software design and implementation is also appropriate for this section.
+Finally, this statement should contain a general statement with regard to what
+is "success."*
```
xylar, line 32:
```suggestion
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+Requirement: XX order accuracy of the solution.
+
+Requirement: Performance is not degraded below an acceptable level.
+
+Requirement: Layer thicknesses are non-negative (and above a minimum thickness).
+
+*Possible additional requirements:
+- re. development of instabilities
+- re. conservation/monotonicity of interpolants
+*
```
xylar, line 58: Your list are good requirements that should be fleshed out in full paragraphs if possible.

To me, some main requirements are:
* No calculations are made outside of remapping itself using prognostic (or diagnostic) variables both before and after remapping.  This is because the vertical locations of information before and after remapping may be completely unrelated to one another (separated by many layers).  Once we have a variable number of layers, this information could go from being "valid" to "invalid" or visa versa after remapping.
* Remapping should be supported (eventually) in both the RK4 and the split-explicit time stepping schemes.  For performance reasons, the sooner it can be supported in split-explicit, the better.  (In the design below, we may state that we will start with RK4.)
* To the extent that it is feasible, the user should have the ability to control the order of accuracy and other parameters of the remapping (e.g. via namelist options).
* To the degree possible, the code for performing remapping should be kept in its own Fortran module(s) for better readability.

Happy to keep brainstorming

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+Requirement: A grid that relaxes to an analytical expression for a target grid.
+
+Requirement: The ability to implement hybrid coordinate systems.
+
+Requirement: XX order accuracy of the solution.
+
+Requirement: Performance is not degraded below an acceptable level.
+
+Requirement: Layer thicknesses are non-negative (and above a minimum thickness).
+
```
xylar, line 54: I agree with non-negative.  Above a minimum thickness seems like an algorithmic choice rather than a requirement. to me.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Requirement: A grid that relaxes to an analytical expression for a target grid.
+
+Requirement: The ability to implement hybrid coordinate systems.
+
+Requirement: XX order accuracy of the solution.
+
+Requirement: Performance is not degraded below an acceptable level.
```
xylar, line 51: We can flesh out what the requirement might be as we decide which test cases we will use to determine this but I would presume we would want to keep performance within a few percent of what it currently is.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Date last modified: 2020/11/24
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Requirement: A grid that relaxes to an analytical expression for a target grid.
+
+Requirement: The ability to implement hybrid coordinate systems.
+
+Requirement: XX order accuracy of the solution.
```
xylar, line 49: Presumably we would want to show that layer thicknesses, velocities and other prognostic variables are pretty close after a few time steps (e.g. within 1e-10 of their values with the old method) and that the effect in longer simulations within E3SM is not "climate changing", which @vanroekel and @mark-petersen would be able to help us define.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Requirement: Performance is not degraded below an acceptable level.
+
+Requirement: Layer thicknesses are non-negative (and above a minimum thickness).
+
+*Possible additional requirements:
+- re. development of instabilities
```
xylar, line 56: I think it goes without saying that we don't want to cause numerical instabilities but if there are specific instabilities that this approach is at risk of causing, it would certainly be worth requiring that they be avoided.

We do want the approach to be "robust" in the sense that we'd like it to support quite a variety of vertical coordinate choices without causing numerical instabilities.  But some of that may be beyond the scope of just vertical remapping (e.g. relating to horizontal pressure-gradient force).

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,124 @@
...
+- re. conservation/monotonicity of interpolants
```
xylar, line 57: Ooh, good ones!  Conservation and monotonicity should both be added.  I would think we would require conservation in an integral sense (e.g. column volume is the same to machine precision before and after remapping).  Maybe there would be other requirements (e.g. conservation in a layer-wise sense, or at least the option for it).  @dengwirda, what is possible with PPR?

Regarding monotonicity, @dengwirda would be in a better position than I am to suggest the requirements.  But we would certainly want at least user control over now monotonic the reconstruction is.

xylar (commented):
@cbegeman, thanks for getting this going.  We will just keep this PR open and the conversation going until the design doc is completely finished, so it won't get merged anytime soon.  For now, let's focus on the requirements and maybe start sketching out the Algorithmic Formulation / Design Solution for each of them.  Implementation and testing usually get left until later but it doesn't hurt to start playing around with prototype implementation even before we get all the requirements and algorithms nailed down.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: My understanding of PPR is that there is conservation over the layer integral.

cbegeman: I'm not sure how detailed it makes sense to get about the details of the remapping operation in the requirements section. Maybe just include layer-wise conservation?

dengwirda: @cbegeman, @xylar, yes, the remapping methods should conserve the vertical integrals of whatever is being remapped (sum(h_k * psi_k) for example, where h_k is layer thickness and psi_k is the layer mean value of a given variable).

There are a few different options for order-of-accuracy + monotonicity / slope-limiting in PPR: PLM (2nd-order), PPM (3rd-order), PQM (5th-order), and either monotone or WENO flux / slope limiting.

I agree that there's no need to get into all of these details in the doc / requirements, but just for reference:

- PLM: Piecewise Linear Method
- PPM: Piecewise Parabolic Method
- PQM: Piecewise Quartic Method
- Monotone limiting: No under/overshoots, but perhaps more numerical diffusion than is desired, can limit order-of-accuracy to 2nd-order.
- WENO: Weighted Essentially Non-oscillatory limiting: "eps"-sized under/overshoots, less numerical diffusion, should generally achieve full order-of-accuracy.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: I think this is a place we'll have to be a little careful, and think about interactions with the horizontal dynamics / algorithms: e.g. to ensure h > eps (h = layer thickness), the continuity equation

    dh / dt + div(uh) = 0

will have to be discretised using a monotone scheme (evaluating the u*h fluxes via the FCT-type approach used for tracers, perhaps).

xylar: We can point out these considerations but I think we want anything related to horizontal algorithms to be part of a separate design document.  So maybe the requirement here is a relatively trivial one: that the new vertical coordinate being remapped to doesn't include invalid (zero or negative) thicknesses?

xylar: Hmm, on further consideration, I think my version is too trivial to include and it's presumably not what you had in mind, @cbegeman.  So maybe this requirement is one for a future design process that includes changes to horizontal equations, too.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,302 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Date last modified: 2020/12/04
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+The vertical Lagrangian-remap can be characterized by the following equations, as written in Griffies and Adcroft (2020):
+
+Lagrangian layer motion
+.. math::
+   w^{grid}_{k,*} = -D^n_{k,*}
```
xylar, line 133: rst is picky about spacing [(see the documentation)](https://www.sphinx-doc.org/en/1.0/ext/math.html#directive-math)  I think you need some extra white space for this to work out:
```suggestion
Lagrangian layer motion

.. math::

      w^{grid}_{k,*} = -D^n_{k,*}
```
Even so, it probably won't show up in the GitHub preview and you'd need to do a local build of the documentation to see it.  Let me know if you need instructions for doing that.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,302 @@
...
+
+Requirement: remapping
+^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/04
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Requirement: A grid that relaxes to an analytical expression for a target grid with non-negative layer thicknesses.
+
```
xylar, line 39: I wonder if it might make sense to expand this into two requirements.  One would be for a mechanism to specify the vertical grid after remapping.  The other would be for a way to specify a target grid with relaxation.  The latter might even be moved to follow-on work, since I think we might want to keep the current design focused on the "bare minimum" to get Lagrangian vertical remapping into MPAS-Ocean.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,302 @@
...
+Requirement: hybrid coordinates
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/04
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Requirement: The ability to implement hybrid coordinate systems.
```
xylar, line 47: I'm not 100% clear on what hybrid coordinates means in this context.  Could this be fleshed out?  Is this maybe a requirement for a future design document?

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,302 @@
...
+
+Requirement: user specifications
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/04
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Requirement: To the extent that it is feasible, the user should have the ability to control the order of accuracy and other parameters of the remapping (e.g. via namelist options).
```
xylar, line 56:
```suggestion
To the extent that it is feasible, the user should have the ability to control the order of accuracy and other parameters of the remapping (e.g. via namelist options).
```
Maybe "Requirement: " is redundant here and elsewhere?

mark-petersen:
@cbegeman thanks for starting this design document. I was trying to read a compiled version. @xylar what's the easiest way to do that for a single page like this? I tried the [compass instructions](https://mpas-dev.github.io/compass/stable/developers_guide/building_docs.html), but the `Makefile` is missing here (might be in another PR). Looks like `rst2html` works for a quick html version, but not for equations.

cbegeman:
> @cbegeman thanks for starting this design document. I was trying to read a compiled version. @xylar what's the easiest way to do that for a single page like this? I tried the [compass instructions](https://mpas-dev.github.io/compass/stable/developers_guide/building_docs.html), but the `Makefile` is missing here (might be in another PR). Looks like `rst2html` works for a quick html version, but not for equations.

Thanks @mark-petersen. I rebased onto the @xylar's design doc template branch so the Makefile should be there now.

xylar:
@cbegeman, could you also add `vert-lagrang-remap/vert-lagrang-remap` to `docs/ocean/design_docs/index.rst` so the design doc gets included when the documentation gets built? Thanks!

cbegeman:
@mark-petersen I have a few updates to make to the design doc so you can wait until I ping you to comment

cbegeman:
@mark-petersen I think this is far-enough along for you to start providing input. We've largely been looking at the RK4 implementation and haven't thought too much about the split-explicit implementation, so your recommendations there would also be helpful.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Per the discusion in our 12/8 meeting, hybrid coordinates are left for a future design doc

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Specification of the target grid beyond the initial z-levels is reserved for a future design doc.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+vertical, but we do not require this result for this implementation to be 
+successful.
+
+Generalization of the vertical coordinate is not in the scope of this development. 
+Additionally, the infrastructure for specifying a target grid that differs from 
+the existing z-star and z-tilde options is reserved for future development.
+The minimum criteria for success are not degrading the accuracy of the solution 
+or the performance by more than [an acceptable threshold] for several basic test 
+cases. 
+*[not quite sure how to write the success statement if not "meets requirements"]*
```
cbegeman, line 32: @dengwirda @xylar Thoughts on criteria for success?

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/04
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+There are several choices to make for the remapping operation to balance accuracy, 
+stability and computational cost. To the extent that it is feasible, the user 
+should have the ability to control the order of accuracy of the remapping method,
+[Darren, other options here?] and other remapping options (e.g., via namelist options).
```
cbegeman, line 78: @dengwirda Feel free to elaborate on remapping options here.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+Requirement: not climate changing
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+For the same target grid, time-stepping scheme and problem, the solution for 
+layer thicknesses and key prognostic variables using VLR is within a non-
+climate-changing threshold value [presumably greater than machine precision?]
```
cbegeman, line 99: @vanroekel Is there an agreed-upon threshold for allowable differences in the solution?

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+
+Requirement: performance
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/04
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Performance is not degraded below an acceptable level *which is to be 
+determined*.
```
cbegeman, line 110: @dengwirda @xylar I think we settled on leaving this TBD for now, but jump in if you have thoughts on this.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+The choice between the two is discussed in the Implementation section.
+
+Algorithm Design: Grid is sufficiently conditioned for accuracy and stability
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+*Discuss damping (in time) function here*
```
cbegeman, line 230: @dengwirda This might not be the right place to discuss the damping function for the rate of grid movement, but I wasn't sure it should be a separate requirement either.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
vanroekel: @cbegeman if the changes are nonBFB I think we have to go with expert judgement on if we are okay with the changes.  It would be very interesting to try and use the capability from Salil Mahajan to test for an unchanged climate.  I'm not comfortable putting a specific threshold on this requirement.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+
+*Discuss smoothing (in space) function here*
```
cbegeman, line 232: @dengwirda In your notes you showed something like Laplacian smoothing, but I imagine we want to choose a method that is volume conserving (especially for the ice-shelf basal surface). Thoughts?

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
philipwjones: Generally, any performance degradation should be kept to less than a few percent in overall throughput (SYPD). The smaller, the better - if we keep adding stuff, even small perf hits add up over time. Higher performance degradation can be justified for required changes that have significant (positive) impacts on simulation quality. We're happy to help with the design - catching things early is better so getting a performant design up front will avoid a complete refactor later.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+- *Add more here*
+
+:code:`layerThickness(tlev=2)` is then overwritten with the new layer thickness.
+
+After remapping, :code:`ocn_diagnostic_solve` is called. This is needed to 
+compute the density and pressure fields based on the remapped ocean state and
+the diagnostic field :code:`vertVelocityTop` which is the vertical velocity 
+through the top of the layer. This is only used as a diagnostic variable for 
+computing the MOC streamfunction. None of the mixing parameterizations require
+a vertical velocity (Eulerian or diasurface velocity).
```
cbegeman, line 328: Re. mixing parameterizations, per my conversation with @vanroekel.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+After determining the target grid, perform the following steps:
+
+#. Apply damping function to limit the rate of grid movement in one timestep 
+   [and enforce a maximum movement limit?]. 
+#. Optional: Assign :math:`h_k^{t+1}` to :math:`h_k^{lg}` if 
+   :math:`h_k^{t+1} - h_k^{lg}` is less than a minimum thickness alteration. 
+   This motivated by accuracy considerations, as each remapping may introduce 
+   errors. *Darren, would this improve PPR computational performance?*
```
cbegeman, line 369: @dengwirda Would this provide any benefits for PPR's accuracy/efficiency?

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,540 @@
...
+
+Namelist options:
+
+- frequency with which remapping should be performed (on which timestep)
+- order of the remapping
+- order of edge slope estimates
+- monotone limiter
+- boundary condition option
+- option to output some diagnostics?
+- *Some other remapping options here*
```
cbegeman, line 397: @dengwirda Feel free to contribute here

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Thanks @philipwjones. When we get a little further along with knowing what needs to happen in the implementation phase we'll get in touch with you to hone the design for performance.

dengwirda: I agree with @philipwjones, but I think we should revisit this on a second pass (after first getting the remap working, and ensuring consistency with the flux-form implementation).

With performance in mind, I can imagine rewriting some of the remap functionality from PPR direct in MPAS-O, to ensure compatibility with data-structures, eliminate copies to temp arrays, etc. Will be great to work with you on this @philipwjones, when we get a little further progressed.

philipwjones: That's fine, but an early heads-up - for GPU performance, we want to see local tightly nested loops (with both horiz and vertical loops nested together). So please don't code these as 1-d column-based routines. We've been having to back out that design decision for a lot of other modules. See, for example:

https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1041499662/Fortran+GPU+Refactoring+Guide+Using+Directives

cbegeman: @philipwjones Thanks for the link! That looks very useful.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: In cases where we're remapping back to known standard grids (i.e. z*, etc) there's no need for any smoothing of the target grid in time or space, since z^{t+1} / h^{t+1} can be determined diagnostically from SSH, etc.

In the "full" hybrid coordinate version, where the target grid is determined via heuristics, the spatial and time-based smoothing will come into play. Perhaps these details are best left for future revisions at this stage though?

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: One nice thing about the remap formulation is that it should be conservative irrespective of how layers are distributed within a column (as long as the same SSH + bottomDepth are used). This is due to the integral formulation of the remap (it preserves finite-volume style DoF essentially).

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: This is an interesting idea to investigate! As you say, if there's insufficient difference between the Lagrangian grid and the (pending) target grid z^{t+1}, it may be worthwhile to simply skip the remapping step. This should be both cheaper, and (may) reduce the overall amount of numerical dissipation. 

As above, the time/space smoothing for z^{t+1} should only come into play for non-diagnostic target grid choices.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: I think you've covered all the options that PPR supports here. At this stage, let's not be overly specific about which options are the "right" ones to use. Hopefully initial comparisons against the flux-form variant will offer some guidance.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: I was thinking about an additional horizontal smoothing step outside of remapping, but I think you're right that this horizontal smoothing can be left to future development when we implement hybrid and other grid options.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Good point. Maybe this belongs with the next "hybrid-grid" design doc.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: Totally agree @philipwjones! Perhaps we can arrange a meeting on this at some point early next year, and you can advise on best practice loop/data layouts, etc. The remap code has a somewhat different structure to "standard" advection --- I'd be happy to run through the formulation details too.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Conversation continued below

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Now addressed by https://github.com/MPAS-Dev/MPAS-Model/pull/763/commits/6c1f1a91d466c47f049d3c4db0cc46ad7f9d585d which removes discussion of smoothing from this design doc.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Now addressed by https://github.com/MPAS-Dev/MPAS-Model/commit/6c1f1a91d466c47f049d3c4db0cc46ad7f9d585d which removes discussion of smoothing from this design doc.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: > As above, the time/space smoothing for z^{t+1} should only come into play for non-diagnostic target grid choices.

Smoothing discussion removed by https://github.com/MPAS-Dev/MPAS-Model/commit/6c1f1a91d466c47f049d3c4db0cc46ad7f9d585d

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
xylar: I think what @vanroekel discussed below is correct.  It will be some combination of "expert judgment" and somewhat more quantitative criteria such as Salil's test to determine if the code modifications change the climate outside of ensemble variability.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
+.. role:: red
```
xylar, line 1: @cbegeman, out of curiosity, what does this do?

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+The conservation of momentum, volume, and tracer equations in MPAS-Ocean are:
+
+.. math::
+
+   \frac{\partial u_k}{\partial t} + q_k h_k u_k^{normal} + \overline{w^t \delta z^t u} = -\frac{1}{\rho_0} \nabla p_k - \frac{rho_k g}{\rho_0} \nabla z_k - \nabla K_k + [D_h^u]_k + [D_{\nu}^u]_k + F_k^u
```
xylar, line 149:
```suggestion
   \frac{\partial u_k}{\partial t} + q_k h_k u_k^{normal} + \overline{w^t \delta z^t u} = -\frac{1}{\rho_0} \nabla p_k - \frac{\rho_k g}{\rho_0} \nabla z_k - \nabla K_k + [D_h^u]_k + [D_{\nu}^u]_k + F_k^u
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+   
+   \frac{\partial h_k}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k) + w_k^t - w_{k+1}^t = 0
+
+   \frac{\partial (h_k \phi_k)}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k \phi_k) + \overline{\phi}_k^t w_k^t - \overline{\phi}_{k+1}^t w_{k+1}^t = [D_h^{\phi}]_k + [D_{\nu}^{\phi}]_k + F_k^{\phi}
+   
+For the Lagrangian step, the vertical velocity through the top of the cell, :math:`w_k^t`, is set to zero in all of the above equations. Thus, these equations simplify to:
+
+.. math::
+
+   \frac{\partial u_k}{\partial t} + q_k h_k u_k^{normal} = -\frac{1}{\rho_0} \nabla p_k - \frac{rho_k g}{\rho_0} \nabla z_k - \nabla K_k + [D_h^u]_k + [D_{v}^u]_k + F_k^u
```
xylar, line 159:
```suggestion
   \frac{\partial u_k}{\partial t} + q_k h_k u_k^{normal} = -\frac{1}{\rho_0} \nabla p_k - \frac{\rho_k g}{\rho_0} \nabla z_k - \nabla K_k + [D_h^u]_k + [D_{v}^u]_k + F_k^u
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+---------------------------
+
+Algorithm Design: Ability to perform vertical Lagrangian-remapping
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+The conservation of momentum, volume, and tracer equations in MPAS-Ocean are:
```
xylar, line 145:
```suggestion
The conservation of momentum, volume, and tracer equations in MPAS-Ocean `(Petersen et al. 2014) <https://www.sciencedirect.com/science/article/abs/pii/S1463500314001796>`_ are:
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+   
+   \frac{\partial h_k}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k) = 0
+
+   \frac{\partial (h_k \phi_k)}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k \phi_k) = [D_h^{\phi}]_k + [D_{v}^{\phi}]_k + F_k^{\phi}
+   
+The time-stepping algorithm (RK4 or split-explicit) yields the updated 
+variables :math:`u_k^{lg},h_k^{lg},\phi_k^{lg}`, where the superscript
+*lg* is used to designate the values after the Lagrangian step.
+
+Note that the vertical mixing terms :math:`D_h^u, D_v^h, D_h^{\phi}, D_v^{\phi}`
```
xylar, line 169:
```suggestion
Note that the vertical mixing terms :math: D_v^h, D_v^{\phi}` 
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
xylar: I personally find the square bracket notation that @mark-petersen used on the `D_h` and `D_v` terms to be confusing because those square brackets are used in Ringler et al. (2013) to mean a vertical average, whereas here they are just used to handle the fact that there is already a subscript (`h` or `v`) on the terms.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+
+The conservation of momentum, volume, and tracer equations in MPAS-Ocean are:
+
+.. math::
+
+   \frac{\partial u_k}{\partial t} + q_k h_k u_k^{normal} + \overline{w^t \delta z^t u} = -\frac{1}{\rho_0} \nabla p_k - \frac{rho_k g}{\rho_0} \nabla z_k - \nabla K_k + [D_h^u]_k + [D_{\nu}^u]_k + F_k^u
+   
+   \frac{\partial h_k}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k) + w_k^t - w_{k+1}^t = 0
+
+   \frac{\partial (h_k \phi_k)}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k \phi_k) + \overline{\phi}_k^t w_k^t - \overline{\phi}_{k+1}^t w_{k+1}^t = [D_h^{\phi}]_k + [D_{\nu}^{\phi}]_k + F_k^{\phi}
```
xylar, line 153:
```suggestion
   \frac{\partial (h_k \phi_k)}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k \phi_k) + \overline{\phi}_k^t w_k^t - \overline{\phi}_{k+1}^t w_{k+1}^t = [D_h^{\phi}]_k + [D_v^{\phi}]_k + F_k^{\phi}
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+   
+For the Lagrangian step, the vertical velocity through the top of the cell, :math:`w_k^t`, is set to zero in all of the above equations. Thus, these equations simplify to:
+
+.. math::
+
+   \frac{\partial u_k}{\partial t} + q_k h_k u_k^{normal} = -\frac{1}{\rho_0} \nabla p_k - \frac{rho_k g}{\rho_0} \nabla z_k - \nabla K_k + [D_h^u]_k + [D_{v}^u]_k + F_k^u
+   
+   \frac{\partial h_k}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k) = 0
+
+   \frac{\partial (h_k \phi_k)}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k \phi_k) = [D_h^{\phi}]_k + [D_{v}^{\phi}]_k + F_k^{\phi}
```
xylar, line 163:
```suggestion
   \frac{\partial (h_k \phi_k)}{\partial t} + \nabla \cdot (h_k \mathbf{u}_k \phi_k) = [D_h^{\phi}]_k + [D_v^{\phi}]_k + F_k^{\phi}
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
xylar: My suggestion would be `D_{h,k}^u`, etc.

xylar: I take that back. After a closer look, I think Ringler et al. (2013) is using square brackets in the same way as Petersen et al. (2014), so it's fine as it is.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+computation follows remapping due to remapping errors in vertical gradients of 
+prognostic variables. We do not intend to test this at this time.
+
+The target grid needs to be determined after the solution for prognostic 
+variables so that the vertical Lagrangian-remapping method is general enough to
+be used with coordinate systems that depend on the ocean state (this includes 
+the z-star coordinate system in which SSH perturbations are vertically 
+distributed between layers). We do not present an algorithmic design for 
+regridding to coordinate systems not already supported in MPAS-Ocean, as this 
+will be the subject of future development. For now, the target grid based on a
```
xylar, line 186:
```suggestion
will be the subject of future development. For now, the target grid is based on a 
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+
+.. math::
+
+   w = - \nabla \cdot (h_k \mathbf{u}_k) - (h_k^{t+1} - h_k^t)/dt
+
+or
+
+.. math::
+
+   w = (h_k^{t+1} - h_k^lg)/dt
```
xylar, line 219:
```suggestion
   w = (h_k^{t+1} - h_k^{lg})/dt
```

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+   an analytical expression for the grid. 
+   The simplest case is constant z-levels, :math:`z_k^1 = z_k^{init}`.
+   Since :math:`z_k^1` can be a function of the ocean state (e.g., :math:`\rho` 
+   for isopycnal coordinates, regridding doesn't begin until after the solution 
+   for prognostic variables.
+#. Superimpose SSH perturbations according to one of the existing depth-
+   dependent functions, :math:`z_k^2 = z_k^1 + c(z) \: \eta`. As in 
+   :code:`ocn_ALE_thickness`, layer thicknesses are adjusted from the seafloor 
+   upwards. Ideally, there is a single function that is used for both ALE
+   implementations, with and without VLR.
```
xylar, line 270: I'm a bit confused by this notation.  Since `k` is the vertical index, shouldn't this be `z_1`, `z_2`, etc. instead of `z_k^1`, `z_k^2`, etc?

Separate from that confusion, I'm not clear on why the top layer is handled differently than subsequent layers.  I would expect one would have a list of layer-interface locations `z_k^\textrm{target}` that could z-level, z-star or z-tilde for now (and more general in the future).  The top layer would not be special.

xylar: >  Ideally, there is a single function that is used for both ALE implementations, with and without VLR.

I think this should be a requirement (maybe to add to the requirement section even).  Redundant code would be quite risky in my opinion.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+#. Update :code:`layerThicknessEdge` from the updated :code:`layerThickness`
+   using routine :code:`ocn_diagnostic_solve_layerThicknessEdge`. Note that 
+   :code:`zTop, zMid` are updated later when :code:`ocn_diagnostic_solve` is 
+   called.
+
+All of the regridding steps will be performed from a separate module.
+This topic is further addressed in section Implementation: modularity.
+
+
+Remapping step:
```
xylar, line 281: At least for purposes of this design, we should probably settle on a single term ("regridding" or "remapping").  My sense is that, in general, "remapping" is more general and includes the potential for unstructured meshes.  Either term is correct in this context but I might lean toward the more restrictive term "regridding".

xylar: Reading further, I gather that "regridding" here refers to determining the new target grid, while "remapping" is putting the variables onto that target grid.  These uses of the terms would not be consistent with my understanding but may have been adopted in the ALE literature.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+- *Add more here*
+
+:code:`layerThickness(tlev=2)` is then overwritten with the new layer thickness.
+
+After remapping, :code:`ocn_diagnostic_solve` is called. This is needed to 
+compute the density and pressure fields based on the remapped ocean state and
+the diagnostic field :code:`vertVelocityTop` which is the vertical velocity 
+through the top of the layer. This is only used as a diagnostic variable for 
+computing the MOC streamfunction. None of the mixing parameterizations require
+a vertical velocity (Eulerian or diasurface velocity).
```
xylar, line 316: If this is the only usage, it would be worth discussing with @vanroekel, @milenaveneziani and @mark-petersen whether the MOC itself should be computing the this variable as needed when we're in VLR mode.

One consideration is that the MOC also really needs to be computed on an Eulerian (fixed-in-time) or isopicnal grid.  We have treated z-star as sufficiently Eulerian up to now but more complex ALE coordinates would not be.  The vertical velocity through ALE surfaces is not what is needed to compute an accurate MOC.  Instead, the MOC would want to compute the Eulerian vertical velocity (`w = - \nabla \cdot (h_k \mathbf{u}_k)` if I'm not mistaken).  This work is outside the scope of the current design doc but might be worth mentioning as a `.. note::`.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
xylar: I could imagine that remapping at irregular intervals could also induce strange behavior like larger shocks to the system each time remapping does occur.

I wonder if this might be something to put on the back burner and explore for future improvements in performance and/or accuracy.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Options for improving performance:
+
+- Using the split-explicit scheme
+- Splitting the scalar and momentum timesteps
+- Only remapping when the change in thickness exceeds given threshold
+- Optimizing/parallelizing PPR?
```
xylar, line 410: I would think all of these should be noted as avenues for exploration in a future project but would not be in the scope of the current design.  Instead, I would think our fallback for now if performance isn't great would be that VLR would only be used in contexts where it is required to support a new vertical coordinate, and where the performance hit would be deemed acceptable.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Ensure that no calculations are made outside of remapping itself using 
+prognostic (or diagnostic) variables both before and after remapping to avoid 
+introducing additional errors.
+
+Look for places in the code where prognostic variables are used at the previous 
+timestep.
```
xylar, line 424: One way of detecting these would be to run a test where you set all the prognostic variables at the previous time level to garbage (`-1e34` is a popular choice in MPAS-Ocean) and make sure nothing breaks.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,524 @@
...
+Implementation: modularity
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Remapping operations (PPR) are performed in a separate routine. 
+
+Target grid levels should be determined in a separate routine.
```
xylar, line 435: 👍🏻👍🏻👍🏻

And this separate routine should be shared with the non-VLR implementation as well.

xylar:
@cbegeman, things are looking really good.  As far as I'm concerned, it's time to jump into a draft implementation (as I think you already are).

As I noted, one requirement that might need to be made explicit (in an existing requirements section or a new one) is that the function where the target grid is computed should be shared with the non-VLR implementation of the same vertical coordinate (if there is one).  So we should only have one place where we define z-star, for example, and it should be shared.  Otherwise, there is the potential for bug fixes to be applied to either the VLR version of the non-VLR but not both.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: @xylar, I believe the surface layer is currently treated differently for coastal wetting + drying (as I understand it, this is the only layer that can flood the coastal zone), but agree that we want the general formulation for VLR --- setting a target grid for all layers.

xylar: I see, the coastal wetting-and-drying case is certainly one we should support but maybe not one that needs special attention in this design document, except to say that it would be another target grid in addition to z-star and z-tilde that we would want to support.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
dengwirda: @xylar, yes, I think the "regrid" + "remap" terms are consistent with e.g. HYCOM, MOM6 terminology, etc. See [Griffies et al](https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2019MS001954), for example:
- "regrid": select new target grid levels.
- "remap": do the integral interpolation.

I'd personally be happy to use different terminology (others have also found the regrid / remap combination to be a little inscrutable in the past!) Perhaps: "grid selection" and "remap"?

xylar: Yes, I'd prefer "grid selection" and "remap".

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: Yes, I agree. I wrote something along these lines in the testing section.

xylar: Ah, sorry, I didn't get that far.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,525 @@
...
+
+Testing and Validation: no interference
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Temporarily set prognosticVariable(tlev=1) to unrealistic value after remapping 
+so that any errors due to interference will be detectable?
```
xylar, line 525: Ah, yep, that's exactly what I was thinking, perfect!

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: That's leftover from my personal markup. I'll be removing it. If you add these lines at the top or in a stylesheet (building with rst2html) then you can use `:red:` to get red text:

`.. raw:: html`
` `
`    <style> .red {color:red} </style>`

xylar: Ah, good to know.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: > Since k is the vertical index, shouldn't this be z_1, z_2, etc. instead of z_k^1, z_k^2, etc?
This isn't the best notation choice. In general the superscript is the timestep marker. These numbers are just referring to temporary, intermediate grid choices, but they're maybe better expressed as updates to `z_k^{target}` prior to `h_k` actually being reassigned to be consistent with `z_k^{target}`.

xylar: I guess I'm still not following but I agree that the numbers are not a good choice for provisional variables.  The POP convention was to use a `*` or some similar symbol to indicate a provisional variable.  Perhaps something like that would be more typical.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: I'm fine with changing the terminology as suggested. However, I think this does raise the question of where we want the `layerThickness` update to take place because in Griffies et al. this is at least conceptually associated with the regridding step. I think it does make sense to have the remapping module update both prognostic variables and layer thickness and just do grid selection in the regridding module. Sound good to you @xylar and @dengwirda ?

xylar: Yes, I think grid selection would merely determine what the target vertical grid is.  This would probably be specified in terms of the target layer thickness on cells.  But this would not necessarily involve actually updating `layerThickness` or `layerThicknessEdge`.

I agree, then, that it would make sense to update `layerThickness` (and make the call to update `layerThicknessEdge`) within the remapping step.

So there would not be a "regridding module" but rather something like a "vertical grid" module, I would think.  We have something like that already in `mode_init` but it's nothing to use as a starting point.

dengwirda: Yes, I think this makes sense: "grid selection" would determine the position of the new grid levels, and "remap" would update the prognostic variables, thicknesses, etc for the next time step.

philipwjones: Keeping thickness and tracer remapping together also helps if you require tracer consistency, so another reason to do it this way.

mark-petersen:
@cbegeman and others, I added a commit with some text, I thought that was easier than marking up with comments. Feel free to revise.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,544 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2020/12/15
+
+Contributors: Darren Engwirda, Xylar Asay-Davis, Carolyn Begeman
+
+Namelist options:
+
+- To turn VLR on/off: 
+  :code:`ALE_vertical_grid, config_vert_lagrangian_remap = .true. or .false.`
```
mark-petersen, line 258: I see this as a choice, so I would prefer being more explicit about `false`:
`config_tracer_vert_advection_method = 'transport' or 'remap'`
Then in the code, `if (config_tracer_vert_advection_method.eq.'transport')` is more clear.

The problem is that this list is now confusing:
```
&advection
    config_vert_tracer_adv = 'stencil'
    config_vert_tracer_adv_order = 3
    config_horiz_tracer_adv_order = 3
    config_coef_3rd_order = 0.25
    config_monotonic = .true.
/
```
because the first two flags look like they apply to remapping. I see the first is not even used. So I would do this:
```
&advection
    config_vert_tracer_advection_method = 'transport' or 'remap' (new)
    config_vert_tracer_adv = 'stencil' --- delete
    config_vert_tracer_adv_order = 3 --> config_vert_tracer_transport_order = 3
    config_vert_tracer_remap_order = 3 (new)
    config_horiz_tracer_adv_order = 3
    config_coef_3rd_order = 0.25
    config_monotonic = .true.
/
```
I think the new flags belong in the advection list, not the ALE list.

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
```diff
@@ -0,0 +1,544 @@
...
+- *Something related to target grid, for now just z_initial*
```
mark-petersen, line 259: Add to `ALE_vertical_grid, config_ALE_target_grid = 'z_initial'`

philipwjones:
While this is a design/req doc, just a reminder that for the implementation, while string choices are fine/encouraged in the input file or namelist, once we're outside the initialization, we should avoid the use of string comparisons and replace with enums (integer params in Fortran) as the choice

docs/ocean/design_docs/vert-lagrang-remap/vert-lagrang-remap.rst
cbegeman: @mark-petersen I see. That seems appropriate to me.

dengwirda: I think we may actually need the single option to toggle lagrangian-remap on / off, since this'll be changing the vertical advection scheme / terms for tracers, thickness and momentum (rather than just tracers) and it'll be necessary for everything to be done either as `remap` or `transport` consistently.

No problem to opt for more explicitness in the options if that's desired, though `config_vert_tracer_adv_order = {1,2,3,5}` would definitely make direct sense (to me) in the `remap` case too.

mark-petersen: @dengwirda I see what you mean about thickness and momentum. We could instead use
```
config_vert_advection_method = 'transport' or 'remap'
```
for tracers, thickness and momentum.  

My point was just that if you use a Boolean, it does not communicate what happens for `false`. A Boolean is appropriate if it is really on/off, but here false implies that you turn on transport instead.

cbegeman: > I think we may actually need the single option to toggle lagrangian-remap on / off,since this'll be changing the vertical advection scheme / terms for tracers, thickness and momentum (rather than just tracers)

Is this an appropriate fix, make some of the remap-related names more general:
```
&advection
    config_vert_tracer_advection_method = 'transport' or 'remap' --> config_vert_advection_method = 'transport' or 'remap'
    config_vert_tracer_transport_order = 3
    config_vert_tracer_remap_order = 3 --> config_vert_remap_order = 3
    config_horiz_tracer_adv_order = 3
    config_coef_3rd_order = 0.25
    config_monotonic = .true.
```
and should we add all of the remap options here:
```
   config_vert_remap_method = 'pqm'
   config_vert_remap_edge_interp_method = `p5e`
   etc.
```

dengwirda: @cbegeman, yes I think `config_vert_advection_method =` `transport` or `remap` makes sense, although what does everyone think about using `flux-form` in place of `transport`?

In terms of the remap options, we may be able to get away with simplifying a little, perhaps something like:
```
(method = pcm) or (order = 1) --> pcm_method
(method = plm) or (order = 2) --> plm_method
(method = ppm) or (order = 3) --> ppm_method + p3e_method
(method = pqm) or (order = 5) --> pqm_method + p5e_method
```

It is possible to mix and match some of these options in `PPR`, but I think I'd expect the combinations above to be the ones typically used in practice.

cbegeman: @dengwirda `flux-form` sounds clear to me.

I'm in favor of simplifying. Are you amenable to cutting `config_vert_remap_method` and only using `config_vert_remap_order` since we don't expect folks to mix and match?

If not, do you expect that logic so be applied sequentially so that if order is higher than method, the method is changed; and if method is higher than order, the method stays the same? Given this logic it would be pretty hard to mix and match anyhow, right?

dengwirda: Oh, I just meant that I think the edge interp scheme can be inferred, so that `ppm+p3e` and `pqm+p5e` always go together, and there's not a separate edge option necessary. 

I don't have strong feelings either way in terms of the `order` vs `method` question. As you say, I think `config_vert_remap_order` would work well, and is probably more consistent with the way other MPAS options are set up.

mgduda:

PR #764, opened 2020-12-07
Parallel Interpolation of monthly vegetation fraction and monthly climatological albedo

MiCurry:
This merge allows the init_atmosphere model to safely interpolate monthly vegetation fraction (greenfrac) and monthly climatological albedo (albedo12m) in parallel on non-cvt partitions. However, the multiprocessor check in init_atm_cases is still in place.


src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1230,108 +1265,131 @@ subroutine init_atm_static(mesh, dims, configs)
...
-       end if
+    do iCell = 1, nCells
+        if (nhs(iCell) == 0) then
+            albedo12m(:,iCell) = fillVal
+        else
+            albedo12m(:,iCell) = real(albedo12m_int(:,iCell), kind=R8KIND) / real(nhs(iCell), kind=R8KIND)
+        end if
+        if (lu_index(iCell) == isice_lu) then
+            albedo12m(:,iCell) = 70.0 ! TODO: Where does this come from?
+        endif
```
MiCurry, line 1382: We need to scale the `albedo12m` values using `scalefactor` somewhere.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1031,104 +1035,135 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                    ! This field only matters for land cells, and for all but the outermost boundary cells,
+                    ! we can safely assume that the nearest model grid cell contains the pixel (else, a different
+                    ! cell would be nearest)
+                    !
+                    if (landMask(res % id) == 1 .and. bdyMaskCell(res % id) < nBdyLayers) then
+                        do k = 1, tile_nz
+                            if (tile % tile(i, j, k) / scale_factor == missing_value) then
+                                gval = fillval
+                            else
+                                gval = tile % tile(i,j,k) / scale_factor
```
MiCurry, line 1102: I think we need to scale these values somewhere along lines 1147

MiCurry:
This PR needs to be reworked and shouldn't me reviewed until it has been done so. It also shouldn't be merged/reviewed until after PR #750.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1079,104 +1082,134 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                        end do
+                        nhs(res % id) = nhs(res % id) + 1
+
+                        if (res % id <= nCellsSolve) then
+                            all_pixels_mapped_to_halo_cells = .false.
+                        end if
+
+                    ! For outermost land cells, additional work is needed to verify that the pixel
+                    ! actually lies within the nearest cell
+                    else if (landMask(iPoint) == 1) then
```
MiCurry, line 1160: `iPoint` should be `res % id`

MiCurry:
@mgduda I think we are now good to remove the 'Don't Merge' flag on this PR. As well I think you're good to start your review.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1079,104 +1082,134 @@ subroutine init_atm_static(mesh, dims, configs)
...
-    wordsize = 1
-    scalefactor = 1.0
-    msgval = 200.0
-    fillval = 0.0
-    allocate(rarray(nx,ny,nz))
+    geog_sub_path = 'greenfrac_fpar_modis/'
+
+    ierr = mgr % init(trim(config_geog_data_path)//trim(geog_sub_path))
+    if (ierr /= 0) then
+       call mpas_log_write('Error occurred when initalizing the interpolation of montly vegetation fraction (greenfrac)', &
```
mgduda, line 1089: montly -> monthly

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1079,104 +1082,134 @@ subroutine init_atm_static(mesh, dims, configs)
...
+        shdmax(iCell) = maxval(greenfrac(:,iCell))
     end do
-    deallocate(rarray)
+
     deallocate(nhs)
+    deallocate(greenfrac_int)
+
+    ierr = mgr % finalize()
+    if (ierr /= 0) then
+       call mpas_log_write('Error occurred when finalizing the interpolation of montly vegetation fraction (greenfrac)', &
```
mgduda, line 1210: montly -> monthly

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -539,21 +541,54 @@ end subroutine read_geogrid
...
+        ! to be between tile_z_start:tile_z_end incase a future dataset describes its z dimension as
+        ! anything other than 1:tile_z or 1:tile_z_end.
+        !
+        if (.not. associated(tile_nz)) then
+            allocate(tile_nz)
+            tile_nz = (tile_z_end - tile_z_start) + 1
+            c_tile_nz = tile_nz
+        else
+            allocate(tile_z_start)
+            allocate(tile_z_end)
```
MiCurry, line 587: @mgduda I think you're correct that something not right is going on. Taking a second look at this this would always allocate `tile_z_start` and `tile_z_end` when `read_geogrid` is called... Perhaps I wrote this function before my first cup of coffee?

I'll move this logic up to the initialization function.

MiCurry:
Okay, I have pushed updates that change montly to monthly and reorganize the allocation of tile_z, tile_z_start and tile_z_end.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1278,108 +1311,136 @@ subroutine init_atm_static(mesh, dims, configs)
...
-                   end do
-                   nhs(iPoint) = nhs(iPoint) + 1
+            end do
+
+            tile % is_processed = .true.
+
+            if (.not. all_pixels_mapped_to_halo_cells) then
+                ierr = mgr % push_neighbors(tile)
+                if (ierr /= 0) then
+                    call mpas_log_write("error pushing the tile neighbors of: "//trim(tile%fname), messagetype=mpas_log_crit)
```
mgduda, line 1417: For consistency with other code, let's capitalize "mpas_log_crit".

mgduda:
I don't think the `albedo12m` field is being correctly computed. There are a few issues with this field that I can see when comparing against the results from the v7.0 code:
1. The albedo values over water are 0.08 rather than 8.0.
2. The albedo values over permanent snow and ice (e.g., Greenland and Antarctica) are 0.7 rather than 70.0.
3. The albedos over non-ice land for all months other than January are zero.

MiCurry:
@mgduda Okay, the code that scaled the albedo field has been removed as well the calls to `mpas_pool_get_config` have been changed to actually grab `tile_z_start` and `tile_z_end` rather than `tile_x`!

MiCurry:
@mgduda I have just pushed an update to this branch that puts all 12 month climatological albedo between `0.0` and `100.0` sets values over ice to 70.0 and sets values over water to `8.0` and non-ice values for months other than January are correct.

src/core_init_atmosphere/mpas_geotile_manager.F
```diff
@@ -263,6 +265,39 @@ function mpas_geotile_mgr_init(mgr, path) result(ierr)
...
+
+        call mpas_pool_get_config(mgr % pool, 'tile_z', tile_nz)
+        call mpas_pool_get_config(mgr % pool, 'tile_z_start', tile_z_start)
+        call mpas_pool_get_config(mgr % pool, 'tile_z_end', tile_z_end)
+
+        if (associated(tile_nz)) then
+            ! Here we are assuming that if tile_z is specified then tile_z_start and tile_z_end
+            ! are not. This is safe currently as no dataset that MPAS uses specifies both.
+            allocate(tile_z_start)
+            allocate(tile_z_end)
```
mgduda, line 289: Do these allocations result in a memory leak? If so, should we address those in this PR?

MiCurry: It does. I have a branch, [init_atmsophere/geotile_memleak_fix](https://github.com/MiCurry/MPAS-Model/tree/init_atmosphere/geotile_memleak_fix), which has two commits that fix this issue and other memleaks in `mpas_geotile_manager` and in `mpas_parse_geoindex`. 

Previously I was thinking about making a PR for that branch after this one was made, but perhaps it would be easiest to pull those commits onto this one.

mgduda: If the `init_atmosphere/geotile_memleak_fix` branch fixes memory leaks besides this one, then I think a separate PR make sense. Otherwise, it might be cleanest to not introduce memory leaks here in the first place.

MiCurry: It does. Sounds good. I'll push an update that fixes the `MPAS_LOG_CRIT` comment and removes these memory leaks.

MiCurry:
@mgduda okay, this last update has capitalized `MPAS_LOG_CRIT` and has removed the memory leak! Please let me know if there is anything else.

mgduda (approved):
This all looks good to me. Thanks, @MiCurry !

PR #767, opened 2020-12-09
Initial working version of cmake build

brobey:
This pull request adds the capability of a cmake build while leaving the current make system intact. The build now is:

mkdir build && cd build
cmake ..
make -j 8

The new cmake build has been tested with Intel, GCC, and PGI compilers. The Cori system with Intel has been partly incorporated into the cmake build, but more testing is still needed.


mgduda:
Before this PR is merged into any branch of MPAS-Dev/MPAS-Model, I think it might be good to discuss whether adding a couple of `.tar.gz` files to the repository is prudent. It's probably also worth having a conversation about a strategy for CMake builds in general. For example, we on the MPAS-Atmosphere side have plans to incorporate a CMake build for MPAS-Atmosphere at some point as well, so some coordination is probably in order.

brobey:
Let’s consider this as a proposal and get feedback. We may want to incorporate the feedback and rework the changes to incorporate the MPAS-Atmosphere cmake mods.

Can we make this a branch? I couldn’t figure out a way to do that.

brobey:
I’ve modified the cmake build to use downloads rather than tar files. It also will first test for existing local versions and preferentially use those.

It may be best to look at the MPAS-Atmosphere cmake build before considering a merge. With the cmake build system on MPAS-Ocean, we get a 15% speedup of the code on Skylake-gold processors because we can get better vectorization flags.

PR #768, opened 2020-12-11
Added vertexDegree back into variational velocity solver.

akturner:
A previous performance PR changed vertexDegree in the variational velocity solver for the hardwired value of "3". This prevents the code from working correctly for quadrilateral meshes, that are used in idealized test cases and comparisons with structured models such as CICE.

### DIFF
--- a/src/core_seaice/shared/mpas_seaice_mesh_pool.F
+++ b/src/core_seaice/shared/mpas_seaice_mesh_pool.F
@@ -22,7 +22,8 @@ module seaice_mesh_pool
 
    integer, public :: &
       nCells,         &
-      nVerticesSolve
+      nVerticesSolve, &
+      vertexDegree
 
    integer, public, dimension(:), pointer :: & 
       nEdgesOnCell,   &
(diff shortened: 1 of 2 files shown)


akturner:
@amametjanov: Hi Az, the previous performance PR changed "vertexDegree" for "3" in the variational velocity solver. This prevents the code from working for quadrilateral (vertexDegree=4) meshes that are used for test cases and for comparisons with structured sea ice models. Hopefully this doesn't affect performance too much.

amametjanov (approved):
Yes, I was trying to help a compiler optimize the inner loop (with the constant number of iterations of the middle loop). I didn't see vertexDegree other than 3 in mesh specification document and was too optimistic. Thanks for the fix.

akturner:
@amametjanov : Thanks! Yeah, its not used in E3SM and the ocean people who wrote the spec. don't use it either, so no reason for you to know. I should have seen this when I reviewed the PR.
Cheers Ade

mgduda:

PR #770, opened 2020-12-16
Switch ocean freezing temperature to variable in MPAS-SI

proteanplanet:
This PR fixes a problem in the column package that also exists in [Icepack](https://github.com/CICE-Consortium/Icepack), in that the ocean freezing temperature has previously been set to a fixed constant rather than the correct value as determined by the input flag tfrz_option (config_sea_freezing_temperature_type in the MPAS dycore).  It affects the code in three places:

1. https://github.com/MPAS-Dev/MPAS-Model/blob/51d5624709f75182130a2fd94715e1bb13310172/src/core_seaice/column/ice_itd.F90#L1191

2. https://github.com/MPAS-Dev/MPAS-Model/blob/51d5624709f75182130a2fd94715e1bb13310172/src/core_seaice/column/ice_algae.F90#L537

3. https://github.com/MPAS-Dev/MPAS-Model/blob/51d5624709f75182130a2fd94715e1bb13310172/src/core_seaice/column/ice_colpkg_tracers.F90#L218

In addition, the default value of tocnfrz being used in the fully coupled model was being set assuming a salinity of the ocean surface was fixed at 34 PSU, which was incorrect:

https://github.com/MPAS-Dev/MPAS-Model/blob/51d5624709f75182130a2fd94715e1bb13310172/src/core_seaice/column/constants/cesm/ice_constants_colpkg.F90#L44

This is instead of using the same fixed value of -1.8C in the default setting.   

**The core problem this PR solves is to ensure consistent coupling throughout the code in both stand-alone and coupled configurations of the sea ice model when config_sea_freezing_temperature_type='mushy' as used in E3SM Versions 1 and 2.**   The code in this branch compiles and runs in G-case simulations in E3SM, and now needs the scrutiny of @akturner  and @njeffery in the stand-alone version.    In the G-case simulation, the mushy freezing temperature was also enacted in the ocean model and coupled model according to https://github.com/E3SM-Project/E3SM/pull/3861 and https://github.com/E3SM-Project/E3SM/pull/3983. Existing results indicate this makes a significant impact on Arctic sea ice thickness. The change implemented here should not affect the coupled model in G- and B-case simulations, but is necessary to ensure that testing the stand-alone MPAS-SI and potentially F-cases corresponds to the fully coupled model and ice-ocean coupled simulations.  

_@eclare108213 is working to implement similar changes to Icepack for the CICE Consortium to keep ColPkg and Icepack aligned._

### DIFF
--- a/src/core_seaice/column/constants/cesm/ice_constants_colpkg.F90
+++ b/src/core_seaice/column/constants/cesm/ice_constants_colpkg.F90
@@ -38,11 +38,12 @@
          dragio    = 0.00536_dbl_kind ,&! ice-ocn drag coefficient
 #endif
 
-         albocn    = 0.06_dbl_kind    ,&! ocean albedo
+         albocn    = 0.06_dbl_kind   ,&! ocean albedo
          gravit    = SHR_CONST_G     ,&! gravitational acceleration (m/s^2)
          viscosity_dyn = 1.79e-3_dbl_kind, & ! dynamic viscosity of brine (kg/m/s)
-         Tocnfrz= -34.0_dbl_kind*depressT,&! freezing temp of seawater (C),
-                                           ! used as Tsfcn for open water
+         Tocnfrz   = -1.8_dbl_kind   ,&! freezing temp of seawater (C), used 
+                                       ! as Tsfcn for open water only when 
+                                       ! tfrz_option is 'minus1p8' or null
          rhofresh  = SHR_CONST_RHOFW ,&! density of fresh water (kg/m^3)
          zvir      = SHR_CONST_ZVIR  ,&! rh2o/rair - 1.0
          vonkar    = SHR_CONST_KARMAN,&! von Karman constant
--- a/src/core_seaice/column/constants/cesm/ice_constants_colpkg.F90
+++ b/src/core_seaice/column/constants/cesm/ice_constants_colpkg.F90
@@ -41,6 +41,8 @@
          albocn    = 0.06_dbl_kind    ,&! ocean albedo
          gravit    = SHR_CONST_G     ,&! gravitational acceleration (m/s^2)
          viscosity_dyn = 1.79e-3_dbl_kind, & ! dynamic viscosity of brine (kg/m/s)
+         Tocnfrz= -34.0_dbl_kind*depressT,&! freezing temp of seawater (C),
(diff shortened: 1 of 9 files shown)


jonbob:
Assigned this PR to @akturner since he does the PRs into MPAS

jonbob:
A test merge of this PR passes ERS.T62_oEC60to30v3.GMPAS-IAF.anvil_intel

akturner (approved):
Looks good. Passes all tests except regression as expected.

proteanplanet:
Just to confirm that the two physical changes here are for coding consistency, and don't change the results in coupled simulations _providing_  the ocean is coupled at the mushy freezing temperature. This is because only sea ice cover, not open water cover, is used when coupling with the ocean.  Hence in this test case, the second and third traces are perfectly overlaid, where "Freeze Fix" adjusts the MCT coupling in E3SM (PR https://github.com/E3SM-Project/E3SM/pull/3983), and "Final Changes" includes this PR change in addition to that one.  So this is confirmation that what we think should be happening is in fact happening:

Freezing_Same_Freezing_Fix_Freeze_Coupling3_arctic_seaice_volume_graph_0001_0012_0

The only change that is non-BFB in E3SM B-case and G-case simulations is the BGC term - @njeffery - please check this,

njeffery (approved):
Passes parallelism and restartability checks for all stand-alone testsuites. Regressions fail.  I compared ice volume in a five year MPAS-SI DCASE run with seaice/develop and found that by year 5, ice thickness differences in the Arctic are  < 1 cm while Southern Ocean differences are still small but as high as 4 cm.  No differences in computation expense.

njeffery:
@proteanplanet :This PR doesn't impact E3SM BGC.  The Tin_bot change to the bgc code is in the skeletal bgc model and not the vertically resolved model used in E3SM.

proteanplanet:
> @proteanplanet :This PR doesn't impact E3SM BGC. The Tin_bot change to the bgc code is in the skeletal bgc model and not the vertically resolved model used in E3SM.

Perfect. Thank you. I am ignorant of such things, as you can tell.  So the end result of this PR is that it won't change E3SM except if future changes are made to the code assuming that freezing temperature is consistent as set by user flags, which is leaving V2 code in a good state for university developers.

PR #771, opened 2020-12-16
Switch to blackbody coupling for sea ice in E3SM

proteanplanet:
This one-line change to the code ensures physically correct coupling in E3SM in that the cloud feedback occurs on the radiation timestep.  Emissivity is set to unity for sea ice. In reality, it is 0.985 or thereabouts, but we do not have the correct coupling infrastructure at present to institute that properly, and in any case, the sensitivity to it is minimal in the coupled model.


jonbob:
Re-assigned to @akturner since he is responsible for PR's into MPAS

jonbob:
@proteanplanet - this passed an ERS.T62_oEC60to30v3.GMPAS-IAF.anvil_intel test. Please let me know if there's anything else you want to check

akturner (approved):

njeffery (approved):
Passed all testsuites for MPAS stand-alone including regressions.

PR #773, opened 2020-12-17
Fixed bug with high frequency coupling

akturner:
Step therm1 used vertex ice velocity instead of cell ice velocity
Fixed minor bug with testing system

### DIFF
--- a/src/core_seaice/Registry.xml
+++ b/src/core_seaice/Registry.xml
@@ -3661,6 +3661,8 @@
 		<var name="elasticTimeStep"			type="real"	dimensions=""				name_in_code="elasticTimeStep"/>
 		<var name="uVelocity"				type="real"	dimensions="nVertices Time"		name_in_code="uVelocity"/>
 		<var name="vVelocity"				type="real"	dimensions="nVertices Time"		name_in_code="vVelocity"/>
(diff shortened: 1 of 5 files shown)


mattdturner (approved):
Approved via visual inspection.  Thanks @akturner for looping me into this bugfix.

akturner:
This bug involves Step therm 1 using the vertex ice velocity rather than the interpolated cell ice velocity. This has been there since the code was written and follows what CICE does:
https://github.com/CICE-Consortium/CICE/blob/master/cicecore/cicedynB/general/ice_step_mod.F90#L327
This is much less of a problem with CICE since the vertex velocity of the same indices is always close (one of the cell vertices) to the cell velocity of the same indices. That is not true for MPAS-Seaice where a vertex velocity of a given index might be on the other side of the partition from the cell velocity of the same index.  A consequence of this for CICE is that CICE is not invariant for a transformation of the grid that leaves the physical cell positions unchanged but alters the indices used to label those cells.

proteanplanet (approved):
The source of this bug makes perfect sense. I've reviewed the changes and they look fine to me by visual inspection. Thanks @akturner for acting on this so quickly.

PR #774, opened 2020-12-18
Global limiting options for FCT vertical tracer advection

mark-petersen:
This PR adds the flag `config_activeTracer_vert_limiter` which has four options:
- `standard_FCT`: default, same as before this PR: limits wherever needed to preserve monotonicity
- `fixed_global_limit`: limits T&S to global bounds, using these flags:
```
    config_vert_limiter_temperature_min = -1.8
    config_vert_limiter_temperature_max = 34
    config_vert_limiter_salinity_min = 0
    config_vert_limiter_salinity_max = 45
```
- `always_high_order`: Uses high-order advection everywhere. Reduces spurious diffusion, but results in oscillations and violates local and global bounds. For example, T can hit -20 and S as high as 80. For experimentation, not recommended for simulations.
- `always_low_order`: Uses low-order upwind everywhere. This is always bounds preserving, but very diffusive. For experimentation, not recommended for simulations.

The most likely new setting for E3SM is `fixed_global_limit`, if tests proceed well. This option will use high-order advection almost everywhere, and will only use low-order where T&S could violate the global bounds. It should significantly lower vertical heat transport due to numerical diffusion.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_advection_mono_vert_hor.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_advection_mono_vert_hor.F
@@ -1,0 +1,1015 @@
+! Copyright (c) 2013,  Los Alamos National Security, LLC (LANS)
+! and the University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
+! Additional copyright and license information can be found in the LICENSE file
+! distributed with this code, or at http://mpas-dev.github.com/license.html
+!
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+!
+!  ocn_tracer_advection_mono
+!
+!> \brief MPAS monotonic tracer advection with FCT
+!> \author Mark Petersen, David Lee, Doug Jacobsen, Phil Jones
+!> \date   October 2017, updated May 2019
+!> \details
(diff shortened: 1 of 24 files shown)


mark-petersen:
Passes nightly regression suite with all new options using gnu 6 debug on grizzly.

mark-petersen:
@mattdturner please double check the new loops I added for correct openMP pragmas.

src/core_ocean/shared/mpas_ocn_tracer_advection_mono_vert_hor.F
```diff
@@ -0,0 +1,1109 @@
...
+
+#ifdef _ADV_TIMERS
+        call mpas_timer_stop('horiz flux')
+        call mpas_timer_start('scale factor build')
+#endif
+        ! Initialize flux arrays for all cells
+        nCells = nCellsArray(size(nCellsArray))
+
+        !$omp parallel
+        !$omp do schedule(runtime)
```
mattdturner, line 747: This should include `private(k)`

mark-petersen:
Thanks @mattdturner I added that `private(k)` and also the original line it was copied from.

mattdturner (approved):

mark-petersen:
Closing. We tested these different options in E3SM G cases, and none made a substantial improvement with the heat transport or MOC, so we are not merging it in.

PR #775, opened 2020-12-23
Fix issues with thermal solver

matthewhoffman:
This merge fixes subtle issues with how temperature is handled:
* Fixes a bug in basalThickness on restarts by adding it as a new restart field.
* Forces the interface to use the same definition of which cells have valid temperatures as MPAS is using.

### DIFF
--- a/src/core_landice/mode_forward/Interface_velocity_solver.cpp
+++ b/src/core_landice/mode_forward/Interface_velocity_solver.cpp
@@ -46,6 +46,7 @@ const double unit_length = 1000;
 const double T0 = 273.15;
 const double secondsInAYear = 31536000.0;  // This may vary slightly in MPAS, but this should be close enough for how this is used.
 double minThickness = 1e-3; //[km]
+double thermal_thickness_limit; //[km]
 const double minBeta = 1e-5;
 double rho_ice;
 double rho_ocean;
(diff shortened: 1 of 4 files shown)


src/core_landice/mode_forward/Interface_velocity_solver.cpp
```diff
@@ -1157,7 +1160,7 @@ void importFields(std::vector<std::pair<int, int> >& marineBdyExtensionMap,  dou
       int v = verticesOnTria[iVertex + 3 * index];
       int iCell = vertexToFCell[v];
       //compute temperature by averaging temperature values of triangles vertices where ice is present
-      if (cellsMask_F[iCell] & ice_present_bit_value) {
+      if (thickness_F[iCell] > thermal_thickness_limit * unit_length) {
```
mperego, line 1163: You do not have to scale this by `unit_length`. `thickness_F` is the original MPAS thickness.

src/core_landice/mode_forward/Interface_velocity_solver.cpp
```diff
@@ -113,11 +114,13 @@ void velocity_solver_set_parameters(double const* gravity_F, double const* ice_d
                          double const* sea_level_F, double const* flowParamA_F,
                          double const* flowLawExponent_F, double const* dynamic_thickness_F,
                          double const* clausius_clapeyron_coeff,
+                         double const* thermal_thickness_limit_F,
```
mperego, line 117: Isn't there a mask for the thickness? If so you could replace `li_mask_ValueIce`  with that mask. If not this is fine, but should we get rid of `li_mask_valueIce`?

mperego (commented):
Looks good, but you have to remove the scaling when you compare the thickness with the thermal_thickness_limit.

src/core_landice/mode_forward/Interface_velocity_solver.cpp
matthewhoffman: I did this because when I set the value of thermal_thickness_limit I decided to have it use Albany units (km) for consistency with the other things we import from MPAS.  So here I'm converting it back to meters because the thickness is in m.

src/core_landice/mode_forward/Interface_velocity_solver.cpp
matthewhoffman: In MPAS we don't have a mask value for the thermal thickness limit and it can differ from the dynamic thickness limit.  Doing it this way is kind of inconsistent with the way we otherwise use mask values, but this was easier than adding the thermal thickness limit to the mask field.

src/core_landice/mode_forward/Interface_velocity_solver.cpp
mperego: Ah, OK!  Maybe it's worth putting a commet to explaining that.

mperego (approved):

PR #779, opened 2021-01-07
Optimization drydynamics shortloop

evanmacbride:
This is to align code in this branch with what is being used in scaling runs.
Optimized inner loops with shortloop.

Running at 30km resolution on 6 GPUs for a 30 minute simulation gave an average integration time of 1.210891429 seconds using shortloop compared to 1.224177143 seconds without. We expect a greater performance increase at higher resolutions.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -27,9 +27,9 @@ module atm_time_integration
 
    use mpas_atm_iau  
 
-#ifdef MPAS_OPENACC
-   use openacc
-#endif
+!#ifdef MPAS_OPENACC
+!   use openacc
+!#endif
 
    integer :: timerid, secs, u_secs
 
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -27,9 +27,9 @@ module atm_time_integration
 
    use mpas_atm_iau  
 
-!#ifdef MPAS_OPENACC
-!   use openacc
-!#endif
+#ifdef MPAS_OPENACC
+   use openacc
+#endif
 
    integer :: timerid, secs, u_secs
 
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -6231,9 +6231,9 @@ if (config_horiz_mixing == "2d_smagorinsky") then
       integer :: iEdge, iCell, iVertex, k, cell1, cell2, vertex1, vertex2, eoe, i, j, iq, iAdvCell
 
       !real (kind=RKIND), parameter :: c_s = 0.125
-      real (kind=RKIND), dimension( 59 ) :: d_diag, d_off_diag,flux_arr1, flux_arr, ru_edge_w, ru_save_temp , tend_wk, delsq_wk, wduz, wdwz, wdtz, we_w, u_mix, h_wk
+      real (kind=RKIND), dimension( 64 ) :: d_diag, d_off_diag,flux_arr1, flux_arr, ru_edge_w, ru_save_temp , tend_wk, delsq_wk, wduz, wdwz, wdtz, we_w, u_mix, h_wk
       integer, dimension(15) ::  iadv_cell_w
-      integer, dimension(59) :: eoe_w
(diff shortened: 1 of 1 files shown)


mgduda:
@supreethms1809 Is this one of the sets of changes that you mentioned during our meeting this past Wednesday (1/6)?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -27,9 +27,9 @@ module atm_time_integration
 
    use mpas_atm_iau  
 
-!#ifdef MPAS_OPENACC
-!   use openacc
-!#endif
+#ifdef MPAS_OPENACC
+   use openacc
+#endif
```
mgduda, line 32: @evanmacbride How is it that the code was working with this `use` statement commented-out? What error message were you getting with this commented-out? This would be useful information to include in the commit message for e4ea3627.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -5162,9 +5162,9 @@ subroutine atm_compute_dyn_tend_work(nCells, nEdges, nVertices, nVertLevels_dumm
       integer :: iEdge, iCell, iVertex, k, cell1, cell2, vertex1, vertex2, eoe, i, j, iq, iAdvCell
 
       !real (kind=RKIND), parameter :: c_s = 0.125
-      real (kind=RKIND), dimension( 64 ) :: d_diag, d_off_diag, flux_arr, ru_edge_w, tend_wk, delsq_wk, wduz, wdwz, wdtz, we_w, u_mix, h_wk
+      real (kind=RKIND), dimension( 59 ) :: d_diag, d_off_diag, flux_arr, ru_edge_w, tend_wk, delsq_wk, wduz, wdwz, wdtz, we_w, u_mix, h_wk
       integer, dimension(15) ::  iadv_cell_w
-      integer, dimension(64) :: eoe_w
+      integer, dimension(59) :: eoe_w
```
mgduda, line 5167: What is the reasoning behind the change from 64 to 59 here and on line 5165?

mgduda (commented):
@evanmacbride I think it would be valuable to document the performance improvement that you're seeing with these `shortloop` optimizations. This information could either be added to the commit message for c7b7e4a9, or it could be added to the PR description (which we will use as the merge commit message).

evanmacbride:
> @supreethms1809 Is this one of the sets of changes that you mentioned during our meeting this past Wednesday (1/6)?

Yes.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6229,9 +6231,9 @@ subroutine atm_compute_dyn_tend_work_rk23(nCells, nEdges, nVertices, nVertLevels
       integer :: iEdge, iCell, iVertex, k, cell1, cell2, vertex1, vertex2, eoe, i, j, iq, iAdvCell
 
       !real (kind=RKIND), parameter :: c_s = 0.125
-      real (kind=RKIND), dimension( 64 ) :: d_diag, d_off_diag,flux_arr1, flux_arr, ru_edge_w, ru_save_temp , tend_wk, delsq_wk, wduz, wdwz, wdtz, we_w, u_mix, h_wk
+      real (kind=RKIND), dimension( 59 ) :: d_diag, d_off_diag,flux_arr1, flux_arr, ru_edge_w, ru_save_temp , tend_wk, delsq_wk, wduz, wdwz, wdtz, we_w, u_mix, h_wk
       integer, dimension(15) ::  iadv_cell_w
-      integer, dimension(64) :: eoe_w
+      integer, dimension(59) :: eoe_w
```
mgduda, line 6236: Here and on line 6234, too, it looks like `59` should be changed back to `64`.

mgduda:
@evanmacbride Thanks very much for adding some details of the performance improvement in the PR description. It looks like the short-loop optimizations give about a 1 percent improvement for the test case you described. You mentioned that the times represent average times; I'm curious whether you know the standard deviations of the times before the optimization and after, and whether the improvement in performance is statistically significant.

You also mentioned that you expect the performance improvement to be more significant at higher resolution, and I'd be curious to know the reasons for this (speculative or otherwise).

evanmacbride:
Thanks for your suggestions, Michael! We learned some new information about these optimizations while preparing for a presentation recently. Namely, we discovered a few discrepancies that we need to address. So we'd like to close this pull request for now and revisit it at a later date. Thanks again!

PR #780, opened 2021-01-07
Widen Redi halo in hmix, correct OpenMP private variables

mark-petersen:
This fixes https://github.com/E3SM-Project/E3SM/issues/3958.  A computation in the Redi hmix needed to be widened by one halo ring. Otherwise, partition tests failed - but only in E3SM and only with DIB-IAF-ISMF (data iceberg, ice sheet melt fluxes). 

I also removed incorrect variables in an OpenMP private statement, and removed an unused variable.

This PR is BFB with previous.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
@@ -189,7 +189,7 @@ contains
       !$omp end do
       !$omp end parallel
 
-      nCells = nCellsArray(1)
+      nCells = nCellsArray(2)
       nCellsP1 = nCellsArray(size(nCellsArray)) + 1
 
       ! Term 1: this is the "standard" horizontal del2 term, but with RediKappa coefficient.
--- a/src/core_ocean/shared/mpas_ocn_gm.F
+++ b/src/core_ocean/shared/mpas_ocn_gm.F
@@ -132,7 +132,7 @@ contains
       integer, dimension(:, :), pointer :: cellsOnEdge, edgesOnCell
       integer                          :: i, k, iEdge, cell1, cell2, iCell, N, iter, iCellSelf, maxLocation
       real(kind=RKIND)                 :: h1, h2, areaEdge, c, BruntVaisalaFreqTopEdge, rtmp
-      real(kind=RKIND)                 :: kappaGMEdge, sumN2, countN2, maxN, kappaSum, ltSum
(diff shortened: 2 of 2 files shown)


mark-petersen:
Passes MPAS-O nightly regression with gnu and both debug and optimized. Passes E3SM test that caused the original problem in E3SM-Project/E3SM#3958:
```
PEM_P480_Ld5.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_gnu
PEM_P480_Ld5.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel
```
as well as a variety of others tested:
```
PEM_Ld5_P512.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_intel
PEM_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-haswell_gnu
PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-haswell_intel
```

vanroekel:
Thanks @mark-petersen.  DO you know why this wasn't a problem for non ISMF runs?  It's unclear to me why the PEM test only failed for ISMF.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -189,7 +189,7 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, layerThickness, layerThicknessEdg
       !$omp end do
       !$omp end parallel
 
-      nCells = nCellsArray(1)
```
mark-petersen, line 192: @vanroekel the Redi hmix was only computing on the owned cells here, but involves T&S on neighboring cells. So the ice sheet melt fluxes must alter T&S on halo 1 and caused the PEM fail, while non-ISMF did not alter. We almost never compute *only* on owned cells, so extending the computation to halo 1 here seems quite reasonable.

vanroekel: Thanks, that makes sense.  I agree extending one is reasonable.  I was just curious why this fix did it.

mattdturner (approved):
Approved based in visual inspection, @mark-petersen testing, and testing within E3SM (detailed https://github.com/E3SM-Project/E3SM/pull/4025#issuecomment-756439592)

vanroekel (approved):
Approved by visual inspection and testing from @mark-petersen and @mattdturner

PR #782, opened 2021-01-08
Add two new upper absorbing layer options [atmosphere/cam]

mgduda:
This PR adds two new upper absorbing layer options:

1. A fixed, second-order horizontal diffusion in the top-most three layers on the dynamics variables u, w, and theta.  This option is often used when the MPAS-A dycore is used in CAM. This option is controlled by the namelist option `config_mpas_cam_coef`.

2. A Rayleigh damping on the horizontal momentum with a user-specified damping timescale and a linear ramp over a number of levels from the top. This option is controlled by the namelist options `config_rayleigh_damp_u`, `config_rayleigh_damp_u_timescale_days`, and `config_number_rayleigh_damp_u_levels`.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -256,22 +256,17 @@
                      description="Maximum w-damping coefficient at model top"
                      possible_values="0 $\leq$ config_xnutr $\leq$ 1"/>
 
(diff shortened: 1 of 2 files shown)


mgduda:
Note that this PR is identical to PR #168, except that this PR targets the `atmosphere/cam` branch.

mgduda:
Because the changes in this PR have previously been approved and merged to the `develop` branch, I'll not add a reviewer to this PR.

PR #792, opened 2021-01-20
Impulse Response Function addition

mark-petersen:
This PR adds a new tracer group for Impulse Response Functions (IRFs), which are used to record the effects of each tendency term. The IRFs are reset to their original values of 1s and 0s at the end of each timestep. The time averages of the tendency terms for each IRF are saved for later analysis, to be used by implicit solvers to find equilibrium solutions and to speed up the spin-up process.


mark-petersen:
Rebased. Passes nightly regression suite with `config_use_IRFTracers = .true.` with both intel 17 debug and gnu 5.3 debug.

mark-petersen:
Moved to https://github.com/E3SM-Ocean-Discussion/E3SM/pull/2 after MPAS transplant.

PR #794, opened 2021-01-22
Do not zero N under floating ice in Interface

matthewhoffman:
This commit removes code that zeros effective pressure beneath floating
ice in the Interface to Albany.  This is unneeded because Albany zeros
beta at quadrature points, and zeroing it here at FEM nodes interferes
with that.  At best it's redundant, and at worst (which is if N were to
be assigned nonzero values under floating ice), it results in
significant errors in the solve.  This is also redundant because MPAS
hydro model and ocean_connection_N parameterization will already zero N
under floating ice.


mgduda:

Issue #777, opened 2021-01-04
Ocean RK4 fails bfb restart test


mark-petersen:
Testing e4dbf57d2 on branch `ocean/develop`, and compass test `ocean/global_ocean/QU240/restart_test` but with flags changed to:
```
 24 &time_integration
 25     config_dt = '00:10:00'
 26     config_time_integrator = 'RK4'
```
the bfb comparison fails. I also turned off surface fluxes, surface restoring, and set to fixed vertical layers, and problem persists, i.e.:
```
config_vert_coord_movement = 'fixed'
config_use_bulk_wind_stress = .false.
config_use_activeTracers_surface_bulk_forcing = .false.
config_use_activeTracers_surface_restoring = .false.
```
This changes the mismatch from 1e-4 to 1e-15. So main problem appears to be in the tracer time stepping:
```
    config_disable_thick_all_tend = .false.
    config_disable_vel_all_tend = .false.
    config_disable_tr_all_tend = .true.
```

Note that the split explicit passes bfb restarts. Tendency routines are identical for split explicit and RK4. RK4 passes partition test, so this is not a halo update problem.

mark-petersen:
@xylar thanks for pointing this out. I tried the easy things that came to mind, as noted in the first comment. I was hoping to find some option that required an update within the split-explicit module, that was not added to RK4. Surface fluxes and forcing are obvious candidates, but do not appear to be related.

xylar:
Thanks @mark-petersen!  As I mentioned at Coffee today, I'm seeing this problem as well in COMPASS 1.0.  Given that some of the wetting-and-drying work will begin with RK4, it would be really helpful if we can track this down and fix it.

cc @dengwirda and @cbegeman

mark-petersen:
OK, I found the module (at least). If I turn off GM:
```
 70 &GM_eddy_parameterization
 71     config_use_GM = .false.
 72     config_GM_closure = 'EdenGreatbatch'
```
Then I get an exact bfb match on restart. I flipped every other flag possible - cvmix, redi, tracer vmix, bottom drag, etc, and GM is definitely the problem. We just added this `EdenGreatbatch` method last fall. Maybe that is it.

cbegeman:
[...]

PR #313, opened 2019-07-02
Add Wright (1997) equation of state

xylar:
This merge adds a new equation of state (EOS) from [Wright (1997)](https://doi.org/10.1175/1520-0426(1997)014%3C0735:AEOSFU%3E2.0.CO;2):
  * This formulation of the new EOS is linear with depth (using the Boussinesq approximation of pressure as `p = p_s - rho0*g*(z - ssh)`), which will allow computation of the full pressure in the horizontal pressure gradient force (HPGF) to be computed analytically.  This is in contrast to the more complex formulation of the reference pressure used in the current Jacket and McDougall (JM) EOS.
  * The EOS accounts for surface pressure (for in situ density, but not potential density) from atmosphere, sea ice, frazil and ice shelves, unlike the JM EOS.
  * Potential density is referenced to `p = config_eos_wright_ref_pressure` (which is zero by default), not in the middle of the top reference grid cell as in JM.

COMPASS changes that were previously in this PR will instead be added to the `compass` package when the time comes.

PR #796, opened 2021-01-26
Move surface pressure and zMid before EOS

xylar:
This merge adds a ``surfacePressure ``variable to the forcing pool
and adds a subroutine to compute it separately from the full
3D pressure.  The computation  of ``zMid``, ``zTop`` and ``ssh`` is 
also moved to its own subroutine within  diagnostics, separate
from the pressure.

These two new subroutines are called before the equation of state
 (EOS) with the idea that future EOSes could depend on each of
these variables, none of which depends on the density computed
by the EOS itself.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -192,6 +192,12 @@ contains
                 seaIcePressure, surfacePressure)
 
       !
+      ! Z-coordinates
+      ! This section must be placed in the code before computing the density.
+      !
+      call ocn_diagnostic_solve_z_coordinates(forcingPool, layerThickness, zMid, zTop, ssh)
+
+      !
       ! equation of state
       !
 
@@ -220,7 +226,7 @@ contains
       ! This section must be placed in the code after computing the density.
       !
       call ocn_diagnostic_solve_pressure(forcingPool, layerThickness, density, &
-                surfacePressure, montgomeryPotential, pressure, zMid, zTop, ssh)
+                surfacePressure, montgomeryPotential, pressure)
 
       nCells = nCellsAll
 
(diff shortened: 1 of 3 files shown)


xylar:
This is precursory work for reviving #313.  It is important to show that these changes, which cannot feasibly be wrapped in new namelist options so that they are disabled with the existing EOSes, produced bit-for-bit identical results to the previous code.

xylar:
## Testing

I have run the `nightly` and `quwisc240` regression suites from `compass` 1.0 on my laptop with gnu compilers.  All tests succeed (except the new RK4 restart test, which fails before and after these changes due to #777) and are bit-for-bit when compared with the same tests before these changes.

xylar:
@mattdturner, I'm asking you to review because you were the most recent developer to touch this code, and I want to make sure my changes are consistent with the performance improvements you made.  I'm pretty sure they're fine.

@mark-petersen and @mattdturner, I'm hoping you can both teach me what your techniques are for doing bit-for-bit testing like this in E3SM so I can be sure this PR won't cause any surprises there.  @mark-petersen, I'd also appreciate it if you could run a test or 2 with legacy COMPASS to see if you notice anything concerning.

xylar:
@cbegeman and @dengwirda, I just want you both to be aware of this work because we almost certainly want to add better EOSes for ice-shelf cavities.  Wright will probably be the first one but the option would be open for others appropriate for the boussinesq approximation.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -185,6 +185,19 @@ subroutine ocn_diagnostic_solve(dt, statePool, forcingPool, meshPool, diagnostic
...
+      ! This section must be placed in the code before computing the density.
+      !
+      call ocn_diagnostic_solve_surface_pressure(forcingPool, atmosphericPressure, &
+                seaIcePressure, surfacePressure)
+
+      !
+      ! Z-coordinates
+      ! This section must be placed in the code before computing the density.
+      !
+      call ocn_diagnostic_solve_z_coordinates(forcingPool, layerThickness, zMid, zTop, ssh)
```
mattdturner, line 199: Does this call need to happen every time `ocn_diagnostics_solve` is called?  Or could this be wrapped in an `if ( full_compute ) then` conditional since (as far as I can tell) `zMid`, `zTop`, and `ssh` are only used in subroutines wrapped in a similar conditional.

mattdturner:
Generally speaking, I rely on the Compass `nightly` ocean regression suite for most of the BFB testing.  I'll usually run it for both Intel and GNU optimized, but run debug if there is non-BFB results for optimized.

I run quite a bit of E3SM tests for each PR that I review, but most of them are not BFB checks against `master` (just the usual `PEM`, `PET`, and `SMS` tests, both with and without debug).  The only E3SM test I run to check BFB for every PR is `SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi` using the following command (from `<E3SM_root_dir>/cime/scripts`):
```
./create_test -p e3sm -q slurm --walltime 01:00:00 SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi -c -b /compyfs/e3sm_baselines/pgi/master
```

That test is usually sufficient for the performance-related PRs that I have been reviewing.  

Occasionally, I'll want to run a BFB test for another compiler on another system.  When that is the case, I typically will create my own baseline using `master` and then test against that generated baseline.   There are baselines on all servers that could be compared against, but its good to know how to generate your own to compare against.

To generate a baseline:
```
# From E3SM_root_dir
git fetch origin
git checkout master
git reset --hard origin/master
cd cime/scripts
./create_test -q debug --walltime 00:30:00 SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-haswell_gnu -g -b master_20210126 --baseline-root $WORKDIR/e3sm_baselines
```
Note that the `--baseline-root <directory>` is wherever you want the new baselines written to, and `-b <tag>` is the tag for the baselines.  The `-g` is used to generate baselines.  Then compare to those baselines:
```
cd <E3SM_root_dir>/components/mpas-source
git merge <branch/to/merge>
cd ../../cime/scripts
./create_test -q debug --walltime 00:30:00 SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-haswell_gnu -c -b master_20210126 --baseline-root $WORKDIR/e3sm_baselines
```

Note that this is the same `create_test` command as above, but `-g` is replaced with `-c`.

src/core_ocean/shared/mpas_ocn_diagnostics.F
xylar: @mattdturner, that's a good thing to discuss.  I don't fully understand the difference between ``full_compute`` and not yet but I'll try to get caught up.

The purpose of this work is so that ``surfacePressure``, ``zMid`` and ``ssh`` can become arguments to the equation of state, and these calls appear to be outside of ``if (full_compute) then`` conditionals.  We could add another flag like
``eos_depends_on_zmid`` and then compute ``zMid`` if either ``full_compute`` or ``eos_depends_on_zmid``.  Then, as new EOSes get added, we would have to make sure ``eos_depends_on_zmid`` gets updated accurately.  Manageable but a little clumsy.

xylar:
@mattdturner, that's, that seems easy enough and is incredibly helpful!

src/core_ocean/shared/mpas_ocn_diagnostics.F
mattdturner: The `full_compute` conditionals were added to try to reduce the cost of running `ocn_diagnostic_solve`, since all of the calculations in the routine are not needed each time `ocn_diagnostic_solve` is run in `mpas_ocn_time_integration_split.F`.

It looks like there is additional room for improvement, since I see before some of the calls to the routine there are notes like:
```
      ! For Richardson vertical mixing, only density,                                     
      ! layerThicknessEdge, and kineticEnergyCell need to be computed
```

For now, I would say to just leave it as is.  I will be porting the diagnostics module to GPUs soon, and will likely reassess how to effectively break up the module to allow only necessary computations to be run for each call.

xylar: @mattdturner, looking at the one place in the code where I could find `full=.false.`:
https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/mode_forward/mpas_ocn_time_integration_split.F#L1708-L1711
there's a note indicating that SSH is needed even when `full=.false.`.  So that would be a pretty firm indication that we can't wrap this in `if (full_compute) then`. One could compute only ``ssh`` and not the rest but that really does sound like the point of diminishing returns to me.

xylar:
@mark-petersen, I'm satisfied that these changes are bit-for-bit.  This is ready to be added to the queue for an E3SM merge once you're happy.

mattdturner (approved):
Approved based on visual inspection and @xylar testing

mark-petersen:
@mattdturner could you check visually if the added code should be modified due to #800 that was just merged? I just rebased locally and I'm testing now. Will push up rebased code if it passes.

mattdturner (approved):
The code looks good, and shouldn't need any modification due to #800.

mark-petersen (approved):
Rebased. Passes nightly regression on grizzly with optimized and debug, both gnu 5.2 and intel 17. Compares bfb with `ocean/develop` with optimized gnu and intel for nightly regression suite.

mark-petersen:
This is next in line for E3SM. All the older ones need more rebasing and testing.

xylar:
Thanks for the reviews!

PR #797, opened 2021-01-27
Remove unneeded uses of specific ESMF Fortran modules in MPAS infrastructure

mgduda:
This PR removes unneeded uses of specific ESMF Fortran modules in MPAS infrastructure.

The `mpas_derived_types` and `mpas_timekeeping` modules contained not only a 'use'
of the main ESMF module, but also uses of specific ESMF timekeeping modules
(`ESMF_BaseMod`, `ESMF_Stubs`, `ESMF_CalendarMod`, `ESMF_ClockMod`, `ESMF_TimeMod`, and
`ESMF_TimeIntervalMod`). Since all of the specific ESMF modules are ultimately
included in the main ESMF module, all that is needed in the `mpas_derived_types`
and `mpas_timekeeping` modules is a 'use ESMF' statement.

It's not clear why specific ESMF modules were being used originally, or why
commit dab101b5 didn't remove the use of specific ESMF modules when the use
of the main ESMF module was added.

### DIFF
--- a/src/framework/mpas_derived_types.F
+++ b/src/framework/mpas_derived_types.F
@@ -25,12 +25,6 @@ module mpas_derived_types
    use pio_types
 
    use ESMF
-   use ESMF_BaseMod
-   use ESMF_Stubs
-   use ESMF_CalendarMod
-   use ESMF_ClockMod
-   use ESMF_TimeMod
-   use ESMF_TimeIntervalMod
 
 #include "mpas_attlist_types.inc"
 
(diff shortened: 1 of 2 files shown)


mark-petersen (approved):
Local merge into ocean/develop. Compiled with intel 17 and gnu 5.3. Passes nightly regression suite with both, using debug.

@matthewhoffman that testing might be sufficient for you.

@mgduda thanks for the cleanup.

matthewhoffman (approved):
Approved by inspection and based on Mark's testing.  Looks like obvious and straightforward cleanup.

PR #798, opened 2021-01-27
Add missing err initialization

xylar:
This merge adds a missing err initialization to
li_setup_vertical_grid.  Without this fix and using gnu
compilers under Linux, the err had a garbage value that triggered
MALI to quit with a generic error:

```
ERROR: An error has occurred in landice_init_block.
CRITICAL ERROR: An error has occurred in li_core_init. Aborting...
```

### DIFF
--- a/src/core_landice/shared/mpas_li_setup.F
+++ b/src/core_landice/shared/mpas_li_setup.F
@@ -194,6 +194,8 @@ contains
       integer :: k
       real (kind=RKIND) :: fractionTotal
 
+      err = 0
+
       ! Get pool stuff
       call mpas_pool_get_config(liConfigs, 'config_do_restart', config_do_restart)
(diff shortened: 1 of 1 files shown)


xylar:
@matthewhoffman, except for this little hiccup, I was able to build MALI in SIA mode and run the full SIA regression suite from "legacy COMPASS".  With this fix, all the tests pass for me running on my Linux laptop with gnu.

xylar:
I'm guessing compilers elsewhere must be more generous about initializing memory to zeros?

matthewhoffman:
@xylar , thanks for making a PR for this.  Interesting we haven't run into this.  We primarily use gnu, but maybe you are using a different (probably newer) version.

matthewhoffman (commented):
Thanks for fixing, this @xylar .

PR #799, opened 2021-01-28
Implementation of a unified static mapping routine

MiCurry:
This merge factors out shared code from the parallized implementation that was first introduced in 91d8539 and subsequently added to other fields. The changes implemented in 91d8539 (and other merges), enabled static terrestrial fields to be interpolated safely on non-cvt partitions. 

This merge, factors out that shared code into a new subroutine, init_atm_map_static_data. Doing so, enables easier maintenance of code, the ability for all static fields to be easily supersampled, and allows for future enhancements to be added easily to all fields that use the unified mapping function.

This merge also moves the interpolation of all fields into their own separate functions. If one field has different methods for its interpolation (i.e. using the unified mapping function or interp_sequence) both are placed in a single, separate function.

The unified mapping function takes in two procedures as arguments, which allow each field to accumulate values as necessary; however, because of this, the interpolated static field variables need to be declared as module variables, so that they can be accessed in these passed in procedures, and in the function(s) that calls the mapping function.

A better alternative than having these global variables, would be declaring these procedures internal routines, which would allow these procedures and their parent routine, to access them without having to declare them as module variables. However, passing internal subroutines as an actual argument was added in the Fortran 2008 standard and at the time of this commit, the PGI does not support this 2008 feature.


MiCurry:
This PR successfully interpolates the terrain field. A few current questions/things to resolve before moving forward:

* Where should the alternate interpolation methods (those that use `interp_sequence`) be located?
*  Do you have a better name to use for the `interp_criteria_function` abstract interface?
*  How should we handle [sub-sample factoring](https://github.com/MiCurry/MPAS-Model/blob/2d3b9fd1b9aaa8e512278b5e4e3159daedc73353/src/core_init_atmosphere/mpas_init_atm_static.F#L470)?

Let me know if you have ideas for above and of course if you have any comments, questions or suggestions on this PR!

MiCurry:
The code at [lines 511-514 ](https://github.com/MiCurry/MPAS-Model/blob/2d3b9fd1b9aaa8e512278b5e4e3159daedc73353/src/core_init_atmosphere/mpas_init_atm_static.F#L511-L514) don't actually work for three dimensional fields. 

``` Fortran
                    do k = tile_z_start, tile_z_end
                        i8val = int(tile % tile(ii,jj,k), kind=I8KIND)
                        ierr = accumulation_method(res % id, i8val)
                    end do
```

The `interp_accumulation_function` abstract function will need to be updated to take a (perhaps optional?) `k` argument.

PR #802, opened 2021-02-02
Minor updates and clean-up to 'gfortran' build target

mgduda:
This PR introduces several minor updates and clean-up changes to the 'gfortran' build target:
1) Remove the `-m64` flag, which is architecture-specific
2) Change `-O3` to `-g` in the definition of `CXXFLAGS_DEBUG`
3) Enable all available run-time checks with `-fcheck=all` for debug builds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,7 +225,7 @@ gfortran:
 	"LDFLAGS_OPT = -O3 -m64" \
 	"FFLAGS_DEBUG = -g -m64 -ffree-line-length-none -fconvert=big-endian -ffree-form -fbounds-check -fbacktrace -ffpe-trap=invalid,zero,overflow" \
 	"CFLAGS_DEBUG = -g -m64" \
-	"CXXFLAGS_DEBUG = -O3 -m64" \
+	"CXXFLAGS_DEBUG = -g -m64" \
 	"LDFLAGS_DEBUG = -g -m64" \
 	"FFLAGS_OMP = -fopenmp" \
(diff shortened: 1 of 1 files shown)


mgduda:
Note: I've targeted the `hotfix-v7.1` branch because I'd like the `-m64` fix to appear in the v7.1 release so that the MPAS-Atmosphere core will build out-of-the-box with the GNU compilers on AArch64 systems.

mark-petersen:
Local merge into ocean/develop. Compiled with gnu 5.3. Passes nightly regression suite using debug.

@philipwjones @mattdturner or @amametjanov Do you have any comments on changing these gnu flags? It's for stand-alone only, as E3SM uses cmake files.

philipwjones:
@mark-petersen - no comment from me, I'm fine with these.

matthewhoffman (approved):
@mgduda , these changes look good.  Thanks for updating this build target.

mark-petersen (approved):

mattdturner:
I also have no comments.  These changes look fine to me

mgduda:
@amametjanov If you wanted more time to consider these changes, just let me know. Otherwise, I'll try to merge it in the next couple of days.

amametjanov:
Yes, I also don't see any issues. E3SM uses `-fcheck=bounds` instead of `-fcheck=all` for debug FFLAGS, but I don't recall any difficulties with the more inclusive `-fcheck=all`.

PR #803, opened 2021-02-03
Update Makefile for LAPACK comments and library directories

mark-petersen:
Change format of LAPACK library flag, error message and help message to match PIO and PAPI. Check both the `$(LAPACK)` and `$(LAPACK)/lib` paths for library locations, as either is possible for different LAPACK versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -624,10 +624,20 @@ endif
 	LIBS += -L$(PNETCDF)/$(PNETCDFLIBLOC) -lpnetcdf
 endif
 
-ifneq "$(LAPACK)" ""
-        LIBS += -L$(LAPACK)
-        LIBS += -llapack
-        LIBS += -lblas
+ifeq "$(USE_LAPACK)" "true"
+ifndef LAPACK
+$(error LAPACK is not set.  Please set LAPACK to the LAPACK install directory when USE_LAPACK=true)
+endif
+ifneq ($(wildcard $(LAPACK)/liblapack.*), )
+	LIBS += -L$(LAPACK)
+else ifneq ($(wildcard $(LAPACK)/lib/liblapack.*), )
+	LIBS += -L$(LAPACK)/lib
+else
+$(error liblapack.* does NOT exist in $(LAPACK) or $(LAPACK)/lib)
+endif
+	LIBS += -llapack
+	LIBS += -lblas
+	override CPPFLAGS += -DUSE_LAPACK
 endif
 
 RM = rm -f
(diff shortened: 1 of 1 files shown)


mark-petersen:
Compiles successfully on LANL IC with 
```
export LAPACK=/usr/projects/climate/mpeterse/repos/lapack/build/lapack_gnu

make gfortran CORE=ocean OPENMP=false DEBUG=true GEN_F90=true USE_LAPACK=true
```
Throws the error correctly if `liblapack.*` does not exist in `$LAPACK` or `$LAPACK/lib`.

mark-petersen:
@mgduda, hopefully this is a quick review. We are just cleaning up the LAPACK checks and messaging here.

mgduda (commented):
Was there a reason for using the `shell` function rather than the `wildcard` function as we do with other libraries? As a test, if I move my LAPACK and BLAS libraries into a `lib/` subdirectory, I get the following message when building:
```
ls: cannot access /sysdisk1/duda/lapack-3.8.0/liblapack.*: No such file or directory
```
which can be avoided by using the `wildcard` function. It also seems preferable to me to not have to invoke the shell (and assume an `ls` command exists) unless absolutely necessary.

mark-petersen:
For LAPACK I'm checking for the file `liblapack.*`, not a directory. I was not able to use the wildcard command to search for `$(LAPACK)/liblapack.*`. I tried a few versions of this:
```
 631 ifneq ($(wildcard $(LAPACK)/liblapack.), )
 632     LIBS += -L$(LAPACK)
 633 else ifneq ($(wildcard $(LAPACK)/lib/liblapack.), )
 634     LIBS += -L$(LAPACK)/lib
```
but I can't get the bash wildcard to work. I don't understand why it goes first. Anyway, I'm willing to change from `script ls` to `wildcard` if it can find `$(LAPACK)/liblapack.*`.

mgduda:
It may be that I have a peculiar version of `make` on my system, but the following seems to work for me:
```
ifeq "$(USE_LAPACK)" "true"
ifndef LAPACK
$(error LAPACK is not set.  Please set LAPACK to the LAPACK install directory when USE_LAPACK=true)
endif
ifneq ($(wildcard $(LAPACK)/liblapack.*), )
        LIBS += -L$(LAPACK)
else ifneq ($(wildcard $(LAPACK)/lib/liblapack.*), )
        LIBS += -L$(LAPACK)/lib
else
$(error liblapack.* does NOT exist in $(LAPACK) or $(LAPACK)/lib)
endif
        LIBS += -llapack
        LIBS += -lblas
        override CPPFLAGS += -DUSE_LAPACK
endif
```

mark-petersen:
@mgduda I added the asterisk and it works. Thanks for the pointer.

mgduda:
@mark-petersen I'd like to squash the two commits in this PR into a single commit. If you don't object, I can take care of this and push the result back to your fork (automatically updating this PR).

mark-petersen:
No problem, it's done.

mgduda (approved):
This looks good to me. Thanks, @mark-petersen!

PR #804, opened 2021-02-05
Fixes shortwave absorption for shallow cells

vanroekel:
Currently for very shallow cells some of the incident shortwave
radiation is not applied to the column.  This adds that missing heating
back to the bottom cell.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
@@ -126,7 +126,7 @@ contains
 
       integer, dimension(:), pointer :: maxLevelCell
 
-      real (kind=RKIND) :: depth
+      real (kind=RKIND) :: depth, fluxRemaining
       real (kind=RKIND), dimension(:), pointer :: refBottomDepth
       real (kind=RKIND), dimension(:), allocatable :: weights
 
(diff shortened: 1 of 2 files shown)


vanroekel:
@maltrud could you post a quick figure here showing this fixes the non conservation of heat?

@mark-petersen this PR is needed as soon as we can.  The ocean has a non-conservation of heat right now, it is small but definitely there.  If we could make this PR high priority I'd appreciate it.

maltrud:
i think the OMP loop needs to have fluxRemaining as a private variable?

       !$omp do schedule(runtime) private(depth, k, depLev) &
should be
       !$omp do schedule(runtime) private(fluxRemaining, depth, k, depLev) &

maltrud:
in my version, i did the bottom check the same way as the other heat fluxes.  i'm not sure it matters, but i thought i'd point it out:
Luke's:
<          k = maxLevelCell(iCell)
<          tend(index_temperature, k, iCell) = tend(index_temperature, k, iCell) + penetrativeTemperatureFlux(iCell) &
<            * fluxRemaining
---
Mat's
>          if(maxLevelCell(iCell) > 0 .and. fluxRemaining > 0.0_RKIND) then
>            tend(index_temperature, maxLevelCell(iCell), iCell) = tend(index_temperature, maxLevelCell(iCell), iCell) &
>                                                      + penetrativeTemperatureFlux(iCell) * fluxRemaining
>          end if

maltrud:
here is a little blurb that demonstrates the scale of the problem and the improvement with this fix:
Screen Shot 2021-02-15 at 5 02 19 PM

vanroekel:
@maltrud thanks for catching the missing thread private variable.  I will fix that.  

On the other form for fluxRemaining.  Do we have places where maxLevelCell(iCell) = 0?  I don't think that is allowed.  The other check on fluxRemaining > 0 also seems unnecessary to me, for the exponential it should never be less than 0 and if it is zero it seems fine to just do the calculation, perhaps more performant?  @mark-petersen what do you think?  I can add it back if you'd like

xylar:
> Do we have places where maxLevelCell(iCell) = 0? I don't think that is allowed. 

Our current meshes cull out any such cells but MPAS-Ocean is supposed to run in situations where `maxLevelCell < 1` is used to indicate an inactive ocean column.  This could become very important for us with grounding line motion, so please do try to check for that.  I'm guessing we should add a test case to COMPASS with an unculled mesh (say, QU240) to make sure this works.

vanroekel:
Great point @xylar I will add that check back in.

vanroekel:
I just added a check on `maxLevelCell` as in the heat flux case and what @maltrud posts here.  @mark-petersen could you please take a look as well?  I think this is ready to go.

mark-petersen (approved):
This looks good. Tested with gnu and intel, both debug, on nightly regression suite.

PR #808, opened 2021-02-09
Fix core_ocean submodule checks

jgfouca:
Need FATAL_ERROR instead of FATAL. Fix paths to submodules.

I recently built an E3SM compset with MPAS components and I was checking some CMake stuff, so I ran the CMake command by hand and the output surprised me. CMake claimed "FATAL Missing core_ocean submodule ..." but still returned a 0 return code. Also, I confirmed that my submodules were initialized. A little digging and I found that both the existence checks and the corresponding error messages were broken. This PR fixes them.


jgfouca:
@mark-petersen

jgfouca:
Note, both these mistakes were mine and were introduced in this PR: https://github.com/MPAS-Dev/MPAS-Model/pull/420

mark-petersen (approved):
Passes
```
```

PR #809, opened 2021-02-10
Enable mpas_timekeeping module to work with the full, external ESMF library

mgduda:
This PR enables the mpas_timekeeping module to work with the full, external ESMF
library.

While the MPAS repository includes an old, Fortran-only implementation of
the ESMF time manager, coupled systems in which MPAS is a component may link
with the full (and more recent) ESMF library. This PR makes changes in
the mpas_timekeeping module to enable it to use the full, external ESMF library
in place of the built-in, Fortran-only time manager library by compiling
the MPAS framework with the MPAS_EXTERNAL_ESMF_LIB pre-processing macro defined.

Note that this PR does not include changes to Makefiles to set the module
include paths to an external ESMF library, nor does it include changes to handle
linking with an external ESMF library. The intended use case for these changes 
is a coupled model (e.g., CESM with the ESMF-based NUOPC driver) in which MPAS
models are components. In this case, the stand-alone MPAS Makefiles are unlikely
to be used, and all MPAS code is compiled with the build system of the coupled
model.

### DIFF
--- a/src/framework/mpas_timekeeping.F
+++ b/src/framework/mpas_timekeeping.F
@@ -1576,6 +1576,7 @@ module mpas_timekeeping
       integer :: days, sn, sd, hours, minutes
       integer (kind=I8KIND) :: seconds
       real (kind=R8KIND) :: seconds_real
+      character (len=1) :: neg
       integer :: local_ierr
 
 
@@ -1663,12 +1664,18 @@ module mpas_timekeeping
                                    d=days, h=hours, m=minutes, s_i8=seconds, sN=sn, sD=sd, rc=ierr)
          seconds_real = real(seconds,kind=R8KIND) + (real(sn, kind=R8KIND) / real(sd, kind=R8KIND))
 
-         ! Seconds should not be negative, so there is no need to handle negative cases, below
-         if (seconds_real < 10.0_R8KIND) then
-            write(timeString,'(i9.9,a,i2.2,a,i2.2,a,f11.9)') days, '_', hours, ':', minutes, ':0', seconds_real
-         else
-            write(timeString,'(i9.9,a,i2.2,a,i2.2,a,f12.9)') days, '_', hours, ':', minutes, ':', seconds_real
+         !
+         ! If the time interval is negative, the ESMF library will return negative values for
+         ! non-zero components of the interval, so days, hours, minutes, and seconds_real must
+         ! be checked in order to determine whether a '-' needs to be prepended to the timeString
+         !
+         neg = ''
+         if (days < 0 .or. hours < 0 .or. minutes < 0 .or. seconds_real < 0.0_R8KIND) then
+            neg = '-'
          end if
+
+         write(timeString,'(a,i9.9,a,i2.2,a,i2.2,a,i2.2,f10.9)') trim(neg), abs(days), '_', abs(hours), ':', abs(minutes), ':', &
+                                                               int(abs(seconds_real)), abs(seconds_real) - int(abs(seconds_real))
 #endif
       end if
 
--- a/src/framework/mpas_timekeeping.F
+++ b/src/framework/mpas_timekeeping.F
@@ -1309,7 +1309,11 @@ module mpas_timekeeping
 
       call ESMF_TimeGet(curr_time % t, YY=YYYY, MM=MM, DD=DD, H=H, M=M, S=S, Sn=S_n, Sd=S_d, rc=ierr)
       call ESMF_TimeGet(curr_time % t, dayOfYear=DoY, rc=ierr)
+#ifndef MPAS_EXTERNAL_ESMF_LIB
       call ESMF_TimeGet(curr_time % t, timeString=dateTimeString, rc=ierr)
+#else
+      call ESMF_TimeGet(curr_time % t, timeStringISOFrac=dateTimeString, rc=ierr)
+#endif
 
       !
       ! In case an external ESMF library that returns ISO timestamps is being used,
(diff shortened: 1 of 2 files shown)


mgduda:
The commit message for aa910935 summarizes the changes to the `mpas_timekeeping` module in approximately the same order in which those changes occur in the code.

mgduda:
@matthewhoffman and @mark-petersen Just for some background, the CESM ESMF/NUOPC driver requires the use of the external, standalone ESMF library, and this PR is designed to handle the difference between that library and the built-in ESMF timekeeping code that we have. Other than the removal of some unneeded `USE` statements, the changes in this PR should only be activated when the MPAS framework is compiled with `-DMPAS_EXTERNAL_ESMF_LIB`. Please feel free to add other E3SM reviewers if you'd like, since this does affect shared code.

src/framework/mpas_timekeeping.F
```diff
@@ -98,18 +92,31 @@ subroutine mpas_timekeeping_init(calendar)
...
+#ifndef MPAS_EXTERNAL_ESMF_LIB
          call ESMF_Initialize(defaultCalendar=ESMF_CALKIND_NOLEAP)
+#else
+         call ESMF_Initialize(defaultCalKind=ESMF_CALKIND_NOLEAP)
+#endif
 #endif
-!     else if (trim(calendar) == '360day') then
-!        TheCalendar = MPAS_360DAY
-!#ifndef MPAS_NO_ESMF_INIT
+     else if (trim(calendar) == '360day') then
```
MiCurry, line 110: The indentation of this else if statement does not appear to match match the indentation of the else if on line 103 and the if statement on line 92.

src/framework/mpas_timekeeping.F
```diff
@@ -1606,7 +1651,25 @@ subroutine mpas_get_timeInterval(interval, StartTimeIn, DD, H, M, S, S_n, S_d, S
...
+         call ESMF_TimeIntervalGet(interval % ti, StartTimeIn=StartTimeIn%t, &
+                                   d=days, h=hours, m=minutes, s_i8=seconds, sN=sn, sD=sd, rc=ierr)
+         seconds_real = real(seconds,kind=R8KIND) + (real(sn, kind=R8KIND) / real(sd, kind=R8KIND))
+
+         ! Seconds should not be negative, so there is no need to handle negative cases, below
+         if (seconds_real < 10.0_R8KIND) then
+            write(timeString,'(i9.9,a,i2.2,a,i2.2,a,f11.9)') days, '_', hours, ':', minutes, ':0', seconds_real
+         else
+            write(timeString,'(i9.9,a,i2.2,a,i2.2,a,f12.9)') days, '_', hours, ':', minutes, ':', seconds_real
+         end if
```
MiCurry, line 1671: Just to be sure I understand whats going on here, the above if statement is being used to determine if the seconds in timeString should be padded with a leading 0 or not?

mgduda: That's correct.

mgduda: I don't know of a standard way to guarantee padding with leading zeros in Fortran, so if anyone has suggestions, here, I'd be glad to hear them.

MiCurry:
Other than the indentation in `mpas_timekeeping_init`, this PR looks good to go.

src/framework/mpas_timekeeping.F
mgduda: I've just added a commit to this PR that corrects the indentation.

src/framework/mpas_timekeeping.F
philipwjones: I think this works:
write(timeString,'(i9.9,a,i2.2,a,i2.2,a,i2.2,f10.9)') days, '_', hours, ':', minutes, ':', int(seconds_real),seconds_real-int(seconds_real)

mgduda: That seems to work well -- thanks for the suggestion! I'll add a new commit later this afternoon that takes this approach rather than having to select between two `write` statements depending on the magnitude of `seconds_real`.

matthewhoffman:
@mgduda , this all makes sense to me.  I was a little surprised at how many 'features' of the inline version of ESMF timekeeping are not supported by the external one, but so it goes sometimes.  

@rljacob, you might want to take a look at this MPAS Framework PR.  Do you envision a situation where we would want to be using an external ESMF timekeeping module as Michael is enabling here?

rljacob:
E3SM will be keeping the fortran-based esmf-time-manager around for a while.

MiCurry (approved):

src/framework/mpas_timekeeping.F
mgduda: I've just pushed a new commit, 1d82840, that makes use of this, in addition to correcting an issue with negative time intervals.

mgduda:
After more extensive testing, I've pushed two additional commits, 1d82840 and b4db8cf, that correct problems with negative time intervals and fractional seconds in date-time strings.

philipwjones (approved):
For some reason the last two commits aren't showing up in the PR diffs or in the Conversation, so you should probably make sure they're included and the PR is pointing to the right hash. But I did look at the commits individually and the other mods and it looks good to me.
So approve based on visual inspection and developer testing.

mgduda:
@philipwjones Thanks for pointing out the commit issue -- I had apparently pushed the updated branch to the wrong fork, but all should be good, now.

src/framework/mpas_timekeeping.F
MiCurry: Looks good.

matthewhoffman (approved):

mark-petersen (approved):

PR #810, opened 2021-02-10
Add landice ismip6 calving & melting params., phase 1

trhille:
This is part 1 of a multi-part merge of the branch that includes ISMIP6 submarine face melting and calving parameterizations for Greenland. We are breaking into parts to attempt a cleaner merge, since this branch has been under development for almost 10 months.

The ISMIP6 submarine face melting parameterization is based on Xu et al. (2013) and Rignot et al. (2016), which defines a linear melt velocity below the water-line as a function of water depth, subglacial discharge, and ocean thermal forcing. Subglacial discharge and ocean thermal forcing are defined by input fields ismip6Runoff and ismip6_2dThermalForcing, respectively. Alternatively, submarine face melting can be set to a uniform value rate.

The calving parameterization is the von Mises stress calving relationship from Morlighem et al. (2016), which calculates a calving rate from the ice velocity at the calving front, the depth-averaged von Mises stress, and a threshold stress set in the namelist. If the von Mises stress exceeds the threshold stress, then the calving velocity exceeds the ice velocity and the calving front retreats. Different thresholds can be set for calving and floating ice; both default to 1 MPa. This calving law should generally be used where the calving front tends to be grounded.

Melting and calving velocities are converted to thickness changes and applied at the marine margin.

Morlighem M and others (2016) Modeling of Store Gletscher’s calving dynamics, West Greenland, in response to ocean thermal forcing. Geophysical Research Letters 43(6), 2659–2666. doi:10.1002/2016GL067695.
Rignot E and others (2016) Modeling of ocean‐induced ice melt rates of five west Greenland glaciers over the past two decades. Geophysical Research Letters 43(12), 6374–6382. doi:10.1002/2016GL068784.
Xu, Y., E. Rignot, I. Fenty, D. Menemenlis, and M. Mar Flexas (2013), Subaqueous melting of Store Glacier, West Greenland from three-dimensional, high-resolution numerical modeling and ocean observations, Geophys. Res. Lett., 40, 4648–4653, doi:10.1002/grl.50825.


src/core_landice/Registry.xml
```diff
@@ -289,6 +301,35 @@
 		            description="The fraction of the deepest grounding line depths over which the grounding line depth is averaged to obtain a reference grounding line depth for the 'temperature_profile' melt parameterization."
 		            possible_values="any positive real value"
 		/>
+                <!-- Options related to grounded glacier front submarine melt param. -->
+                <nml_option name="config_front_mass_bal_grounded" type="character" default_value="none" units="unitless"
+                            description="Selection of the method for computing the frontal mass balance of submarine grounded ice.  'none' sets the frontMassBal field to 0 everywhere.  'file' uses without modification whatever value was read in through an input or forcing file or the value set by an ESM coupler.  'constant' uses a prescribed constant melt rate. 'ismip6' uses the parameterization provided by ISMIP6 and requires an ocean thermal forcing field and subglacial hydrology."
```
matthewhoffman, line 306: In the description, indicate what field is used for the 'file' option.

src/core_landice/Registry.xml
```diff
@@ -289,6 +301,35 @@
...
+                />
+		<nml_option name="config_add_ocean_thermal_forcing" type="real" default_value="0.0" units="deg. C"
+			    description="Adjust ismip6_2dThermalForcing by a scalar value"
+			    possible_values="any real value"
+		/>
+                <nml_option name="config_alpha_subglacial_discharge" type="real" default_value="0.39" units="unitless"
+                            description="Exponent of the subglacial discharge term in ISMIP6 grounded ice front melting parameterization"
+                            possible_values="any real value"
+                />
+                <nml_option name="config_subglacial_discharge_coefficient" type="real" default_value="3.0e-4" units="unitless"
```
matthewhoffman, line 321: Are you sure this is unitless?

src/core_landice/Registry.xml
```diff
@@ -289,6 +301,35 @@
...
+                            description="Coefficient of the subglacial discharge term in ISMIP6 grounded ice front melting parameterization"
+                            possible_values="any real value"
+                />
+                <nml_option name="config_subglacial_discharge_intercept" type="real" default_value="0.15" units="unitless"
```
matthewhoffman, line 325: Are you sure this is unitless?

src/core_landice/Registry.xml
```diff
@@ -289,6 +301,35 @@
...
+                            description="Added value B in ISMIP6 grounded ice front melting parameterization"
+                            possible_values="any real value"
+                />
+		<nml_option name="config_uniform_face_melt_rate" type="real" default_value="0.0" units="m s^{-1}"
+			    descrption="Apply a uniform linear submarine melt rate at all grounded marine margins. If set to 0.0, it will not be applied."
```
matthewhoffman, line 330: Elaborate that this needs to be 0 for the ismip6 method to work.

src/core_landice/Registry.xml
```diff
@@ -482,7 +523,9 @@
 
                 <package name="observations" description="This package includes variables required for observations in Albany."/>
 
-                <package name="ismip6Melt" description="This package includes variables required for the ISMIP6 melt parameterization."/>
+                <package name="ismip6ShelfMelt" description="This package includes variables required for the ISMIP6 ice-shelf melt parameterization."/>
```
matthewhoffman, line 526: add "developed for Antarctica"

src/core_landice/Registry.xml
```diff
@@ -482,7 +523,9 @@
...
+
+		<package name="ismip6GroundedFaceMelt" description="This package includes variables required for the ISMIP6 grounded glacier front melt parameterization" />
```
matthewhoffman, line 528: add "developed for Greenland"

src/core_landice/Registry.xml
```diff
@@ -582,13 +625,20 @@
...
-                        <var name="ismip6shelfMelt_zOcean" packages="ismip6Melt"/>
-                        <var name="ismip6shelfMelt_offset" packages="ismip6Melt"/>
+                        <!-- these variables just for ISMIP6 ice shelf forcing -->
+                        <var name="ismip6shelfMelt_basin" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_gamma0" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_3dThermalForcing" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_deltaT" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_zOcean" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_offset" packages="ismip6ShelfMelt"/>
+			<!-- these variables just for ISMIP6 grounded glacier forcing -->
```
matthewhoffman, line 635: Do all of these need to be input-able?

src/core_landice/Registry.xml
```diff
@@ -660,12 +710,19 @@
...
-                        <var name="ismip6shelfMelt_deltaT" packages="ismip6Melt"/>
-                        <var name="ismip6shelfMelt_zOcean" packages="ismip6Melt"/>
-                        <var name="ismip6shelfMelt_offset" packages="ismip6Melt"/>
+                        <var name="ismip6shelfMelt_basin" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_gamma0" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_3dThermalForcing" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_deltaT" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_zOcean" packages="ismip6ShelfMelt"/>
+                        <var name="ismip6shelfMelt_offset" packages="ismip6ShelfMelt"/>
+                        <!-- these variables just for ISMIP6 grounded glacier forcing -->
```
matthewhoffman, line 719: Are all of these required in the restart file?

src/core_landice/Registry.xml
```diff
@@ -1033,29 +1090,54 @@ is the value of that variable from the *previous* time level!
...
+                <var name="subglacialWaterFlux" type="real" dimensions="nCells Time" units="m day^{-1}" packages="ismip6GroundedFaceMelt" default_value="0.0"
+                     description="Catchment-wide subglacial runoff normalized by modern submerged face area, in units used by ismip6 grounded marine melt parameterization, with zeros where ISMIP6 forcing file has negative runoff"
+                />
+                <var name="ismip6Runoff" type="real" dimensions="nCells Time" units="kg m^{-2} s^{-1}" packages="ismip6GroundedFaceMelt" default_value="0.0"
+                     description="Runoff forcing for ISMIP6 grounded ice melting method, input to model" 
+                />
+		<var name="ismip6_2dThermalForcing" type="real" dimensions="nCells Time" units="deg. C" packages="ismip6GroundedFaceMelt" default_value="2.0"
+	 	     description="thermal forcing for ISMIP6 grounded marine melt method, input to model"
+		/>
+		<var name="meltingThickness" type="real" dimensions="nCells Time" units="m" time_levs="1"
```
matthewhoffman, line 1135: I suggest renaming this facemeltingThickness or something to avoid confusion with surface melt, in the SMB sense.

src/core_landice/Registry.xml
```diff
@@ -1033,29 +1090,54 @@ is the value of that variable from the *previous* time level!
...
+                     description="thickness of ice that melts on a given timestep (less than or equal to ice thickness)"
```
matthewhoffman, line 1136: "equivalent plan-view area-averaged thickness ..." (or similar)

src/core_landice/Registry.xml
```diff
@@ -1249,7 +1337,7 @@ is the value of that variable from the *previous* time level!
 	       />
 	  <var name="iceCellMask2" type="integer" dimensions="nCells" units="none"
 	       description="mask set to 1 in cells where some criterion is satisfied and 0 otherwise"
-	       persistence="scratch"
+	       persistence="persistent"
```
matthewhoffman, line 1340: Change this back to scratch.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1170,6 +1174,7 @@ subroutine eigencalving(domain, err)
 
          calvingVelocity(:) = 0.0_RKIND
          ! First calculate the front retreat rate (Levermann eq. 1)
+         !calvingVelocity(:) = eigencalvingParameter(:) * max(0.0_RKIND, eMax(:)) * max(0.0_RKIND, eMin(:)) ! m/s
```
matthewhoffman, line 1177: I don't understand the diff shown here, but double check that the eigencalving calculation of calvingVelocity matches what is in landice/develop here.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1236,7 +1241,144 @@ subroutine eigencalving(domain, err)
...
+          call mpas_pool_get_array(geometryPool, 'calvingThickness', calvingThickness)
+          call mpas_pool_get_array(geometryPool, 'calvingVelocity', calvingVelocity) 
+          call mpas_pool_get_array(geometryPool, 'thickness', thickness)
+          call mpas_pool_get_array(thermalPool, 'temperature', temperature)
+          
+          vonMisesStress(:) = 0.0_RKIND
+
+          ! get flowParamA from MPAS or use Albany-like equation
+          if ( config_use_Albany_flowA_eqn ) then
+          !calculate Albany-type flowParamA
```
matthewhoffman, line 1333: Add an error here.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1236,7 +1241,144 @@ subroutine eigencalving(domain, err)
...
+          else
+             call li_calculate_flowParamA(meshPool, temperature, thickness,flowParamA,err) ! Get MPAS flowParamA
+          endif
+    
+         !Using a depth-averaged ice viscosity parameter B_depthAvg
```
matthewhoffman, line 1338: Does this bit of code need to remain for any reason?

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1236,7 +1241,144 @@ subroutine eigencalving(domain, err)
...
+         !thickness(iCell)
+         ! Calculate effective von  Mises stress.
+          calvingVelocity(:) = 0.0_RKIND
+
+          do iCell = 1,nCells
+             if ( thickness(iCell) > 0.0_RKIND ) then
+                vonMisesStress(iCell) = sqrt(3.0_RKIND) * sum(layerThickness(:,iCell) * &
+                              flowParamA(:,iCell)**(-1.0_RKIND/config_flowLawExponent),dim=1) / &
+                              thickness(iCell) * ( 0.5_RKIND * ( (max(0.0_RKIND, eMax(iCell)))**2 + &
+                              (max(0.0_RKIND, eMin(iCell)))**2) )**(1 / (2 * config_flowLawExponent))
```
matthewhoffman, line 1350: Make the exponent integers into floating point _RKIND types (and check if it changes your answers - I don't think it will because I believe there will be automatic typecasting going on already).

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1128,6 +1140,345 @@ subroutine iceshelf_melt_ismip6(domain, err)
...
+            ! OR is adjacent to an inactive floating margin cell
+            nEmptyNeighbors = 0
+            oceanEdgeLength = 0.0_RKIND
+            ! Check if neighboring cells contain ice and have bed topo below sea
+            ! level
+            do iEdge = 1, nEdgesOnCell(iCell)
+               iNeighbor = cellsOnCell(iEdge, iCell)
+                 if ( ((thickness(iNeighbor) == 0.0_RKIND) & 
+                    .and. iNeighbor > 0 & ! make sure this cell exists
+                    .and. iNeighbor <= nCellsSolve & ! make sure this cell exists
```
matthewhoffman, line 1266: I don't think this check should be needed.  Do you remember it being necessary?

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1128,6 +1140,345 @@ subroutine iceshelf_melt_ismip6(domain, err)
...
+                    .or. (li_mask_is_floating_ice(cellMask(iNeighbor)) & 
+                    .and. li_mask_is_margin(cellMask(iNeighbor)) &
+                    .and. (.not. li_mask_is_albany_active(cellMask(iNeighbor)))) ) then
+                     nEmptyNeighbors = nEmptyNeighbors + 1
+                     oceanEdgeLength = oceanEdgeLength + dvEdge(edgesOnCell(iEdge, iCell))
+                 endif
+            enddo 
+            
+            if ( nEmptyNeighbors > 0 &
+               .and. thickness(iCell) > 0.0_RKIND &
```
matthewhoffman, line 1277: No need to check if thickness>0 if you are already checking the li_mask_is_grounded_ice mask value.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1128,6 +1140,345 @@ subroutine iceshelf_melt_ismip6(domain, err)
...
+                
+                if (config_uniform_face_melt_rate == 0.0) then                
+                ! Calculate ice front melt rate at each cell
+                   faceMeltSpeed(iCell) = (aSubglacial * waterDepth * & ! m s^-1
+                         (ismip6Runoff(iCell) / rhow * secPerDay)**alphaSubglacial &
+                         + B) * (TFocean(iCell) + addTFocean)**betaTF / secPerDay
+                else
+                   faceMeltSpeed(iCell) = config_uniform_face_melt_rate
+                endif
+                !calvingVelocity(iCell) = calvingVelocity(iCell) + iceFrontMeltRate ! m s^-1
```
matthewhoffman, line 1294: Do you feel comfortable removing this now?  Or want to keep it in awhile longer until things are fully settled?

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1128,6 +1140,345 @@ subroutine iceshelf_melt_ismip6(domain, err)
...
+                faceMeltRateApplied(iCell) = 0.0_RKIND     
+                faceMeltSpeed(iCell) = 0.0_RKIND
+            endif           
+         enddo
+
+         ! Distribute melt among neighbors
+         call distribute_face_melt_flux(meshPool, geometryPool, scratchPool, groundedMarineMarginMask, err_tmp)
+         err = ior(err, err_tmp)
+
+         ! Apply facemelt: open the Ark
```
matthewhoffman, line 1307: Classic

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1128,6 +1140,345 @@ subroutine iceshelf_melt_ismip6(domain, err)
...
+!> and adapted for grounded ice.
+!> The melt rate (faceMeltRateApplied) is calculated in subroutine grounded_face_melt_ismip6.
+!> The melt volume needs to be distributed in three ways:
+!> 1. We need to first remove any "thin" ice in front of this cell
+!> 2. Then we remove ice from this cell
+!> 3. If there is still additional ice to be removed, we have to recursively
+!>    remove ice "inland" of this cell until the total required mass is removed
+!-----------------------------------------------------------------------
+
+   subroutine distribute_face_melt_flux(meshPool, geometryPool, scratchPool, groundedMarineMarginMask, err)
```
matthewhoffman, line 1338: I'm not bothering to review this routine, because we know it will go away in the next PR.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1128,6 +1140,345 @@ subroutine iceshelf_melt_ismip6(domain, err)
...
+         call mpas_pool_get_array(geometryPool, 'meltingThickness', meltingThickness)
+         call mpas_pool_get_array(geometryPool, 'bedTopography', bedTopography)
+         call mpas_pool_get_array(geometryPool, 'lowerSurface', lowerSurface) 
+         call mpas_pool_get_array(geometryPool, 'groundedMarineMarginMask', groundedMarineMarginMask)
+         call mpas_pool_get_array(geometryPool, 'faceMeltRateApplied', faceMeltRateApplied)
+         call mpas_pool_get_array(geometryPool, 'faceMeltSpeed', faceMeltSpeed)
+         
+       ! Calculate retreat rate for each cell
+         do iCell = 1, nCellsSolve
+            ! Define marine marginal cells as those that are (1) at the ice
```
matthewhoffman, line 1254: If (1) is true, isn't (2) automatically true?

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1128,6 +1140,345 @@ subroutine iceshelf_melt_ismip6(domain, err)
...
+            oceanEdgeLength = 0.0_RKIND
+            ! Check if neighboring cells contain ice and have bed topo below sea
+            ! level
+            do iEdge = 1, nEdgesOnCell(iCell)
+               iNeighbor = cellsOnCell(iEdge, iCell)
+                 if ( ((thickness(iNeighbor) == 0.0_RKIND) & 
+                    .and. iNeighbor > 0 & ! make sure this cell exists
+                    .and. iNeighbor <= nCellsSolve & ! make sure this cell exists
+                    .and. bedTopography(iNeighbor) < seaLevel) &
+                    .or. (li_mask_is_floating_ice(cellMask(iNeighbor)) &
```
matthewhoffman, line 1268: I think you need parentheses around all the terms after the `.or.`.

matthewhoffman (changes requested):
@trhille , I made a quick review through this ISMIP6-GIS phase 1 PR.    Most of my comments should be trivial to fix, but let me know if you disagree with any of it or want to discuss further.  I tried not to look too carefully at parts I know will change in the next phase, but if any of my comments are better addressed in the next phase, let me know.  
Very nice work on getting all of this into the code, and sorry it has sat for so long!  Feel free to address the issues and/or respond as you have time.

src/core_landice/Registry.xml
trhille: I don't actually know the answer to that. I stole the description from config_basal_mass_bal_float before modifying, which doesn't include that information.

src/core_landice/Registry.xml
trhille: Units changed to m day^{-1} deg. C^{-beta} after Choi et al. (2018)

src/core_landice/Registry.xml
trhille: Units changes to m^{-alpha} day^{alpha-1} deg. C^{-beta} after Choi et al. (2018).

src/core_landice/mode_forward/mpas_li_calving.F
trhille: This change was to make eigencalving work for grounded ice. This PR has eigencalving completely overridden by a uniform scalar calving rate in order to test the apply_calving_velocity routine. I will change this back to normal eigencalving (i.e., calculated from eMax and eMin, only for floating ice).

src/core_landice/mode_forward/mpas_li_calving.F
trhille: This was always a comment, just to make it easier to understand when comparing the von Mises calculation here to the equations in publications that use B. I'd prefer to leave it, but I can rephrase it to be more clearly just an explanatory note if you would like.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: I do have that, but the spacing made it hard to tell. I've adjusted the spacing to make it easier to see.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: I don't remember one way or the other, but I agree that it doesn't seem necessary.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Currently running this test.

src/core_landice/Registry.xml
matthewhoffman: It looks like 'none' and 'ismip6' are the only implemented option here.   Can you remove the other options from the possible_values and description?

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -179,7 +184,8 @@ subroutine li_basal_melt_floating_ice(domain, err)
       endif
 
       ! Simple options handled here
-      if (trim(config_basal_mass_bal_float) == 'none') then
+      if (trim(config_basal_mass_bal_float) == 'none' &
+                .and. config_front_mass_bal_grounded == 'none') then
```
matthewhoffman, line 188: I'm not seeing how the value of `config_front_mass_bal_grounded` should affect this option of `config_basal_mass_bal_float`.  Unless you have an explanation I'm not thinking of, this change should be removed.

src/core_landice/Registry.xml
matthewhoffman: Can you point out (either in this option or in config_front_mass_bal_grounded), that if this is not 0, it will override the prognostic calculation based on the ISMIP6 method?  Actually a better solution would be to add an option to config_front_mass_bal_grounded that is 'uniform' that uses this value.

src/core_landice/mode_forward/mpas_li_calving.F
matthewhoffman: I think it's ok to change it to apply to grounded ice as well.  If so, add that in a separate commit.  The main thing is to avoid leaving it commented out when this PR gets merged.

src/core_landice/mode_forward/mpas_li_calving.F
matthewhoffman: My confusion is that what's commented is identical to what's in the vonMisesStress calculation a few lines down.  So if that's the version you think is better to use, I don't see a reason to keep the commented out version.

trhille: Right, the reason I had that was to show that we are replacing the term B used in publications with a calculation using our available variables. It is not identical to the entire vonMisesStress calculation, just one term in it.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
matthewhoffman: Thinking about it more, this is not required the way the MPAS mesh information is arranged.  So go ahead and delete it.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: I've confirmed that vonMisesStress and calvingVelocity are BFB with integers and floating point _RKIND types here.

src/core_landice/Registry.xml
trhille: I've added the 'uniform' option to config_front_mass_bal_grounded and specified that it must be set for config_uniform_face_melt_rate to work.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Added in a separate commit

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: Removed

src/core_landice/Registry.xml
trhille: 'file' has been removed. 'constant' has been replaced with 'uniform', which sets the face melt speed to the value given by config_uniform_face_melt_rate.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: Removed

src/core_landice/mode_forward/mpas_li_calving.F
matthewhoffman: Ok, thanks for checking that out and for keeping the explicit floating point version.

src/core_landice/Registry.xml
```diff
@@ -127,7 +127,11 @@
 	<nml_record name="calving" in_defaults="true">
 		<nml_option name="config_calving" type="character" default_value="none" units="unitless"
 		            description="Selection of the method for calving ice (as defined further below)."
-		            possible_values="'none', 'floating', 'topographic_threshold', 'thickness_threshold', 'eigencalving', 'mask'"
+		            possible_values="'none', 'floating', 'topographic_threshold', 'thickness_threshold', 'eigencalving', 'von_Mises_stress', 'mask'"
+		/>
+                <nml_option name="config_use_Albany_flowA_eqn" type="logical" default_value=".false." units="unitless"
```
matthewhoffman, line 132: Let's change this to something like 'config_use_Albany_flowA_eqn_for_vM'.   Nobody like overly verbose option names, but unfortunately the way Fortran handles namelist options internally (and how they are written into the netCDF file metadata), there is no indication of which group an option belongs to.  So I think it could prevent confusion to somehow make it clear that this option is not a general thing but just applies to the von Mises stress calculation.

trhille: Changed from config_use_Albany_flowA_eqn to config_use_Albany_flowA_eqn_for_vM

matthewhoffman:
Passed all COMPASS tests.

PR #812, opened 2021-02-12
Simplify computing landIcePressure and vertical grid in ocean init mode

xylar:
This merge affects only `init` mode.  It modifies the main registry and diagnostics variables to rename an init-mode variable.  All other changes are in the `mode_init` directory.

While we have supported 3 options for initializing the ice draft (SSH) and `landIcePressure`, we now only use the simplest one, which uses the density in the top layer as a guess for the effective density of the ocean.

This merge significantly simplifies and clarifies the process for creating the vertical coordinate and later initializing `landIcePressure` in several test cases by getting rid of the unused options.  The subroutine for creating the vertical coordinate `ocn_init_vertical_grid()` has been moved to `ocn_init_vertical_grids` (where it clearly belongs).

The namelist option `config_iterative_init_variable` is no longer needed and has been removed because we always initialize
`landIcePressure` (not SSH) and we always use the density in the top layer.

The init mode diagnostic variable `modifySSHMask` has been renamed `modifyLandIcePressureMask` to be consistent with these changes.   We now only modify `landIcePressure`, not SSH.

The 4 test cases with ice-shelf cavities:
* `global_ocean`
* `isomip`
* `isomip_plus`
* `sub_ice_shelf_2D`
have been updated with these changes.

### DIFF
--- a/src/core_ocean/mode_init/mpas_ocn_init_ssh_and_landIcePressure.F
+++ b/src/core_ocean/mode_init/mpas_ocn_init_ssh_and_landIcePressure.F
@@ -73,7 +73,7 @@ contains
 !>  This routine produces an initial guess at land-ice pressure using the
 !>  density of the topmost layer as the effective density of seawater within
 !>  the land ice. The resulting land-ice pressure is approximately consistent
-!>  with the in the sense that the horizontal pressure-gradient force (HPGF)
+!>  with the ice draft in the sense that the horizontal pressure-gradient force (HPGF)
 !>  should be small at the ocean surface.
 !>  ocn_init_vertical_grid() should be called to produce the appropriate
 !>  vertical grid before calling this subroutine and activeTracers should be
--- a/src/core_ocean/mode_init/mpas_ocn_init_ssh_and_landIcePressure.F
+++ b/src/core_ocean/mode_init/mpas_ocn_init_ssh_and_landIcePressure.F
@@ -30,8 +30,6 @@ module ocn_init_ssh_and_landIcePressure
    use ocn_constants
    use ocn_config
    use ocn_diagnostics_variables
-   use ocn_init_interpolation
-   use ocn_init_vertical_grids
 
    use ocn_equation_of_state
 
@@ -51,8 +49,7 @@ module ocn_init_ssh_and_landIcePressure
    !
    !--------------------------------------------------------------------
 
-   public :: ocn_init_ssh_and_landIcePressure_vertical_grid, &
-             ocn_init_ssh_and_landIcePressure_balance
+   public :: ocn_init_ssh_and_landIcePressure_balance
 
 
    !--------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


xylar:
I'm getting non-bit-for-bit results in two tests:
```
sub-ice-shelf 2D - default - Smoke Test
sub-ice-shelf 2D - Haney-number constrained - Smoke Test
```
It appears that the salinity in the initial condition is off.  I'll try to find out why.

xylar:
## Testing

I ran the full nightly and land-ice-fluxes regression suites on my Ubuntu laptop with gnu.  All tests passed bit-for-bit except:
```
ocean/sub_ice_shelf_2D/5km/default
ocean/sub_ice_shelf_2D/5km/Haney_number_init
```
It turns out that these test have previously been using `config_iterative_init_variable = 'landIcePressure'` (the default), not `'landIcePressure_from_top_density'` as other test cases with ice-shelf cavities do and as they should have.

When I manually fix this config option and rerun these basline simulations, the results are now bit-for-bit.  So I think the changes are desired.

xylar:
## Further testing

### COMPASS on Anvil

I tested the legacy COMPASS `nightly` and `land_ice_fluxes` test suites on Anvil with Intel.  I manually modified `config_iterative_init_variable = 'landIcePressure'` to be `config_iterative_init_variable = 'landIcePressure_from_top_density'` in every test case the baseline.  This ensures that the old configuration matches what should now be the default.

With these changes, all tests in both suites pass.

### E3SM on Anvil

I ran a test with current `master` as the baseline against a test merge of this branch into `mpas-source` with:
```
```
it passed (see `/lcrc/group/e3sm/ac.xylar/scratch/anvil/SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_intel.C.20210216_061527_4cgf5z`).

xylar:
@mark-petersen, as I will bring up at today's Seven Samurai meeting, this is a priority for @cbegeman, @dengwirda and me because our work on adding `minLevelCell` relies on this.  I needed to make these simplifications in order to add a version of `sub_ice_shelf_2D` with a z-level coordinate.

src/core_ocean/mode_init/mpas_ocn_init_ssh_and_landIcePressure.F
```diff
@@ -65,67 +62,23 @@ module ocn_init_ssh_and_landIcePressure
...
-!>  or visa versa (if config_iterative_init_variable = 'landIcePressure' or 'landIcePressure_from_top_density').
-!>  The routine produces an initial guess at land-ice pressure or SSH using either the density of the topmost layer
-!>  ('landIcePressure_from_top_density') or of all layers above the SSH ('ssh' or 'landIcePressure')
-!>  to determine the effective density of seawater within the land ice.
-!>  The resulting land-ice pressure and SSH are approximately consistent with one another
-!>  in the sense that the horizontal pressure-gradient force (HPGF)
+!>  This routine produces an initial guess at land-ice pressure using the
+!>  density of the topmost layer as the effective density of seawater within
+!>  the land ice. The resulting land-ice pressure is approximately consistent
+!>  with the in the sense that the horizontal pressure-gradient force (HPGF)
```
cbegeman, line 76:
```suggestion
!>  with the ice draft in the sense that the horizontal pressure-gradient force (HPGF)
```

xylar:
Thanks @cbegeman

cbegeman (approved):
The PR looks good to me. I've successfully run the same regression suite @xylar mentioned above with this branch.

mark-petersen (approved):
Looks good. Passes nightly test suite with gnu debug and intel debug.

PR #813, opened 2021-02-16
Port ocean diagnostics to GPU - Phase 1

mattdturner:
This PR is the first phase of porting the ocean diagnostics code to GPUs using OpenACC.  It contains a subset of the necessary changes for running on the GPU that retain bit-for-bit accuracy with the current version when running on the CPU.

A phase 2 PR will contain the loop structure changes for optimal GPU performance that result in roundoff-level changes for the CPU code.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -1302,7 +1302,7 @@ contains
       !$acc                  normalTransportVelocity, normalGMBolusVelocity, vertVelocityTop, &
       !$acc                  vertTransportVelocityTop, vertGMBolusVelocityTop, invAreaCell) &
       !$acc          private(invAreaCell1, i, iEdge, edgeSignOnCell_temp, dcEdge_temp, &
-      !$acc                  dvEdge_temp, k, r_tmp) 
+      !$acc                  dvEdge_temp, k, r_tmp)
 #else
       !$omp parallel
       !$omp do schedule(runtime) &
@@ -1348,7 +1348,7 @@ contains
          end do
       end do
 #ifdef MPAS_OPENACC
-      !$acc exit data delete(div_hu, div_huTransport, div_huGMBolus) 
+      !$acc exit data delete(div_hu, div_huTransport, div_huGMBolus)
 #else
       !$omp end do
       !$omp end parallel
@@ -1361,7 +1361,7 @@ contains
       !$acc parallel loop gang vector &
       !$acc          present(tangentialVelocity, nEdgesOnEdge, edgesOnEdge, weightsOnEdge, &
       !$acc                  maxLevelEdgeTop, normalVelocity) &
-      !$acc          private(i, eoe, weightsOnEdge_temp, k) 
+      !$acc          private(i, eoe, weightsOnEdge_temp, k)
 #else
       !$omp parallel
       !$omp do schedule(runtime) private(i, eoe, weightsOnEdge_temp, k)
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -985,11 +985,20 @@ contains
       if(config_use_cvmix_kpp) then
 
         nCells = nCellsOwned
+        surfaceLayerDepth = config_cvmix_kpp_surface_layer_averaging
 
+#ifdef MPAS_OPENACC
+        !$acc enter data copyin(layerThickness, activeTracers)
+
+        !$acc parallel loop gang vector &
+        !$acc          present(tracersSurfaceLayerValue, indexSurfaceLayerDepth, maxLevelCell, &
+        !$acc                  layerThickness, activeTracers) &
+        !$acc          private(sumSurfaceLayer, k, rSurfaceLayer)
+#else
         !$omp parallel
         !$omp do schedule(runtime) private(surfaceLayerDepth, sumSurfaceLayer, k, rSurfaceLayer)
+#endif
         do iCell=1,nCells
-          surfaceLayerDepth = config_cvmix_kpp_surface_layer_averaging
           sumSurfaceLayer=0.0_RKIND
           tracersSurfaceLayerValue(:,iCell) = 0.0_RKIND
           indexSurfaceLayerDepth(iCell) = -9.e30
@@ -1003,7 +1012,6 @@ contains
              exit
            endif
           end do
-          tracersSurfaceLayerValue(:, iCell) = 0.0_RKIND
           do k=1,int(rSurfaceLayer)
             tracersSurfaceLayerValue(:,iCell) = tracersSurfaceLayerValue(:,iCell) + activeTracers(:,k,iCell) &
                                               * layerThickness(k,iCell)
@@ -1012,8 +1020,13 @@ contains
           tracersSurfaceLayerValue(:,iCell) = (tracersSurfaceLayerValue(:,iCell) + fraction(rSurfaceLayer) &
                                             * activeTracers(:,k,iCell) * layerThickness(k,iCell)) / surfaceLayerDepth
         enddo
+#ifdef MPAS_OPENACC
+        !$acc update host(tracersSurfaceLayerValue, indexSurfaceLayerDepth)
+        !$acc exit data delete(layerThickness, activeTracers)
+#else
         !$omp end do
         !$omp end parallel
+#endif
 
         nEdges = nEdgesOwned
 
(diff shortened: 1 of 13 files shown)


mattdturner:
This isn't ready to merge.  I need to run additional tests to confirm that it is in fact BFB on the CPU.  I also need to nullify the pointers in `ocn_diagnostics_variables_destroy`.

When running on GPUs and comparing `ocean/develop` to this PR, only roundoff differences are observed.

@philipwjones could you do a general review of how I've done the OpenACC directives to see if there is anything glaringly wrong?  Based on my testing, everything looks to be correct but I wanted to check to make sure.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -1560,26 +1722,41 @@ subroutine ocn_vert_transport_velocity_top(meshPool, verticalMeshPool, scratchPo
...
       ! See Ringler et al. (2010) jcp paper, eqn 19, 21, and fig. 3.
+#ifdef MPAS_OPENACC
+      !$acc enter data create(div_hu, projectedSSH, ALE_Thickness)
+
+! DEBUG, MDT :: need to add OpenACC
+!      !$acc parallel loop &
+!      !$acc          present(div_hu, invAreaCell, nEdgesOnCell, edgesOnCell, maxLevelEdgeTop, &
+!      !$acc                  layerThicknessEdge, normalVelocity, dvEdge, edgeSignOnCell, &
+!      !$acc                  oldSSH, projectedSSH) &
+!      !$acc          private(div_hu_btr, invAreaCell1, i, iEdge, flux, k)
```
mattdturner, line 1733: @philipwjones If I uncomment this directive (and the directives around line 1754), then I get a segfault when running on GPUs.  The error looks to be the result of an incorrect calculation of the values in this subroutine.  I've deleted and re-generated this directive numerous times and always end up with the same error.  Could you check the directive to see if you can notice a problem?

src/core_ocean/shared/mpas_ocn_mesh.F
```diff
@@ -116,6 +116,9 @@ module ocn_mesh
       meshDensity, &! density of mesh
       angleEdge        ! angle the edge normal makes with local east
 
+   real(kind=RKIND), public, dimension(:), allocatable :: &
+      invAreaCell   ! inverse of area of each cell
```
mattdturner, line 120: I added an array for `invAreaCell` to pre-calculate this value since it is used throughout the model.  I have yet to propagate this to any routines aside from diagnostics.

src/core_ocean/shared/mpas_ocn_diagnostics.F
philipwjones: looking into this now - I have had trouble occasionally with PGI getting confused about these reductions so we'll see if that's the issue and come up with a workaround

mattdturner:
After rebase, this is BFB w/ the current `ocean/develop` for Intel w/ Debug in MPAS-O standalone nightly regression suite.  I'm in the process of setting up a BFB test w/ GNU on Grizzly

src/core_ocean/shared/mpas_ocn_diagnostics.F
philipwjones: Update - the loop itself is computing correctly. There is a transfer bug later in the routine. Now that you've rebased, I'll update to the latest and then continue the review and let you know

mattdturner: @philipwjones Thanks for looking at this.  I had gone through the subroutines loop by loop, verifying that the overall output from the test was within roundoff of the baseline.  Are you saying that there is an incorrect data transfer elsewhere that is only showing up as a bug when I add OpenACC to this loop?  If so, any suggestions on how to debug such an issue?

philipwjones: Actually it isn't really a bug. When I uncommented that loop, I forgot to deal with a later data update of div_hu that's only needed if this loop in done on cpu. When I took care of that, this routine runs fine for me in a QU240 test on Summit.  However, there must still be some issues elsewhere because this same QU240 test is leading to some very wrong answers with OPENACC on. Still trying to track that down. Sorry for the confusion.

src/core_ocean/shared/mpas_ocn_equation_of_state_jm.F
```diff
@@ -393,12 +393,14 @@ subroutine ocn_equation_of_state_jm_density_only(nVertLevels,      &
...
       !$acc update host(density)
-      !$acc exit data delete(density, p, p2, tracerTemp, tracerSalt)
+      !$acc exit data delete(p, p2, tracerTemp, tracerSalt)
 #else
       !$omp end do
       !$omp end parallel
 #endif
 
+      call mpas_log_write('leaving eq of state')
+
```
mattdturner, line 403: I don't recall adding this.  I'm not sure if it showed up during rebase?  Unless there is a need for this log statement, I would rather remove it...

philipwjones: Part of our equation of state contact tracing ;^)    definitely remove

philipwjones:
I found and fixed a number of issues including the problematic loop above and created a PR into @mattdturner fork/branch with the changes. These fixes agreed within roundoff between CPU/GPU runs on summit with PGI. Note that the PGI compiler was non-bfb in optimized mode for even the CPU version compared to latest (couldn't test in debug mode because PGI complains about an uninitialized pointer for the GM arguments to hmix routines - a bug that existed prior to this branch). But those diffs were also within roundoff.

mattdturner:
@philipwjones Thanks for looking into this.  I'll have to spend a little bit of time looking at the changes you made in the PR to my fork, to make sure that I understand what my mistakes were.

philipwjones:
@mattdturner  - no problem. some were just bookkeeping things, tracking when a variable is updated by the device or host and that's why I ended up adding a bunch of "safety" transfers to keep everything synchronized on host/device - we'll need to remove/consolidate some of those later for efficiency. The one that was causing the biggest problem was in the split explicit driver where the first calls to vert transport vel were using a state variable (normVel) and needed a copyin/delete, so you had also used the same copyin/delete on a second set of calls but in that case the variable was a diagnostic that was already on the device, so needed updates rather than copyin. It took me a while to find the problem because I made the exact same assumption and didn't see it either.

mattdturner:
I just rebased.  After rebase, I see round-off level errors on the GPU.  The nightly regression suite for MPAS-O is BFB (on CPU) for Intel (w/ and w/o Debug), and GNU (w/ and w/o Debug).

This should be ready to go.

mattdturner:
Rebased.  After rebase, still matches BFB to `ocean/develop` baseline in MPAS standalone w/ Intel + Debug

mattdturner:
Pushed a new commit that changes some loop indices similar to what was done in https://github.com/MPAS-Dev/MPAS-Model/pull/851.  This PR now (once again) runs successfully on Summit w/ the `pgi-summit` build target.

mattdturner:
PASSes MPAS-O nightly regression suite (BFB compared to `ocean/develop`) for Intel (DEBUG and Optimized) and GNU (DEBUG and Optimized).

From E3SM...
```
PASS:
SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_gnu
SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_gnu
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_gnu
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel
PEM_Ln9.T62_oQU240.GMPAS-IAF.anvil_gnu
PEM_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_gnu
PEM_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_intel
PEM_Ln9.T62_oQU240.GMPAS-IAF.anvil_intel
PET_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_intel
PET_Ln9.T62_oQU240.GMPAS-IAF.anvil_gnu
PET_Ln9.T62_oQU240.GMPAS-IAF.anvil_intel
PET_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_gnu
SMS_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_gnu
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_gnu
PET_Ln3_D.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_gnu
PEM_Ln9_D.T62_oQU240.GMPAS-IAF.anvil_intel
PEM_Ln9_D.T62_oQU240.GMPAS-IAF.anvil_gnu
PASS:
PEM_Ln9.ne30_oECv3.A_WCYCL1850S.chrysalis_intel.C
  - PASS PEM_Ln9.ne30_oECv3.A_WCYCL1850S.chrysalis_intel BASELINE /lcrc/group/e3sm/baselines/chrys/intel
/master:
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel.C
  - PASS SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel BASELINE /lcrc/group/e3sm/baselines/chrys/
intel/master:


--
FAIL:
FAIL SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi BASELINE /compyfs/e3sm_baselines/pgi/master: DIFF
  - FAIL SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi BASELINE /compyfs/e3sm_baselines/pgi/master: DIFF

```

mattdturner:
Also FAILs BFB check for `ERS.ne11_oQU240.A_WCYCL1850.compy_pgi` and `SMS_P12x2.ne4_oQU240.A_WCYCL1850.compy_pgi.allactive-mach_mods`.  I'll have to look into why PGI fails but Intel passes.

mattdturner:
Adding the `edgeSignOnCell` pointer retrieval back to diagnostics (see below) allows this to be BFB.

This might be resolved by https://github.com/MPAS-Dev/MPAS-Model/pull/852.  If not, I'll apply the below patch to ensure BFB for this PR.

```
diff --git a/src/core_ocean/shared/mpas_ocn_diagnostics.F b/src/core_ocean/shared/mpas_ocn_diagnostics.F
index 9eddaec3b..6214aa084 100644
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -1813,6 +1813,7 @@ contains
       integer :: iEdge, iCell, k, i, nCells, kmin, kmax
 
       real (kind=RKIND) :: flux, invAreaCell1, div_hu_btr
+      integer, dimension(:,:), pointer :: edgeSignOnCell
 
       ! Scratch Arrays
       ! projectedSSH: projected SSH at a new time
@@ -1835,6 +1836,8 @@ contains
 
       ncells = nCellsAll
 
+      call mpas_pool_get_array(meshPool, 'edgeSignOnCell', edgeSignOnCell)
+
       allocate(div_hu(nVertLevels, nCells), &
                projectedSSH(nCells), &
                ALE_Thickness(nVertLevels, nCells))
```

mattdturner:
Replaced by https://github.com/E3SM-Project/E3SM/pull/4346

PR #814, opened 2021-02-17
Add landice ismip6 calving & melting params., phase 2

trhille:
This is part 2 of a multi-part merge of the branch that includes ISMIP6 submarine face melting and calving parameterizations for Greenland. Builds on PR #810. 

The main purpose of this branch is to change the subroutine apply_calving_velocity to a generalized subroutine  called li_apply_front_ablation_velocity, which treats both face-melting (e.g., from config_front_mass_bal_grounded = 'ismip6' or 'uniform') and calving. Calving and melting velocities are calculating in the calling routines and passed to li_apply_front_ablation_velocity to be converted to a calving or melting thickness, which is passed back to the calling routines to be removed from the thickness field. The location at which ablation is applied is determined by logicals applyToGrounded, applyToFloating, and applyToGroundingLine, which are set within the calling routines. Front mass balance (i.e., submarine melting) is set to be applied at the grounding line (last grounded cell if no floating ice; first floating cell if floating ice), while eigencalving and von Mises stress calving are applied to grounded and floating marine margins. 

This branch includes a more rigorous treatment of floating non-dynamic cells, including those 'stranded' cells with only non-dynamic neighbors. There are also a number of critical bug fixes in li_apply_front_ablation (formerly apply_calving_velocity), including copying forward thickness and advective velocity from dynamic to non-dynamic cells, correcting indexing errors, and correcting the edge length scaling function. Also includes a bug fix in mpas_li_mask.F to properly treat edges of the mesh.

There is now an option to remove all dynamic floating ice in a timestep while also using the von Mises stress calving law by setting config_floating_von_Mises_threshold_stress = 0.0 in namelist.

### DIFF
--- a/src/core_landice/Registry.xml
+++ b/src/core_landice/Registry.xml
@@ -1138,7 +1138,7 @@ is the value of that variable from the *previous* time level!
                      description="mask of grid cells at which front ablation will be applied. Determined by settings of applyToGrounded, applyToFloating, and applyToGroundingLine in routines that call li_apply_front_ablation_velocity."
                 />
                 />
-		<var name="faceMeltSpeed" type="real" dimensions="nCells Time" units="m s^{-1}" packages="ismip6GroundedFaceMelt"
+		<var name="faceMeltSpeed" type="real" dimensions="nCells Time" units="m s^{-1}" time_levs="1"
 		     description="linear speed of submarine melting at grounded glacier front"
   		/>
                 <var name="ismip6Runoff" type="real" dimensions="nCells Time" units="kg m^{-2} s^{-1}" packages="ismip6GroundedFaceMelt" default_value="0.0"
--- a/src/core_landice/Registry.xml
+++ b/src/core_landice/Registry.xml
@@ -1050,11 +1050,11 @@ is the value of that variable from the *previous* time level!
                 <var name="calvingVelocity" type="real" dimensions="nCells Time" units="m s^{-1}" time_levs="1"
                      description="rate of calving front retreat due to calving, represented as a velocity normal to the calving front (in the x-y plane)."
                 />
-                <var name="uncalvedVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
-                     description="volume of ice that was left uncalved from required calving flux at non-dynamic cells"
+                <var name="unablatedVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                     description="volume of ice that was left uncalved or unmelted from required calving or melting flux at non-dynamic cells"
                 />
-                <var name="uncalvedVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
-                     description="volume of ice that was left uncalved from required calving flux at dynamic cells"
+                <var name="unablatedVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                     description="volume of ice that was left uncalved or unmelted from required calving or melting flux at dynamic cells"
                 />
                 <var name="calvedVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
                      description="volume of ice that was calved from non-dynamic cells"
--- a/src/core_landice/Registry.xml
+++ b/src/core_landice/Registry.xml
@@ -1134,8 +1134,8 @@ is the value of that variable from the *previous* time level!
                      description="Optional field of offset that should be applied to the melt field calculated by the ISMIP6 melt method.  Can be used to run in anomaly mode, but this is not the anomaly.  See Registry.xml for details."
                 />
                 <!-- Variables needed for ISMIP6 grounded face melt parameterization -->
-                <var name="groundedMarineMarginMask" type="integer" dimensions="nCells Time" units="none" time_levs="1" 
-                     description="mask of grid cells that are grounded and adjacent to the ocean. 0=interior, 1=grounded ice margin"
+                <var name="frontAblationMask" type="integer" dimensions="nCells Time" units="none" time_levs="1" 
+                     description="mask of grid cells at which front ablation will be applied. Determined by settings of applyToGrounded, applyToFloating, and applyToGroundingLine in routines that call li_apply_front_ablation_velocity."
                 />
                 />
 		<var name="faceMeltSpeed" type="real" dimensions="nCells Time" units="m s^{-1}" packages="ismip6GroundedFaceMelt"
--- a/src/core_landice/Registry.xml
+++ b/src/core_landice/Registry.xml
@@ -1133,10 +1133,10 @@ is the value of that variable from the *previous* time level!
                 <var name="ismip6shelfMelt_offset" type="real" dimensions="nCells" units="kg m^{-2} s^{-1}" packages="ismip6ShelfMelt" default_value="0.0"
                      description="Optional field of offset that should be applied to the melt field calculated by the ISMIP6 melt method.  Can be used to run in anomaly mode, but this is not the anomaly.  See Registry.xml for details."
                 />
-                <var name="groundedMarineMarginMask" type="integer" dimensions="nCells Time" units="none" time_levs="1"
+                <!-- Variables needed for ISMIP6 grounded face melt parameterization -->
+                <var name="groundedMarineMarginMask" type="integer" dimensions="nCells Time" units="none" time_levs="1" 
                      description="mask of grid cells that are grounded and adjacent to the ocean. 0=interior, 1=grounded ice margin"
                 />
-                <!-- Variables needed for ISMIP6 grounded face melt parameterization -->
                 <var name="faceMeltRateApplied" type="real" dimensions="nCells Time" units="m^{3} s^{-1}" packages="ismip6GroundedFaceMelt"
                      description="Volumetric submarine melt rate applied at grounded marine margin"
                 />
@@ -1155,7 +1155,6 @@ is the value of that variable from the *previous* time level!
                <var name="unmeltedVolume" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
                      description="volume of ice that was left unmelted from required facemelt flux due to only applying flux over immediate neighbors (diagnostic field to assess if this limitation is a problem)"
 		/>
->>>>>>> trevor/landice/ismip6_calving_melting_phase1
                 <!--  description="Optional field of offset that should be appliped to the melt field calculated by the ISMIP6 melt method.  This field is added directly to the melt field calculated by the method and can be used to use the method in anomaly mode.  Note this field is NOT the anomaly.  Let m be the final melt field generated, m_t be the ISMIP6 melt calculation at any time, m_t0 the ISMIP6 melt at time 0, and m_0 a reference melt field (e.g. observed melt field used for initialization).  The melt anomaly would be defined as (delta m)=m_t-m_t0. Then, m=(delta m) + m_0.  Rearranging, m=m_t+(m_0-m_t0).  ismip6shelfMelt_offset is the term in parentheses.  Note that the general usage requires that the first time step be run with ismip6shelfMelt_offset=0 in order to determine m_t0, and then NCO or similar tool used to generate the (m_0-m_t0) field to be input as ismip6shelfMelt_offset in the forward run." -->
                 />
                 <var name="upliftRate"            type="real"     dimensions="nCells Time"
--- a/src/core_landice/Registry.xml
+++ b/src/core_landice/Registry.xml
@@ -1056,22 +1056,22 @@ is the value of that variable from the *previous* time level!
                 <var name="unablatedVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
                      description="volume of ice that was left uncalved or unmelted from required calving or melting flux at dynamic cells"
                 />
-                <var name="calvedVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
-                     description="volume of ice that was calved from non-dynamic cells"
+                <var name="ablatedVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                     description="volume of ice that was ablated from non-dynamic cells"
                 />
-                <var name="calvedVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
-                     description="volume of ice that was calved from dynamic cells"
+                <var name="ablatedVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                     description="volume of ice that was ablated from dynamic cells"
                 />
-                <var name="requiredCalvingVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                <var name="requiredAblationVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
                      description="bookkeeping field for how much volume should be removed from a non-dynamic cell"
                 />
-                <var name="requiredCalvingVolumeNonDynEdge" type="real" dimensions="nEdges Time" units="m^3" time_levs="1"
+                <var name="requiredAblationVolumeNonDynEdge" type="real" dimensions="nEdges Time" units="m^3" time_levs="1"
                      description="bookkeeping field for how much volume should be removed from a non-dynamic edge"
                 />
-                <var name="requiredCalvingVolumeDynEdge" type="real" dimensions="nEdges Time" units="m^3" time_levs="1"
+                <var name="requiredAblationVolumeDynEdge" type="real" dimensions="nEdges Time" units="m^3" time_levs="1"
                      description="bookkeeping field for how much volume should be removed from a dynamic edge"
                 />
-                <var name="requiredCalvingVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                <var name="requiredAblationVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
                      description="bookkeeping field for how much volume should be removed from a dynamic edge"
                 />
                 <var name="basalWaterThickness" type="real" dimensions="nCells Time" units="m" time_levs="1"
--- a/src/core_landice/Registry.xml
+++ b/src/core_landice/Registry.xml
@@ -1137,8 +1137,6 @@ is the value of that variable from the *previous* time level!
                 <var name="groundedMarineMarginMask" type="integer" dimensions="nCells Time" units="none" time_levs="1" 
                      description="mask of grid cells that are grounded and adjacent to the ocean. 0=interior, 1=grounded ice margin"
                 />
-                <var name="faceMeltRateApplied" type="real" dimensions="nCells Time" units="m^{3} s^{-1}" packages="ismip6GroundedFaceMelt"
-                     description="Volumetric submarine melt rate applied at grounded marine margin"
                 />
 		<var name="faceMeltSpeed" type="real" dimensions="nCells Time" units="m s^{-1}" packages="ismip6GroundedFaceMelt"
 		     description="linear speed of submarine melting at grounded glacier front"
--- a/src/core_landice/Registry.xml
+++ b/src/core_landice/Registry.xml
@@ -181,7 +181,7 @@
 			    possible_values="Any positive real value"
 		/>
                 <nml_option name="config_floating_von_Mises_threshold_stress" type="real" default_value="1.0e6" units="Pa"
-                            description="Threshold von Mises stress value required for calving velocity to exceed ice velocity on floating ice. Generally seems to need to be low (1e5 to 2e5 for Humboldt Glacier) to allow floating ice to retreat. sigma_max in Morlighem et al. (2016) eq. 4. 1 MPa default value is from Morlighem et al.'s calibration for Store Glacier."
+                            description="Threshold von Mises stress value required for calving velocity to exceed ice velocity on floating ice. Generally seems to need to be low (1e5 to 2e5 for Humboldt Glacier) to allow floating ice to retreat. If set to 0.0, floating ice will be removed from dynamic cells each timestep; calvingThickness will be correct, but calvingVelocity will not be correct for those cells. sigma_max in Morlighem et al. (2016) eq. 4. 1 MPa default value is from Morlighem et al.'s calibration for Store Glacier."
                             possible_values="Any positive real value"
                 />
 	</nml_record>
--- a/src/core_landice/mode_forward/mpas_li_calving.F
+++ b/src/core_landice/mode_forward/mpas_li_calving.F
@@ -1758,10 +1758,9 @@ module li_calving
 
           ! a2.Set thicknessForAblation and ablationVelocity of stranded non-dynamic cells to
           ! average of non-dynamic neighbors. Stranded non-dynamic cells are
-          ! identified as those that still have thicknessForAblation == 0.0
+          ! identified as those with frontAblationMask == 3
       do iCell = 1, nCells
-         if ( thicknessForAblation(iCell) == 0.0_RKIND .and. frontAblationMask(iCell)>0 &
-              .and. (.not. li_mask_is_dynamic_ice(cellMask(iCell)))) then
+         if ( frontAblationMask(iCell) == 3.0 ) then
             ablationVelSum = 0.0_RKIND
             thkSum = 0.0_RKIND
             uvelSum = 0.0_RKIND
--- a/src/core_landice/mode_forward/mpas_li_calving.F
+++ b/src/core_landice/mode_forward/mpas_li_calving.F
@@ -1124,7 +1124,6 @@ module li_calving
       real(kind=RKIND) :: calvingSubtotal
       integer :: err_tmp
       logical :: applyToGrounded, applyToFloating, applyToGroundingLine
-      logical :: applyCalving, applyMelting
 
       err = 0
 
@@ -1132,8 +1131,6 @@ module li_calving
       applyToGrounded = .true.
       applyToFloating = .true.
       applyToGroundingLine = .false.
-      applyCalving = .true.
-      applyMelting = .false.
 
       call mpas_pool_get_config(liConfigs, 'config_print_calving_info', config_print_calving_info)
       call mpas_pool_get_config(liConfigs, 'config_calving_eigencalving_parameter_scalar_value', &
@@ -1195,10 +1192,20 @@ module li_calving
          ! First calculate the front retreat rate (Levermann eq. 1)
          calvingVelocity(:) = eigencalvingParameter(:) * max(0.0_RKIND, eMax(:)) * max(0.0_RKIND, eMin(:)) ! m/s
 
-         ! Convert calvingVelocity to calvingThickness
-          call li_apply_front_ablation_velocity(meshPool, geometryPool, velocityPool, applyToGrounded, &
-                                              applyToFloating, applyToGroundingLine, applyCalving, &
-                                              applyMelting, domain, err)
+         call mpas_log_write("calling li_apply_front_ablation_velocity from eigencalving")
+          ! Convert calvingVelocity to calvingThickness
+          call li_apply_front_ablation_velocity(meshPool, geometryPool, velocityPool, &
+                                              calvingThickness, calvingVelocity, applyToGrounded, &
+                                              applyToFloating, applyToGroundingLine, domain, err)
+         ! Update halos on calvingThickness or faceMeltingThickness before
+         ! applying it.
+         ! Testing seemed to indicate this is not necessary, but I don't
+         ! understand
+         ! why not, so leaving it.
+         ! NOTE: THIS WILL NOT WORK ON MULTIPLE BLOCKS PER PROCESSOR
+         call mpas_timer_start("halo updates")
+         call mpas_dmpar_field_halo_exch(domain, 'calvingThickness')
+         call mpas_timer_stop("halo updates")
 
          ! === apply calving ===
          thickness(:) = thickness(:) - calvingThickness(:)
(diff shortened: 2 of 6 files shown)


src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -45,7 +44,7 @@ module li_calving
    !
    !--------------------------------------------------------------------
 
-   public :: li_calve_ice, li_restore_calving_front
+   public :: li_calve_ice, li_restore_calving_front, apply_front_ablation_velocity
```
matthewhoffman, line 47: The MPAS convention is to proceed all public routine with `li_`, so now that this routine is public, that should be added everywhere.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1187,10 +1195,10 @@ subroutine eigencalving(domain, err)
          ! First calculate the front retreat rate (Levermann eq. 1)
          calvingVelocity(:) = eigencalvingParameter(:) * max(0.0_RKIND, eMax(:)) * max(0.0_RKIND, eMin(:)) ! m/s
 
-
          ! Convert calvingVelocity to calvingThickness
-         call apply_calving_velocity(meshPool, geometryPool, velocityPool, domain, err)
-
+          call apply_front_ablation_velocity(meshPool, geometryPool, velocityPool, applyToGrounded, &
```
matthewhoffman, line 1199: If you don't want all the new variables above, you can just pass the values into the subroutine.  If you want to pass the values but make it clear what they are, you can do it like `applyToGrounded=.true.`.  You don't need to remove these new variables, but thought I'd mention it.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1366,14 +1384,25 @@ subroutine von_Mises_calving(domain, err)
              if ( .not. li_mask_is_floating_ice(cellMask(iCell)) ) then
                 calvingVelocity(iCell) = sqrt(xvelmean(iCell)**2.0_RKIND + yvelmean(iCell)**2.0_RKIND) * &
                                          vonMisesStress(iCell) / config_grounded_von_Mises_threshold_stress
-             elseif ( li_mask_is_floating_ice(cellMask(iCell)) ) then
-                calvingVelocity(iCell) = sqrt(xvelmean(iCell)**2.0_RKIND + yvelmean(iCell)**2.0_RKIND) * &
+             elseif ( li_mask_is_floating_ice(cellMask(iCell)) .and. config_floating_von_Mises_threshold_stress .ne. 0.0_RKIND) then
+                calvingVelocity(iCell) = sqrt(xvelmean(iCell)**2 + yvelmean(iCell)**2) * &
                                          vonMisesStress(iCell) / config_floating_von_Mises_threshold_stress
+             !if gloating VM threshold is zero, remove floating ice
```
matthewhoffman, line 1390: 'gloating' typo sounds like a Lewis Carroll line.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1366,14 +1384,25 @@ subroutine von_Mises_calving(domain, err)
              if ( .not. li_mask_is_floating_ice(cellMask(iCell)) ) then
                 calvingVelocity(iCell) = sqrt(xvelmean(iCell)**2.0_RKIND + yvelmean(iCell)**2.0_RKIND) * &
                                          vonMisesStress(iCell) / config_grounded_von_Mises_threshold_stress
-             elseif ( li_mask_is_floating_ice(cellMask(iCell)) ) then
-                calvingVelocity(iCell) = sqrt(xvelmean(iCell)**2.0_RKIND + yvelmean(iCell)**2.0_RKIND) * &
+             elseif ( li_mask_is_floating_ice(cellMask(iCell)) .and. config_floating_von_Mises_threshold_stress .ne. 0.0_RKIND) then
```
matthewhoffman, line 1387: Can you add a comment for why the mask might indicate floating but the vM stress is 0?  Does the vM stress==0 case need to be handled differently?

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1366,14 +1384,25 @@ subroutine von_Mises_calving(domain, err)
...
+                calvingVelocity(iCell) = sqrt(xvelmean(iCell)**2 + yvelmean(iCell)**2) * &
                                          vonMisesStress(iCell) / config_floating_von_Mises_threshold_stress
+             !if gloating VM threshold is zero, remove floating ice
+             elseif ( li_mask_is_floating_ice(cellMask(iCell)) .and. config_floating_von_Mises_threshold_stress .eq. 0.0_RKIND &
+                       .and. li_mask_is_dynamic_ice(cellMask(iCell)) ) then
+                thickness(iCell) = 0.0_RKIND
```
matthewhoffman, line 1393: The convention in the calving module has been to never remove thickness directly, but instead populate the calvingThickness field first and then apply that to thickness later.  Is that possible to do here?

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -48,7 +48,7 @@ module li_iceshelf_melt
    !
    !--------------------------------------------------------------------
 
-   public :: li_basal_melt_floating_ice
+   public :: li_basal_melt_floating_ice, grounded_face_melt_ismip6
```
matthewhoffman, line 51: Please add `li_` prefix to this new public routine.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1166,6 +1166,7 @@ end subroutine iceshelf_melt_ismip6
 
     subroutine grounded_face_melt_ismip6(domain, err)
 
+      use li_calving
```
matthewhoffman, line 1169: This dependency should also be added to the Makefile.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1253,63 +1266,65 @@ subroutine grounded_face_melt_ismip6(domain, err)
...
-            endif           
+            endif
+
          enddo
 
+         call mpas_log_write("calling apply_front_ablation_velocity")
          ! Distribute melt among neighbors
-         call distribute_face_melt_flux(meshPool, geometryPool, scratchPool, groundedMarineMarginMask, err_tmp)
-         err = ior(err, err_tmp)
+       !  call distribute_face_melt_flux(meshPool, geometryPool, scratchPool, groundedMarineMarginMask, err_tmp)
```
matthewhoffman, line 1313: This routine should be deleted.  RIP

matthewhoffman (changes requested):
I did a cursory review and identified a handful of small things.  I need to do another, more careful review of the more extensive changes in the PR, but I thought I'd pass these along now.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Fixed while improving vM stress == 0.0 implementation.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: calvingThickness is now populated for dynamic floating cells after calling apply_ablation_velocity if config_floating_von_Mises_threshold_stress is set to 0.0

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1687,37 +1765,36 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
                ! No need to check what type of edge this is - only required edges are nonzero
-               requiredCalvingVolumeDynCell(iCell) = requiredCalvingVolumeDynCell(iCell) + &
-                          requiredCalvingVolumeDynEdge(iEdge) ! Keep running total
+               requiredAblationVolumeDynCell(iCell) = requiredAblationVolumeDynCell(iCell) + &
+                          requiredAblationVolumeDynEdge(iEdge) ! Keep running total
             enddo
             ! c. Apply calving here
-            removeVolumeHere = min(cellVolume(iCell), requiredCalvingVolumeDynCell(iCell))  ! Don't use more than available
-            calvingThickness(iCell) = removeVolumeHere / areaCell(iCell)
-            calvedVolumeDynCell(iCell) = removeVolumeHere
```
matthewhoffman, line 1703: The corresponding rename of this line is missing.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1779,27 +1856,39 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
-                 "  Try using a smaller timestep or apply_calving_velocity may need improvements for this simulation.")
+         call mpas_log_write("Deleting stranded floating cells accounts for more than 1% of ablation loss." // &
+                 "  Try using a smaller timestep or apply_front_ablation_velocity may need improvements for this simulation.")
          err_tmp = 1
          err = ior(err, err_tmp)
       endif
 
 
-      ! Update halos on calvingThickness before applying it.
+      ! Update halos on calvingThickness or meltingThickness before applying it.
```
matthewhoffman, line 1866: If we  modify the calling signature as I suggested above, then I think this halo update will have to be performed by the calling routine.  So this halo update should be deleted and there should be a note added to the subroutine description that a halo update should be made after this routine.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1366,16 +1384,28 @@ subroutine von_Mises_calving(domain, err)
...
-          call apply_calving_velocity(meshPool, geometryPool, velocityPool, domain, err)
-
+          call apply_front_ablation_velocity(meshPool, geometryPool, velocityPool, applyToGrounded, &
+                                      applyToFloating, applyToGroundingLine, applyCalving, applyMelting, domain, err)
+
+
+          ! If floating VM threshold is zero, remove floating ice.
+          if ( config_floating_von_Mises_threshold_stress .eq. 0.0_RKIND ) then
+             do iCell = 1,nCells
+                if ( li_mask_is_floating_ice(cellMask(iCell)) .and. &
```
matthewhoffman, line 1401: Thinking about this more carefully, we may need to be more careful at high res.  We sometimes get 'subglacial lakes' a bit behind the GL where ice goes afloat in isolated pockets.  We wouldn't want to zap those spots.  The existing float kill routine also suffers from this issue.  We could add a TODO item for this for now and deal with it in a later PR if/when we run into the issue operationally.

I think the solution would be to do a flood-fill algorithm from all open ocean cells through all floating cells.  Because that is a lot more complicated than the code here, we might want to write that in a way that we implement that in the float-kill routine and call the float-kill routine from here.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1207,8 +1209,18 @@ subroutine grounded_face_melt_ismip6(domain, err)
...
       real (kind=RKIND), dimension(:), pointer :: faceMeltRateApplied, faceMeltSpeed
       integer :: err_tmp
-
+      logical :: applyToFloating, applyToGrounded, applyToGroundingLine
+      logical :: applyMelting, applyCalving
+       
       err = 0
+      call mpas_log_write('Starting face melt routine')
+
+      ! Define logicals for call to apply_front_ablation_velocity
```
matthewhoffman, line 1218: As stated above, these explicit variables are not necessary.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1166,6 +1166,7 @@ end subroutine iceshelf_melt_ismip6
 
     subroutine grounded_face_melt_ismip6(domain, err)
```
matthewhoffman, line 1167: Variables no longer used in this routine that should be deleted:
```
scratchPool
nEdgesSolve
maxEdges
calvingVelocity
areaCell is used but is retrieved twice
cellsOnEdge
lowerSurface
groundedMarineMarginMask
faceMeltRateApplied & deltat are currently used but only for the now retired distribute_face_melt_flux routine, so should be deleted.  faceMeltRateApplied will need to be deleted from Registry as well.
oceanEdgeLength & submergedFaceArea also appear to be vestigial calculations from the old distribute... routine
```

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1253,63 +1266,65 @@ subroutine grounded_face_melt_ismip6(domain, err)
...
-            else 
-                groundedMarineMarginMask(iCell) = 0
-                faceMeltRateApplied(iCell) = 0.0_RKIND     
-                faceMeltSpeed(iCell) = 0.0_RKIND
-            endif           
+            endif
+
          enddo
 
+         call mpas_log_write("calling apply_front_ablation_velocity")
```
matthewhoffman, line 1311: Add something like "...from grounded_face_melt_ismip6" to this log message.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: I'm not entirely sure what you mean by this. This is referring to the VM stress threshold for floating ice, not the calculated VM stress. If config_floating_von_Mises_threshold_stress is set to 0.0, then we simply remove floating ice. Otherwise, we calculate a calving velocity. 
Or are you asking why there is ever floating ice when we use floatkill?

matthewhoffman (changes requested):
@trhille , this second review takes care of a much more careful review of the more significant revisions in the routines `grounded_face_melt_ismip6` and `apply_front_ablation_velocity`.

src/core_landice/mode_forward/mpas_li_calving.F
matthewhoffman: Oh my bad, I was reading it as the vM stress itself, but it's clearly the config threshold.  In that case just add a short comment that floating vM threshold=0 is handled below.

Also, at the top of this routine, add a check and error if config_grounded_von_Mises_threshold_stress==0.0.

trhille: Comment and check/error added.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: li_ prefix added everywhere

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Noted, but not changed.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: Added

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: This routine has been deleted.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1501,15 +1520,15 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
       call mpas_pool_get_array(geometryPool, 'calvingThickness', calvingThickness)
-      call mpas_pool_get_array(geometryPool, 'uncalvedVolumeNonDynCell', uncalvedVolumeNonDynCell)
-      call mpas_pool_get_array(geometryPool, 'uncalvedVolumeDynCell', uncalvedVolumeDynCell)
-      call mpas_pool_get_array(geometryPool, 'calvedVolumeNonDynCell', calvedVolumeNonDynCell)
-      call mpas_pool_get_array(geometryPool, 'calvedVolumeDynCell', calvedVolumeDynCell)
+      call mpas_pool_get_array(geometryPool, 'meltingThickness', meltingThickness)
+      call mpas_pool_get_array(geometryPool, 'unablatedVolumeNonDynCell', unablatedVolumeNonDynCell)
+      call mpas_pool_get_array(geometryPool, 'unablatedVolumeDynCell', unablatedVolumeDynCell)
       call mpas_pool_get_array(geometryPool, 'groundedMarineMarginMask', groundedMarineMarginMask)
       call mpas_pool_get_array(geometryPool, 'requiredCalvingVolumeNonDynEdge', requiredCalvingVolumeNonDynEdge)
```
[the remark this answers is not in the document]
trhille: Renamed

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1457,33 +1476,33 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
+      real (kind=RKIND), dimension(:), allocatable :: requiredAblationVolumeNonDynEdge, requiredAblationVolumeNonDynCell
+      real (kind=RKIND), dimension(:), allocatable :: requiredAblationVolumeDynEdge, requiredAblationVolumeDynCell
+      real (kind=RKIND), dimension(:), pointer :: unablatedVolumeNonDynCell, unablatedVolumeDynCell
       real (kind=RKIND), dimension(:), allocatable :: cellVolume
-      real(kind=RKIND) :: calvingSubtotal1, calvingSubtotal2, calvingSubtotal3
+      real(kind=RKIND) :: ablationSubtotal1, ablationSubtotal2, ablationSubtotal3
       real(kind=RKIND) :: thkSum
       integer :: thkCount
       real(kind=RKIND) :: removeVolumeHere
-      real(kind=RKIND) :: volumeAvailableToPass !< a temp. var. for accounting purposes that indicates how much volume
```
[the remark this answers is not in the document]
trhille: Added back

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1520,56 +1539,71 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
+            else
+               groundedMarineMarginMask(iCell) = 0
+            endif
+         enddo
+      endif
 
 
       ! Init fields for accounting
       calvingThickness(:) = 0.0_RKIND
-      calvedVolumeNonDynCell(:) = 0.0_RKIND
```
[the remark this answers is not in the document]
trhille: Added back

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1598,67 +1678,65 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
+                  requiredAblationVolumeNonDynEdge(iEdge) = ablationVelocity(iCell) * &
+                          edgeLengthScaling * dvEdge(iEdge) * thicknessForAblation(iCell) * deltat
+                  requiredAblationVolumeNonDynCell(iCell) = requiredAblationVolumeNonDynCell(iCell) + &
+                          requiredAblationVolumeNonDynEdge(iEdge) ! Keep running total
                endif
             enddo
-            ! c. Calculate calvingThickness here
-            removeVolumeHere = min(cellVolume(iCell), requiredCalvingVolumeNonDynCell(iCell))  ! Don't use more than available
-            calvingThickness(iCell) = removeVolumeHere / areaCell(iCell)
-            calvedVolumeNonDynCell(iCell) = removeVolumeHere
```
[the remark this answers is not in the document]
trhille: Added back

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Added back

src/core_landice/Registry.xml
```diff
@@ -1050,11 +1050,11 @@ is the value of that variable from the *previous* time level!
...
-                     description="volume of ice that was left uncalved from required calving flux at non-dynamic cells"
+                <var name="unablatedVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                     description="volume of ice that was left uncalved or unmelted from required calving or melting flux at non-dynamic cells"
                 />
-                <var name="uncalvedVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
-                     description="volume of ice that was left uncalved from required calving flux at dynamic cells"
+                <var name="unablatedVolumeDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
+                     description="volume of ice that was left uncalved or unmelted from required calving or melting flux at dynamic cells"
                 />
                 <var name="calvedVolumeNonDynCell" type="real" dimensions="nCells Time" units="m^3" time_levs="1"
```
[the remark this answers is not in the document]
trhille: Renamed

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1429,14 +1459,16 @@ end subroutine von_Mises_calving
 !> by the calling routine.
```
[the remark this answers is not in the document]
trhille: Information added to routine description.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1520,56 +1539,71 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
+      allocate(requiredAblationVolumeDynCell(nCells+1))
+      requiredAblationVolumeNonDynEdge(:) = 0.0_RKIND
+      requiredAblationVolumeNonDynCell(:) = 0.0_RKIND
+      requiredAblationVolumeDynEdge(:) = 0.0_RKIND
+      requiredAblationVolumeDynCell(:) = 0.0_RKIND
+      allocate(thicknessForAblation(nCells+1))
+      thicknessForAblation = 0.0_RKIND
+      allocate(ablationVelocity(nCells+1))
+
+      if ( applyCalving ) then
```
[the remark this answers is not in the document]
trhille: This has been updated. We now pass the desired calving or melting thickness and velocity arrays to li_apply_ablation_velocity, which no longer differentiates between calving and melting. We vertically average melting over the whole ice face before calling li_apply_ablation_velocity, since faceMeltSpeed is only below the waterline. Halo updates are now applied in the calling routines.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Halo updates now in calling routines.

src/core_landice/mode_forward/mpas_li_calving.F
matthewhoffman: Please add the TODO item mentioned above.  Also, looking at this again, I think this requires more careful treatment of the dynamic ice mask.  We don't want to leave a stranded band of nondynamic floating ice at the former calving front, but I guess we also don't want to delete nondynamic floating ice attached to grounded ice.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: Added

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
trhille: These have been removed, except for lowerSurface, which is used to calculate faceMeltSpeedVertAvg. dvEdge also removed.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1598,67 +1678,65 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
-      calvingSubtotal2 = 0.0_RKIND
+      ablationSubtotal2 = 0.0_RKIND
 
       ! a. Calculate calving on dynamic margin edges
       do iCell = 1, nCellsSolve
          if ( ( (li_mask_is_floating_ice(cellMask(iCell)) .and. li_mask_is_dynamic_ice(cellMask(iCell)) ) & ! floating dynamic
-                .or.  (groundedMarineMarginMask(iCell) == 1) ) &
-              .and. li_mask_is_margin(cellMask(iCell)) ) then
-            ! only consider dynamic margins adjacent to open ocean (need to exclude dynamic margins adjacent to land)
+              .and. li_mask_is_margin(cellMask(iCell)) ) .or.  (groundedMarineMarginMask(iCell) == 1) ) then
```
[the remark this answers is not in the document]
trhille: Yes, this change is necessary because groundedMarineMarginMask will not always correspond to li_mask_is_margin when we use applyToGroundingLine.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1598,67 +1691,66 @@ subroutine apply_calving_velocity(meshPool, geometryPool, velocityPool, domain,
...
+                  requiredAblationVolumeNonDynCell(iCell) = requiredAblationVolumeNonDynCell(iCell) + &
+                          requiredAblationVolumeNonDynEdge(iEdge) ! Keep running total
                endif
             enddo
-            ! c. Calculate calvingThickness here
-            removeVolumeHere = min(cellVolume(iCell), requiredCalvingVolumeNonDynCell(iCell))  ! Don't use more than available
-            calvingThickness(iCell) = removeVolumeHere / areaCell(iCell)
-            calvedVolumeNonDynCell(iCell) = removeVolumeHere
-            uncalvedVolumeNonDynCell(iCell) = requiredCalvingVolumeNonDynCell(iCell) - removeVolumeHere
+            ! c. Apply calving here
```
matthewhoffman, line 1700: Return this comment to "c. Calculate ablationThickness here".  The distinction being we don't actually apply calving until the end of this routine when thickness is modified.

src/core_landice/mode_forward/mpas_li_calving.F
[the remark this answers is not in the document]
matthewhoffman: Let's remove this change for now, and revisit the issue when we implement the 3 masks.  That will allow the changes up to this point to not change answers.

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Done

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1660,30 +1660,66 @@ subroutine li_apply_front_ablation_velocity(meshPool, geometryPool, velocityPool
...
                jCell = cellsOnCell(iNeighbor, iCell)
                if (li_mask_is_dynamic_ice(cellMask(jCell))) then
                   thkSum = thkSum + thickness(jCell)
+                  ablationVelSum = ablationVelSum + ablationVelocity(jCell)
                   thkCount = thkCount + 1
                endif
             enddo
             if (thkCount == 0) then
-               !call mpas_log_write("Found a stranded non-dynamic floating cell: cell $i with thickness=$r m.", MPAS_LOG_WARN, &
+                !call mpas_log_write("Found a stranded non-dynamic floating cell: cell $i with thickness=$r m.", MPAS_LOG_WARN, &
```
matthewhoffman, line 1714: picky comment: undo the indentation change here and in the `else` clause immediately following.

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -1183,7 +1184,7 @@ subroutine li_grounded_face_melt_ismip6(domain, err)
       !-----------------------------------------------------------------
       integer :: iCell, iNeighbor, iEdge, nEmptyNeighbors
       real (kind=RKIND), pointer :: rhoi
-      integer, pointer :: nCellsSolve
+      integer, pointer :: nCells, nCellsSolve
```
matthewhoffman, line 1187: Should nCellsSolve (and its 'get' routine below) be entirely deleted?  (if it's not used elsewhere)

trhille: It's used for the loop in which faceMeltSpeed is calculated. I don't know whether it's more appropriate to loop over nCells or nCellsSolve there.

src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1613,13 +1613,21 @@ subroutine li_apply_front_ablation_velocity(meshPool, geometryPool, velocityPool
                .and. bedTopography(iCell) < config_sea_level &
                .and. li_mask_is_dynamic_ice(cellMask(iCell)) ) then
 
-               groundedMarineMarginMask(iCell) = 1
-            else
-               groundedMarineMarginMask(iCell) = 0
+               frontAblationMask(iCell) = 1
+               ! Include adjacent non-dynamic cells
```
matthewhoffman, line 1617: Can you explain why adjacent non-dynamic cells need to be included here and in the applyToFloating logic?  Also, if they do need to be included, do we need multiple passes of finding them to include those 'horn' locations of non-dynamic cells that don't have dynamic neighbors?

src/core_landice/mode_forward/mpas_li_calving.F
trhille: Fixed

src/core_landice/mode_forward/mpas_li_iceshelf_melt.F
```diff
@@ -60,8 +60,65 @@ module li_iceshelf_melt
...
+      err = 0
+      err_tmp = 0
+
+      call mpas_pool_get_config(liConfigs, 'config_front_mass_bal_grounded', config_front_mass_bal_grounded)
+
+      if ( trim(config_front_mass_bal_grounded) == 'ismip6' &
+          .or. trim(config_front_mass_bal_grounded) == 'uniform' ) then
+         call grounded_face_melt_ismip6(domain, err_tmp)
+         err = ior(err, err_tmp)
+      endif
```
matthewhoffman, line 116: Add an 'elseif' for the 'none' case and an 'else' with an error for all over strings.  This is useful for catching typos in the namelist.

trhille: This has been updated.

matthewhoffman:
Testing on Humboldt domain using a specified calving velocity of 0.5 km/yr run for 50 years.  Target retreat is 25 km along a flowline.  This is with fixed velocity field and no advection. 
image

PR #818, opened 2021-02-23
Bug fixes to enable aerosols & dust in multi-layer snow with Snicar-ad

njeffery:
This PR allows for radiative transfer calculations of aerosols and dust in seaice.  The capability will currently remain off as default.
1. Corrects the snow surface scattering layer and bgc snow grid for aerosols with multiple snow layers
2. adds a minimum snow thickness for direct deposition of aerosols in snow.
3. adds a length scale for penetration of surface atm flux into the snow interior for thin snow layers.
4. Alters the aerosol testing suite configuration to test multiple snow layers and snicar-ad
5. defines diagnostic verticalAerosolSnowCell
6. corrects unit definitions for aerosol fields in Registry


njeffery:
This code is BFB in the WC configuration.  All five testsuites performed and passed except the standard_bgc and the aerosol_shortwave regression tests.  This is expected because the PR has changes to the aerosol transport and snow grid are aerosols are enabled in the standard_bgc test.

proteanplanet:
@njeffery "This code is BFB in the WC configuration":  Does this mean you've run a B-case with this?  Was it BFB for a full annual cycle?

njeffery:
Hi Andrew,


The code changes are BFB in WC because the aerosols are off by default and there's no plans to use them for this phase.  We're not sure yet if we want them on for BGC either.   That would be a separate PR.


  I have done a couple of 38-year fully coupled runs to test the feature.  I added some results to the feature page: https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1907818531/B12+Activate+Aerosol+Shortwave+Feedbacks+in+Sea+Ice


  The code base was a pre-v2 configuration from November that Jon had been testing.   I'll do another simulation once we are closer to having a CBGCv2 code-base.  then we can discuss whether it makes sense to activate.


Thanks for looking at this.

Nicole



proteanplanet (approved):
Approved by visual inspection and evidence of testing.

PR #819, opened 2021-02-25
Fixes missing edgeSignOnCell in redi_term2

vanroekel:
Currently `redi_term2` is missing a factor of edgeSignOnCell in its calculation.  It is being correctly applied in the limiter, but not the original calculation.


vanroekel:
This bug was first noted by @ambrad.

vanroekel:
@jonbob and @mark-petersen this is a new bug brought to my attention.  It should go in as soon as we can get it in, but it is not the PR I mentioned in the email to you both.

PR #822, opened 2021-03-03
Simplifies the shortwave fixer formulation

vanroekel:
makes the shortwave fixer add energy at bottom of column instead of
spreading throughout.  This is more physically correct

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
+++ b/src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
@@ -130,7 +130,7 @@ contains
 
       integer, dimension(:), pointer :: maxLevelCell
 
-      real (kind=RKIND) :: depth, fluxRemaining, layerFraction
+      real (kind=RKIND) :: depth, fluxRemaining
       real (kind=RKIND), dimension(:), pointer :: refBottomDepth
       real (kind=RKIND), dimension(:), allocatable :: weights
 
@@ -169,6 +169,8 @@ contains
          k = maxLevelCell(iCell)
          if(k > 0 .and. fluxRemaining > 0.0_RKIND) then
             bottomLayerShortwaveTemperatureFlux(iCell) = penetrativeTemperatureFlux(iCell) * fluxRemaining
+         else
+            bottomLayerShortwaveTemperatureFlux(iCell) = 0.0_RKIND
          end if
 
          depth = 0.0_RKIND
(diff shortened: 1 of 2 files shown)


vanroekel:
Initial testing of this implementation seemed to suggest strong sensitivity but this was due to a faulty configuration of the atmosphere diagnostics.  Since this implementation is more physical this PR reimplements the bottom addition as an option.

vanroekel:
@mark-petersen @maltrud could you take a look at this as soon as you can?  We'd like to push this for WC v2

src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
```diff
@@ -193,13 +195,12 @@ subroutine ocn_tracer_short_wave_absorption_jerlov_tend(meshPool, forcingPool, i
...
+         !$omp do schedule(runtime) private(k)
 #endif
          do iCell = 1, nCells
-            layerFraction = bottomlayerShortwaveTemperatureFlux(iCell) / (maxLevelCell(iCell) + 1.0E-15_RKIND)
-            do k=1,maxLevelCell(iCell)
-               tend(index_temperature, k, iCell) = tend(index_temperature, k, iCell) + layerFraction
-            end do
+            k = maxLevelCell(iCell)
+            tend(index_temperature, k, iCell) = tend(index_temperature, k, iCell) + &
+                      bottomLayerShortwaveTemperatureFlux(iCell)
```
mark-petersen, line 203: Is it possible that this line could heat a bottom layer at 5000m? Are we sure we won't heat the deep ocean? This could be tested by writing out an instantaneous `bottomLayerShortwaveTemperatureFlux`, make a log color bar, and make sure it is non-zero only at shallow depths - I would expect 100m or less.

mark-petersen:
passes nightly regression with gnu 5.3 on badger

src/core_ocean/shared/mpas_ocn_tracer_short_wave_absorption_jerlov.F
vanroekel: I have indeed done this, there is no heating below 200 m.  Further down in the file (get_weights function) there is a cap at 200 m that sets weight to zero at 200m capping the effect of this change.  I can also post a figure too if you’d like to see that analysis

vanroekel:
@mark-petersen here is a plot of where the solar fixer is greater than 1e-14.  It is definitely only targeting the shallowest depths.

mark-petersen (approved):
This all looks good. Thanks for the plots.

maltrud (approved):

PR #823, opened 2021-03-05
Ignore the water thickness head in the channel model

matthewhoffman:
For calculations of hydropotential gradient in the channel model, this
commit swtiches from using the full hydropotential to hydropotentialBase
(which ignores the hydropotential contribution from the thickness of the
water).

This seems to prevent channel blowups around lakes.  It also might not
be modeling channel interactiosn with lakes properly.  This requires
further investigation.


mgduda:

PR #827, opened 2021-03-13
Remove unused sea ice modules

mark-petersen:
It turns out we had a sea ice module in the ocean core that was completely unused. I think this was added for early testing in 2013 and then transferred to sea ice or E3SM.


mark-petersen:
Passes nightly regression suite with gnu debug.

eclare108213:
Just a suggestion: code coverage tools identify which lines of code are never hit in a suite of tests, and are quite useful for cleaning up things like this as well as improving the test suites to exercise more of the code.  The CICE Consortium uses lcov, whose user interface is a little clunky but it works well for complex codes and big test suites.

proteanplanet (approved):
This PR now leaves only frazil production in MPAS-Ocean within the module ocn_frazil_forcing, which is correct.

PR #828, opened 2021-03-15
Landice/ismip6 calving melting phase3

trhille:
This fixes the treatment of calving at grounded marine margins that have a floating fringe on non-dynamic cells. The issue arises from advection being implemented before calving, which can cause a double layer of floating non-dynamic cells. This led to grounded margins retreating significantly too slowly (by around ~25% in our tests). The new treatment is to use the mean thickness, advective velocity, and ablation velocity from two cells upstream, to avoid using potentially bad (i.e., low) ice thicknesses in the last grounded cell. The cells used are defined as the dynamic neighbors of the ice-containing neighbors of the non-dynamic cell if that non-dynamic cell has at least one neighbor that is grounded or the grounding line.

### DIFF
--- a/src/core_landice/mode_forward/mpas_li_calving.F
+++ b/src/core_landice/mode_forward/mpas_li_calving.F
@@ -1763,7 +1763,7 @@ module li_calving
                nOneCellBack = 0
                do iNeighbor = 1, nEdgesOnCell(iCell)
                   jCell = cellsOnCell(iNeighbor, iCell)
-                  if ( li_mask_is_dynamic_ice(cellMask(jCell)) ) then
+                  if ( li_mask_is_ice(cellMask(jCell)) ) then
                         nOneCellBack = nOneCellBack + 1
                         oneCellBackList(nOneCellBack) = jCell
                   endif
--- a/src/core_landice/mode_forward/mpas_li_calving.F
+++ b/src/core_landice/mode_forward/mpas_li_calving.F
@@ -1530,9 +1530,9 @@ module li_calving
       !-----------------------------------------------------------------
       integer, intent(out) :: err !< Output: error flag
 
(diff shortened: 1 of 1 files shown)


src/core_landice/mode_forward/mpas_li_calving.F
```diff
@@ -1733,18 +1738,71 @@ subroutine li_apply_front_ablation_velocity(meshPool, geometryPool, velocityPool
...
                   uvelSum = uvelSum + uvelForAblation(jCell)
                   vvelSum = vvelSum + vvelforAblation(jCell)
                   ablationVelSum = ablationVelSum + ablationVelocity(jCell)
                   thkCount = thkCount + 1
+               else if ( li_mask_is_grounding_line(cellMask(jCell)) .or. li_mask_is_grounded_ice(cellMask(jCell)) ) then
+                  nGroundedNeighbors = nGroundedNeighbors + 1
                endif
             enddo
+
+            ! if floating non-dynamic cell has any grounded neighbors, use ice
```
matthewhoffman, line 1755: The current logic ignores the case where the cell being evaluated has a mix of floating dynamic neighbors and grounded neighbors.  In that case, the code will ignore the floating dynamic neighbor and only evaluate the grounded neighbors two-back.  Is that what we want?  What I'm describing seems like an edge case.  It may occur at the lateral margin of an ice shelf in an embayment.

trhille: The current logic does not in fact ignore that case. That section is called if the non-dynamic cell has one grounded (and/or grounding line) neighbor, but then there is no logic about grounded ice in the actual implementation.
oneCellBackList is defined as any cell with ice, then twoCellsBackList is defined as any dynamic cell that is not is oneCellBackLIst (or already in twoCellsBackList).

PR #835, opened 2021-03-31
Fix subtle problems in hydro timestepper 

matthewhoffman:
This merge cleans up some obscure issues in the hydrology model's adaptive timestepper.  It ignores edges on the grounding line in the diffusive CFL check, because the diffusive flux is not actually applied on those edges and they can otherwise be overly restrictive.  It also makes the model die if the hydrology dt becomes less than 1 second.  This should avoid the model hanging when unstable cavity growth occurs.


mgduda:

PR #837, opened 2021-03-31
 Add freshwater temperature flux field to pass to the atm

jonbob:
Adds the calculation of a freshwater temperature flux, with the intention of passing it through the coupler to the atm. This is to help track some terms missing from the heat budget, due to water coming in to the ocean with a specific temperature that the rest of the system does not know about.

[non-BFB] due to some order of operations changes

### DIFF
--- a/src/core_ocean/Registry.xml
+++ b/src/core_ocean/Registry.xml
@@ -3091,6 +3091,22 @@
 			 description="Short wave flux at cell centers from coupler. Positive into the ocean."
 			 packages="activeTracersBulkRestoringPKG;ecosysTracersPKG"
 		/>
+		<var name="rainTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with rain at cell centers sent to coupler. Positive into the ocean."
+		/>
+		<var name="evapTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with Evaporation at cell centers sent to coupler. Positive into the ocean."
+		/>
+		<var name="seaIceFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with sea ice melt water at cell centers sent to coupler. Positive into the ocean."
+		/>
+		<var name="icebergFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with iceberg melt at cell centers sent to coupler. Positive into the ocean."
+		/>
+
+		<var name="totalFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Sum of heat fluxes associated with water fluxes cell centers sent to coupler. Positive into the ocean."
+		/>
 
 
 		<!-- Coupling fields associated with mass or salinity fluxes -->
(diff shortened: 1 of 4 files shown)


xylar:
@jonbob, I think the destination here needs to be `ocean/develop`, not `e3sm/develop`.  You can just edit the PR to change that, no need to open a new one.

xylar:
@maltrud, could you talk a little bit about why it makes sense to have a temperature flux as opposed to a heat flux?  Presumably the ocean temperature flux isn't useful to any other components without first being converted to a heat flux, right?

jonbob:
Thanks @xylar -- I haven't made an MPAS PR in a long time, and it shows!

jonbob:
@xylar - I added you as a reviewer, if only because it looks like you're already doing that

jonbob:
@maltrud, @mark-petersen - can you please look at this today? The WaterCycle group would like to test this in its next set of runs

mark-petersen:
Passes visual inspection and nightly regression suite, gnu and intel, debug for both. 

I tried to put the new arrays in a package, but they are used in KPP lines where it's better not to add separate cases, so I left them outside of packages. They are 1D arrays and will only add unneeded memory outside of E3SM.

mark-petersen (approved):

src/core_ocean/Registry.xml
```diff
@@ -3091,6 +3091,22 @@
...
+			 description="Heat flux associated with Evaporation at cell centers sent to coupler. Positive into the ocean."
+		/>
+		<var name="seaIceFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with sea ice melt water at cell centers sent to coupler. Positive into the ocean."
+		/>
+		<var name="icebergFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with iceberg melt at cell centers sent to coupler. Positive into the ocean."
+		/>
+
+		<var name="totalFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
```
mark-petersen, line 3107: @maltrud and @vanroekel it's unusual to have a temperature flux in degC m/s rather than a heat flux in W/m^2. I'm guessing these are the units received from the coupler, so we are conforming to units from the other components?

xylar: Yes, essentially the same as my question above.

jonbob: In the driver, this field gets multiplied by rho_sw*cp_sw before being sent to the cpl -- so it does end up in W/m^2 there.

maltrud: my rationale is that the variables are called "*TemperatureFlux" so should have units of degC m/s.  it seems kind of silly to calculate these as a heat flux by multiplying by rho*cp, then dividing by it again before converting to a tendency.

maltrud: don't know why git decided to italicize part of my comment....

maltrud (approved):

src/core_ocean/Registry.xml
xylar: @maltrud, okay, I guess that makes sense.

Your comment got italicized because you have two `*` symbols in there and you didn't use back-ticks to make them code.

xylar:
I took myself off as a reviewer because I don't have the time or expertise to test properly.

PR #825, opened 2021-03-10
Add the option for inactive top cells

cbegeman:
Inactive top cells are introduced to add flexibility to the vertical coordinate below ice shelves. 

This PR introduces phase 1 in the development documented in https://github.com/MPAS-Dev/MPAS-Model/pull/811, which touches only the code needed to run a basic ice shelf test case.

The variable `minLevelCell` and related `minLevelEdge*` and `minLevelVertex*` variables are introduced by analogy with `maxLevel*` variables. By default, there are no inactive top cells and `minLevel*` = 1. The main code changes are to k-loop limits and designating surface quantities as those at `minLevel*`.

PR #840, opened 2021-04-04
Enhance inactive top cells feature

cbegeman:
Here, we extend the capability for Inactive top cells introduced by https://github.com/MPAS-Dev/MPAS-Model/pull/825 to apply to additional model configurations besides those needed to run a basic ice shelf test case. More detail about the scope of this development is documented in https://github.com/MPAS-Dev/MPAS-Model/pull/811.

The variable `minLevelCell` and related `minLevelEdge*` and `minLevelVertex*` variables are added by analogy with `maxLevel*` variables. By default, there are no inactive top cells and `minLevel*` = 1. The main code changes are to k-loop limits and designating surface quantities as those at `minLevel*`.

### DIFF
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -242,7 +242,7 @@ module ocn_time_integration_rk4
          !$omp parallel
          !$omp do schedule(runtime) private(k)
          do iEdge = 1, nEdges
-            do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)
+            do k = 1, maxLevelEdgeTop(iEdge)
                normalVelocityNew(k, iEdge) = normalVelocityCur(k, iEdge)
             end do
          end do
@@ -250,7 +250,7 @@ module ocn_time_integration_rk4
 
          !$omp do schedule(runtime) private(k)
          do iCell = 1, nCells
-            do k = minLevelCell(iCell), maxLevelCell(iCell)
+            do k = 1, maxLevelCell(iCell)
                layerThicknessNew(k, iCell) = layerThicknessCur(k, iCell)
             end do
          end do
@@ -269,7 +269,7 @@ module ocn_time_integration_rk4
                   !$omp parallel
                   !$omp do schedule(runtime) private(k)
                   do iCell = 1, nCells  ! couple tracers to thickness
-                     do k = minLevelCell(iCell), maxLevelCell(iCell)
+                     do k = 1, maxLevelCell(iCell)
                         tracersNew(:, k, iCell) = tracersCur(:, k, iCell) * layerThicknessCur(k, iCell)
                      end do
                   end do
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -141,7 +141,7 @@ module ocn_time_integration_rk4
       ! Diagnostics Indices
 
       ! Mesh array pointers
-      integer, dimension(:), pointer :: minLevelCell, maxLevelCell, minLevelEdgeBot, maxLevelEdgeTop
+      integer, dimension(:), pointer :: minLevelCell, maxLevelCell, maxLevelEdgeTop
       real (kind=RKIND), dimension(:), pointer :: bottomDepth
 
       ! Provis Array Pointers
@@ -234,10 +234,11 @@ module ocn_time_integration_rk4
          call mpas_pool_get_array(statePool, 'lowFreqDivergence', lowFreqDivergenceCur, 1)
          call mpas_pool_get_array(statePool, 'lowFreqDivergence', lowFreqDivergenceNew, 2)
 
-         call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
          call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
-         call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
          call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
+
+         ! Lower k-loop limit of 1 rather than minLevel* needed in *New = *Cur
+         ! assignments below are needed to maintain bit-for-bit results
 
          !$omp parallel
          !$omp do schedule(runtime) private(k)
@@ -622,6 +623,8 @@ module ocn_time_integration_rk4
          call mpas_pool_get_dimension(meshPool, 'nCells', nCells)
          call mpas_pool_get_dimension(meshPool, 'nEdges', nEdges)
          call mpas_pool_get_array(meshPool, 'bottomDepth', bottomDepth)
+         call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
+         call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
 
          call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocityCur, 1)
          call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocityNew, 2)
@@ -721,8 +724,8 @@ module ocn_time_integration_rk4
          !$omp parallel
          !$omp do schedule(runtime) 
          do iCell = 1, nCells
-            surfaceVelocity(indexSurfaceVelocityZonal, iCell) = velocityZonal(minLevelEdgeBot(iEdge), iCell)
-            surfaceVelocity(indexSurfaceVelocityMeridional, iCell) = velocityMeridional(minLevelEdgeBot(iEdge), iCell)
+            surfaceVelocity(indexSurfaceVelocityZonal, iCell) = velocityZonal(minLevelCell(iCell), iCell)
+            surfaceVelocity(indexSurfaceVelocityMeridional, iCell) = velocityMeridional(minLevelCell(iCell), iCell)
 
             SSHGradient(indexSSHGradientZonal, iCell) = gradSSHZonal(iCell)
             SSHGradient(indexSSHGradientMeridional, iCell) = gradSSHMeridional(iCell)
@@ -1033,7 +1036,7 @@ module ocn_time_integration_rk4
 
       real (kind=RKIND), dimension(:, :, :), pointer :: tracersGroupCur, tracersGroupProvis, tracersGroupTend
 
-      integer, dimension(:), pointer :: minLevelCell, maxLevelCell, minLevelEdgeBot, maxLevelEdgeTop
+      integer, dimension(:), pointer :: minLevelCell, maxLevelCell, maxLevelEdgeTop
 
       logical, pointer :: config_use_tracerGroup
       type (mpas_pool_iterator_type) :: groupItr
@@ -1075,7 +1078,6 @@ module ocn_time_integration_rk4
 
       call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
-      call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
       call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
 
       !$omp parallel
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -1081,7 +1081,7 @@ module ocn_time_integration_rk4
       !$omp parallel
       !$omp do schedule(runtime) private(k)
       do iEdge = 1, nEdges
-         do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)
+         do k = 1, maxLevelEdgeTop(iEdge)
             normalVelocityProvis(k, iEdge) = normalVelocityCur(k, iEdge) + rkSubstepWeight &
                                            * normalVelocityTend(k, iEdge)
             normalVelocityProvis(k, iEdge) = normalVelocityProvis(k, iEdge) * (1.0_RKIND - wettingVelocity(k, iEdge))
@@ -1091,7 +1091,7 @@ module ocn_time_integration_rk4
 
       !$omp do schedule(runtime) private(k)
       do iCell = 1, nCells
-         do k = minLevelCell(iCell), maxLevelCell(iCell)
+         do k = 1, maxLevelCell(iCell)
             layerThicknessProvis(k, iCell) = layerThicknessCur(k, iCell) + rkSubstepWeight &
                                            * layerThicknessTend(k, iCell)
          end do
@@ -1115,7 +1115,7 @@ module ocn_time_integration_rk4
                   !$omp parallel
                   !$omp do schedule(runtime) private(k)
                   do iCell = 1, nCells
-                     do k = minLevelCell(iCell), maxLevelCell(iCell)
+                     do k = 1, maxLevelCell(iCell)
                         tracersGroupProvis(:, k, iCell) = ( layerThicknessCur(k, iCell) * tracersGroupCur(:, k, iCell)  &
                                                  + rkSubstepWeight * tracersGroupTend(:, k, iCell) &
                                                    ) / layerThicknessProvis(k, iCell)
--- a/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -141,7 +141,7 @@ module ocn_time_integration_rk4
       ! Diagnostics Indices
 
       ! Mesh array pointers
-      integer, dimension(:), pointer :: maxLevelCell, maxLevelEdgeTop
+      integer, dimension(:), pointer :: minLevelCell, maxLevelCell, minLevelEdgeBot, maxLevelEdgeTop
       real (kind=RKIND), dimension(:), pointer :: bottomDepth
 
       ! Provis Array Pointers
@@ -234,13 +234,15 @@ module ocn_time_integration_rk4
          call mpas_pool_get_array(statePool, 'lowFreqDivergence', lowFreqDivergenceCur, 1)
          call mpas_pool_get_array(statePool, 'lowFreqDivergence', lowFreqDivergenceNew, 2)
 
+         call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
          call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
+         call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
          call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
 
          !$omp parallel
          !$omp do schedule(runtime) private(k)
          do iEdge = 1, nEdges
-            do k = 1, maxLevelEdgeTop(iEdge)
+            do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)
                normalVelocityNew(k, iEdge) = normalVelocityCur(k, iEdge)
             end do
          end do
@@ -248,7 +250,7 @@ module ocn_time_integration_rk4
 
          !$omp do schedule(runtime) private(k)
          do iCell = 1, nCells
-            do k = 1, maxLevelCell(iCell)
+            do k = minLevelCell(iCell), maxLevelCell(iCell)
                layerThicknessNew(k, iCell) = layerThicknessCur(k, iCell)
             end do
          end do
@@ -267,7 +269,7 @@ module ocn_time_integration_rk4
                   !$omp parallel
                   !$omp do schedule(runtime) private(k)
                   do iCell = 1, nCells  ! couple tracers to thickness
-                     do k = 1, maxLevelCell(iCell)
+                     do k = minLevelCell(iCell), maxLevelCell(iCell)
                         tracersNew(:, k, iCell) = tracersCur(:, k, iCell) * layerThicknessCur(k, iCell)
                      end do
                   end do
@@ -658,12 +660,12 @@ module ocn_time_integration_rk4
              if (tidalInputMask(iCell) == 1.0_RKIND) then
                ! compute total depth for relative thickness contribution
                totalDepth = 0.0_RKIND
-               do k = 1, maxLevelCell(iCell)
+               do k = minLevelCell(iCell), maxLevelCell(iCell)
                  totalDepth = totalDepth + restingThickness(k,iCell)
                end do
 
                ! only modify layer thicknesses on tidal boundary
-               do k = 1, maxLevelCell(iCell)
+               do k = minLevelCell(iCell), maxLevelCell(iCell)
                  layerThicknessNew(k, iCell) = tidalInputMask(iCell)*(tidalBCValue(iCell) + bottomDepth(iCell))*(restingThickness(k,iCell)/totalDepth)
                  !(1.0_RKIND - tidalInputMask(iCell))*layerThicknessNew(k, iCell)  ! generalized tappered assumption code
                end do
@@ -719,8 +721,8 @@ module ocn_time_integration_rk4
          !$omp parallel
          !$omp do schedule(runtime) 
          do iCell = 1, nCells
-            surfaceVelocity(indexSurfaceVelocityZonal, iCell) = velocityZonal(1, iCell)
-            surfaceVelocity(indexSurfaceVelocityMeridional, iCell) = velocityMeridional(1, iCell)
+            surfaceVelocity(indexSurfaceVelocityZonal, iCell) = velocityZonal(minLevelEdgeBot(iEdge), iCell)
+            surfaceVelocity(indexSurfaceVelocityMeridional, iCell) = velocityMeridional(minLevelEdgeBot(iEdge), iCell)
 
             SSHGradient(indexSSHGradientZonal, iCell) = gradSSHZonal(iCell)
             SSHGradient(indexSSHGradientMeridional, iCell) = gradSSHMeridional(iCell)
@@ -1031,7 +1033,7 @@ module ocn_time_integration_rk4
 
       real (kind=RKIND), dimension(:, :, :), pointer :: tracersGroupCur, tracersGroupProvis, tracersGroupTend
 
-      integer, dimension(:), pointer :: maxLevelCell, maxLevelEdgeTop
+      integer, dimension(:), pointer :: minLevelCell, maxLevelCell, minLevelEdgeBot, maxLevelEdgeTop
 
       logical, pointer :: config_use_tracerGroup
       type (mpas_pool_iterator_type) :: groupItr
@@ -1071,13 +1073,15 @@ module ocn_time_integration_rk4
       call mpas_pool_get_array(tendPool, 'layerThickness', layerThicknessTend)
       call mpas_pool_get_array(tendPool, 'lowFreqDivergence', lowFreqDivergenceTend)
 
+      call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
+      call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
       call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
 
       !$omp parallel
       !$omp do schedule(runtime) private(k)
       do iEdge = 1, nEdges
-         do k = 1, maxLevelEdgeTop(iEdge)
+         do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)
             normalVelocityProvis(k, iEdge) = normalVelocityCur(k, iEdge) + rkSubstepWeight &
                                            * normalVelocityTend(k, iEdge)
             normalVelocityProvis(k, iEdge) = normalVelocityProvis(k, iEdge) * (1.0_RKIND - wettingVelocity(k, iEdge))
@@ -1087,7 +1091,7 @@ module ocn_time_integration_rk4
 
       !$omp do schedule(runtime) private(k)
       do iCell = 1, nCells
-         do k = 1, maxLevelCell(iCell)
+         do k = minLevelCell(iCell), maxLevelCell(iCell)
             layerThicknessProvis(k, iCell) = layerThicknessCur(k, iCell) + rkSubstepWeight &
                                            * layerThicknessTend(k, iCell)
          end do
@@ -1111,7 +1115,7 @@ module ocn_time_integration_rk4
                   !$omp parallel
                   !$omp do schedule(runtime) private(k)
                   do iCell = 1, nCells
-                     do k = 1, maxLevelCell(iCell)
+                     do k = minLevelCell(iCell), maxLevelCell(iCell)
                         tracersGroupProvis(:, k, iCell) = ( layerThicknessCur(k, iCell) * tracersGroupCur(:, k, iCell)  &
                                                  + rkSubstepWeight * tracersGroupTend(:, k, iCell) &
                                                    ) / layerThicknessProvis(k, iCell)
@@ -1208,7 +1212,7 @@ module ocn_time_integration_rk4
 
       real (kind=RKIND), dimension(:, :, :), pointer :: tracersGroupNew, tracersGroupTend
 
-      integer, dimension(:), pointer :: maxLevelCell
+      integer, dimension(:), pointer :: minLevelCell, maxLevelCell
 
       logical, pointer :: config_use_tracerGroup
       type (mpas_pool_iterator_type) :: groupItr
@@ -1243,12 +1247,13 @@ module ocn_time_integration_rk4
       call mpas_pool_get_array(tendPool, 'highFreqThickness', highFreqThicknessTend)
       call mpas_pool_get_array(tendPool, 'lowFreqDivergence', lowFreqDivergenceTend)
 
+      call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
 
       !$omp parallel
       !$omp do schedule(runtime) private(k)
       do iCell = 1, nCells
-         do k = 1, maxLevelCell(iCell)
+         do k = minLevelCell(iCell), maxLevelCell(iCell)
             layerThicknessNew(k, iCell) = layerThicknessNew(k, iCell) + rkWeight * layerThicknessTend(k, iCell)
          end do
       end do
@@ -1277,7 +1282,7 @@ module ocn_time_integration_rk4
                   !$omp parallel
                   !$omp do schedule(runtime) private(k)
                   do iCell = 1, nCells
-                     do k = 1, maxLevelCell(iCell)
+                     do k = minLevelCell(iCell), maxLevelCell(iCell)
                         tracersGroupNew(:, k, iCell) = tracersGroupNew(:, k, iCell) + rkWeight &
                                                 * tracersGroupTend(:, k, iCell)
                      end do
@@ -1326,7 +1331,7 @@ module ocn_time_integration_rk4
       real (kind=RKIND), dimension(:, :), pointer :: layerThicknessNew, normalVelocityNew
       real (kind=RKIND), dimension(:, :, :), pointer :: tracersGroupNew
 
-      integer, dimension(:), pointer :: maxLevelCell
+      integer, dimension(:), pointer :: minLevelCell, maxLevelCell
 
       logical, pointer :: config_use_tracerGroup
       type (mpas_pool_iterator_type) :: groupItr
@@ -1355,6 +1360,7 @@ module ocn_time_integration_rk4
       call mpas_pool_get_dimension(tracersPool, 'index_temperature', indexTemperature)
       call mpas_pool_get_dimension(tracersPool, 'index_salinity', indexSalinity)
 
+      call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
 
       call mpas_pool_begin_iteration(tracersPool)
@@ -1365,7 +1371,7 @@ module ocn_time_integration_rk4
                !$omp parallel
                !$omp do schedule(runtime) private(k)
                do iCell = 1, nCells
-                 do k = 1, maxLevelCell(iCell)
+                 do k = minLevelCell(iCell), maxLevelCell(iCell)
                    tracersGroupNew(:, k, iCell) = tracersGroupNew(:, k, iCell) / layerThicknessNew(k, iCell)
                  end do
                end do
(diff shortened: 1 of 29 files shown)


cbegeman:
@vanroekel Do you know of a test case I can use to test the changes I made to `ocn_tracer_short_wave_absorption_variable_tend` with `config_sw_absorption_type='ohlmann0'`? It seems that this option requires forcing files. If not, do you know who would know about this?

cbegeman:
@caozd999 With the changes introduced by this PR, the `drying_slope/zstar_variableManningn/1km` test case is not bit-for-bit with `ocean/develop`. However, the changes to the `ocn_vel_vmix_tend_spatially_variable_manning` seem to be the same as `ocn_vel_vmix_tend_spatially_variable` and `drying_slope/zstar_variableCd/1km` is bit-for-bit. 

Do you have time to take a look at [this routine](https://github.com/MPAS-Dev/MPAS-Model/pull/840/files#diff-826faacb65f63d6ede35709da4eee7f8c6b5990ea48c0e3bd905f770fa3b052e) and let me know if anything jumps out at you? You could checkout my branch and run these tests for yourself, or you can look at my output here: `/lustre/scratch3/turquoise/cbegeman/MPAS-test-cases/minlevelcell-vmix-opts/ocean/drying_slope/zstar_variableManningn/1km`. Maximum `normalVelocity` differences between `ocean/develop` and this branch are O(1e-7) and maximum `ssh` differences are O(1e-8). If you don't see any code issues then the next question would be whether you prefer that these changes are included or excluded. Thanks a bunch!

src/core_ocean/shared/mpas_ocn_tracer_ecosys.F
```diff
@@ -1188,6 +1190,7 @@ subroutine ocn_tracer_ecosys_surface_flux_compute(activeTracers, ecosysTracers,
          BGC_forcing%SST(column) = activeTracers(indexTemperature,iLevelSurface,iCell)
          BGC_forcing%SSS(column) = activeTracers(indexSalinity,iLevelSurface,iCell)
 
+         ! the 1 indices may need to be changed to iLevelSurface
```
cbegeman, line 1193: @maltrud Do you know whether the 1 index in `ecosysTracers(ecosysIndices%*,1,iCell)` should refer to the surface index (i.e., `minLevelCell`)?

vanroekel:
@cbegeman the shortwave absorption variable code is never tested (not since I. added it a number of years ago).  I'd like to switch it on but it hasn't been high priority.  For testing we would have to rerun the init phase of the COMPASS global ocean test case for each resolution and move them to the input data repo.  Then we would have to do the testing by hand.  I don't think that should be part of your responsibility in this PR.   I'd propose a few possibilities
```
1. remove the minlevelcell changes from that file for now and revisit if we turn that functionality on later.
2. I could review the changes by visual inspection and approve that way.  I imagine that the changes in that module will be straight forward
3. We develop a test for this PR, I could help with this
```
I think my choice would be (2).  What do you think?

cbegeman:
### TESTING

**E3SM**

Baseline: `e3sm/master` at 6db49a63ae

PASS comparison with baseline
* `SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-knl_gnu`
* `PEM_Ln9.ne30_oECv3.A_WCYCL1850S.cori-knl_intel`
* `PET_Ln9.ne30_oECv3.A_WCYCL1850S.cori-knl_gnu`
* `SMS_Ld1.ne30pg2_r05_EC30to60E2r2.A_WCYCL1850S_CMIP6.compy_intel.allactive-wcprod`
* `SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi`
* `SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel`
* `PET_Ln9_P1024.ne30_oECv3.A_WCYCL1850S.compy_intel`

PASS (without comparison with baseline)
* `PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel`
* `PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel`
* `PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-haswell_gnu`
* `PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.compy_pgi`
* `PET_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi`
* `PEM_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi`

**Standalone**

`z_level_dev`: encompasses the nightly regression suite and four additional ice-shelf test cases. 
Baseline: `ocean/develop` and `MPAS-Dev/compass/legacy` + `c1650df1 REMOVE LATER!!! Add z-leve dev. test suite`

PASS
* `z_level_dev`, Grizzly, intel

cbegeman:
@vanroekel Thanks for this context. I think that you'll find option (2) to be reasonable, as the code changes are pretty straightforward and aren't extensive in these routines. The changes to shortwave_absorption_variable can be found in this commit: https://github.com/MPAS-Dev/MPAS-Model/pull/840/commits/9230b88bac4b0199286ec8da680bb894960f09cb. Let me know what you think.

vanroekel:
The changes to `shortwave_absorption_variable` look good to me.  I'm happy to approve those changes by visual inspection once the PR is ready to go.

Also, thanks for running the new Watercycle configuration on compy, I appreciate it!

xylar:
@cbegeman, I know I still owe you some testing on Anvil and Chrysalis.  I'll get to that as soon as I can, hopefully this evening.

src/core_ocean/shared/mpas_ocn_vmix.F
```diff
@@ -282,30 +283,31 @@ subroutine ocn_vel_vmix_tend_implicit_rayleigh(meshPool, dt, kineticEnergyCell,
...
                / layerThicknessEdge(k,iEdge)
          enddo
 
          ! B is diagonal term
-         B(1) = 1.0_RKIND - C(1) &
-           ! added Rayleigh terms
-           + dt*rayleighDampingCoef/((1.0_RKIND - rayleighDepthVariable) + rayleighDepthVariable*edgeThicknessTotal)
-         do k = 2, N-1
+         B(1:Nsurf-1)=0.0_RKIND
+         B(Nsurf) = 1.0_RKIND - C(Nsurf)
```
mattdturner, line 309: I'm probably missing something, but is the calculation here missing some terms?
```suggestion
         B(Nsurf) = 1.0_RKIND - C(Nsurf) &
           + dt*rayleighDampingCoef/((1.0_RKIND - rayleighDepthVariable) + rayleighDepthVariable*edgeThicknessTotal)
```

mattdturner (approved):
Approved via visual inspection and testing by @cbegeman and @xylar

cbegeman:
@mattdturner Thanks for your review! I made that fix to `mpas_ocn_vmix` and pushed one more commit that resolves a merge conflict with the freshwater flux treatment in this commit https://github.com/MPAS-Dev/MPAS-Model/commit/5aeccaa606e43bf69a0209cffcc2a7f7a26a396c. I'll test that out shortly.

src/core_ocean/Registry.xml
```diff
@@ -3091,6 +3091,22 @@
...
+		<var name="seaIceFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with sea ice melt water at cell centers sent to coupler. Positive into the ocean."
+		/>
+		<var name="icebergFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Heat flux associated with iceberg melt at cell centers sent to coupler. Positive into the ocean."
+		/>
+
+		<var name="totalFreshWaterTemperatureFlux" type="real" dimensions="nCells Time" units="degC m s^{-1}"
+			 description="Sum of heat fluxes associated with water fluxes cell centers sent to coupler. Positive into the ocean."
+		/>
```
xylar, line 3109: Mostly a note to myself to make sure this goes away after a rebase.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -239,12 +246,22 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, layerThickness, layerThicknessEdg
...
+               if (minLevelCell(cell1) < minLevelEdgeBot(iEdge)) then
+                  flux_term2 = flux_term2 + coef*RediKappa(iEdge)* &
+                               layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadUp(k, 1, iEdge)* &
+                               (tracers(iTr, k - 1, cell1) - tracers(iTr, k, cell1))/(zMid(k - 1, cell1) - zMid(k, cell1))
+               endif
+               if (minLevelCell(cell2) < minLevelEdgeBot(iEdge)) then
+                  flux_term2 = flux_term2 + coef*RediKappa(iEdge)* &
+                               layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadUp(k, 2, iEdge)* &
+                               (tracers(iTr, k - 1, cell2) - tracers(iTr, k, cell2))/(zMid(k - 1, cell2) - zMid(k, cell2))
+               endif
```
xylar, line 258: @mark-petersen (or maybe @vanroekel) could you take a careful look at these Redi changes, mostly because there among the more complex in this PR and I don't know Redi well enough to check them?

xylar (commented):
Wow, @cbegeman, this is some extensive and quite careful work.

As noted above, it seems like RK4 hasn't been fully transitioned yet.  Is that right?

And there are some freshwater-flux changes that I want to make sure disappear after a rebase. 

Beyond that, the only other thing I'd like is to have someone familiar with Redi check those 3 major computations out (@mark-petersen?).

I have done standalone testing with intel and gun on Anvil and Chrysalis.  The only test that is non-bit-for-bit is `sub_ice_shelf_2D/5km/iterative_init` with intel but not gnu.  Changes are order `1e-16` so not concerning.  Also, this is not a configuration we typically run with or use in any other tests besides this one.  I'm rerunning with intel in debug just to make sure the results are bit-for-bit there.

I can either perform E3SM testing on Anvil and Chrysalis now or wait for an E3SM PR, whichever is more useful.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
mark-petersen: Yes. I have a convergence test and idealized Redi test I can rerun.

cbegeman: Thanks @mark-petersen. So far I have only run the `Redi_verification/SouthernOceanSlice40` test case.

src/core_ocean/shared/mpas_ocn_tracer_DMS.F
```diff
@@ -172,18 +172,18 @@ subroutine ocn_tracer_DMS_compute(activeTracers, DMSTracers, nTracersDMS, ecosys
 
       numColumns = 1
       column = 1
-      iLevelSurface = 1
+      iLevelSurface = minLevelCell(iCell)
```
xylar, line 175: This needs to move inside the `iCell` loop.

xylar:
My testing in debug mode ran into some other issues.

`Global Ocean 240km - RK4 Partition Test` and `Global Ocean 240km - RK4 Thread Test` failed with this stack trace:
```
forrtl: severe (408): fort: (3): Subscript #1 of the array VELOCITYZONAL has value 0 which is less than the lower bound of 1

libpnetcdf.so.3.0  0000155551067332  for_emit_diagnost     Unknown  Unknown
ocean_model        000000000119FB93  ocn_time_integrat         724  mpas_ocn_time_integration_rk4.F
ocean_model        0000000001086CBD  ocn_time_integrat         108  mpas_ocn_time_integration.F
ocean_model        000000000108538A  ocn_forward_mode_         597  mpas_ocn_forward_mode.F
ocean_model        000000000107E7C4  ocn_core_mp_ocn_c         111  mpas_ocn_core.F
ocean_model        0000000000413467  mpas_subdriver_mp         358  mpas_subdriver.F
ocean_model        000000000040E20A  MAIN__                     20  mpas.F
ocean_model        000000000040E162  Unknown               Unknown  Unknown
libc-2.28.so       000015554DE527B3  __libc_start_main     Unknown  Unknown
ocean_model        000000000040E06E  Unknown               Unknown  Unknown

```
It isn't immediately clear to me what it causing this but it didn't occur in the baseline.  The line referred to is:
```
surfaceVelocity(indexSurfaceVelocityZonal, iCell) = velocityZonal(minLevelEdgeBot(iEdge), iCell)
```
which suggests that `minLevelEdgeBot` may not yet have been initialized here?

In ``, I'm seeing:
```
forrtl: severe (194): Run-Time Check Failure. The variable 'ocn_tracer_dms_mp_ocn_tracer_dms_compute_$ICELL' is being used in 'mpas_ocn_tracer_DMS.F(175,23)' without being defined
ocean_model        00000000014D580A  ocn_tracer_dms_mp         175  mpas_ocn_tracer_DMS.F
ocean_model        000000000154576B  ocn_tendency_mp_o         704  mpas_ocn_tendency.F
ocean_model        00000000010B463B  ocn_time_integrat        1493  mpas_ocn_time_integration_split.F
ocean_model        0000000001086CE2  ocn_time_integrat         110  mpas_ocn_time_integration.F
ocean_model        000000000108538A  ocn_forward_mode_         597  mpas_ocn_forward_mode.F
ocean_model        000000000107E7C4  ocn_core_mp_ocn_c         111  mpas_ocn_core.F
ocean_model        0000000000413467  mpas_subdriver_mp         358  mpas_subdriver.F
ocean_model        000000000040E20A  MAIN__                     20  mpas.F
ocean_model        000000000040E162  Unknown               Unknown  Unknown
libc-2.28.so       000015554DE527B3  __libc_start_main     Unknown  Unknown
ocean_model        000000000040E06E  Unknown               Unknown  Unknown
```
Here, the relevant line is:
```
iLevelSurface = minLevelCell(iCell)
```
I have noted that this line needs to move inside the `iCell` loop in my comment above.

The rest of my COMPASS testing in debug mode was fine (and bit-for-bit).

src/core_ocean/mode_forward/mpas_ocn_time_integration_rk4.F
```diff
@@ -1071,7 +1073,9 @@ subroutine ocn_time_integrator_rk4_diagnostic_update(block, dt, rkSubstepWeight,
       call mpas_pool_get_array(tendPool, 'layerThickness', layerThicknessTend)
       call mpas_pool_get_array(tendPool, 'lowFreqDivergence', lowFreqDivergenceTend)
 
+      call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
+      call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
       call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
```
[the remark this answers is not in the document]
cbegeman: @xylar I had to revert these changes to maintain bit-for-bit results. However, this was only necessary in the `*New = *Cur` assignments so my reasoning was that it's worth paying the additional cost in the `minLevelCell != 1` cases to keep bit-for-bit. Anyone is welcome to disagree with me on that though. You are right that we don't need those lines and I'll remove them.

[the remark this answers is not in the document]
xylar: I see.  That makes sense to me.  Maybe make a comment in the code about that.  Otherwise, someone might be tempted to "fix" it later.

src/core_ocean/shared/mpas_ocn_gm.F
```diff
@@ -671,9 +673,9 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
             gmStreamFuncTopOfEdge(:, iEdge) = 0.0_RKIND
 
             ! Construct the tridiagonal matrix
-            if (maxLevelEdgeTop(iEdge) .GE. 3) then
+            if ((maxLevelEdgeTop(iEdge) - minLevelEdgeBot(iEdge)) .GE. 2) then
```
vanroekel, line 676: this change seems to reproduce the original behavior, but is there a reason to switch the form instead of 
```
maxLevelEdgeTop(iEdge) .GE. minLevelEdgeBot(iEdge) + 2
```
?

src/core_ocean/shared/mpas_ocn_tendency.F
```diff
@@ -1069,15 +1071,16 @@ subroutine ocn_tend_freq_filtered_thickness(tendPool, statePool, meshPool, timeL
...
             flux = layerThickEdge(k, iEdge) * normalVelocity(k, iEdge) * dvEdge(iEdge) * edgeSignOnCell(i, iCell) * invAreaCell
             div_hu(k) = div_hu(k) - flux
             div_hu_btr = div_hu_btr - flux
           end do
         end do
 
-        totalThickness = sum(layerThickness(1:maxLevelCell(iCell),iCell))
-        do k = 1, maxLevelCell(iCell)
+        totalThickness = sum(layerThickness(minLevelCell(iCell):maxLevelCell(iCell),iCell))
+        do k = minLevelCell(iCell), maxLevelCell(iCell)
```
vanroekel, line 1083: not necessarily related to this place but what is the plan for cases when there are fewer than three active cells (minLevelCell within 3 of maxLevelCell). Will this be allowed?  If so, will the calls to ocn_tendency be skipped?  Something else?

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -221,6 +223,11 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, layerThickness, layerThicknessEdg
             r_tmp = dvEdge(iEdge)/dcEdge(iEdge)
             coef = dvEdge(iEdge)
 
+            k = minLevelEdgeBot(iEdge)
+            kappaRediEdgeVal = 0.25_RKIND*(RediKappaScaling(k, cell1) + RediKappaScaling(k, cell2) + &
+                                           RediKappaScaling(k + 1, cell1) + RediKappaScaling(k + 1, cell2))* &
+                               0.25_RKIND*(RediKappaSfcTaper(k, cell1) + RediKappaSfcTaper(k, cell2) + &
+                                           RediKappaSfcTaper(k + 1, cell1) + RediKappaSfcTaper(k + 1, cell2))
             k = 1
```
vanroekel, line 231: why not minLevelEdgeBot here?

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
vanroekel: I think this will be fine for `minLevelCell = 1`, but not for other cases. I'll try make suggestions what I think it should be above.  Also I believe `k=1` at this point, so this will give out of bounds errors if this if is triggered.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -239,12 +246,22 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, layerThickness, layerThicknessEdg
                              /(zMid(k, cell1) - zMid(k + 1, cell1)) &
                              + slopeTriadDown(k, 2, iEdge)*(tracers(iTr, k, cell2) - tracers(iTr, k + 1, cell2)) &
                              /(zMid(k, cell2) - zMid(k + 1, cell2)))
+               if (minLevelCell(cell1) < minLevelEdgeBot(iEdge)) then
+                  flux_term2 = flux_term2 + coef*RediKappa(iEdge)* &
+                               layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadUp(k, 1, iEdge)* &
+                               (tracers(iTr, k - 1, cell1) - tracers(iTr, k, cell1))/(zMid(k - 1, cell1) - zMid(k, cell1))
```
vanroekel, line 252:
```suggestion
                 k = minLevelCell(cell1)
                  flux_term2 = flux_term2 + coef*RediKappa(iEdge)* &
                               layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadUp(k+1, 1, iEdge)* &
                               (tracers(iTr, k, cell1) - tracers(iTr, k+1, cell1))/(zMid(k, cell1) - zMid(k+1, cell1))
```

vanroekel: @mark-petersen could you check this?  I'm not sure this suggestion is exactly right.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -239,12 +246,22 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, layerThickness, layerThicknessEdg
...
+               endif
+               if (minLevelCell(cell2) < minLevelEdgeBot(iEdge)) then
+                  flux_term2 = flux_term2 + coef*RediKappa(iEdge)* &
+                               layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadUp(k, 2, iEdge)* &
+                               (tracers(iTr, k - 1, cell2) - tracers(iTr, k, cell2))/(zMid(k - 1, cell2) - zMid(k, cell2))
```
vanroekel, line 257:
```suggestion
                  k = minLevelCell(cell2)
                  flux_term2 = flux_term2 + coef*RediKappa(iEdge)* &
                               layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadUp(k+1, 2, iEdge)* &
                               (tracers(iTr, k, cell2) - tracers(iTr, k+1, cell2))/(zMid(k, cell2) - zMid(k+1, cell2))
```

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -507,17 +509,17 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, err, tim
...
-                   w_s = turbulentScalarVelocityScale(1:maxLevelCell(iCell)))
+                   w_s = turbulentScalarVelocityScale(minLevelCell(iCell):maxLevelCell(iCell)))
 
               ! averaging over a surface layer assuming that BLdepth is cell bottom
               ! Build deltaVelocitySquared
               do i = 1, nEdgesOnCell(iCell)
                  iEdge = edgesOnCell(i, iCell)
 
-                 deltaVelocitySquared(1:minLevelCell(iCell)) = 0.0_RKIND
+                 deltaVelocitySquared(minLevelCell(iCell):minLevelCell(iCell)) = 0.0_RKIND
```
vanroekel, line 522: It's not clear to me why this should be changed.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -668,7 +670,7 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, err, tim
             ! intent out of BoundaryLayerDepth is boundary layer depth measured in meters and vertical index
             indexBoundaryLayerDepth(iCell) = cvmix_variables % kOBL_depth
 
-            do k = 1, maxLevelCell(iCell) + 1
+            do k = minLevelCell(iCell), maxLevelCell(iCell) + 1
```
vanroekel, line 673: I'm not sure this will work as expected for minLevelCell not equal to 1.  I could imagine since you declare `cvmix_variables % Mdiff_iface` to be nVertLevels+1, you will get the first values in the area entered by cvmix, but with this change you will start at minLevelCell(iCell).  You may need a different counter for that array access below

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
```diff
@@ -685,7 +687,7 @@ subroutine ocn_vmix_coefs_cvmix_build(meshPool, statePool, forcingPool, err, tim
          if ( kpp_stage == 2) then
          ! call convective mixing scheme
          if (config_use_cvmix_convection) then
-            do k = 1, maxLevelCell(iCell) + 1
+            do k = minLevelCell(iCell), maxLevelCell(iCell) + 1
```
vanroekel, line 690: similar comment to above.  I think you'll want to keep this as 1 to zero out the entire array.

vanroekel (changes requested):
@cbegeman really impressive and careful work here.  I took a really close look at gm, redi, and vmix and a quicker look at other places.  I think there are some issues for when minLevelCell is not 1 in redi and cvmix.  But could be missing something.

src/core_ocean/shared/mpas_ocn_gm.F
cbegeman: The former seemed a little more intuitive to me, but I don't mind changing it to your version.

src/core_ocean/shared/mpas_ocn_tendency.F
xylar: @cbegeman can respond but my feeling is that it's up to the code that actually computes the vertical coordinate (either at init or as it evolves in time) to enforce the 3-cell minimum.  That would be beyond the scope of this PR, of course.

cbegeman: It's a great question, and one we haven't settled. We should continue the conversation [here](https://github.com/MPAS-Dev/MPAS-Model/pull/811/files#r574745570) and flesh out the design doc further on this point. It'll have to be more nuanced than skipping all calls to ocn_tendency.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
cbegeman: I duplicated these lines first with `k=minLevelEdgeBot` and then `k=1` because I found that this was needed to preserve bit-for-bit results in standalone cases with ecosystem tracers. I can try removing these lines again after the following changes and see if I get the same result.

src/core_ocean/shared/mpas_ocn_tendency.F
vanroekel: I totally agree this is way beyond this PR, so no need to address farther in this space.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
vanroekel: I see, thanks for the explanation.  I doubt my suggestions below will help though.  I don't believe they will make a difference if minLevelCell(:) = 1

Just for my understanding L226-230 are extraneous right now correct?

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
cbegeman: @vanroekel Thanks for catching this. I think we want k to be set back to `minLevelEdgeBot` as it was above before I added the reassignment to 1. And then the `k-1` index will never be 0 because we never allow `minLevel*` to be 0.

src/core_ocean/shared/mpas_ocn_gm.F
vanroekel: actually thinking about it, I agree the new change does make more intuitive sense, but we may want to make a comment that this ensures that this ensures there are at least three active cells in the column. Since we have that requirement elsewhere some may get tripped up by seeing a '2' here. But is also probably fine to leave as is.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
cbegeman: I can change it back. I don't think we ever use values `k<minLevelCell(iCell)`

vanroekel: good point.  Could we instead change to
```
deltaVelocitySquared(minLevelCell(iCell)) = 0.0_RKIND
```
I don't see the need for array indexing.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
vanroekel: Ah yes, that is a much better fix.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
cbegeman: @vanroekel Can you say a little more about what you mean? The strategy I implemented was to keep a correspondence between k-levels in cvmix%* and the other variables in this routine. Are you suggesting it would be better to shift the k-levels in cvmix%* so that they start at minLevelCell?

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
cbegeman: Yeah. I think it comes down to the relative importance of having clean code (delete L231-235) and retaining bit-for-bit results in ecosystem tracers cases (keep L231-235). Maybe @maltrud has an opinion on this?

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
cbegeman: Oh, I totally missed the fact that the start and end indices were the same. Of course.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
vanroekel: yes of course.  I think the issue is that in the `cvmix_coeffs_kpp` call you pass things like
```
cvmix_variables % Mdiff_iface(minLevelCell(iCell):maxLevelCell(iCell)+1)
```
but have `max_nlev = nVertLevels`.  I think what will happen is that cvmix will be expecting an array that is nVertLevels long and if minLevelCell is not 1 the lengths are inconsistent.  I think you may trip an error in that case or perhaps cvmix will just fill the first maxLevel - minLevel + 1 slots of the array.  I doubt it will fill the correct way.  

Thinking about this, I think the best fix is to change `max_nlev` in L666 to `maxLevelCell(icell) - minLevelCell(iCell) + 1`. I think this will then make the loop variables right here. 

Does that make sense?

vanroekel: looking at the code I think we also want `nlev = maxLevelCell(iCell) - minLevelCell(iCell) + 1` on L665

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
xylar: Actually, I think the most likely thing going on here is that the original code is from phase 1 and you're modifying it again in phase 2, but should probably leave it as it was in phase 1.

xylar: I would assume anything that already has a `minLevel*` has already been handled.

cbegeman: Good point. I'll leave it as it was.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
cbegeman: I'll try changing `nlev` as you suggest. However, I do wonder if we actually need `max_nlev = maxLevelCell(iCell) - minLevelCell(iCell)+2` since the array limits of the other inputs are `minLevelCell(iCell):maxLevelCell(iCell)+1`

vanroekel: oh yes, oops.  You are completely right.  It should be +2, i think they should both be the same.

mark-petersen:
I ran the Redi test in compass legacy, 
```
193: -o ocean -c tendency_verification -r all -t Redi
```
which computes the Redi tendency term with 9 different polynomials in temperature, salinity, and layer thickness with an ideal equation of state and compares error against the analytic Redi terms (I had to use sympy). Solutions are bfb identical between `ocean/develop` and this PR. Using faf28979d and intel optimized. Here are the convergence plots from this PR:
image
The right column shows the nine error terms with error differences from one resolution to the next. Columns 1 and 2 are other error normalizations.

cbegeman:
@mark-petersen Thank you for running that verification test!

mark-petersen:
With the current head faf28979d it fails RK4, grizzly intel-mpi_19
```
make intel-mpi_19 CORE=ocean USE_PIO2=true OPENMP=true DEBUG=true GEN_F90=true
   ** FAIL (See case_outputs/Global_Ocean_240km_-_RK4_Partition_Test for more information)

fails to run with:
  1  * Running 4partition_run
  2 forrtl: severe (408): fort: (3): Subscript #1 of the array MINLEVELEDGEBOT has value -858993460 which is less than the lower bound of 1
  3
  4 Image              PC                Routine            Line        Source
  5 ocean_model        0000000002D89F46  Unknown               Unknown  Unknown
  6 ocean_model        00000000024C7DD0  ocn_time_integrat         736  mpas_ocn_time_integration_rk4.f90
  7 libiomp5.so        00002B8DB7AC9ED3  __kmp_invoke_micr     Unknown  Unknown
  8 libiomp5.so        00002B8DB7A8CDF7  __kmp_fork_call       Unknown  Unknown
  9 libiomp5.so        00002B8DB7A4F020  __kmpc_fork_call      Unknown  Unknown
 10 ocean_model        00000000024BE000  ocn_time_integrat         733  mpas_ocn_time_integration_rk4.f90
 11 ocean_model        0000000001B0CEBA  ocn_time_integrat         119  mpas_ocn_time_integration.f90
 12 ocean_model        00000000014EEE2C  ocn_forward_mode_         602  mpas_ocn_forward_mode.f90
```

cbegeman:
@mark-petersen I'm aware of that issue. It seems to have been introduced by the recent commits and I'm working on it.

cbegeman:
@mark-petersen RK4 is now fixed. It passes nightly + z_level_dev standalone cases on Grizzly with intel17. I'll be retesting cvmix since we had some changes there.

cbegeman:
cvmix tests `single_column_model/sphere` and `single_column_model/planar` are bit-for-bit with `ocean/develop`. 

@xylar, it's ready for any testing you want to do. I'll start with Cori and Compy E3SM tests.

cbegeman:
@xylar This branch still passes all compy and cori tests listed above as well as the nightly ocean test suite on Grizzly. I got a build error at `[ 74%] Build target lnd` on blues.

src/core_ocean/shared/mpas_ocn_vmix_cvmix.F
cbegeman: @vanroekel I kept the number of vertical levels specified in arguments to cvmix consistent with `ocean/develop` in order to pass WC tests (https://github.com/MPAS-Dev/MPAS-Model/pull/840/commits/f9e02c36acd725c76cb8acdde8bc5949642d5728).

vanroekel: @cbegeman very interesting, I'm not sure why that would be, but definitely not your job to figure out.  If the current iteration passes WC testing, I'll be happy to approve as is.

src/core_ocean/shared/mpas_ocn_tracer_hmix_redi.F
cbegeman: @vanroekel @maltrud I haven't come up with a good way to have Redi work correctly for the `minLevel* != 1` cases while preserving bit-for-bit results in MPAS Standalone QU240/bgc_ecosys_test. Commit 
19dc7d9 preserves bit-for-bit results in bgc_ecosys_test but is incompatible with `minLevel* != 1` while commit https://github.com/cbegeman/MPAS-Model/commit/71663bf5ac71d361c5b03f3891851b857b62e401 is compatible with `minLevel* != 1` and introduces non-bfb results in bgc_ecosys_test. The differences are small and affect only the tracers shown below. Both branches pass all the E3SM tests [here](https://github.com/MPAS-Dev/MPAS-Model/pull/840#issuecomment-813427479). 

Let me know what you think. An alternative could be duplicating the block of code that solves for `flux_term2` but that's not very readable. 

Thanks!

```
Beginning variable comparisons for all time levels of field 'spChl'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
0:  l1: 4.37707780575496e-23  l2: 2.42434975903054e-19  linf: 2.16840434497101e-19
 ** FAIL Comparison of spChl between forward/output.nc and
    /home/ac.cbegeman/MPAS-Ocean-output/develop_i18/ocean/global_ocean/QU240/bgc_ecosys_test/forward/output.nc
Beginning variable comparisons for all time levels of field 'spC'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
0:  l1: 7.47021278848847e-21  l2: 5.55111512312578e-17  linf: 5.55111512312578e-17
 ** FAIL Comparison of spC between forward/output.nc and
    /home/ac.cbegeman/MPAS-Ocean-output/develop_i18/ocean/global_ocean/QU240/bgc_ecosys_test/forward/output.nc
Beginning variable comparisons for all time levels of field 'diatFe'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
0:  l1: 7.12415007447097e-27  l2: 5.29395592033938e-23  linf: 5.29395592033938e-23
 ** FAIL Comparison of diatFe between forward/output.nc and
    /home/ac.cbegeman/MPAS-Ocean-output/develop_i18/ocean/global_ocean/QU240/bgc_ecosys_test/forward/output.nc
Beginning variable comparisons for all time levels of field 'phaeoChl'. Note any time levels reported are 0-based.
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
0:  l1: 1.02131815467616e-22  l2: 4.47028007668187e-19  linf: 3.25260651745651e-19
 ** FAIL Comparison of phaeoChl between forward/output.nc and
    /home/ac.cbegeman/MPAS-Ocean-output/develop_i18/ocean/global_ocean/QU240/bgc_ecosys_test/forward/output.nc
```

xylar (approved):
My testing all when great (see https://github.com/MPAS-Dev/MPAS-Model/pull/840#issuecomment-813427479).  The one standalone test where there are diffs (`ocean/sub_ice_shelf_2D/5km/iterative_init/`) is clearly an optimization thing and it's not a configuration representative of anything we run production runs with so I'm fine with just "blessing" the changes.

cbegeman:
@xylar Thank you for all your testing efforts!

src/core_ocean/shared/mpas_ocn_tracer_ecosys.F
mark-petersen: The following six lines are a rearrangement of index order:
`(k,iCell,iTracer) <--- (iTracer,k,iCell)`
but only for a single column, and only for the top layer. I'm sure this is for the BGC interface API.

@cbegeman I agree that all 1 indices in the next six lines should be replaced with `iLevelSurface`. I think a shorter index name like `kTop` would be more readable. Perhaps `kSurface`.

cbegeman: @mark-petersen Thanks! These changes are in 9d189a2. With this commit, it passes the nightly test suite.

mark-petersen (approved):
Reviewed visually and read past comments. Passes nightly regression suite on last commit 
9d189a2  with intel 19 and gnu 5.3, both debug and optimized on grizzly, and bfb match with `ocean/develop` on both. Also, analytic comparison and convergence test for Redi terms are correct and match bfb with previous.

@cbegeman and @xylar thanks for your detailed work on this.

vanroekel (approved):
approving based on visual inspection and extensive testing from @xylar and @cbegeman

cbegeman:
@vanroekel Thanks for all your input!

philipwjones (approved):
Previous comments addressed. Approve based on visual inspection, testing by others.

PR #843, opened 2021-04-06
Fixes nonlocal temp flux for KPP

vanroekel:
Currently the nonLocalTemperatureFlux includes the river flux as 
```
surfaceThicknessFluxRunoff*max(0,SST)/rho_sw
```
however `surfaceThicknessFluxRunoff` is already divided by rho_sw in `mpas_ocn_surface_bulk_forcing.F` so this second division is incorrect and does not yield units of m C /s. The extra `rho_sw` is removed in this PR.

The effect should be small but the PR will be nonBFB

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_diagnostics.F
+++ b/src/core_ocean/shared/mpas_ocn_diagnostics.F
@@ -2020,6 +2020,10 @@ contains
                               -100.0_RKIND) )
          ! Store the total tracer flux below in nonLocalSurfaceTemperatureFlux for use in the CVMix nonlocal
          ! transport code.  This includes tracer forcing due to thickness
+         ! Note: rainFlux, evaporationFlux, seaIceFreshWaterFlux, icebergFreshWaterFlux 
+         !       all have units of kg m^{-2} s^{-1}, so require division by rho_sw.
+         !       surfaceThicknessFlux, surfaceThicknessFluxRunoff have units of m s^{-1},
+         !       so has no division by rho_sw is needed.
 
          nonLocalSurfaceTracerFlux(indexTempFlux, iCell) = activeTracersSurfaceFlux(indexTempFlux,iCell) &
                  + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell)  &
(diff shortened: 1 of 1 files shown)


vanroekel:
@maltrud here is the KPP fix we talked about.  Could you approve?  And @mark-petersen can you add this to the list to merge as soon as there is an opening?

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -2020,11 +2020,15 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, timeLe
...
+         ! Note: rainFlux, evaporationFlux, seaIceFreshWaterFlux, icebergFreshWaterFlux 
+         !       all have units of kg m^{-2} s^{-1}, so require division by rho_sw.
+         !       surfaceThicknessFlux, surfaceThicknessFluxRunoff have units of m s^{-1},
+         !       so has no division by rho_sw is needed.
 
          nonLocalSurfaceTracerFlux(indexTempFlux, iCell) = activeTracersSurfaceFlux(indexTempFlux,iCell) &
                  + penetrativeTemperatureFlux(iCell) - penetrativeTemperatureFluxOBL(iCell)  &
                  - fracAbsorbed * (rainFlux(iCell) + evaporationFlux(iCell)) * activeTracers(indexTempFlux,1,iCell)/rho_sw &
-                 - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell)* min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND)/rho_sw &
+                 - fracAbsorbedRunoff * surfaceThicknessFluxRunoff(iCell)* min(activeTracers(indexTempFlux,1,iCell),0.0_RKIND) &
```
mark-petersen, line 2031: @vanroekel I added the note above because someone in the future might think `/rho_sw` is missing again. I checked the Registry, and units are all correct there and agree with this correction.

src/core_ocean/shared/mpas_ocn_diagnostics.F
```diff
@@ -2020,11 +2020,15 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, meshPool, timeLe
                               -100.0_RKIND) )
          ! Store the total tracer flux below in nonLocalSurfaceTemperatureFlux for use in the CVMix nonlocal
          ! transport code.  This includes tracer forcing due to thickness
+         ! Note: rainFlux, evaporationFlux, seaIceFreshWaterFlux, icebergFreshWaterFlux 
+         !       all have units of kg m^{-2} s^{-1}, so require division by rho_sw.
+         !       surfaceThicknessFlux, surfaceThicknessFluxRunoff have units of m s^{-1},
+         !       so has no division by rho_sw is needed.
```
mark-petersen, line 2026: Note here, I mean.

mark-petersen:
@vanroekel and @maltrud please compare these lines to the PR #837 I just merged, here:
https://github.com/MPAS-Dev/MPAS-Model/pull/837/files#diff-8095f6b5d5976cc3adcbbbea7455b974cf97a7364ba7062622349fd58b322002R2026-R2028

Those divide by rho_sw are alrady gone, so it looks like this current PR is not needed. Everything is now in units of `m s^{-1}`. If you agree, we can just close this PR.

vanroekel:
Thanks @mark-petersen I did not track that PR and made this PR off an earlier tag.  I agree this PR is not needed.  I am closing

PR #853, opened 2021-04-21
Include river runoff term in totalFreshWaterTemperatureFlux

jonbob:
Adds the river runoff term to totalFreshWaterTemperatureFlux, which is passed to the coupler and atm in the E3SM driver.


jonbob:
@maltrud - can you please review and make sure I got what you intended included in this PR? The file had been updated since you checked it out so I had to merge changes

maltrud (approved):
thanks for getting this going so quickly, @jonbob .

mark-petersen (approved):
passes nightly regression suite with gnu and intel, both debug. Looks good.

PR #854, opened 2021-04-21
Enable mpas_timekeeping module to work with the full, external ESMF library [atmosphere/cam]

mgduda:
This PR enables the mpas_timekeeping module to work with the full, external ESMF
library.

While the MPAS repository includes an old, Fortran-only implementation of
the ESMF time manager, coupled systems in which MPAS is a component may link
with the full (and more recent) ESMF library. This PR makes changes in
the mpas_timekeeping module to enable it to use the full, external ESMF library
in place of the built-in, Fortran-only time manager library by compiling
the MPAS framework with the MPAS_EXTERNAL_ESMF_LIB pre-processing macro defined.

Note that this PR does not include changes to Makefiles to set the module
include paths to an external ESMF library, nor does it include changes to handle
linking with an external ESMF library. The intended use case for these changes
is a coupled model (e.g., CESM with the ESMF-based NUOPC driver) in which MPAS
models are components. In this case, the stand-alone MPAS Makefiles are unlikely
to be used, and all MPAS code is compiled with the build system of the coupled
model.

### DIFF
--- a/src/framework/mpas_derived_types.F
+++ b/src/framework/mpas_derived_types.F
@@ -31,12 +31,6 @@ module mpas_derived_types
    use pio_types
 
    use ESMF
-   use ESMF_BaseMod
-   use ESMF_Stubs
-   use ESMF_CalendarMod
-   use ESMF_ClockMod
-   use ESMF_TimeMod
-   use ESMF_TimeIntervalMod
 
 #include "mpas_attlist_types.inc"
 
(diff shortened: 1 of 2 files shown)


mgduda:
This PR mirrors PR #809, but targets the `atmosphere/cam` branch rather than the `develop` branch.

mgduda:
The changes in this PR were tested and found to be working by CESM developers. Note, too, that the changes in this PR will be reviewed before inclusion in the `develop` branch through PR #809 .

PR #862, opened 2021-05-04
Fix bad gotm path in ocean build

### LINKED ISSUE #861: ocean build fails due to GOTM dependency with PGI on Summit
While attempting to build latest MPAS on Summit with PGI, the compiler can't find the mod file for mtridiagonal (a GOTM module) when trying to build one of the top-level drivers mpas_subdriver.  There's no explicit include at this level and no reason for PGI to search for it. GOTM paths are (correctly) not in the search path at this level. I suspect PGI is following the module dependencies recursively further down.  Working on it...

philipwjones:
The current path in the ocean core build_options.mk for the gotm package points to the wrong directory for module files, causing build failures for PGI compiler. This PR points to the proper directory.
bit for bit
Fixes #861


qingli411 (approved):
This looks good to me. Thanks, @philipwjones!

Just curious, why it was working fine with GNU and Intel?

philipwjones:
@qingli411 - Within core_ocean the paths appear in two places. In the Makefile they are added to OCEAN_SHARED_INCLUDES and these are used to build the dycore library and were set correctly.  But for some reason, PGI was also looking for module files at the higher mpas_subdriver level, even though there are no explicit use statements for the modules. PGI seems to follow the recursive "use" statements down the call stack? In that case, the include paths added in the build_options are being used by the higher-level makefile and the correct path wasn't there.  Apparently the other compilers are only going off the explicit use statements so this isn't an issue. It's probably not great to have these set in multiple places and the build system needs some work, but this fix works for now.

qingli411:
@philipwjones OK I see. Thanks!

mattdturner:
It looks like the change in this PR is no longer necessary, as the include path for GOTM within E3SM is already correct.  This can be closed.

mark-petersen:
Yes, I added this change in the transplant, as I was messing with the makefiles anyway. Thanks for closing.

PR #863, opened 2021-05-05
Enable bootstrapping from an externally provided PIO file descriptor

mgduda:
This PR adds the ability for both phase1 and phase2 of the boostrapping process
to use an externally provided PIO file descriptor.

When MPAS is a component in a larger system, it may be necessary to "bootstrap"
the MPAS infrastructure from an externally provided file descriptor. This PR
adds an optional PIO file_desc_t argument to mpas_bootstrap_framework_phase1
and to mpas_bootstrap_framework_phase2.

The MPAS_io_open routine now uses the optional PIO file_desc_t if provided, and
opens the file indicated by the filename argument as before if no PIO
file_desc_t argument is provided.

In order to avoid closing a PIO file_desc_t that was not opened by MPAS_io_open
(because this external file_desc_t was provided as an optional argument in
the call to MPAS_io_open), this PR also adds a new member, external_file_desc,
to the MPAS_IO_Handle_type type. This member is set to .true. if an external
PIO file_desc_t was provided, and if .true. in the call to MPAS_io_close,
PIO_closefile is not called.

In the second bootstrapping phase, when a PIO file descriptor is provided,
rather than scanning through all input streams from the streams.<core> file
to find definitions of dimensions, the routine searches for dimensions only in
the file corresponding to the externally provided PIO file descriptor.

### DIFF
--- a/src/framework/mpas_bootstrapping.F
+++ b/src/framework/mpas_bootstrapping.F
@@ -441,14 +441,16 @@ module mpas_bootstrapping
    !>  and allocating all fields and structs.
    !
    !-----------------------------------------------------------------------
-   subroutine mpas_bootstrap_framework_phase2(domain) !{{{
+   subroutine mpas_bootstrap_framework_phase2(domain, pio_file_desc) !{{{
 
       use mpas_stream_manager
       use mpas_stream_list
+      use pio, only : file_desc_t
 
       implicit none
 
       type (domain_type), pointer :: domain
+      type (file_desc_t), pointer, optional :: pio_file_desc
 
       type (mpas_pool_type), pointer :: readableDimensions
       type (mpas_pool_type), pointer :: streamDimensions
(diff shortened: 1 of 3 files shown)


mark-petersen (approved):

src/framework/mpas_bootstrapping.F
```diff
@@ -474,105 +480,173 @@ subroutine mpas_bootstrap_framework_phase2(domain) !{{{
...
-            call mpas_pool_create_pool(streamDimensions)
+         ! If to determine if file was opened or not.
+         if ( err_local == MPAS_IO_NOERR ) then
 
-            do while ( mpas_stream_mgr_get_next_field(domain % streamManager, streamName, fieldName, isActive=fieldActive) )
+            call mpas_log_write(' ')
 
-               if (fieldActive) then
-                  call get_dimlist_for_field(domain % blocklist % allFields, fieldName, dimNames)
+            ! Iterate over list of dimensions we determined we need from the above loop
```
MiCurry, line 510: I think this comment would be a bit better if it explained why it was iterating: `! Iterate over the list of dimensions we determined we need from the above loop to read dimension values`.

What do you think?

mgduda: In my opinion, the comments in the loop body sufficiently explain what is happening when we iterate over the list of dimensions (which is more than just reading dimension values); for example, see the comments on [L.514](https://github.com/MPAS-Dev/MPAS-Model/blob/0b145f239cd54090bfba0c486017289ffaf6955d/src/framework/mpas_bootstrapping.F#L514) and [L.517](https://github.com/MPAS-Dev/MPAS-Model/blob/0b145f239cd54090bfba0c486017289ffaf6955d/src/framework/mpas_bootstrapping.F#L517).

MiCurry: Sounds good.

MiCurry (approved):

PR #865, opened 2021-05-05
Make several stream manager routines public: postread_reindex, prewrite_reindex, and postwrite_reindex

mgduda:
The postread_reindex, prewrite_reindex, and postwrite_reindex routines in
the mpas_stream_manager module were previously private. However, code that
builds ad hoc output streams at the mpas_io_streams level that contain
indexing fields could benefit from the use of these routine to reindex
indexing fields (e.g., cellsOnCell).
    
The use of prewrite_reindex and postwrite_reindex requires some care, since
they set module variables within the mpas_stream_manager module to save and
retrieve pointers to the original indexing arrays. Problems can arise, for
example, if external code calls prewrite_reindex, then makes a call to write
output streams via the stream manager: in this case, pointers set by the
external call to prewrite_reindex will be overwritten by the internal call to
prewrite_reindex within mpas_stream_mgr_write.

Unlike the prewrite_reindex and postwrite_reindex routines, which rely on
module state, the postread_reindex routine does not rely on any module
variables in the mpas_stream_manager module, and can therefore be safely used
by external code.

### DIFF
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -47,6 +47,8 @@ module mpas_stream_manager
               MPAS_get_stream_filename, &
               MPAS_build_stream_filename
 
+    public :: postread_reindex
+
     private
 
     interface MPAS_stream_mgr_set_property
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -47,7 +47,9 @@ module mpas_stream_manager
               MPAS_get_stream_filename, &
               MPAS_build_stream_filename
 
-    public :: postread_reindex
+    public :: prewrite_reindex, &
+              postwrite_reindex, &
+              postread_reindex
 
     private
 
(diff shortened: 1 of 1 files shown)


mark-petersen (approved):
Passes stand-alone ocean test suite, intel and gnu. Please merge. I thought I'd review these, so we have the option to tag along with the framework a bit longer.

src/framework/mpas_stream_manager.F
```diff
@@ -47,7 +47,9 @@ module mpas_stream_manager
               MPAS_get_stream_filename, &
               MPAS_build_stream_filename
 
-    public :: postread_reindex
+    public :: prewrite_reindex, &
+              postwrite_reindex, &
+              postread_reindex
```
MiCurry, line 52: Should we make a comment here for `prewrite_reindex` and `postwrite_reindex` that says that these routines should be only be used when calling `MPAS_writeStream` and not `mpas_stream_mgr_write`? I think that would clarify to an unfamiliar user that these routines are for ad-hoc streams.

mgduda: That seems like a good idea. I've pushed a new commit that adds notes in several places.

MiCurry: Those notes look great!

MiCurry (approved):

PR #867, opened 2021-05-06
Refactor and clean up calls to scalar transport routines

mgduda:
This PR refactors and cleans up the calls to the MPAS-Atmosphere scalar transport
routines. Now, rather than querying subpools for arguments and calling `atm_advance_scalars`
and `atm_advance_scalars_mono`, the `atm_srk3` routine calls a new, higher-level
routine, `advance_scalars`, which in turn makes calls to `atm_advance_scalars` and
`atm_advance_scalars`. Additionally, the `advance_scalars` routine takes as an argument
the name of the scalar `var_array` with constituents to be transported, making it easier
to advect, e.g., additional var_arrays of chemical species.


mgduda:
Based on my testing, we are able to get bit-identical results between the branch in this PR and the current `develop` branch, and there is no negative impact on performance.

skamaroc (approved):

PR #868, opened 2021-05-06
Fix a bug in computing bottom drag coeffcient using Manning equation

caozd999:
This is an obvious bug: in Manning equation:
`Cd = g * manningN**2 * water_depth**(-1/3)`.
The original code missed a negative sign in the last parentheses of the equation.

Meanwhile, some exponents in the real form, as they are in fact integers, were changed to integers for computational efficiency purpose.

### DIFF
--- a/src/core_ocean/shared/mpas_ocn_vmix.F
+++ b/src/core_ocean/shared/mpas_ocn_vmix.F
@@ -782,19 +782,19 @@ contains
           endif
           if (vegetationMask(iCell) .eq. 1) then
             total_h = bottomDepth(iCell) + ssh(iCell)
-            old_bottom_Cd = gravity * bottomDrag(iCell)**2.0_RKIND * total_h**(1.0_RKIND/3.0_RKIND)
+            old_bottom_Cd = gravity * bottomDrag(iCell)**2 * total_h**(-1.0_RKIND/3.0_RKIND)
             inundation_depth = MIN(vegetationHeight(iCell), total_h)
             inundation_depth = MAX(inundation_depth, 1e-6)
             lambda = vegetationDiameter(iCell) * vegetationDensity(iCell)
             alpha = (config_vegetation_drag_coefficient*lambda/ &
-                   (4.0_RKIND*von_karman**2.0_RKIND*inundation_depth**2.0_RKIND))**(1.0_RKIND/3.0_RKIND)
+                   (4.0_RKIND*von_karman**2 * inundation_depth**2))**(1.0_RKIND/3.0_RKIND)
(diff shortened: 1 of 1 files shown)


xylar (approved):
Looks good.  @caozd999, what test would you suggest I run for this?

caozd999:
> Looks good. @caozd999, what test would you suggest I run for this?

Currently there is only one following test that uses this piece of code,
`76: -o ocean -c drying_slope -r marsh_flooding -t idealized_transect`

xylar:
Great, I'll give that one a try before and after the fix.

sbrus89 (commented):
Good catch @caozd999. The only extremely minor change I'd suggest here while we're looking at this is to change the `bottomDrag(iCell)**2.0_RKIND` to `bottomDrag(iCell)**2`. See @philipwjones's comment on #606.

caozd999:
"Second, as a general rule, don't use real numbers as exponents if they are in fact integers. The algorithm for computing real exponentiation is more expensive than the integer form so you want the compiler to choose the right form. This is especially true if the exponent is 2 or 3 where the compiler will often replace with a simple multiply (or even better, for a simple square, do the replacement yourself and avoid the exponentiation). - @philipwjones " 

Got it. Thanks @sbrus89

xylar:
@caozd999, I'm running this test but it would be nice to have a regression test for this code that takes less than several hours to run, similar to what we discussed in https://github.com/MPAS-Dev/compass/pull/87.

xylar:
@caozd999, I ran that test with the current `ocean/develop` and this branch.

`ocean/develop`:
Waterlevel40

`ocean/fix_bug`:
Waterlevel40

I don't see any difference at all.  Is that expected?

caozd999:
@xylar This fix should only affect the case that uses vegetation characteristics to determine the vegetation-induced Manning coefficient, i.e., the `vegManningEquation ` case. When running that test case, I output the ` old_bottom_drag`  and found the values before and after we fixed the bug were actually different. The reason that this fix had little effect on the water level of the `vegManningEquation ` case is because that the ` old_bottom_drag` is just an intermediate variables in calculating the vegetation Manning, which is largely determined by the vegetation characteristics.

sbrus89 (approved):
These changes look good. Thanks for catching this @caozd999.

mark-petersen (approved):
This looks great. Thanks @caozd999. Approved based on visual inspection and posted test results. We will get this in as soon as we can. We are moving the MPAS code into E3SM this week, I'll help you move this PR over as well after that.

mark-petersen:
Transplanted to E3SM: https://github.com/E3SM-Project/E3SM/pull/4328

PR #869, opened 2021-05-07
Enable building of "dynamics-only" MPAS-Atmosphere

mgduda:
This PR enables the MPAS-Atmosphere model to be built in a "dynamics-only"
configuration, in which no physics or physics interface code is compiled. This
is useful, e.g., in situations where physics parameterizations are provided by
another model that is hosting the MPAS-Atmosphere dycore (e.g., CAM-MPAS).

Building a dynamics-only stand-alone MPAS-Atmosphere model is accomplished
by commenting-out or deleting the definition of the PHYSICS make variable in
the atmosphere core's main Makefile.


MiCurry (approved):
Looks good! I was able to run a stand-alone dynamics-only simulation from this as well!

PR #870, opened 2021-05-11
Support runtime definition of scalars when MPAS-A is a CAM dycore

mgduda:
This PR makes changes to enable CAM to define the set of scalar constituents
for the MPAS-A dycore completely at runtime.

When MPAS-A is compiled as a CAM dycore, the macro `MPAS_CAM_DYCORE` is defined,
in which case the `scalars` and `scalars_tend` var_arrays will be pre-processed out
of the `Registry.xml` file so that no Registry-defined scalars will exist.

Further, CAM will add a config, `cam_pcnst`, which will cause the `atm_setup_block`
routine to call a new routine, `atm_allocate_scalars`, in the `atm_core_interface`
module. The `atm_allocate_scalars` routine will allocate the `scalars` and
`scalars_tend` var_arrays in the state and tend pools, respectively, but the actual
scalar constituents to appear in specific indices of these var_arrays is defined
only later by the CAM-MPAS interface layer.


mgduda:
@MiCurry I just realized that the changes in d9b2c6c might be relevant to this PR. I'll see how we can best incorporate those changes into this PR; due to the repository history leading up to d9b2c6c, I may end up just cherry-picking those changes.

mgduda:
@MiCurry Since the `develop` branch includes default initialization for the pointer components of field types (see bb729b8), perhaps the changes in d9b2c6c are not actually necessary for the changes in this PR to work correctly once merged to the `develop` branch. Does that sound right to you?

src/core_atmosphere/mpas_atm_core_interface.F
```diff
@@ -333,19 +333,178 @@ end function atm_setup_decompositions
...
 
       call atm_generate_structs(block, block % structs, block % dimensions, block % packages)
 
+      !
+      ! When MPAS-A is operating as a dycore in CAM, the scalars/scalars_tend var_arrays are
+      ! allocated by the call to atm_allocate_scalars, below. The CAM-MPAS interface layer
+      ! should have added a config, cam_pcnst, to the configs pool to indicate how many scalars
+      ! are to be allocated.
+      !
+      if (associated(block % domain % configs)) then
```
MiCurry, line 398: Whats the reason for this `if` test here? 

My question may come from my unfamiliarity with when `atm_setup_block` is called. In standalone MPAS, is `atm_setup_block` called before `block % domain % configs` is allocated/created?

src/core_atmosphere/mpas_atm_core_interface.F
```diff
@@ -333,19 +333,178 @@ end function atm_setup_decompositions
...
+         return
+      end if
+
+      timeLevs = 2
+
+      call mpas_pool_add_dimension(statePool, 'num_scalars', num_scalars)
+
+      allocate(scalarsField(timeLevs))
+
+      do i = 1, timeLevs
```
MiCurry, line 476: Can you explain to me the need to loop over `timeLevs` when creating  the `scalarsField`?

MiCurry:
> @MiCurry Since the `develop` branch includes default initialization for the pointer components of field types (see [bb729b8](https://github.com/MPAS-Dev/MPAS-Model/commit/bb729b8b4f07d3636f929f2224f88af35c069138)), perhaps the changes in [d9b2c6c](https://github.com/MPAS-Dev/MPAS-Model/commit/d9b2c6cdc2d48849862c3a5a3e8dddfb2f8cbc49) are not actually necessary for the changes in this PR to work correctly once merged to the `develop` branch. Does that sound right to you?

That sounds right. The fields initialized in bb729b8 should be initialized by the changes in d9b2c6c, so there isn't a need for bb729b8.

src/core_atmosphere/mpas_atm_core_interface.F
mgduda: Since the state variables have more than one time level (see the definition of the [state var_array](https://github.com/MPAS-Dev/MPAS-Model/blob/v7.0/src/core_atmosphere/Registry.xml#L1412)), we need to initialize all time levels of the scalars.

src/core_atmosphere/mpas_atm_core_interface.F
mgduda: Having looked through the call trees in both CAM-MPAS and stand-alone MPAS-Atmosphere, I think we don't actually need the if-test. Here's the relevant call tree in CAM-MPAS:
```
cam_init
  read_namelist
    dyn_readnl
      cam_mpas_init_phase1
        mpas_allocate_domain
  dyn_grid_init
    cam_mpas_init_phase3
      mpas_bootstrap_framework_phase1
        mpas_block_creator_finalize_block_phase1
          setup_block
```
and here the call tree in stand-alone MPAS-Atmosphere:
```
mpas_init
  mpas_allocate_domain
  setup_namelists
  mpas_bootstrap_framework_phase1
    mpas_block_creator_finalize_block_phase1
      setup_block
```
In both cases, we allocate the domain and setup namelist configs before calling `setup_block`.

I'll do some testing and update this PR with a commit to remove the if-test.

mgduda:
@MiCurry I've added an additional commit, 3cf3c33, to remove the unnecessary if-test in `atm_setup_block`.

MiCurry:
@mgduda This looks good to me. Noticing the 'In Progress' and 'Don't merge' labels, are there more changes coming to this PR?

mgduda:
@MiCurry No more changes are coming -- I've just removed the "in progress" and "Don't merge" labels.

MiCurry (approved):

PR #873, opened 2021-05-12
fEdge divide by zero fix in GM

mark-petersen:
This bug creates a divide by zero error, but **only** if latEdge is very close to zero, so it is dependent on the mesh. The exact jigsaw output varies by compiler and machine, so we did not see this error until now.


mark-petersen:
@vanroekel I think visual inspection is fine. This bug appeared with intel19 debug, QU240 on grizzly, as well as a few other machines. The fix takes care of it.

vanroekel (approved):
Approved by visual inspection.  Thanks for the catch and fix!

mark-petersen:
Transplanted to https://github.com/E3SM-Project/E3SM/pull/4304

PR #874, opened 2021-05-18
Update MPAS Frazil to match E3SM conservation

proteanplanet:
Adds a frazil mass flux correction to the ocean coupling pool in MPAS-SeaIce, and removes an artificial limit on enthalpy of new sea ice created from frazil. These changes are provided in concert with updates to E3SM PR https://github.com/E3SM-Project/E3SM/pull/4295 to conserver mass and heat in frazil coupling. Testing has taken place via a B-case E3SM simulation on Chrysalis.


mark-petersen:
This was brought into E3SM after the MPAS transplant, see https://github.com/E3SM-Project/E3SM/pull/4295#issuecomment-844350021.

PR #876, opened 2021-05-19
Correct order of argument declarations based on NAG compiler errors

mgduda:
This PR corrects the order of argument declarations in the atm_time_integration module.

The NAG compiler generates errors like the following when compiling
the atm_time_integration module:
```
  Error: mpas_atm_time_integration.F, line 2760: Implicit type for NEDGES
         detected at NEDGES@+
  Error: mpas_atm_time_integration.F, line 2764: Symbol NEDGES has already been implicitly typed
         detected at NEDGES@<end-of-statement>
````
These errors are eliminated by moving the declaration of the nEdges dummy
argument above the declarations of the other dummy arguments that are
dimensioned by nEdges.

### CODE REVIEW

**Comment by MiCurry** [comment]
This change looks fine; however, I'm having trouble compiling it with the NAG compiler. Is the atmosphere core only intended to be built with the NAG compiler within CESM at this moment?

**Comment by mgduda** [comment]
@MiCurry The physics schemes may not build with the NAG compiler, but fixing those is outside the scope of this PR (and it would require coordination with the WRF group, too). For now, you can disable physics (now possible with the merge of PR #869), and you'll also need to add a couple of search paths in `src/driver/Makefile`:
```diff
--- a/src/driver/Makefile
+++ b/src/driver/Makefile
@@ -23,7 +23,7 @@ clean:
        $(RM) $@ $*.mod
 ifeq "$(GEN_F90)" "true"
        $(CPP) $(CPPFLAGS) $(CPPINCLUDES) $< > $*.f90
-       $(FC) $(FFLAGS) -c $*.f90 $(FCINCLUDES) -I../framework -I../operators -I../core_$(CORE) -I../external/esmf_time_f90
+       $(FC) $(FFLAGS) -c $*.f90 $(FCINCLUDES) -I../framework -I../operators -I../core_$(CORE) -I../external/esmf_time_f90 -I../core_$(CORE)/diagnostics -I../core_$(CORE)/dynamics
 else
-       $(FC) $(CPPFLAGS) $(FFLAGS) -c $*.F $(CPPINCLUDES) $(FCINCLUDES) -I../framework -I../operators -I../core_$(CORE) -I../external/esmf_time_f90
+       $(FC) $(CPPFLAGS) $(FFLAGS) -c $*.F $(CPPINCLUDES) $(FCINCLUDES) -I../framework -I../operators -I../core_$(CORE) -I../external/esmf_time_f90 -I../core_$(CORE)/diagnostics -I../core_$(CORE)/dynamics
 endif
```
Apparently, the NAG compiler looks for modules that are used indirectly, so we need to provide the paths of modules in the `diagnostics` and `dynamics` subdirectories of `src/core_atmosphere`.

MiCurry:
@mgduda ah gotcha. Yes I built with the changes from #869, but good to know the Makefile needs to be updated as well.

MiCurry (approved):

Issue #344, opened 2019-08-23
Need to allocate tracerObject indexes in seaice


amametjanov:
Intel v19 raises run-time errors when passing unallocated ciceTracerObjectType's index arrays into ice_colpkg:
```
143: forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable INDEX_VERTICALAEROSOLSCONC when it is not allocated
143: 
143: Image              PC                Routine            Line        Source             
143: e3sm.exe           00000000034DFB26  Unknown               Unknown  Unknown
143: e3sm.exe           0000000000EA2749  seaice_column_mp_        9195  mpas_seaice_column.f90
143: e3sm.exe           0000000000E98BE5  seaice_column_mp_        8548  mpas_seaice_column.f90
143: e3sm.exe           0000000000D22812  seaice_column_mp_         233  mpas_seaice_column.f90
143: e3sm.exe           0000000000C5178A  seaice_initialize         102  mpas_seaice_initialize.f90
143: e3sm.exe           0000000002AA7F35  seaice_core_mp_mp         248  mpas_seaice_core.f90
143: e3sm.exe           0000000002AA6BD7  seaice_core_mp_se         138  mpas_seaice_core.f90
143: e3sm.exe           000000000089AA30  ice_comp_mct_mp_i         569  ice_comp_mct.f90
143: e3sm.exe           0000000000468EB7  component_mod_mp_         257  component_mod.F90
143: e3sm.exe           0000000000428C33  cime_comp_mod_mp_        1356  cime_comp_mod.F90
143: e3sm.exe           000000000045F97B  MAIN__                    122  cime_driver.F90
143: e3sm.exe           00000000004167A2  Unknown               Unknown  Unknown
```
for this chunk
```
    use ice_colpkg, only: &
         colpkg_init_tracer_indices

    type(ciceTracerObjectType), intent(in) :: &
         tracerObject

    call colpkg_init_tracer_indices(&
         tracerObject % index_surfaceTemperature, &
         tracerObject % index_iceEnthalpy, &
         tracerObject % index_snowEnthalpy, &
         tracerObject % index_iceSalinity, &
         tracerObject % index_brineFraction, &
         tracerObject % index_iceAge, &
         tracerObject % index_firstYearIceArea, &
         tracerObject % index_levelIceArea, &
         tracerObject % index_levelIceVolume, &
         tracerObject % index_pondArea, &
         tracerObject % index_pondDepth, &
[...]

PR #878, opened 2021-06-15
Sea ice index allocation bug

akturner:
Fix to issue https://github.com/MPAS-Dev/MPAS-Model/issues/344 for the e3sm/maint-1.0 branch.

Allocates bgc tracerObject indices to dimension 1 when not in use. Corrects a compile time error that appears with intel19 compiler versions. BFB with previous version compilers.


njeffery (approved):
This change allocates with dimension 1 vector indices for several BGC/aerosol tracers when they are not used and corrects issue #344 .  Approved.

PR #781, opened 2021-01-08
cleanup of ocean split-explicit driver

philipwjones:
This is a significant cleanup of the ocean split-explicit time integration driver to prepare for future GPU and other performance changes. Modifications include:

   - removal of block loops
   - removal of pointer retrievals for variables accessed
     through modules (config, mesh, diagnostics)
   - consolidation of remaining pointer retrievals
   - fixed misaligned indents and consistent indentation throughout
   - moved some static flags/masks to initialization and simplified
     some logic
   - eliminated a scratch pool variable and replaced with local allocate
   - reduced line lengths for better readability
   - changed some masks to reals to avoid unnecessary type conversion

Bit-for-bit on Summit with PGI and QU240. I expect it to be bit-for-bit on other machines/configurations too.

PR #879, opened 2021-06-16
Modifications of the semi-implicit barotropic mode solver code

hyungyukang:
This PR modifies codes for the semi-implicit barotropic mode solver.
Here are main changes I've made for this PR compared to the former version.

- Significant cleanup as done in #781
- Change of the linear iterative solver algorithm (reducing global and local communications)
    - from the preconditioned pipelined BiCGStab to the single-reduction BiCGStab
    - removal of local halo exchanges before preconditioning
- Change of scalar and array names used in the semi-implicit solver
    - from 'CG***' to 'SI***'
- Addition of the 'config_n_btr_si_large_iter' flag
    - The large barotropic system iteration loop is added.
    - Any positive integer. Default is 1, but less than 2.
    - Higher value of the  'config_n_btr_si_large_iter' can make the model more stable and accurate, but runtime for advance of the barotropic system will increase.
- Other changes
    - removal of the 'config_n_btr_si_outer_iter' flag
           - Instead, the outer iteration tolerance (1.0e-2) is hard coded.
    - more detailed notes on the semi-implicit solver
    - small bugs fix

Note: 
- This PR is not bit-for-bit compared to the previous semi-implicit code.
- By setting `config_btr_si_partition_match_mode = .true.`, this PR has the same pass/fail results of the nightly test suite with the split-explicit code.
  - For `config_btr_si_partition_match_mode = .false.` (default), partition tests fail, which is expected for the semi-implicit barotropic mode solver.


mark-petersen:
Transplanted to https://github.com/E3SM-Project/E3SM/pull/4354. 

@hyungyukang please go to the new pull request. There you can `git clone` the E3SM repo for that PR, then `cd components/mpas-ocean` and it is exactly like before. Please run your standard tests and comment there.

PR #880, opened 2021-06-23
Remove layerThickness from ocn_thick_surface_flux_tend

mark-petersen:
Currently, `layerThickness` is passed into `ocn_thick_surface_flux_tend` but is not allocated. This causes a failure on some compilers. Since `layerThickness` is never used in this routine, we can just remove it.


mark-petersen:
See E3SM review here: https://github.com/E3SM-Project/E3SM/pull/4352#issuecomment-867207653

PR #883, opened 2021-06-30
Enable CAM to override constants when MPAS-A is compiled as a CAM dycore

mgduda:
This PR enables CAM to override constants when MPAS-A is compiled as a CAM dycore.
    
In order to allow MPAS-Atmosphere to use identical values for constants to CAM 
when it is compiled for use as a CAM dycore, the definitions of parameters in
the mpas_constants module now depend on whether the preprocessing macro
MPAS_CAM_DYCORE is defined at build time. If this macro is not defined,
the parameters in mpas_constants are defined as before; otherwise,
the definitions in the mpas_constants module come from CAM's physconst module,
in which case the "constants" are actually variables (CAM allows for runtime
modification of constants).

Additionally, to provide derived constants that are used in the MPAS-A dycore
but not defined in CAM's physconst module, the dummy routine in
the mpas_constants module has been replaced with a routine,
mpas_constants_compute_derived, that can be called to compute constants that
are derived from others after CAM defines constants in its physconst module
at runtime.


src/framework/mpas_constants.F
```diff
@@ -1,4 +1,4 @@
-! Copyright (c) 2013,  Los Alamos National Security, LLC (LANS)
+! Copyright (c) 2013-2020  Los Alamos National Security, LLC (LANS)
```
MiCurry, line 1: I think we should update the copyright date in another PR. One reason would be to have consistent dates, for instance the [copyright date in the top level LICENSE](https://github.com/MPAS-Dev/MPAS-Model/blob/d5a656a90e0fb9a40d40547d4c4f5b832e139518/LICENSE#L1) is from `2013-2019`.

 Perhaps we could have a PR that updates the copyright date across the code base.

src/framework/mpas_constants.F
```diff
@@ -22,35 +22,57 @@ module mpas_constants
...
 !-----------------------------------------------------------------------
-   subroutine dummy()
+   subroutine mpas_constants_compute_derived()
+
+       implicit none
+
+#ifdef MPAS_CAM_DYCORE
+       rvord   = rv / rgas
+       cv      = cp - rgas
+       cvpm    = -cv / cp
```
MiCurry, line 73: Is it not possible to use variables from a use statement in the deceleration section? For example, could we duplicate line 45 to line 33 where CAM-MPAS constants are defined?

src/framework/mpas_constants.F
mgduda: That sounds reasonable. For some reason I was inspired to update the copyright date when I was making the changes to this file, but I generally agree that PRs should not make extraneous code changes.

src/framework/mpas_constants.F
mgduda: The reason why we do this isn't obvious. Essentially, some of these constants (in a physical sense, not so much a code sense) can change based on namelist options, so we can only compute some physical constants later on in the initialization process. Take `rair` (which we import as `rgas`) for example: it may be modified at runtime by the `physconst_readnl` routine, so we can only compute physical constants that depend on `rair` after that modification point. Here's what the relevant parts of the call tree look like in CAM-MPAS:
```
cam_init
   read_namelist
      physconst_readnl
   dyn_init
      mpas_constants_compute_derived
```

mgduda:
@MiCurry I'll add a new commit to this PR to explain the need for computing some constants in the `mpas_constants_compute_derived` routine. If you don't object, in the same commit, I'll revert the modification to the copyright date with a note in the commit message.

MiCurry:
@mgduda that sounds good! Thanks for the explanation on why the routine is needed.

mgduda:
@MiCurry I've just updated this PR to revert the changes to the copyright data and to add more comments in the `mpas_constants.F` code.

MiCurry (approved):

PR #889, opened 2021-07-12
Extend Makefile support to include Ampere GPU

cponder:
Only a matter of adding 5 characters to one line:
```
    117 pgi:
                      ...
    136         "FFLAGS_ACC = -Mnofma -acc -ta=tesla:cc70,cc80 -Minfo=accel" \
```
This enables code-generation to support Ampere as well as Volta.
It will not break compatibility with previous use-cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ pgi:
 	"LDFLAGS_DEBUG = -O0 -g -Mbounds -Mchkptr -Ktrap=divz,fp,inv,ovf -traceback" \
 	"FFLAGS_OMP = -mp" \
 	"CFLAGS_OMP = -mp" \
-	"FFLAGS_ACC = -Mnofma -acc -ta=tesla:cc70 -Minfo=accel" \
+	"FFLAGS_ACC = -Mnofma -acc -ta=tesla:cc70,cc80 -Minfo=accel" \
 	"CFLAGS_ACC =" \
 	"PICFLAG = -fpic" \
 	"BUILD_TARGET = $(@)" \


mgduda:
What is the oldest PGI/NVHPC compiler release that will accept "-ta=tesla:cc70,cc80"? In other words, are there PGI/NVHPC compilers that will throw an error if compute capability 8.0 is requested (perhaps because those compilers are too old to recognize this compute capability)?

cponder:
The 21.1 compiler supports the cc80 option. I don't see it documented with the 19.10 compiler but it may work also.
I don't feel any compulsion to support compilers prior to 21.x given that updates are freely available for download.
Also all the new systems are Ampere-based and I think it would be a mistake not to support them.

mgduda:
@cponder I definitely agree that we should support Ampere-based systems "out-of-the-box", and I agree that asking users to update to the newest NVHPC compilers isn't too much given that they're freely downloadable and usable. I just think it's a good idea to establish what we expect to work versus what we expect to no longer work after a PR.

cponder:
More info -- cc80 is in the 20.5 compiler but not the 20.1 compiler. So it goes back a year.
We could put some kind of variable in the Makefile, like
```
GPU_ARCH?=cc70,cc80
```
which would enable the user to override on the build command-line.
I don't think they should be using old compilers, but this would also give us some flexibility for when future GPU models come available.

mgduda:
@cponder Thanks for that additional testing with other versions. We have the PGI 20.4 compilers available, and I can confirm that they do not accept `cc80`; I get the following when trying to compile with `-ta=tesla:cc70,cc80`:
```
pgfortran-Warning-Switch -ta with unknown keyword cc80
```
I would generally agree that anyone doing work with the OpenACC-accelerated MPAS-Atmosphere code should probably be using up-to-date compilers. In that regard, perhaps we don't need to define additional variables (e.g., `GPU_ARCH`), and the changes in this PR are fine as they are. I can include some details on supported compiler versions (>= 20.5) in the merge commit message.

raghuraj19 (approved):

mgduda (approved):

PR #888, opened 2021-07-12
Fix role in mpas_pool_routines.F file

Pranay-Reddy-Kommera:
ROLE 3 includes ROLE_INTEGRATE but the role is not ROLE_INTEGRATE.

The title above should be a 1 line short summary of the pull request (i.e. what the project the PR represents is intended to do).

Enter a description of this PR. This should include why this PR was created, and what it does.

Testing and relations to other Pull Requests should be added as subsequent comments.

See the below examples for more information.
https://github.com/MPAS-Dev/MPAS/pull/930
https://github.com/MPAS-Dev/MPAS/pull/931

PR #891, opened 2021-07-13
Modify role condition in the mpas_pool_get_array_gpu routines

Pranay-Reddy-Kommera:
The mpas_pool_get_array functions allocate memory on GPU if and only if the executing MPI rank's role is ROLE_INTEGRATE. Hence, running dycore runs such as baroclinic instability test, which is a hybrid role, leads to failure. This commit addresses the issue by replacing the condition to allocate memory if the role includes ROLE_INTEGRATE, which is the case in hybrid roles, instead of an absolute role.

### DIFF
--- a/src/framework/mpas_pool_routines.F
+++ b/src/framework/mpas_pool_routines.F
@@ -4235,7 +4235,7 @@ module mpas_pool_routines
       if (associated(field)) then 
               scalar => field % scalar
 #ifdef CORE_ATMOSPHERE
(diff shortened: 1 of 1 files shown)


mgduda:
@Pranay-Reddy-Kommera It looks like the changes in this PR are a superset of those from PR #888 . In future, it's probably preferable to fix up an existing PR than to close it and create a nearly identical PR, if only because it can help to avoid clutter in our PR list.

mgduda (approved):

PR #892, opened 2021-07-14
Add a generic function to map terrestrial datasets to MPAS meshes

MiCurry:
This commit adds in a generic function which can be used map terrestrial
datasets to MPAS meshes. The function refactors and abstracts the algorithm
that was first introduced in 8e3859, which enables terrestrial datasets to be
safely parallelized without the need for a CVT partition file.

It also adds in two abstract routines which will need to be implemented by any
new dataset interpolation which wishes to use the unified function.

At this time, the unified function is not used and no implementation of the
abstract routines are created, because of this, a warning is generated for the
mpas_init_atm_static module.

Finally, it also makes `nBdyLayers` a private, module level variable so that it
can be used by both init_atm_static and init_atm_map_static_data.


src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -48,6 +53,36 @@ subroutine read_geogrid(fname, rarray, nx, ny, nz, isigned, endian, &
     end subroutine read_geogrid
  end interface
 
+ ! Abstract interface for determining if the cell, iCell, needs to be added to the stack
+ ! for processing. If an interface returns .true., it will indicate to the calling code
+ ! that the cell has not recived any mappings and needs to be processed. Returning .false.
```
mgduda, line 58: There are a few typos in the code comments that might be worth correcting:
recived
recivied
procssed
proccseed
proccssed
proccessing
cloest
conditiona
futher

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -48,6 +53,36 @@ subroutine read_geogrid(fname, rarray, nx, ny, nz, isigned, endian, &
...
+ ! will indicate that the cell has recived mappings and does not need to be procssed.
+ abstract interface
+    function interp_criteria_function(mgr, iCell)
+        import mpas_geotile_mgr_type
```
mgduda, line 62: Would it be better to change the `import` here (and also in the `interp_accumulation_function` interface) to a `use` statement?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1534,6 +1564,209 @@ subroutine init_atm_static(mesh, dims, configs)
...
+ !  routine init_atm_map_static_data
+ !
+ !> \brief   Map values of static datasets to cells
+ !> \author  Miles Curry
+ !> \date    25 January 2020
+ !> \details
+ !>  Given a initialized geotile manager object, an initialized KD tree of cell centers
+ !> (xCell, yCell, zCell), and two function pointers, this subroutine maps pixels of
+ !> a Geogrid binary format dataset (as described in the WRF User's Guide 3-41) to the
+ !> cells they reside in. The geogrid binary format is described in the WRF User's Guide
```
mgduda, line 1579: The reference to the WRF User's Guide 3-41 is repeated from the previous sentence. Since page numbers to change from one version of the  User's Guide to another, we might be safer simply referencing Chapter 3.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1534,6 +1564,209 @@ subroutine init_atm_static(mesh, dims, configs)
...
+
+    call mpas_pool_get_config(mgr % pool, 'tile_bdr', tile_bdr)
+    call mpas_pool_get_config(mgr % pool, 'tile_x', tile_nx)
+    call mpas_pool_get_config(mgr % pool, 'tile_y', tile_ny)
+    call mpas_pool_get_config(mgr % pool, 'tile_z_start', tile_z_start)
+    call mpas_pool_get_config(mgr % pool, 'tile_z_end', tile_z_end)
+    call mpas_pool_get_config(mgr % pool, 'scale_factor', scale_factor)
+
+    ! Currently, the namelist option config_supersample_factor is only for max snow albedo and
+    ! monthly climatological albedo.
```
mgduda, line 1670: Since this interpolation routine is generic, perhaps we don't need this comment that identifies specific fields, whose supersampling factor is the concern of the code that calls this routine.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1534,6 +1564,209 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                    ! lies within the cell returned by mpas_kd_search
+                    !
+                    if (bdyMaskCell(res % id) == nBdyLayers) then
+                        ! mpas_in_cell could be included in the if statement above, but calling it is expensive and most
+                        ! compilers execute both sides of a conditiona (a .and. b), even if the first conditional is
+                        ! false.
+                        if (.not. mpas_in_cell(xPixel, yPixel, zPixel, xCell(res % id), yCell(res % id), zCell(res % id), &
+                                               nEdgesOnCell(res % id), verticesOnCell(:,res % id), xVertex, yVertex, zVertex)) then
+                            !
+                            ! This pixel lays outside of res % cell and outside of the limited-area region, no futher processing is
```
mgduda, line 1736: I think "lays" should be "lies".

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1534,6 +1564,209 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                    call mpas_latlon_to_xyz(xPixel, yPixel, zPixel, sphere_radius, lat, lon)
+                    call mpas_kd_search(kdtree, (/xPixel, yPixel, zPixel/), res)
+
+                    !
+                    ! For outermost boundary cells, extra work is needed to be done to determine if a pixel actually
+                    ! lies within the cell returned by mpas_kd_search
+                    !
+                    if (bdyMaskCell(res % id) == nBdyLayers) then
+                        ! mpas_in_cell could be included in the if statement above, but calling it is expensive and most
+                        ! compilers execute both sides of a conditiona (a .and. b), even if the first conditional is
```
mgduda, line 1731: Stating that *most* compilers execute both sides of a conditional would require some substantiation. Instead, perhaps we could simply state that Fortran compilers are not required to employ short-circuit evaluation of compound conditionals?

MiCurry:
@mgduda thanks for the review. I have just pushed an update that addresses the comments that you listed above.

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1534,6 +1564,205 @@ subroutine init_atm_static(mesh, dims, configs)
...
+
+    do iCell = 1, nCells
+        !
+        ! Insure all cells receive values by loading tiles that have not received values and
+        ! pushing them onto the stack.
+        !
+        if (interp_criteria(mgr, iCell)) then
+            tile => null()
+            ierr = mgr % get_tile(latCell(iCell), lonCell(iCell), tile)
+            ierr = mgr % push_tile(tile)
```
mgduda, line 1689: Should we only try to push the tile after verifying that tile is associated (as in the if-test below this line)?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1534,6 +1564,205 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                    call mpas_latlon_to_xyz(xPixel, yPixel, zPixel, sphere_radius, lat, lon)
+                    call mpas_kd_search(kdtree, (/xPixel, yPixel, zPixel/), res)
+
+                    !
+                    ! For outermost boundary cells, extra work is needed to be done to determine if a pixel actually
+                    ! lies within the cell returned by mpas_kd_search
+                    !
+                    if (bdyMaskCell(res % id) == nBdyLayers) then
+                        ! mpas_in_cell could be included in the if statement above, but calling it is expensive and the Fortran
+                        ! standard does require compilers to short circuit compound conditionals.
```
mgduda, line 1728: Should "Fortran standard does require" be "Fortran standard does not require"?

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1534,6 +1564,205 @@ subroutine init_atm_static(mesh, dims, configs)
...
+                    if (res % id <= nCellsSolve) then
+                        all_pixels_mapped_to_halo_cells = .false.
+                    end if
+                end do
+            end do
+
+            tile % is_processed = .true.
+
+            !
+            ! If a single pixel value maps to an owned cell (i.e. <= nCellsSolve) then
```
mgduda, line 1753: It might be clearer here to say "If at least one pixel" rather than "If a single pixel value", as the latter statement could be interpreted to mean "If exactly one pixel value".

MiCurry: I agree and fixed. Thanks for the suggestion.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: That is correct. Fixed. Thank you!

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: Yes! Fixed.

MiCurry:
@mgduda I just pushed up another commit which addresses your last review comments you made on this branch.

PR #902, opened 2021-07-29
Modified MPAS_to_physics to run on CPUs during an OpenACC run.

Pranay-Reddy-Kommera:
This is the 3rd PR submitted in a pool of approximately 12 PRs.
The code changes include appropriate role checks and data transfers to/from CPU and GPU. The code changes allow the 'MPAS_to_physics' subroutine to execute on the CPU while any other dynamics/physics schemes are executing on the GPU. 'MPAS_to_physics' is duplicated as 'MPAS_to_physics_gpu', where the former is used in radiation role execution and latter in dynamics role execution. These temporary data transfer changes are required to keep track of any intermediate bugs originating as part of the Physics porting efforts. As the physics porting efforts progress, these changes will be removed if the data transfers become redundant.

Code Execution Status:

1. The code compiles and executes successfully on CPU.
2. Code compiles successfully with Dynamics on GPU and Physics on CPU. Code is not executed as the arrays/variables are not yet updated appropriately (will result in NaNs or incorrect answers) in the Physics. The code should be working by the end of all the PRs.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver.F
@@ -226,7 +226,7 @@
     ! For all timesteps, we transfer MPAS state to physics
     !
 !$OMP PARALLEL DO
-    if (mpas_cpl%role_includes(ROLE_RADIATION)) then
+    if (mpas_cpl%role_is(ROLE_RADIATION)) then
     do thread=1,nThreads
        call MPAS_to_physics(block%configs,mesh,state,time_lev,diag,diag_physics, &
                             mpas_cpl, cells_to_rad_handle, cells_from_rad_handle, &
--- a/src/core_atmosphere/physics/mpas_atmphys_driver.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver.F
@@ -226,11 +226,19 @@
     ! For all timesteps, we transfer MPAS state to physics
     !
 !$OMP PARALLEL DO
+    if (mpas_cpl%role_includes(ROLE_RADIATION)) then
     do thread=1,nThreads
        call MPAS_to_physics(block%configs,mesh,state,time_lev,diag,diag_physics, &
                             mpas_cpl, cells_to_rad_handle, cells_from_rad_handle, &
                             cellSolveThreadStart(thread), cellSolveThreadEnd(thread))
     end do
+    else
+    do thread=1,nThreads
+       call MPAS_to_physics_gpu(block%configs,mesh,state,time_lev,diag,diag_physics, &
+                            mpas_cpl, cells_to_rad_handle, cells_from_rad_handle, &
+                            cellSolveThreadStart(thread), cellSolveThreadEnd(thread))
+    end do
+    end if
 !$OMP END PARALLEL DO
 
     !call to cloud scheme:
--- a/src/core_atmosphere/physics/mpas_atmphys_interface.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_interface.F
@@ -316,8 +316,8 @@
  call mpas_pool_get_field(state,'scalars',scalarsField,time_lev)
 
  if (l_radtlw .or. l_radtsw) then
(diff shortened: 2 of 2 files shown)


cenamiller (approved):

src/core_atmosphere/physics/mpas_atmphys_driver.F
```diff
@@ -226,11 +226,19 @@ subroutine physics_driver(domain,itimestep,xtime_s,mpas_cpl,cells_to_rad_handle,
     ! For all timesteps, we transfer MPAS state to physics
     !
 !$OMP PARALLEL DO
+    if (mpas_cpl%role_includes(ROLE_RADIATION)) then
```
mgduda, line 229: If we have MPI ranks doing both radiation and integration, this conditional would lead to the `MPAS_to_physics` routine being called all of the time (and the `MPAS_to_physics_gpu` routine would never be called). Is this what we want to happen?

cponder:
Code-changes look good from my end.
Also I ran the `7.x/benchmark/30km` case with these PRs
```893 895 896 902 903 904 905 906 907 908 909 910 911```
and didn't see any failure issue.

**Review by cponder — APPROVED** [review]

**Comment by raghuraj19** [comment]
Looks good!

**Review by raghuraj19 — APPROVED** [review]

File: src/core_atmosphere/physics/mpas_atmphys_driver.F (hunk 1 above)
Comment (raghuraj19) [reply src/core_atmosphere/physics/mpas_atmphys_driver.F:229]: If each rank is running radiation and integration, then shouldn't it  run every timestep? Did I misunderstand this? Or in ROLE03, the ranks iterate through this twice for each timestep?

Comment (mgduda) [reply src/core_atmosphere/physics/mpas_atmphys_driver.F:229]: If I'm thinking about this correctly, if we have MPI ranks performing both ROLE_INTEGRATION and ROLE_RADIATION roles, then the if-test in the code, below, will always be true; consequently, the `MPAS_to_physics` routine will always be called by all MPI tasks, and the `MPAS_to_physics_gpu` routine will never be called by any tasks. Since the [MPAS_to_physics_gpu](https://github.com/MPAS-Dev/MPAS-Model/blob/5a7bc4186cf6b377545f08a2784a15fae2fb26b1/src/core_atmosphere/physics/mpas_atmphys_interface.F#L523) routine is being added in this PR, my concern is that there are situations in which this routine would never be called. If we're doing a CPU-only run, then I think the code is correct, but what about future situations where the radiation and integration are both running on GPUs?
```fortran
    if (mpas_cpl%role_includes(ROLE_RADIATION)) then
    do thread=1,nThreads
       call MPAS_to_physics(block%configs,mesh,state,time_lev,diag,diag_physics, &
                            mpas_cpl, cells_to_rad_handle, cells_from_rad_handle, &
    end do
    else
    do thread=1,nThreads
       call MPAS_to_physics_gpu(block%configs,mesh,state,time_lev,diag,diag_physics, &
                            mpas_cpl, cells_to_rad_handle, cells_from_rad_handle, &
    end do
    end if
```

mgduda: Actually, since the corresponding block of code in the `atmosphere/v6.x-openacc` branch contains the same logic, perhaps we don't need to worry about this right now? I may have confused myself with another discussion where the possibility of running both radiation and integration on the GPUs was mentioned (specifically, [this discussion](https://github.com/MPAS-Dev/MPAS-Model/pull/893#discussion_r677933996) on PR #893).

raghuraj19: I think you do have a valid point. But yes, we will need make a separate PR when GPU is doing radiation. We can address this at that time. But you can accept this PR with a note that this will require future amendments.

mgduda: Addressing this later (at such time as we have GPU-enabled radiation) sound fine to me. Thanks!

mgduda (approved):

PR #913, opened 2021-08-05
Add empty mpas_atm_dynamics_init and mpas_atm_dynamics_finalize routines

mgduda:
This merge adds empty `mpas_atm_dynamics_init` and `mpas_atm_dynamics_finalize`
routines to the MPAS-Atmosphere core.

The `mpas_atm_dynamics_init` and `mpas_atm_dynamics_finalize` routines provide a way
for data structures and modules that are needed across all MPAS-Atmosphere
dynamics integration steps to be initialized and finalized. Currently these two
routines contain no executable code, but they serve as placeholders.

The `mpas_atm_dynamics_init` routine is called from the `atm_core_init` routine once
before the first dynamics step, and the `mpas_atm_dynamics_finalize` routine is
called from the `atm_core_finalize` routine after the last dynamics step.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -83,6 +83,56 @@ module atm_time_integration
    contains
 
 
+   !----------------------------------------------------------------------------
+   !  routine MPAS_atm_dynamics_init
+   !
+   !> \brief Initialize the dynamics
+   !> \date  28 July 2021
+   !> \details
+   !>  Prepare the dynamics component of MPAS-Atmosphere for time integration.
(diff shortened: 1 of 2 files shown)


mgduda:
There are two future uses of the new `mpas_atm_dynamics_init` and `mpas_atm_dynamics_finalize` routine that I had planned:
1. To set up grouped halo exchanges, which needs to be done once before the first dynamics step
2. To allocate scratch tendency fields that need to persist across dynamics steps and that are used only in CAM-MPAS

MiCurry (approved):

MiCurry:
This looks good to me. I'm confused by grouped halo exchanges. Can you explain to me what you mean by that? (We can discuss this next week at our weekly meeting on Thursday).

PR #915, opened 2021-08-06
Add new atmosphere initialization case, 13, to create 3-d grids for CAM-MPAS

mgduda:
This PR adds a new atmosphere init case to create 3-d grids for CAM-MPAS

The new initialization case (config_init_case = 13) requires a text file with
a list of zeta levels to be provided by the existing namelist option
`config_specified_zeta_levels`, and it optionally requires an input stream,
`cam_topo`, with a surface geopotential field, `PHIS`. If the `cam_topo` stream
is not present in the `streams.init_atmosphere` file, the resulting 3-d grid
will have zero topography everywhere.


src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -5724,6 +5754,630 @@ subroutine init_atm_case_lbc(timestamp, block, mesh, nCells, nEdges, nVertLevels
...
+         call mpas_log_write('')
+     end if
+
+
+      !
+      ! Set terrain field based on PHIS
+      !
+      call mpas_pool_get_array(mesh, 'ter', ter)
+      call mpas_pool_get_array(mesh, 'PHIS', PHIS)
+
```
MiCurry, line 5991: I'd request that we have an additional warning (Or perhaps an error?) that acts as a check that the file in the `cam_topo` stream is defined with `nCells` or not. 

When the `cam_topo` stream file is dimensioned with something other than `nCells`, a CAM-MPAS grid is produced with the `ter` being all zeros, and no warning is produced. This can be quite a perplexing to a user, as renaming a dimension in a file is not standard in any other MPAS workflows.

Thus, if a `cam_topo` stream is present *and* PHIS is all zero, we can assume that the `cam_topo` stream is not dimensioned by `nCells`, and thus either produce an error or a warning alerting the user that the ter field will be set to zero.

``` Fortran
if (MPAS_stream_mgr_stream_exists(stream_manager, 'cam_topo') .and. all(PHIS(:) == 0.0)) then
    ...
end if
```

MiCurry (changes requested):

src/core_init_atmosphere/mpas_init_atm_cases.F
mgduda: I do agree that the code should provide some indication to the user when the `PHIS` field could not be read from the "cam_topo" stream; however, I think the proper place to handle this would be the stream manager (or a lower level in the MPAS I/O stack). I was surprised to find in a test with `PHIS` dimensioned by `ncol` rather than by `nCells` that there were no warnings or errors at all from the I/O stack. I'll take a look in the I/O code to see how difficult it might be to catch the situation where the field to be read is dimensioned differently in the file compared with its `Registry.xml` definition. I would be fine with waiting to merge this PR until we can fix up the I/O code with a separate PR.

mgduda: PR #923 adds a critical error that will be triggered, e.g., when `PHIS` is dimensioned by `ncol` rather than `nCells` in the `cam_topo` stream.

PR #918, opened 2021-08-11
Changed array-handling of u & v to GPU-compatible form in mpas_atm_time_integration.F

cponder:
In the file
```
src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```
the `u` & `w` arrays are used on both the CPU & GPU and need to be handled using the convention
```
   9308             call mpas_pool_get_array_gpu(state, 'w', w, 2)
   9309             call mpas_pool_get_array_gpu(state, 'u', u, 2)
```
This was an oversight inherited from the` 6.x-openacc` branch
```
   8289 !call to mpas pool get array to get the pointer to the array that needs to be
   8290 !updated on the host
   8291             call mpas_pool_get_array(state, 'w', w, 2)
   8292             call mpas_pool_get_array(state, 'u', u, 2)
   8293 
   8294 !add the variables in the "update host" clause below to update the data on host
   8295         !$acc update host(w, &
   8296         !$acc u)
```
and needs to be fixed here. (Note that the line-numbers are different in these two versions of the file).

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -9305,8 +9305,8 @@ if (config_horiz_mixing == "2d_smagorinsky") then
 
 !call to mpas pool get array to get the pointer to the array that needs to be
 !updated on the host
-            call mpas_pool_get_array(state, 'w', w, 2)
-            call mpas_pool_get_array(state, 'u', u, 2)
+            call mpas_pool_get_array_gpu(state, 'w', w, 2)
+            call mpas_pool_get_array_gpu(state, 'u', u, 2)
 
 !add the variables in the "update host" clause below to update the data on host
         !$acc update host(w, &


cponder:
The PCAST runs gave a lot of messages like these:
```
  12: PCAST Float w in function mpas_update_gpu_data_on_host, /workdir/Source/src/core_atmosphere/dynamics/mpas_atm_time_integration.F:9312
  12:   idx: 0 FAIL ABS  act: 0.00000000e+00 exp: -nan dif: nan
  12:   idx: 1 FAIL ABS  act: 1.17938686e-03 exp: -nan dif: nan
  12:   idx: 2 FAIL ABS  act: 2.67154397e-03 exp: -nan dif: nan
  12:   idx: 3 FAIL ABS  act: 4.49581956e-03 exp: -nan dif: nan
  12:   idx: 4 FAIL ABS  act: 6.66643353e-03 exp: -nan dif: nan
  12:   idx: 5 FAIL ABS  act: 9.12658405e-03 exp: -nan dif: nan
  12:   idx: 6 FAIL ABS  act: 1.17650982e-02 exp: -nan dif: nan
  12:   idx: 7 FAIL ABS  act: 1.43989865e-02 exp: -nan dif: nan
  12:   idx: 8 FAIL ABS  act: 1.65729150e-02 exp: -nan dif: nan
  12:   idx: 9 FAIL ABS  act: 1.78317521e-02 exp: -nan dif: nan
  12:   idx: 10 FAIL ABS  act: 1.76396854e-02 exp: -nan dif: nan
  12:   idx: 11 FAIL ABS  act: 1.66416094e-02 exp: -nan dif: nan
  12:   idx: 12 FAIL ABS  act: 1.49542540e-02 exp: -nan dif: nan
  12:   idx: 13 FAIL ABS  act: 1.25839049e-02 exp: -nan dif: nan
  12:   idx: 14 FAIL ABS  act: 9.23908874e-03 exp: -nan dif: nan
```
The fix here does not make the messages go away.
Note, though, that the use of
```
   9311 !add the variables in the "update host" clause below to update the data on host
   9312         !$acc update host(w, &
   9313         !$acc u)
```
indicates that the Host copy of the array hadn't had a value assigned, instead we're computing new values on the GPU and then using it to update the Host copy. So the Host version is listed as holding un-initialized values here, while the GPU version holds the correctly computed values, which is what we should expect.
Maybe we should file a Enhancement Request that PGI/PCAST report this case differently.

mgduda (approved):
CPU-only test runs are successful.

raghuraj19 (approved):

supreethms1809 (approved):

cponder:
Similar changes to the file were made here
https://github.com/MPAS-Dev/MPAS-Model/pull/921
which fixed a different problem.

PR #921, opened 2021-08-13
Added scalars_1 and scalars_2 variables to acc present clause in atm_rk_integration_setup_work

Pranay-Reddy-Kommera:
Both the scalars_1 and scalars_2 are on GPU and are needed to be included in the '!$acc data present' so as to check their presence on the GPU.


mgduda (approved):

supreethms1809 (approved):
call mpas_pool_get_array_gpu(state, 'scalars', scalars_1, 1) and call mpas_pool_get_array_gpu(state, 'scalars', scalars_2, 2) ensures the data is moved to GPU. This change look good to me.

raghuraj19 (approved):

PR #922, opened 2021-08-20
Adjusted loop-bounds in module_bl_gwdo.F to prevent illegal index operation.

cponder:
In the file
```
src/core_atmosphere/physics/physics_wrf/module_bl_gwdo.F
```
contrast this loop between the 6.x & 7,x versions
```
569  do k = kts+1,kpblmax-1                               |  do k = kpblmin,kpblmax                                  539
570    do i = its,ite                                          do i = its,ite                                        540
571      if (k .lt. kbl(i)) then                                 if (k .lt. kbl(i)) then                             541
572        rdelks = (prsl(i,k)-prsl(i,k+1)) * delks1(i)            rdelks    = (prsl(i,k)-prsl(i,k+1)) * delks1(i)   542
573        bnv2(i,1) = bnv2(i,1) + bnv2(i,k) * rdelks              bnv2(i,1) = bnv2(i,1) + bnv2(i,k) * rdelks        543
574        usqj(i,1) = usqj(i,1) + usqj(i,k) * rdelks              usqj(i,1) = usqj(i,1) + usqj(i,k) * rdelks        544
575      endif                                                   endif                                               545
576    enddo                                                   enddo                                                 546
577  enddo                                                   enddo                                                   547
```
The `kpblmax` boundary is set here in the 7.x code
```
    133    if(present(p_top) .and. present(znu) .and. present(znw)) then
    134       do k = kts,kte
    135         if (znu(k).gt.0.6) kpblmax = k + 1
    136       enddo
    137    else
    138       kpblmax = kte
    139    endif
```
which could end up with a value of `kte` (or possibly `(kte+1)`).
The array `prsl` is only sized-out to `kte` in the second dimension in the subroutine declaration here
``` 
    280    real, dimension(its:ite,kts:kte)  , intent(in   ) :: prsl, del 
```
so the index operation `prsl(i,k+1)` here
```
    539    do k = kpblmin,kpblmax                                                
    540      do i = its,ite                                                    
    541        if (k .lt. kbl(i)) then
    542          rdelks    = (prsl(i,k)-prsl(i,k+1)) * delks1(i)
```
can be out-of-bounds.

### DIFF
--- a/src/core_atmosphere/physics/physics_wrf/module_bl_gwdo.F
+++ b/src/core_atmosphere/physics/physics_wrf/module_bl_gwdo.F
@@ -135,7 +135,7 @@ contains
         if (znu(k).gt.0.6) kpblmax = k + 1
       enddo
    else
-      kpblmax = kte
+      kpblmax = kte - 1
    endif
 !
    do j = jts,jte


cponder:
At the very least, I would expect this to cause some slow data-corruption with the CPU-compiled version of the code.
We saw this trigger a segfault when running the OpenACC versions of some of the Physics models.
Likely due to stricter enforcement of array boundaries on the GPU than with the CPU-compiled code.

raghuraj19 (approved):

raghuraj19:
@mgduda

mgduda:
I'm almost totally unfamiliar with this code, but would an alternative fix be to set `kpblmax` to `kte-1` on line 138?
```fortran
    133    if(present(p_top) .and. present(znu) .and. present(znw)) then
    134       do k = kts,kte
    135         if (znu(k).gt.0.6) kpblmax = k + 1
    136       enddo
    137    else
    138       kpblmax = kte
    139    endif
```
My guess is that the loop on lines 134 - 136 is used by the WRF model and will never result in a value of `kte+1` for `kpblmax` because the 0.6 eta level should be somewhere in the middle of the vertical domain. Therefore, it shouldn't be unreasonable to set `kpblmax` one level below `kte` rather than at `kte` in MPAS, given that the value that `kpblmax` is set to in WRF is probably lower still. Does that make sense?

raghuraj19:
Yes, this does. We will rework this PR shortly.

cponder:
Ok I applied a second commit here that un-does the first and applies the change to line 138.

mgduda:
> Ok I applied a second commit here that un-does the first and applies the change to line 138.
> I expect that "quashing" the two commits would have the same effect as ignoring the first one, I don't know how you intend to handle that.

I can squash the two commits into a single commit and push the result back to your `atmosphere/gwdo-fix` branch if you don't object.

cponder:
Yes please do that.

PR #923, opened 2021-08-23
Stop with a critical error if input stream is missing dims for any of its fields

mgduda:
This PR introduces a critical error if an input stream is missing any dimensions
needed by fields to be read from the stream. Prior to this PR, the phase 2
bootstrapping code would only print a non-error message for missing dimensions.

Note that fields that are deactivated by packages will not trigger the error
condition added in this PR, since deactivated fields are excluded from
streams by logic in the `MPAS_stream_mgr_add_field` routine.


mgduda:
@MiCurry This PR addresses your request in PR #915 for an error when an input stream hasn't defined a dimension for one of its fields.

PR #925, opened 2021-08-30
Fix lagged radiation time

anotherjohnwong:
The original implementation is missing an entire radiation step, and solar zenith angle is a whole 1 radt behind after the first radt. This PR corrects the solar zenith angles by offsetting the xtime going into shortwave radiation driver forward by 1 radt, while adding an extra condition to perform an additional radiation step at +1 dynamic dt.

A shortcoming of this implementation is that when dt and radt are the same, the 2nd and 3rd timesteps would still have the wrong solar zenith angles. Since this use case is extremely rare, a warning message is thrown instead.

Screen Shot 2021-08-30 at 10 56 11 AM

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_sw.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_sw.F
@@ -939,9 +939,11 @@
 !local pointers:
  logical,pointer:: config_o3climatology
  character(len=StrKIND),pointer:: radt_sw_scheme
+ real(kind=RKIND),pointer:: dt
 
 !local variables:
  integer:: o3input
+ logical:: is_lagged
  real(kind=RKIND):: radt,xtime_m
 
 !-----------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/core_atmosphere/physics/mpas_atmphys_driver_radiation_sw.F
```diff
@@ -949,13 +950,34 @@ subroutine driver_radiation_sw(itimestep,configs,mesh,state,time_lev,diag_physic
...
+ call mpas_pool_get_config(configs,'config_dt'            ,dt                  )
 
  xtime_m = xtime_s/60.
 
+ if (itimestep == 1 .and. dt_radtsw == dt .and. &
+     (mpas_cpl%role_is(ROLE_INTEGRATE) .or. mpas_cpl%role_is(ROLE_RADIATION))) then
+     ! When running lagged radiation with dt == radt, the 2nd and 3rd timesteps would receive
+     ! SW radiation computed at 1 radt behind.
+     call mpas_log_write('')
+     call mpas_log_write('WARNING: Lagged radiation with dt==radt is not fully supported.')
```
mgduda, line 962: You can get a "proper" warning message by changing this line to:
```fortran
call mpas_log_write('Lagged radiation with dt==radt is not fully supported.', messageType=MPAS_LOG_WARN)
```
Using `messageType` will ensure that the warning message is counted towards the total number of warning messages printed at the end of the log file, and it will ensure that -- if we make adjustments to the style of warning messages -- this warning will always be consistent in style with other warning messages.

src/core_atmosphere/physics/mpas_atmphys_driver_radiation_sw.F
```diff
@@ -949,13 +950,34 @@ subroutine driver_radiation_sw(itimestep,configs,mesh,state,time_lev,diag_physic
...
+
+!... convert the radiation time_step to minutes:
+ radt = dt_radtsw/60.
+
+!... Adjust effective solar time for lagged radiation
+ if (itimestep == 2) then
+    xtime_m = xtime_m - dt/60. + radt ! Special radiation step after init
+ elseif (itimestep > 2) then
+    xtime_m = xtime_m + radt
+ end if
```
mgduda, line 979: Should this code only be executed for lagged-radiation runs?

mgduda (commented):
Are there additional code changes needed for long-wave radiation? Since the `mpas_atmphys_driver_radiation_sw.F` file has been modified, I'm wondering whether similar changes are needed in `mpas_atmphys_driver_radiation_lw.F`?

src/core_atmosphere/physics/mpas_atmphys_driver_radiation_sw.F
anotherjohnwong: Yes, good catch. I'll correct the code.

As far as LW goes, RRTMG_LW does not seem to be dependent on time.

src/core_atmosphere/physics/mpas_atmphys_driver_radiation_sw.F
anotherjohnwong: Thank you for the suggestion. Missed this one last time I looked at the PR. I've pushed a new commit to implement this suggestion as well as a prior commit to address the other issue raised.

mgduda (approved):

mgduda:
@anotherjohnwong Before I merge this, I would at least like to wrap the commit messages to 80 columns; with some default pagers used by git (e.g., `less -S`), running `git log` doesn't wrap lines, and causes commit messages to be truncated; for example, when my terminal is only 108 columns wide, this is what I see:
```
commit 804eb4ddbba14f1c13f2e1a0c7fe9b284c790858
Author: John Wong <john.wong@weather.com>
Date:   Mon Sep 27 15:44:52 2021 -0400

    Changed lagged radiation warning message to use messageType
    
    Using messageType will ensure that the warning message is counted towards the total number of warning me

commit ea92ef99fb2fbe77bdaae5b769d88019f953a2df
Author: John Wong <john.wong@weather.com>
Date:   Mon Sep 20 10:57:34 2021 -0400

    Bug fix to adjust xtime for radiation only when lagging radiation
    
    This is a bug fix to make sure radiation time is only lagged when lagged radiation is active as indicate

commit 01a93bdb2eac4bdf5b84f4021b275c272a5f8314
Author: John Wong <john.wong@weather.com>
Date:   Thu Aug 26 17:39:12 2021 -0400

    Warning message re: lagged radiation
    
    Adding warning message regarding the shortcoming of the current implementation of lagged radiation. As o

```
Since the later commits on this branch are fixup anyway, I'd also propose simply squashing those fix-up commits. Would you have any objection to this?

anotherjohnwong:
No objections at all!

mgduda:
@anotherjohnwong Thanks!

PR #897, opened 2021-07-28
Added OpenACC directives to Convection and New Tiedtke model, for GPU execution.

cponder:
Added OpenACC directives from 6.x-openacc version to manage Host <-> GPU memory movement, to enable convection model to run on the GPU.

PR #926, opened 2021-09-02
Update the common physics variables to GPU.

Pranay-Reddy-Kommera:
In this PR, all the physics variables common across schemes are updated on the GPUs. As we have started to port each physics scheme, the ported physics scheme would require the common physics variables on the GPU. All the physics variables in MPAS_to_physics_gpu are needed to be on both the CPU and GPU. So, all the common variables are updated.

PR #927, opened 2021-09-04
Added OpenACC directives to Convection and New Tiedtke model, for GPU execution

cponder:
This is a re-working of #897 to be compatible with the host/device mediation in #909 & #910.
I closed #897 in favor of this one.
I ran a test and it looked ok. It does depend on the fix on #926.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
@@ -120,16 +120,21 @@
  if(.not.allocated(pratec_p)   ) allocate(pratec_p(ims:ime,jms:jme)          )
  if(.not.allocated(raincv_p)   ) allocate(raincv_p(ims:ime,jms:jme)          )
 
- do i = its,ite
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(2)
  do j = jts,jte
+ do i = its,ite
     pratec_p(i,j) = 0._RKIND
     raincv_p(i,j) = 0._RKIND
  enddo
  enddo
+!$acc end parallel
 
- do i = its,ite
- do k = kts,kte
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(3)
  do j = jts,jte
+ do k = kts,kte
+ do i = its,ite
     rthcuten_p(i,k,j) = 0._RKIND
     rqvcuten_p(i,k,j) = 0._RKIND
     rqccuten_p(i,k,j) = 0._RKIND
@@ -137,6 +142,7 @@
  enddo
  enddo
  enddo
+!$acc end parallel
 
  convection_select: select case(convection_scheme)
 
@@ -205,22 +211,28 @@
        if(.not.allocated(rucuten_p)    ) allocate(rucuten_p(ims:ime,kms:kme,jms:jme)    )
        if(.not.allocated(rvcuten_p)    ) allocate(rvcuten_p(ims:ime,kms:kme,jms:jme)    )
 
-       do i = its,ite
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(2)
        do j = jts,jte
+       do i = its,ite
           hfx_p(i,j)   = 0._RKIND
           qfx_p(i,j)   = 0._RKIND
           xland_p(i,j) = 0._RKIND
        enddo
        enddo
+!$acc end parallel
 
-       do i = its,ite
-       do k = kts,kte
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(3)
        do j = jts,jte
+       do k = kts,kte
+       do i = its,ite
           rucuten_p(i,k,j)     = 0._RKIND
           rvcuten_p(i,k,j)     = 0._RKIND
        enddo
        enddo
        enddo
+!$acc end parallel
 
        cu_tiedtke_select: select case(convection_scheme)
 
@@ -242,14 +254,17 @@
              if(.not.allocated(rqvften_p)) allocate(rqvften_p(ims:ime,kms:kme,jms:jme))
              if(.not.allocated(rthften_p)) allocate(rthften_p(ims:ime,kms:kme,jms:jme))
 
-             do i = its,ite
-             do k = kts,kte
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(3)
              do j = jts,jte
+             do k = kts,kte
+             do i = its,ite
                 rqvften_p(i,k,j) = 0._RKIND
                 rthften_p(i,k,j) = 0._RKIND
              enddo
              enddo
              enddo
+!$acc end parallel
 
            case default
 
@@ -384,7 +399,8 @@
  convection_select: select case(convection_scheme)
 
     case ("cu_kain_fritsch")
-       call mpas_pool_get_array(diag_physics,'nca',nca)
+       call mpas_pool_get_array_gpu(diag_physics,'nca',nca)
+!$acc update host(nca)
        if(.not. config_do_restart) then
           do iCell = 1, nCells
              nca(iCell) = -100._RKIND
@@ -456,12 +472,14 @@
  curr_secs   = -1
  cudtacttime = -1
  adapt_step_flag = .false.
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(2)
  do j = jts, jte
  do i = its, ite
     cu_act_flag(i,j) = .false.
  enddo
  enddo
-
+!$acc end parallel
 !... initialization of initflag needed in the calls cu_tiedtke and cu_ntiedtke:
  initflag = 1
  if(config_do_restart .or. itimestep > 1) initflag = 0
@@ -508,6 +526,7 @@
        else
           ktau = itimestep + 1
        endif
+!$acc update host(cu_act_flag)
 
        call mpas_timer_start('Kain-Fritsch')
        call  kf_eta_cps ( &
@@ -543,6 +562,7 @@
        call mpas_timer_stop('Kain-Fritsch')
 
     case("cu_tiedtke")
+!$acc update host(cu_act_flag)
        call mpas_timer_start('Tiedtke')
        call cu_tiedtke( &
              pcps        = pres_hyd_p    , p8w       = pres2_hyd_p   ,               &
@@ -652,20 +672,30 @@
  call mpas_pool_get_array_gpu(tend_physics,'rqvcuten',rqvcuten)
  call mpas_pool_get_array_gpu(tend_physics,'rqccuten',rqccuten)
  call mpas_pool_get_array_gpu(tend_physics,'rqicuten',rqicuten)
-!$acc update host(cuprec, raincv, rthcuten, rqvcuten, rqccuten, rqicuten)
 
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(2)
  do j = jts,jte
  do i = its,ite
     raincv_p(i,j) = raincv(i)
     pratec_p(i,j) = cuprec(i)
-    do k = kts,kte
-       rthcuten_p(i,k,j) = rthcuten(k,i)
-       rqvcuten_p(i,k,j) = rqvcuten(k,i)
-       rqccuten_p(i,k,j) = rqccuten(k,i)
-       rqicuten_p(i,k,j) = rqicuten(k,i)
-    enddo
  enddo
  enddo
+!$acc end parallel
+
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(3)
+ do j = jts,jte
+ do i = its,ite
+ do k = kts,kte
+    rthcuten_p(i,k,j) = rthcuten(k,i)
+    rqvcuten_p(i,k,j) = rqvcuten(k,i)
+    rqccuten_p(i,k,j) = rqccuten(k,i)
+    rqicuten_p(i,k,j) = rqicuten(k,i)
+ enddo
+ enddo
+ enddo
+!$acc end parallel
 
  convection_select: select case(convection_scheme)
 
@@ -790,26 +820,34 @@
        call mpas_pool_get_array_gpu(tend_physics,'rqvdynten',rqvdynten)
        call mpas_pool_get_array_gpu(tend_physics,'rucuten'  ,rucuten  )
        call mpas_pool_get_array_gpu(tend_physics,'rvcuten'  ,rvcuten  )
-!$acc update host(xland, hfx, qfx, rqvblten, rqvdynten, rucuten, rvcuten)
 
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(2)
        do j = jts,jte
        do i = its,ite
           xland_p(i,j) = xland(i)
           hfx_p(i,j)   = hfx(i)
           qfx_p(i,j)   = qfx(i)
        enddo
+       enddo
+!$acc end parallel
 
-       do k = kts,kte
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(3)
+       do j = jts,jte
        do i = its,ite
+       do k = kts,kte
           rucuten_p(i,k,j) = rucuten(k,i)
           rvcuten_p(i,k,j) = rvcuten(k,i)
        enddo
        enddo
        enddo
+!$acc end parallel
 
        cu_tiedtke_select: select case(convection_scheme)
 
           case ("cu_tiedtke")
+!!$acc update host(dx_p, xland_p, hfx_p, qfx_p, cubot_p, cutop_p, rucuten_p, rvcuten_p)
              do j = jts,jte
              do k = kts,kte
              do i = its,ite
@@ -818,7 +856,7 @@
              enddo
              enddo
              enddo
- 
+
           case("cu_ntiedtke")
              call mpas_pool_get_config(configs,'config_len_disp',len_disp)
              call mpas_pool_get_array_gpu(mesh,'meshDensity',meshDensity)
@@ -829,21 +867,28 @@
              call mpas_pool_get_array_gpu(tend_physics,'rthblten'  ,rthblten  )
              call mpas_pool_get_array_gpu(tend_physics,'rthratenlw',rthratenlw)
              call mpas_pool_get_array_gpu(tend_physics,'rthratensw',rthratensw)
-!$acc update host(meshDensity, rqvdynten, rqvblten, rthdynten, rthblten, rthratenlw, rthratensw)
-             
+
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(2)
              do j = jts,jte
              do i = its,ite
                 dx_p(i,j) = len_disp / meshDensity(i)**0.25
              enddo
+             enddo
+!$acc end parallel
 
-             do k = kts,kte
+!$acc parallel vector_length(32)
+!$acc loop gang vector collapse(3)
+             do j = jts,jte
              do i = its,ite
+             do k = kts,kte
                 rqvften_p(i,k,j) = rqvdynten(k,i) + rqvblten(k,i)
                 rthften_p(i,k,j) = (rthdynten(k,i) + rthblten(k,i) &
-                                 + rthratenlw(k,i) + rthratensw(k,i)) * pi_p(i,k,j) 
+                                 + rthratenlw(k,i) + rthratensw(k,i)) * pi_p(i,k,j)
              enddo
              enddo
              enddo
+!$acc end parallel
 
           case default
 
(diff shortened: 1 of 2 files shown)


mgduda (changes requested):
I'm getting a couple of errors when building with the GNU 11.1.0 compilers:
```
module_cu_ntiedtke.F:3327:12:

 3327 |         if ( flag .eq. .true. ) then
      |            1
Error: Logicals at (1) must be compared with .eqv. instead of .eq.
module_cu_ntiedtke.F:1961:12:

 1961 |       if  ( flag .eq. .false. ) exit
      |            1
Error: Logicals at (1) must be compared with .eqv. instead of .eq.
```
My preference would be to simply use `if (flag) then` and `if (.not. flag) exit` rather than comparing against `.true.` and `.false.`.

cponder:
Ok -- I pushed a second commit, that made these edits.
Do please squash these together for the merge.

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
```diff
@@ -2310,17 +2612,20 @@ subroutine cuascn &
...
+!!!     is = is + klab(jl,jk+1)
+        is = klab(jl,jk+1)
         if ( klab(jl,jk+1) == 0 ) klab(jl,jk) = 0
         if ( (ldcum(jl) .and. klab(jl,jk+1) == 2) .or.   &
              (ktype(jl) == 3 .and. klab(jl,jk+1) == 1) ) then
           loflag(jl) = .true.
-          jlm = jlm + 1
-          jlx(jlm) = jl
+!!!       jlm = jlm + 1
+!!!       jlx(jlm) = jl
```
mgduda, line 2628: Can you explain why it is that these two lines can be safely commented-out?

mgduda (changes requested):

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
```diff
@@ -973,20 +1033,24 @@ subroutine convection_to_MPAS(configs,diag_physics,tend_physics,its,ite)
...
-!$acc update device(rucuten, rvcuten)
+!!!  For the ntiedke we will use the GPU copy. If we actually use the tiedke (before porting it) then we would need the below:
+!!!$acc update host(rucuten,rvcuten)
 
     case default
 
  end select convection_select
 
-!$acc update device(cuprec, raincv, rthcuten, rqvcuten, rqccuten, rqicuten)
+!!$acc update device(cuprec, raincv, rthcuten, rqvcuten, rqccuten, rqicuten)
```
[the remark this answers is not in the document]
cponder: I had problems the first time I tried to remove this, which were fixed by changing declarations in other files.
Please delete it now.

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
```diff
@@ -330,17 +330,20 @@ subroutine cu_ntiedtke(                                    &
...
-      do j=jts,jte
-         do i=its,ite
-            cu_act_flag(i,j)=.true.
-         enddo
-      enddo
+!!!   do j=jts,jte
+!!!      do i=its,ite
+!!!         cu_act_flag(i,j)=.true.
+!!!      enddo
+!!!   enddo
```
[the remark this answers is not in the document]
cponder: Yes please delete this.
I wasn't sure whether we wanted to leave code like this to give an indication why things had changed.
But a plain `sdiff` against the earlier version is good enough.

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
```diff
@@ -406,73 +428,94 @@ subroutine cu_ntiedtke(                                    &
...
         enddo
-        pp = pp + 1
+!!!     pp = pp + 1
       enddo
+!$acc end parallel
 
       if(present(rqccuten))then
         if ( f_qc ) then
-          pp = 0
+!!!       pp = 0
```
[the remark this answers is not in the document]
cponder: Yes please delete this, and the commented-out forms (as per the discussion up above).
Just for the record, we're replacing these accumulated values pp & zz
``` 
   448       if(present(rqccuten))then
    449         if ( f_qc ) then
    450           pp = 0
    451           do k=kts,kte
    452             zz = kte-pp
    453             do i=its,ite
    454               rqccuten(i,k,j)=(q2(i,zz)-qc3d(i,k,j))*rdelt
    455             enddo
    456             pp = pp + 1
    457           enddo
    458         endif
    459       endif
```
with a loop-invariant
```
zz = kte+1-k
```

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
```diff
@@ -1290,57 +1442,69 @@ subroutine cumastrn  &
...
-        ik = jk - 1
+!!!     ik = jk - 1
+!$acc loop gang vector private(zerate,zmfa)
         do jl = 1,klon
           if ( ldcum(jl) ) then
             if ( jk == idtop(jl) ) then
-              zud(jl,jk) = 0.5*(zuu(jl,jk)+puen(jl,ik))
-              zvd(jl,jk) = 0.5*(zvu(jl,jk)+pven(jl,ik))
+              zud(jl,jk) = 0.5*(zuu(jl,jk)+puen(jl,jk+1))
+              zvd(jl,jk) = 0.5*(zvu(jl,jk)+pven(jl,jk+1))
```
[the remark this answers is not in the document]
cponder: Yes please make this change.
The stanza before that used ```jk+1``` so I mistakenly made the same substitution in this stanza as well.

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
```diff
@@ -2310,17 +2612,20 @@ subroutine cuascn &
...
       jlm = 0
+!$acc parallel vector_length(128)
+!$acc loop vector private(zmfmax,zfac) reduction(+:is)
       do jl = 1,klon
         loflag(jl) = .false.
         zprecip(jl) = 0.
         llo1(jl) = .false.
-        is = is + klab(jl,jk+1)
+!!!     is = is + klab(jl,jk+1)
+        is = klab(jl,jk+1)
```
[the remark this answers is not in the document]
cponder: I mis-used the reduction clause in this case.
Yes please put the
```
is = is + klab(jl,jk+1)
```
back.

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
cponder: @Pranay-Reddy-Kommera  I can tell there's something wrong with the way I did this, but also I can't remember what the change was about. Can you look at this with me?

cponder: Ok -- Pranay put me back on track with this. We don't want to collect the compacted index-array here
```
   2319         if ( (ldcum(jl) .and. klab(jl,jk+1) == 2) .or.   &
   2320              (ktype(jl) == 3 .and. klab(jl,jk+1) == 1) ) then
   2321           loflag(jl) = .true.
   2322           jlm = jlm + 1
   2323           jlx(jlm) = jl
   2324         end if
```
since it uses a serial loop to do it.
Instead we'll parallelize on the original Boolean array
```
   2735 ! compute the upfraft speed in cloud layer
   2736 !$acc parallel vector_length(128)
   2737 !$acc loop vector private(jl)
   2738 !!!     do jll = 1 , jlm
   2739 !!!       jl = jlx(jll)
   2740         do jl = 1, klon
   2741         if(loflag(jl)) then
```
and operate if the value had been true.
This adds some extra overhead in the non-GPU case.

I think it would be possible for us to use a prefix-sum operation to generate the index array instead but its a little messy.
I haven't worked out the details of doing this in OpenACC.
Also both the serial approach (in the CPU-only case) and the prefix-sum approach (in the parallel case) probably don't save that much time since they involve scanning the original ```loflag``` array either way.

cponder:
One side-question: as we work on these files, could we insert a comment at the top, describing the indentation-convention to be used in each one?
Some of these files looked junky to me to begin with, but my changes are making them look even worse.
I've adjusted control-flow but left the original indentation just so the file would line up more closely with the original in a side-by-side diff.
The same with the commented-out code, just to keep some correspondence with the original.
But (as you're flagging above) when the file is viewed standalone, these contents just confuse what's going on, so I shouldn't be doing this.

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
cponder: I just pushed up this change, it's the third commit.

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
```diff
@@ -652,19 +672,29 @@ subroutine convection_from_MPAS(dt_dyn,configs,mesh,sfc_input,diag_physics,tend_
  call mpas_pool_get_array_gpu(tend_physics,'rqvcuten',rqvcuten)
  call mpas_pool_get_array_gpu(tend_physics,'rqccuten',rqccuten)
  call mpas_pool_get_array_gpu(tend_physics,'rqicuten',rqicuten)
-!$acc update host(cuprec, raincv, rthcuten, rqvcuten, rqccuten, rqicuten)
 
  do j = jts,jte
```
supreethms1809, line 676: This will launch jte number of kernels. @Pranay-Reddy-Kommera @cponder Should we move the parallel region outside?

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
```diff
@@ -652,19 +672,29 @@ subroutine convection_from_MPAS(dt_dyn,configs,mesh,sfc_input,diag_physics,tend_
...
+ do i = its,ite
+ do k = kts,kte
+    rthcuten_p(i,k,j) = rthcuten(k,i)
+    rqvcuten_p(i,k,j) = rqvcuten(k,i)
+    rqccuten_p(i,k,j) = rqccuten(k,i)
+    rqicuten_p(i,k,j) = rqicuten(k,i)
+ enddo
+ enddo
+!$acc end parallel
  enddo
```
supreethms1809, line 698: This will launch jte number of kernels. @Pranay-Reddy-Kommera @cponder Should we move the parallel region outside?

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
```diff
@@ -922,6 +978,10 @@ subroutine convection_to_MPAS(configs,diag_physics,tend_physics,its,ite)
        call mpas_pool_get_array_gpu(tend_physics,'rucuten',rucuten)
        call mpas_pool_get_array_gpu(tend_physics,'rvcuten',rvcuten)
 
+ !$acc update host(cubot, cutop, xmb_total, xmb_shallow, k22_shallow, &
+ !$acc kbcon_shallow, ktop_shallow, ktop_deep, qc_cu, qi_cu, rucuten, &
+ !$acc rvcuten)
+
        do j = jts,jte
```
supreethms1809, line 985: This seems to be happening on CPU and then updated to GPU. Is there a reason not to do inside parallel region.

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
```diff
@@ -790,26 +820,34 @@ subroutine convection_from_MPAS(dt_dyn,configs,mesh,sfc_input,diag_physics,tend_
        call mpas_pool_get_array_gpu(tend_physics,'rqvdynten',rqvdynten)
        call mpas_pool_get_array_gpu(tend_physics,'rucuten'  ,rucuten  )
        call mpas_pool_get_array_gpu(tend_physics,'rvcuten'  ,rvcuten  )
-!$acc update host(xland, hfx, qfx, rqvblten, rqvdynten, rucuten, rvcuten)
 
        do j = jts,jte
+!$acc parallel vector_length(32)
```
supreethms1809, line 825: @Pranay-Reddy-Kommera @cponder Should we move the parallel region outside? I am concerned about the number of kernel launches

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
cponder: @Pranay-Reddy-Kommera Does this have to do with the assignments coming from the ```*_p``` pointer verisons?
If I recall correctly, these pointers were calculated in the Host memory so we use Host operations to copy the values before updating them to the GPU memory:
```
    981  !$acc update host(cubot, cutop, xmb_total, xmb_shallow, k22_shallow, &
    982  !$acc kbcon_shallow, ktop_shallow, ktop_deep, qc_cu, qi_cu, rucuten, &
    983  !$acc rvcuten)
    984 
    985        do j = jts,jte
    986        do i = its,ite
    987           cubot(i)         = cubot_p(i,j)
    988           cutop(i)         = cutop_p(i,j)
    989           xmb_total(i)     = xmb_total_p(i,j)
    990           xmb_shallow(i)   = xmb_shallow_p(i,j)
    991           k22_shallow(i)   = k22_shallow_p(i,j)
    992           kbcon_shallow(i) = kbcon_shallow_p(i,j)
    993           ktop_shallow(i)  = ktop_shallow_p(i,j)
    994           ktop_deep(i)     = ktop_deep_p(i,j)
    995 
    996           do k = kts,kte
    997              qc_cu(k,i)    = qccu_p(i,k,j)
    998              qi_cu(k,i)    = qicu_p(i,k,j)
    999              rucuten(k,i)  = rucuten_p(i,k,j)
   1000              rvcuten(k,k)  = rvcuten_p(i,k,j)
   1001           enddo
   1002        enddo
   1003        enddo
   1004 !$acc update device(cubot, cutop, xmb_total, xmb_shallow, k22_shallow, &
   1005 !$acc               kbcon_shallow, ktop_shallow, ktop_deep, qc_cu, qi_cu, &
   1006 !$acc               rucuten, rvcuten)
```

cponder: @raghuraj19 
One thing I'd have to add here, from viewing the original 7.* version of the code:
```
    946        do j = jts,jte
    947        do i = its,ite
    948           cubot(i) = cubot_p(i,j)
    949           cutop(i) = cutop_p(i,j)
    950           nca(i)   = nca_p(i,j)
    951           do k = kts, kte
    952              w0avg(k,i)    = w0avg_p(i,k,j)
    953              rqrcuten(k,i) = rqrcuten_p(i,k,j)
    954              rqscuten(k,i) = rqscuten_p(i,k,j)
    955           enddo                          
    956        enddo
    957        enddo
```
It looks like all of the final values *only* depend on the final ```jts=jte``` iteration, since it overwrites all the previous cases for each value of ```i``` & ```k```.
Should I flag this to Michael as an issue? The ```cu_kain_fritsch``` & ```cu_tiedtke``` & ```cu_ntiedtke``` all do that here in the case-list.

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
cponder: It looks like we should do it in both of these compound-loops, though there might be a reason to do a loop-interchange as well:
```
    824        do j = jts,jte
    825 !$acc parallel vector_length(32)
    826 !$acc loop gang vector
    827        do i = its,ite
    828           xland_p(i,j) = xland(i)
    829           hfx_p(i,j)   = hfx(i)
    830           qfx_p(i,j)   = qfx(i)
    831        enddo
    832 !$acc end parallel
    833        enddo
    834 
    835        do j = jts,jte
    836 !$acc parallel vector_length(32)
    837 !$acc loop gang vector collapse(2)
    838        do i = its,ite
    839        do k = kts,kte
    840           rucuten_p(i,k,j) = rucuten(k,i)
    841           rvcuten_p(i,k,j) = rvcuten(k,i)
    842        enddo
    843        enddo
    844 !$acc end parallel
    845        enddo
```

Pranay-Reddy-Kommera: The j loop is always from 1 to 1. So, the parallel region inside should not affect the code. Having it outside also would not affect the code.

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
Pranay-Reddy-Kommera: @supreethms1809 : This do loop is part of the 'cu_grell_freita' scheme which is not ported onto GPU. All the *_p variables related to 'cu_grell_freita' are assumed to be on CPU. So, the do loop is not ported to GPU.
@cponder : The jts and jte values are always equal to 1. So, having j loop does not impact the code. If I am not wrong (@raghuraj19 please correct me if I am wrong), I believe the j loop is left inside to match with the code structure of WRF.

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
raghuraj19: @cponder- @Pranay-Reddy-Kommera is right. jts=jte. This is a trick within MPAS to map WRF routines- or at least that is my understanding. So, no need to file any bugs here.

raghuraj19 (approved):

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
cponder: @supreethms1809 This should answer your question here -- the righthand-side variables (```cubot_p``` etc.) all reside on the CPU here, so we're doing the assignment on the CPU side and pushing the data across to the GPU.
This will change later, when we get the ```cu_grell_freita``` scheme onto the GPU.
Can you please "Resolve converstaion" (or should I)?

src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
cponder: I went ahead and pushed the loops inside, with a corresponding increment of the ```collapse``` parameter.
While the change is redundant, as Pranay explained, the code is "no worse" for the change, and won't raise such questions in the future.
@supreethms1809 Can you please "Resolve conversation" (or should I)?

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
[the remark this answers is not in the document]
cponder: I pushed the edit on line 493 above, it's in the 5th commit.
@mgduda Can you please "Resolve conversation" (or should I)?

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
cponder: Here goes...

mgduda:
With no optimizations and with double-precision reals, I'm seeing what look like larger-than-roundoff differences with the changes in this PR.

@cponder I'll investigate, but in the meanwhile, if you have any ideas on which changes in the nTiedtke scheme might be expected to lead to differences in results, that would be valuable.

src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F
```diff
@@ -353,42 +356,61 @@ subroutine cu_ntiedtke(                                    &
...
       enddo
 
-      pp = 0
+!!!   pp = 0
+!$acc loop gang vector collapse(2) private(zz)
       do k=kts,kte
-        zz = kte-pp
+!!!     zz = kte-pp
         do i=its,ite
+          zz = kte+1-k
```
cponder, line 413: This looks incorrect unless kts is 1 to begin with.
I think this ought to be
zz=kte-(k-kts)

cponder:
One other thing -- on a cursory look, I can't tell whether the code depends on any un-initialized array elements.
As I understand it, the Fortran standard doesn't require they be set to 0.

Also, you're seeing the divergence when you run CPU-only, right?

raghuraj19:
@mgduda : Is the deviation on all compilers? If not, is it only with PGI? If the latter, do you see the errors even after removing "-Mnofma" flag? I can debug if you can let me know which variables should I look at in the routine.

mgduda:
@cponder @raghuraj19 I think I've been able to narrow down the problematic code changes to a relatively small region in `module_cu_ntiedtke.F`. I've just pushed a commit, [8273f113](https://github.com/mgduda/MPAS-Model/commit/8273f113), to a branch named [atmosphere/openacc-convection-mgduda](https://github.com/mgduda/MPAS-Model/commits/atmosphere/openacc-convection-mgduda) on my fork. With the code as it looks in I've just pushed a commit, [8273f113](https://github.com/mgduda/MPAS-Model/commit/8273f113), I'm able to get bit-identical results on CPUs compared with the `atmosphere/develop-openacc` branch.

@cponder Could you take a look at the specific code that I reverted in [8273f113](https://github.com/mgduda/MPAS-Model/commit/8273f113)?

cponder:
This puts the serial flag-calculation back in, we don't want that.
OI may not be using the "flag" test correctly, at least we know that's where the problem is centered.

cponder:
@mgduda 
I think the problem is that I'd tried to merge the logic into the prior loop, which also had some conditions that would suppress the setting of the flag in all cases.
I believe that this last commit fixes the problem. Can you please test it?

mgduda:
> I believe that this last commit fixes the problem. Can you please test it?

I'm still seeing differences after merging this branch.

cponder:
@mgduda 
Please try one more time. I had been assigning the reduction-variable incorrectly, which would change the test-exit cases.
That would also explain why the old code worked while this one didn't.

mgduda:
@cponder Now I'm able to get bit-identical results before and after merging this PR. Would it be alright with you if I squashed the commits of this branch into a single commit, and also fixed this compilation error
```
 2161 |                    if ( loflag(jl) .eq. .true. ) flag = .true.
      |                       1
Error: Logicals at (1) must be compared with .eqv. instead of .eq.
```
before merging?

cponder:
Yes please fix this, and squash the commits.

mgduda (approved):

PR #898, opened 2021-07-28
Added OpenACC directives to Surface Layer model, for GPU excecution. (Superceded by PR #928)

cponder:
Added OpenACC directives from 6.x-openacc version to manage Host <-> GPU memory movement, to enable surface-layer model to run on the GPU.

PR #928, opened 2021-09-08
Added OpenACC directives to Surface Layer model, for GPU excecution 

cponder:
This replaces PR #898.
It has updates to be compatible with the other Physics models being mixed between running on GPU.
I closed the #898 in favor of this one.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_lsm.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_lsm.F
@@ -715,6 +715,7 @@
  enddo
 
  if(config_frac_seaice) then
+!$acc update host(chs_sea,chs2_sea,cqs2_sea,cpm_sea,hfx_sea,lh_sea,qfx_sea,qgh_sea,qsfc_sea,tsk_sea)
     do j = jts,jte
     do i = its,ite
        if(xice_p(i,j).ge.xice_threshold .and. xice_p(i,j).le.1._RKIND) then
--- a/src/core_atmosphere/physics/physics_wrf/module_sf_sfclay.F
+++ b/src/core_atmosphere/physics/physics_wrf/module_sf_sfclay.F
@@ -7,6 +7,7 @@ MODULE module_sf_sfclay
  REAL    , PARAMETER ::  OZO=1.59E-5
 
  REAL,   DIMENSION(0:1000 ),SAVE          :: PSIMTB,PSIHTB
+!$acc declare create(PSIMTB,PSIHTB)
 
 CONTAINS
 
@@ -211,16 +212,18 @@ CONTAINS
 
       INTEGER ::  I,J
 
+!$acc data create(dz8w1d,U1D,V1D,QV1D,P1D,T1D)
       DO J=jts,jte
-
+!$acc parallel vector_length(32)
+!$acc loop gang vector
         DO i=its,ite
            DX2D(i)=DX(i,j)
         ENDDO
-
+!$acc loop gang vector
         DO i=its,ite
           dz8w1d(I) = dz8w(i,1,j)
         ENDDO
-   
+!$acc loop gang vector
         DO i=its,ite
            U1D(i) =U3D(i,1,j)
            V1D(i) =V3D(i,1,j)
@@ -228,6 +231,8 @@ CONTAINS
            P1D(i) =P3D(i,1,j)
            T1D(i) =T3D(i,1,j)
         ENDDO
+!$acc end parallel
+!-!-!$acc update host(dz8w1d,U1D,V1D,QV1D,P1D,T1D)
 
         !  Sending array starting locations of optional variables may cause
         !  troubles, so we explicitly change the call.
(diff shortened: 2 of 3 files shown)


cponder:
This also has a dependency on a change to the file
```
src/core_atmosphere/physics/mpas_atmphys_driver_lsm.F
```

src/core_atmosphere/physics/physics_wrf/module_sf_sfclay.F
```diff
@@ -879,62 +963,77 @@ SUBROUTINE SFCLAY1D(J,UX,VX,T1D,QV1D,P1D,dz8w1d,                &
...
+          CHS(I)=UST(I)*KARMAN/DENOMQ(I)
+!         GZ2OZ0(I)=ALOG(2./ZNT(I))
+!         PSIM2(I)=-10.*GZ2OZ0(I)
+!         PSIM2(I)=AMAX1(PSIM2(I),-10.)
+!         PSIH2(I)=PSIM2(I)
+          CQS2(I)=UST(I)*KARMAN/DENOMQ2(I)
+          CHS2(I)=UST(I)*KARMAN/DENOMT2(I)
+        ENDDO
+!$acc end parallel
+!-!-!$acc update host(QFX,LH,HFX,CHS,CQS2,CHS2)
```
mgduda, line 1034: There are a few places with these deactivated (?) `update host` directives. Will these be needed if we eventually intend to make use of variables like `QFX`, `LH`, `HFX`, etc. on the host, either for computation or to write them to model output files?

src/core_atmosphere/physics/physics_wrf/module_sf_sfclay.F
```diff
@@ -388,22 +393,32 @@ SUBROUTINE SFCLAY1D(J,UX,VX,T1D,QV1D,P1D,dz8w1d,                &
...
+      DO I=its,ite                                   
         TGDSA(I)=TSK(I)                                    
 ! PSFC cb
 !        THGB(I)=TSK(I)*(100./PSFC(I))**ROVCP                
         THGB(I)=TSK(I)*(P1000mb/PSFCPA(I))**ROVCP   
-    5 CONTINUE                                               
+      END DO
+!$acc end parallel
+!-!-!$acc update host(PSFC,TGDSA,THGB)
+!!!i   5 CONTINUE
```
mgduda, line 421: Where old-style loops have been updated to make use of `END DO`, I think it would be fine to just delete the old numbered `CONTINUE` lines to keep the code cleaner.

src/core_atmosphere/physics/physics_wrf/module_sf_sfclay.F
```diff
@@ -388,22 +393,32 @@ SUBROUTINE SFCLAY1D(J,UX,VX,T1D,QV1D,P1D,dz8w1d,                &
...
+!$acc loop gang vector
       DO i=its,ite
 ! PSFC cb
          PSFC(I)=PSFCPA(I)/1000.
       ENDDO
 !                                                      
 !----CONVERT GROUND TEMPERATURE TO POTENTIAL TEMPERATURE:  
 !                                                            
-      DO 5 I=its,ite                                   
+!!!   DO 5 I=its,ite
```
mgduda, line 411: I think it would help maintain cleaner code to just delete the old-style `DO` loops rather than keeping them as comments.

mgduda (changes requested):

cponder:
The commented-out updates, I did that while trying to narrow-down an execution problem.
Frankly, since you'll be controlling the final merge, I'd rather that you remove the commented-out forms that are junking-up the code, since otherwise you & I will end up iterating on it awhile.
I can just sdiff my earlier copies of the files to figure out what had been removed, if I need to figure out what the earlier versions of the code were doing.

src/core_atmosphere/physics/mpas_atmphys_driver_lsm.F
```diff
@@ -717,6 +717,7 @@ subroutine lsm_to_MPAS(configs,mesh,diag_physics,sfc_input,its,ite)
  if(config_frac_seaice) then
     do j = jts,jte
     do i = its,ite
+!$acc update host(chs_sea,chs2_sea,cqs2_sea,cpm_sea,hfx_sea,lh_sea,qfx_sea,qgh_sea,qsfc_sea,tsk_sea)
```
supreethms1809, line 720: the update host directive is inside the loop. Is there a reason?

cponder:
Ok I just fixed it with this commit.

raghuraj19 (approved):

src/core_atmosphere/physics/mpas_atmphys_driver_lsm.F
cponder: @mgduda 
The file-diff on the "Files changed" page shows it as being outside the loop.
My copy of the directory still shows the problem being in the file.
I don't know how to deal with this discrepancy.

mgduda (approved):

PR #933, opened 2021-10-13
Remove redundant -CU -CB from ifort debug flags when -check all is also present

mgduda:
This PR removes the `-CU` and `-CB` options from the definition of `FFLAGS_DEBUG` in
the 'ifort', 'ifort-scorep', 'ifort-gcc', and 'intel-nersc' build targets, since
the `-CU` (`-check uninit`) and `-CB` (`-check bounds`) options are redundant when
`-check all` is specified.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,7 +148,7 @@ ifort:
 	"CFLAGS_OPT = -O3" \
 	"CXXFLAGS_OPT = -O3" \
 	"LDFLAGS_OPT = -O3" \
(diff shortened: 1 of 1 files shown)


mgduda:
Here's the ifort documentation for the `-check` option: https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/compiler-reference/compiler-options/compiler-option-details/language-options/check.html . I didn't include the link in the PR description or commit message, since it's hard to judge how stable the URL might be.

PR #934, opened 2021-10-13
Add missing dtheta_dt_mp computation

mgduda:
This PR adds code for computing 'dtheta_dt_mp', the potential temperature
heating rate from microphysics, that had inadvertently been omitted. Without
the changes in this PR, the 'dtheta_dt_mp' field will always contain a constant
zero value when written to model output files.

Also included in this PR are a few minor cleanup changes surrounding the
computation of dtheta_dt_mp.


mgduda:
This PR has been approved based on visual inspection by @nickszap .

PR #936, opened 2021-10-13
Fix bug in vertical interp of relhum and spechum when levels are given top-to-bottom

mgduda:
This PR fixes a bug in the vertical extrapolation of relative humidity and
specific humidity to model levels below the lowest first-guess level when
first-guess levels are given in top-to-bottom order in the input intermediate
file.

Prior to the changes in this PR, if first-guess levels were given in
top-to-bottom order, the end result was that the vertically interpolated
relative humidity and specific humidity from the highest model level was copied
downward through all other model levels.


smileMchen:
@mgduda 

Michael,
This PR fixes the problem we have before, i.e., when MPAS is initialized by GFS model level data, unreasonable RH and SPECHUMD appear in the MPAS initial condition, which later  leads to MPAS failure. 

The cases driven by GFS model level data can be run successfully with this PR fix. 
Another test case using ERA-I model level data to initialize MPAS was also run successfully.

PR #937, opened 2021-10-14
Correct check on availability of dtheta_dt_mix when computing depv_dt_mix

mgduda:
This PR corrects a check on the availability of the dtheta_dt_mix variable when
computing depv_dt_mix.

The code to compute depv_dt_mix in the calc_pvBudget routine previously checked
that the dtheta_dt_mp variable had been successfully been retrieved from the
diag pool, when instead the check should have been on the dtheta_dt_mix
variable.

Note that the fix in this PR isn't expected to have any impact in practice,
since both dtheta_dt_mp and dtheta_dt_mix should always be found in the diag
pool.


mgduda:
This PR has been approved based on visual inspection by @nickszap .

PR #938, opened 2021-10-24
Added OpenACC directives to WSM6 model, for GPU execution.

cponder:
These are re-introduced from the 6.x-openacc branch, adjusted to accommodate the 6 -> 7 changes.
The updates should be compatible with the other Physics models being added to on the GPU.


mgduda:
@cponder Did you intend to push changes to the nTiedtke scheme to this branch?

cponder:
That must have been an error, I probably pushed from the wrong directory snapshot.
The Dec 19 commit to #927 would have all the nTiedke changes.
I don't see how to delete this commit from #938 here though.

cponder:
Is that what's causing the merge-conflict between 927 & 938? How can I delete this commit from 927?

mgduda:
@cponder If you only want to discard commits from the tip of a branch, you can just hard-reset the branch to the last commit that you want to appear on the branch with
```
git reset --hard XXXXXX
```
where `XXXXXX` is the last commit you want.

Otherwise, if you need to remove commits from before the HEAD of a branch, you can do an [interactive rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).

cponder:
The commit is this one
```
        f8834cbefaeeb40c16c322e6ee8217ac2aab3b0e
```
and it's not the latest one in the sequence, although it should be the only one that introduces the ntiedeke file.
I ran the ```git rebase -i``` as I understand the documentation, but I still see the file & the commit on the lists.

cponder:
I tried running ```filter-branch``` but it looks like it's trying to blast the whole repo rather than the branch in the directory, so I stopped it.
I don't know if that caused any problems.

mgduda:
@cponder Right, I don't think `filter-branch` is what you want. You could try the following:
1. Run `git rebase -i f8834cb^`. This will begin an interactive rebase after the parent of f8834cb.
2. Your default editor should open the following:
```
pick f8834cb Evaluating flag-loop *outside* the other if-conditions.
pick 30e0db3 Added OpenACC directives from 6.x-openacc branch.
pick e5b5f51 Commented-out the debug Checkpoints.
pick 998d4f0 Had to remove corresponding flush statements.
pick ad7e8c7 More variables are treated as GPU-resident.
pick 676756a Removed cycle/exit statements from most loops. Seems to be a compiler issue causing hangs with these.
```
Just delete the first line (corresponding to commit f8834cb) to delete that commit. Save changes and exit.

3. The rebase should succeed. Note that your merge commit has been turned into a regular commit of the changes in the merged branch (such is the nature of rebasing with merges, I suppose).

cponder:
I ran the commands in this order:
```
git clone git@github.com:cponder/MPAS-Model.git .
git pull origin atmosphere/openacc-wsm6
git checkout -b atmosphere/openacc-wsm6
git rebase -i f8834cb^                                                      # Delete line with "f8834cb".
git pull origin atmosphere/openacc-wsm6
git push origin atmosphere/openacc-wsm6
```
I can see the revised lists of commits now. But the ```module_cu_ntiedtke.F``` is still listed as a modified file, and still gives a merge-conflict if I try to apply this PR.

mgduda:
@cponder Running `git log src/core_atmosphere/physics/physics_wrf/module_cu_ntiedtke.F`, it looks like there are multiple commits in this branch that modify `module_cu_ntiedtke.F`.

Is it the case that the modified files in this branch can be classified as either: only contains changes for WSM6; or only contains changes for nTiedtke? If so, it might be easy to just rebuild this branch with a single commit containing only those files from the first group. Otherwise, if there are files with changes for both WSM6 and for nTiedtke, then trying to rebase and select only the WSM6-related commits is probably the easier approach.

cponder:
@mgduda
What we want is everything *except* the ntiedeke file changes here. Can you generate a commit for that?

mgduda:
@cponder I've pushed a reworked version of the `atmosphere/openacc-wsm6` branch to my fork of the repository that I think correctly removes the nTiedtke-related changes from `module_cu_ntiedtke.F` and from `mpas_atmphys_driver_convection.F`. Could you take a look at that branch, and if the code looks like what you want, could you force-push that branch to your fork?

cponder:
@mgduda You mean this branch?
https://github.com/MPAS-Dev/MPAS-Model/tree/atmosphere/develop-openacc/src

cponder:
I have a directory with all the changes, I just can't push the WSM6 because of the ntiedeke conflict.
I think it would be easier if you just merge your branch into the main `atmosphere/develop-openacc` branch, then I can download it and compare the source against what's in my combined directory.

cponder:
I created a cleaned-up PR #952 but it looks like it's trying to change everything in the tree....

mgduda:
@cponder I've updated the `atmosphere/openacc-wsm6` branch in my fork of the repository: https://github.com/mgduda/MPAS-Model/commits/atmosphere/openacc-wsm6 . Could you take a look at this branch and see whether the GPU-related changes look good to you? I've reworked the branch to remove all unnecessary mode and whitespace changes, which, although nice to have, distract from the essential OpenACC changes.

I've you're satisfied with the OpenACC changes in my `atmosphere/openacc-wsm6` branch, you could force-push that branch to your fork, which will then automatically update this PR.

cponder:
Last time I did a force-push I destroyed someone's repo, so no, I don't want to do that. Can you?
Also is there a way I can diff your fork against mine here on the GitHub site?

cponder:
The ```src/core_atmosphere/physics/mpas_atmphys_driver_convection.F``` file  is different between our trees and I don't know which is correct since it depends on the residency of arrays that are controlled in other modules.
My inclination is to take your merge as-is and then see if the merged branch shows any problems.

cponder:
@mgduda -- you saw perturbations in some arrays from the accelerated WSM6 code, right?
Which arrays were they? And were you running with the OpenACC enabled?

cponder:
Comparing these two trees
```
atmosphere/develop-openacc        git@github.com:MPAS-Dev/MPAS-Model.git
atmosphere/openacc-wsm6           git@github.com:cponder/MPAS-Model.git
```
I get these files being different:
```
src/core_atmosphere/inc/core_variables.inc
src/core_atmosphere/physics/mpas_atmphys_driver_microphysics.F
src/core_atmosphere/physics/mpas_atmphys_functions.F
src/core_atmosphere/physics/mpas_atmphys_interface.F
src/core_atmosphere/physics/physics_wrf/module_mp_radar.F
src/core_atmosphere/physics/physics_wrf/module_mp_wsm6.F
src/core_init_atmosphere/mpas_atmphys_functions.F
```
Is this the right list?
It looks like my ```atmosphere/openacc-wsm6``` branch adds onto the latest commits from ```develop-openacc``` in each case.

cponder:
When I run the two for 5 timesteps I see these differences:
```
DIFF (tolerance): variable "hpbl" of type "NC_FLOAT" max at element [0, 1684] of value 70.976639 vs 453.707520
DIFF (tolerance): variable "hfx" of type "NC_FLOAT" max at element [0, 6515] of value 128.867371 vs 28.615496
DIFF (tolerance): variable "gsw" of type "NC_FLOAT" max at element [0, 6515] of value 382.091797 vs 129.568588
```
but I thought they were affected by different modules than the ones listed above.

raghuraj19:
@mgduda : We will need this to be pushed in as well. Looks like Carl made the necessary changes. If we missed anything, please let us know.

mgduda (approved):

PR #939, opened 2021-10-26
Update landuse and soilcat to use refactored mapping function

MiCurry:
This commit updates landuse (lu_index) and soil top category (soilcat_top) to
use the unified mapping function. It also moves the landmask derivation, and
soiltemp interpolation into their own functions outside of init_atm_static.

The interpolation of soiltemp does not need to use the unified mapping function
as it can be safely interpolated in parallel.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -90,6 +90,7 @@
  !
  integer (kind=I8KIND), dimension(:), pointer :: ter_integer
  integer, dimension(:), pointer :: lu_index
+ integer, dimension(:), pointer :: soilcat_top
  integer, dimension(:), pointer :: nhs
  integer, dimension(:,:), allocatable:: ncat
 
@@ -128,7 +129,6 @@
  integer:: iCell,iEdge,iVtx
  integer,dimension(5) :: interp_list
  integer,dimension(:),allocatable  :: nhs
- integer,dimension(:,:),allocatable:: ncat
       
  real(kind=RKIND), pointer :: scalefactor_ptr
  real(kind=RKIND) :: scalefactor
@@ -145,7 +145,6 @@
 
  integer, pointer :: isice_lu, iswater_lu
  integer :: iswater_soil
- integer, pointer :: iswater_soil_ptr
  integer, pointer :: nCells, nCellsSolve, nEdges, nVertices, maxEdges
  logical, pointer :: on_a_sphere
  real (kind=RKIND), pointer :: sphere_radius
@@ -174,7 +173,6 @@
  integer (kind=I8KIND), dimension(:,:), pointer :: albedo12m_int
  real (kind=RKIND) :: fillval
  real (kind=RKIND), pointer :: missing_value
- integer, pointer :: category_min, category_max
  integer, dimension(:), pointer :: lu_index
  integer, dimension(:), pointer :: soilcat_top
  integer, dimension(:), pointer :: landmask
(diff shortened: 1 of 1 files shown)


MiCurry:
@mgduda I just pushed an update that addressed your comments (and mine) above. Feel free to give it another review!

src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -1773,6 +1466,390 @@ subroutine interp_terrain(mesh, kdtree, geog_data_path)
...
+        !
+        ! Currently, the MODIS landuse dataset has zeros at the South Pole, and possibly other
+        ! places, so this return statement does not accumulate those values, which would be below ncat's
+        ! lower bound (aka, category_min). Other categorical datasets may want to use create a new
+        ! accumulation routine that produces an error when pixel(:) is above or below category_min/max.
+        !
+        return
+     else
+        ncat(cat(1), iCell) = ncat(cat(1), iCell) + 1
+     end if
```
mgduda, line 1526: This if/else block still seems awkward to me. If we have an `else` condition for accumulating the category, then the `return` statement isn't necessary; on the other hand, if we keep the `return` statement, then the `else` condition isn't necessary. I would propose either of the following:
```fortran
     !
     ! Currently, the MODIS landuse dataset has zeros at the South Pole, and possibly other
     ! places, so we return early if given an invalid category
     !
     if (cat(1) < category_min .or. cat(1) > category_max) then
        return
     end if

     ncat(cat(1), iCell) = ncat(cat(1), iCell) + 1
```
or
```fortran
     !
     ! Currently, the MODIS landuse dataset has zeros at the South Pole, and possibly other
     ! places, so we need a check on the validity of the category to be accumulated
     !
     if (cat(1) >= category_min .and. cat(1) <= category_max) then
        ncat(cat(1), iCell) = ncat(cat(1), iCell) + 1
     end if
```

Between the two options, I think the second seems better to me.

mgduda (changes requested):

mgduda:
@MiCurry In rebasing this branch, I think you may have inadvertently rewritten the commit whose description begins with "Remove redundant -CU -CB from ifort debug flags when -check all is al…".

MiCurry:
> @MiCurry In rebasing this branch, I think you may have inadvertently rewritten the commit whose description begins with "Remove redundant -CU -CB from ifort debug flags when -check all is al…".

Looks like I did. I have cleaned it up.

src/core_init_atmosphere/mpas_init_atm_static.F
MiCurry: I agree as well. The if statement was a little wonky.

mgduda:
@MiCurry Sorry that it's taking so long to complete my review. What I'm finding is that the static processing for the 60-km mesh doesn't finish in 6 hours with 4 MPI tasks on Cheyenne, while the processing finishes in only ~2700 seconds with a single MPI task. I'm going to see whether this issue existed prior to the changes in this PR, but I thought I'd give a quick update on the testing delays.

MiCurry:
> @MiCurry Sorry that it's taking so long to complete my review. What I'm finding is that the static processing for the 60-km mesh doesn't finish in 6 hours with 4 MPI tasks on Cheyenne, while the processing finishes in only ~2700 seconds with a single MPI task. I'm going to see whether this issue existed prior to the changes in this PR, but I thought I'd give a quick update on the testing delays.

That is very strange. I can look into that problem as well. For reference I (and so did another CESM colleague) run the terrain interpolation that is on develop on the 60-3 km; however, that was with 3 nodes and took 37 mins.

mgduda:
After some preliminary investigation, it looks like the kd-tree searches may be taking quite a bit longer when running in parallel. Perhaps the fact that the set of points in each kd-tree isn't global is somehow a factor in the search time?

mgduda (approved):
The slow runtimes with small MPI task counts >1 seem related to the kd-search routine and not to any changes introduced in this PR, so I think we can merge this.

PR #944, opened 2021-11-23
Port of PBL to GPUs using OpenACC.

Pranay-Reddy-Kommera:
This PR contains PBL code ported to GPUs using OpenACC.
This PR needs to be merged only after #949   has been merged into atmosphere/develop-openacc branch.

- [x]  First #949   has to be merged into [atmosphere/develop-openacc](https://github.com/MPAS-Dev/MPAS-Model/tree/atmosphere/develop-openacc)

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
@@ -297,8 +297,10 @@
  call mpas_pool_get_array_gpu(diag_physics,'wstar' ,wstar )
 
  call mpas_pool_get_array_gpu(sfc_input   ,'xland' ,xland )
-!$acc update host(hfx, hpbl, qfx, ust, wspd, znt, delta, wstar, xland)
+!!!$acc update host(hfx, hpbl, qfx, ust, wspd, znt, delta, wstar, xland)
 
+!$acc parallel vector_length(128)
+!$acc loop gang vector collapse(2)
  do j = jts,jte
  do i = its,ite
     !from surface-layer model:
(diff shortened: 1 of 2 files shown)


mgduda:
@Pranay-Reddy-Kommera Can you explain why the `GET_PBLH` routine has been duplicated as the routine `GET_PBLH_gpu`? The only difference seems to be the addition of the `!$acc routine seq` directive in the latter, and it doesn't look like the `GET_PBLH` routine is called.

Pranay-Reddy-Kommera:
@mgduda I have created `GET_PBLH_gpu` routine as I am passing in different arguments compared to `GET_PBLH`. Both the routines are same except the `!$acc routine seq` line.
In the `module_bl_mynn.F` we have the `GET_PBLH` call and the routine also exists in the same file. So, I was not sure if the `GET_PBLH` will/will not be called anywhere else, so I just duplicated it. 
If the `GET_PBLH` routine is not called then we should be able to remove it. Do you want me to remove it, and commit the change ?

mgduda:
@Pranay-Reddy-Kommera As far as I can tell, the `GET_PBLH` routine that is defined in `module_bl_ysu.F` was only called from within `module_bl_ysu.F`, and the `GET_PBLH` routine that is defined in `module_bl_mynn.F` is only called from within `module_bl_mynn.F`. So, I think it should be safe to remove the `GET_PBLH` routine from `module_bl_ysu.F` (and perhaps to rename `GET_PBHL_gpu` to `GET_PBLH`, though I can see the value in keeping the `_gpu` part of the name if you'd prefer). Thanks!

Pranay-Reddy-Kommera:
@mgduda I have removed the `GET_PBLH` routine from `module_bl_ysu.F` and renamed `GET_PBLH_gpu` to `GET_PBLH` and committed the code.

mgduda:
@Pranay-Reddy-Kommera Thanks very much. I assume you wouldn't mind if I squash the two commits in this PR into a single commit before merging?

Pranay-Reddy-Kommera:
@mgduda I am fine with squashing the two commits into a single commit.

mgduda (approved):

PR #950, opened 2021-12-07
Optimized parallel kernels in the scalars/scalars_mono

Pranay-Reddy-Kommera:
Optimized all the OpenACC kernels in the `atm_advance_scalars` and `atm_advance_scalars_mono` of the `mpas_atm_time_integration.F` file.


mgduda (approved):
I'm finding that the performance on CPUs (using the GNU 11 compilers) is 15% or more slower for the `atm_advance_scalars` and `atm_advance_scalars_mono` routines. But, we can investigate ways to maintain good CPU performance at a later time.

PR #952, opened 2022-01-06
PR 938 changes without the ntiedeke file

cponder:
I pulled down the files from PR #938 and pushed all of them except ntiedke to this PR.
The changes will appear as one commit here.
The mpas_atmphys_driver_microphysics.F may just be a mode-change.
We ought to clean these up...


cponder:
I don't get why 168 files are listed as being changed here....

mgduda:
@cponder It looks like this PR still includes changes to `module_cu_ntiedtke.F` and `mpas_atmphys_driver_convection.F`. I'd suggest we just close this PR and focus on cleaning up PR #938 . I'll take a look at the `atmosphere/openacc-wsm6` branch from PR #938 today.

PR #953, opened 2022-01-07
Add halo exchange to fix flood fill for finding dynamic tropopause

mgduda:
This PR corrects the parallel computation of the `iLev_DT` field by adding 
a halo exchange for a new flag field, `inTropo`, in the flood-fill algorithm
employed by the PV diagnostics module.

Originally, it was assumed that each MPI domain would have at least one cell
with the "correct" DT found by flood filling within the domain, and that would 
be sufficient to connect each domain's troposphere. However, for sufficiently
small domains, say, over the Arctic during winter with strong inversions, the
entire surface can be capped by high PV. So, it is necessary to communicate
between domains during the flood fill to avoid finding an incorrect DT near the
surface. 

Only in very rare cases should more than two halo exchanges be needed.

With a correction to the computation of the lowest vertical layer at or above
the dynamic tropopause, other diagnostic fields at the tropopause level will
be affected: `u_pv`, `v_pv`, `theta_pv`, `vort_pv`, `depv_dt_diab_pv`, and
`depv_dt_fric_pv`.

### DIFF
--- a/src/core_atmosphere/diagnostics/Registry_pv.xml
+++ b/src/core_atmosphere/diagnostics/Registry_pv.xml
@@ -62,5 +62,8 @@
         <var name="iLev_DT" type="integer" dimensions="nCells Time" units="-"
              description="Lowest vertical level at or above dynamic tropopause (.lt.1 if 2 PVU below column; .gt.nLevels if 2PVU above column)"/>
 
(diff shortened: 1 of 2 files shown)


nickszap (approved):
Thanks @mgduda!

PR #962, opened 2022-02-10
Correct Fortran 2008 non-conformance in atmosphere core

mgduda:
This PR makes code changes throughout the atmosphere core to achieve
conformance with the Fortran 2008 standard. Primarily, the non-conformance
issues were related to the use of non-standard kind-types and intrinsic functions.

With the changes in this PR, the atmosphere core can be successfully compiled
by the GNU compilers with `-std=f2008`. Note that there are still warnings about,
e.g., non-conforming tabs or obsolescent features, but these do not prevent
compilation and can therefore be addressed in a future PR.

### DIFF
--- a/src/core_atmosphere/diagnostics/convective_diagnostics.F
+++ b/src/core_atmosphere/diagnostics/convective_diagnostics.F
@@ -8,7 +8,7 @@
 module convective_diagnostics
 
     use mpas_derived_types, only : MPAS_pool_type, MPAS_clock_type, MPAS_LOG_ERR, MPAS_LOG_CRIT
-    use mpas_kind_types, only : RKIND
+    use mpas_kind_types, only : RKIND, R8KIND
     use mpas_log, only : mpas_log_write
 
     type (MPAS_pool_type), pointer :: mesh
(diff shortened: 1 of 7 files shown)


mgduda:
I've tested with the GNU compilers using a 120-km mesh, and the following configurations give bit-identical results after a 24-h simulation:
- [ ] single-precision, 'mesoscale_reference' physics suite
- [ ] double-precision, 'mesoscale_reference' physics suite
- [ ] single-precision, 'convection_permitting' physics suite plus CAM radiation
- [ ] double-precision, 'convection_permitting' physics suite plus CAM radiation

These simulations were run with the standard set of diagnostics fields (some of which may use code that has been modified in this PR).

I'm doubtful that all of the functions that have been modified by this PR are exercises by the above configurations, and it may be that some of the functions (e.g., in libmassv) are not called by any of our physics schemes or diagnostics. But, the fact that our supported physics suites give bit-identical results may be sufficient.

gdicker1 (approved):
I approve after visual inspection and checking the build with GNU and `-std=f2008` succeeds. It would definitely be helpful to update to a newer standard.

PR #971, opened 2022-04-06
Initial support for Cray OpenMP offloading on AMD GPUs

jychang48:
Currently only supported through the cray ftn compiler. 
All OMP directives added on top of existing ACC directives.
Only core dynamics ported. Physics not touched.
Macro definitions (MPAS_AMD_OFFLOAD, GPUOMP, etc) are temporary measures and will be renamed or removed when the AOMP compiler is supported. Should also eventually work with Intel and NVIDIA GPUs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,10 @@ stream_list.*
 *.i90
 
 src/core_*/inc
+
+# cray acc offloading assembly files
+*.acc.s
+# cray intermediate files
+*.opt
+*.cg
+*.lst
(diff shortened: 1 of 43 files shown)


jychang48:
Note, many of the changes in this PR dealt with manually incorporating `!$omp target enter data map ...` for all data structures involving `!$acc declare create` due to the limitations of the compiler at the time. It is my hope today that one can simply use `!$omp declare target` for such data now

PR #972, opened 2022-04-12
corrected the initialization of xice for "cold" sea-surface temperatures

ldfowler58:
This PR adjusts the initialization of xice for cold sea-surface temperatures

Setting `xice` to 1.0 for sea-surface temperatures colder than 271 K has proved to
produce too much sea ice at high latitudes. Furthermore, many datasets provide a
sea ice fraction field. Determination of using actual sea ice fraction input or
setting xice to 1.0 for sea-surface temperatures colder than a pre-defined value
in the case where fractional sea ice is not provided in the input is now moved 
to the subroutine `physics_init_seaice` in module `mpas_atmphys_initialize_real.F`.

Now, the temperature threshold for converting water points to sea ice with a 
fraction of 1.0 is set by a new namelist option, `config_tsk_seaice_threshold`, in 
a new `&physics` namelist group. The default value for this threshold is 100 K,
which effectively shuts off the conversion of water to sea ice, which is
reasonable when ingesting datasets that already provide a fractional sea ice
field.

### DIFF
--- a/src/core_init_atmosphere/Registry.xml
+++ b/src/core_init_atmosphere/Registry.xml
@@ -278,7 +278,12 @@
                      units="-"
                      description="Whether to switch sea-ice threshold from 0.5 to 0.02"
                      possible_values="true or false"/>
+        </nml_record>
 
+        <nml_record name="physics" in_defaults="false">
+                <nml_option name="config_tsk_seaice_threshold" type="real" default_value="100." units="K"
+                     description="surface temperature threshold below which water points are set to seaice points"
+                     possible_values="Positive real values"/>
         </nml_record>
 
         <nml_record name="io" in_defaults="true">
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -4559,9 +4559,6 @@ call mpas_log_write('Done with soil consistency check')
 
       if (allocated(maskslab)) deallocate(maskslab)
 
-      ! Freeze really cold ocean
-      where (sst < 271.0 .and. landmask == 0) xice = 1.0
-
       ! Limit XICE to values between 0 and 1. Although the input meteorological field is between 0.
       ! and 1., interpolation to the MPAS grid can yield values of XiCE less than 0. and greater
       ! than 1.:
--- a/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
@@ -55,6 +55,10 @@
 ! * In subroutine physics_init_seaice, added the initialization of the annual maximum snow albedo over seaice
 !   points to 0.75.
 ! * Laura D. Fowler (laura@ucar.edu) / 2022-03-15).
+! * In subroutine physics_init_seaice, corrected the initialization of seaice points when the surface
+!   temperature was originally colder than 271K. We now use the seaice threshold config_tsk_seaice_threshold
+!   that has a default value set to 100K. This leads to decreased seaice at high latitudes.
+!   Laura D. Fowler (laura@ucar.edu) / 2022-03-25.
 
 
  contains
@@ -599,6 +603,7 @@
 
  real(kind=RKIND):: xice_threshold
  real(kind=RKIND):: mid_point_depth
+ real(kind=RKIND),pointer:: tsk_seaice_threshold
  real(kind=RKIND),dimension(:),pointer  :: vegfra
  real(kind=RKIND),dimension(:),pointer  :: seaice,snoalb,xice
  real(kind=RKIND),dimension(:),pointer  :: skintemp,tmn,xland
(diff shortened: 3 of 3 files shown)


src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -4497,8 +4497,12 @@ subroutine init_atm_case_gfs(block, mesh, nCells, nEdges, nVertLevels, fg, state
 
       if (allocated(maskslab)) deallocate(maskslab)
 
-      ! Freeze really cold ocean
-      where (sst < 271.0 .and. landmask == 0) xice = 1.0
+      ! Freeze really cold ocean: setting xice to 1.0 for sea-surface temperatures colder than 271K
+      ! has proved to produce too much seaice at high latitudes. The line below has been commented
+      ! out and setting xice to 1.0 for sea-surface temperatures colder than 100K is now moved to
+      ! the subroutine physics_init_seaice (in module mpas_atmphys_initialize_real.F).
+      ! Laura D. Fowler (laura@ucar.edu) / 2022-03-25.
```
weiwangncar, line 4504: @ldfowler58 The commented out line is fine. The reason for this change is that in modern met dataset, fractional seaice, which is stored in xice, is typically provided. If one uses a temperature value to reset xice, it will potentially make fractional seaice to solid seaice for a mesh point. The 271 K seaice_threshold value is used in WRF to set seaice only if fractional seaice is not provided. This is the reason the default value for seaice_threshold is set to 100 K.

ldfowler58: Many thanks for the clarification. Makes sense to me.

mgduda: Should we update the PR description and code comments to incorporate the explanation of the 100K threshold from @weiwangncar ?

weiwangncar: @mgduda I agree with that. The comment inside the code can be made clearer too.

ldfowler58: As suggested, I included Wei's explantation to the description of the PR. Does this work for you?

src/core_init_atmosphere/Registry.xml
```diff
@@ -273,6 +273,9 @@
                      description="Whether to switch sea-ice threshold from 0.5 to 0.02"
                      possible_values="true or false"/>
 
+                <nml_option name="config_tsk_seaice_threshold" type="real" default_value="100." units="K"
```
mgduda, line 276: What do you think about moving this namelist option to a different namelist group from `&preproc_stages`? Also, would it be better to keep this option hidden by default to prevent users from changing it, misunderstanding its role and thinking that 100K is far colder than any SST we'd ever encounter on Earth?

src/core_init_atmosphere/mpas_init_atm_cases.F
weiwangncar: @ldfowler58 For this comment, it could say: " Freeze really cold ocean: setting xice to 1.0 for sea-surface temperatures colder than 271K has proved to produce too much seaice at high latitudes. The line below has been commented out.  Determination of using actual seaice fraction input or setting xice to 1.0 for sea-surface temperatures colder than a pre-defined value in the case where fractional seaice is not provided in the input is now moved to the subroutine physics_init_seaice (in module mpas_atmphys_initialize_real.F)."

src/core_init_atmosphere/Registry.xml
ldfowler58: Michael: Do you have a suggestion about moving config_tsk_seaice_threshold outside of preproc_stages? I do not think that the other namelist groups would work as well? Thanks.

src/core_init_atmosphere/mpas_init_atm_cases.F
ldfowler58: It looks like despite of what I write, this unfortunately does not fit your requirements. To avoid further back and forth, can one of you take over the writing as you wish? This would speed up the process.

weiwangncar: @ldfowler58 I'd suggest that we remove the comment as well as the line altogether, and just describe the change in the PR message.

mgduda: @ldfowler58 If it's alright with you, I could push some changes to the branch to update the code comment, commit message, and PR description. I like @weiwangncar's suggestion to make the code comment minimal, and to include in the commit message (and PR description) most of the explanation for why we're changing the code. But I can go with whatever you'd prefer.

src/core_init_atmosphere/Registry.xml
mgduda: Perhaps we could create a new namelist group for options that apply specifically to real-data initialization (mostly sea-ice at present). If it's alright with you, I could push some changes to this PR branch to handle this.

src/core_init_atmosphere/mpas_init_atm_cases.F
ldfowler58: OK with me. I would really appreciate it. As I wrote above, this would speed up resolving this PR.

src/core_init_atmosphere/Registry.xml
ldfowler58: since config_tsk_seaice_threshold is a physics related option, we could call the new namelist group "physics" or "physics_init". Most of those variables would remain "hidden" as they should not be changed.

mgduda: I like the name `&physics`. We can make that namelist group hidden by default.

src/core_init_atmosphere/mpas_init_atm_cases.F
mgduda: I've just removed the commented-out code and comment altogether. The commit message now contains an explanation for the code change here.

src/core_init_atmosphere/Registry.xml
mgduda: The `config_tsk_seaice_threshold` namelist option is now in a new `&physics` namelist group that is hidden by default.

ldfowler58: OK, I can create a new namelist group for config_tsk_seaice_threshold. I like &physics too. I will start working on this right away.

mgduda:
@weiwangncar I've updated the PR description and commits following your comments earlier. Have I mostly captured the idea correctly?

weiwangncar:
@mgduda The message looks good.

mgduda (approved):

PR #974, opened 2022-05-02
Fixes issue in calc_pvBudget subroutine and adapts floodFillStrato subroutine to allow for domain communication

theweathermanda:
This merge fixes an issue in the calc_pvBudget subroutine, which mistakenly called dtheta_dt_mp instead of the diabatic heating tendency from the microphysics scheme. The floodFillStrato subroutine was also modified to be consistent with the domain communication modifications that were previously added to the floodFillTropo subroutine.


theweathermanda:
Sorry - realized that dtheta_dt_mp was not being populated in our runs due to using outdated versions of other source files! Thus, the substitution was no longer necessary.

PR #975, opened 2022-05-12
fixed option for the seaice albedo

ldfowler58:
The parameter seaice_albedo_opt that initializes the seaice albedo was wrongly set to 2 instead of 0 in mpas_atmphys_vars.F. Setting seaice_albedo_opt to 2 assumes that we will read an input array (albsi) of the seaice albedo (for example observed values). This option is currently not available in MPAS. We set seaice_albedo_opt to 0 which initializes the seaice albedo to its default value (seaice_albedo_default) for all grid cells with fractional seaice vales greater than zero.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -758,7 +758,6 @@
 			<var name="noahres"/>
 			<var name="potevp"/>
 			<var name="sfc_albedo"/>
-			<var name="sfc_albedo_seaice"/>
 			<var name="sfc_emiss"/>
 			<var name="sfc_emibck"/>
 			<var name="sfcrunoff"/>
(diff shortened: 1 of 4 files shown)


weiwangncar (approved):
This PR sets the seaice albedo to a constant value of 0.75. This change makes it consistent with that in WRF in the case when albedo input over seaice is not available.

ldfowler58:
In the WRF release 4.4, the value of seaice_albedo_default is set to 0.65 and not 0.75. So I do not understand the change made by Wei.

weiwangncar:
@ldfowler58 Sorry, that was a typo. Also I did not make any change, that was just a comment.

ldfowler58:
OK. Thanks for reviewing the PR.

mgduda:
@ldfowler58 Should this PR to be included in the next MPAS minor release (v7.4)?

weiwangncar:
@mgduda @ldfowler58 I'd vote yes.

mgduda (approved):

PR #978, opened 2022-06-17
Ensure that calls to MPAS_coupler_transfer_field only happen from one thread

mgduda:
This PR addresses a correctness issue when MPAS-A is run with lagged radiation
and OpenMP threading by ensuring that calls to MPAS_coupler_transfer_field are
only made by one thread.

Various calls to MPAS_coupler_transfer_field, which uses MPI to transfer fields
between different roles, are made inside threaded regions in the MPAS-A physics
drivers. However, these calls to MPAS_coupler_transfer_field should only be made
by a single thread. These calls should only happen after all threads have
reached the call site, and furthermore, no thread should continue executing
until after the field transfers have completed.

To ensure correct execution when MPAS-A is running with lagged radiation (which
necessitates calls to transfer fields between radiation and dynamics ranks) and
OpenMP threading enabled, the calls to MPAS_coupler_transfer_field are preceded
by an OMP BARRIER, surrounded by OMP MASTER directives, and followed by another
OMP BARRIER.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_lw.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_lw.F
@@ -411,6 +411,8 @@
  if (mpas_cpl%role_is(ROLE_INTEGRATE) .or. &
      mpas_cpl%role_is(ROLE_RADIATION)) then
 
+!$OMP BARRIER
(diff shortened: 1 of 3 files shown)


gdicker1 (approved):
Makes sense and looks good to me.

PR #979, opened 2022-06-27
Add OpenACC checks to top-level Makefile

gdicker1:
Like in PR#976, this pull request helps to prepare for other OpenACC changes to be added to the develop branch. This PR adds the openacc_test target as a dependency for the mpas_main target which has two types of tests to check for expected behavior when OPENACC=true is provided. The "-ta=tesla:cc70,cc80" flag was removed from the pgi target because it would cause compilation to fail on systems without CUDA.

The first type of test checks if both FFLAGS_ACC and CFLAGS_ACC are blank. Since these variables are expected to contain the compiler specific flags for OpenACC compilation, it is unlikely that OpenACC directives will be interpreted and the *_ACC variables should be edited.

The second type of tests check if the C and FORTRAN compilers (SC, CC, CXX, SFC, and FC variables) can compile a test OpenACC program. The test program attempts to find the openacc header/module file as well as use an "acc kernels" directive around a for-loop. If no error is thrown during compilation, it is assumed that the compiler can compile OpenACC.

NOTE: It is currently difficult to be sure whether a compiler interpreted and compiled an OpenACC directive or if it was treated as a comment and ignored. For example, nvfortran will only output info about OpenACC compilation to STDERR if some equivalent of -Minfo=accel is provided.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ pgi:   # BUILDTARGET PGI compiler suite
 	"LDFLAGS_DEBUG = -O0 -g -Mbounds -Mchkptr -Ktrap=divz,fp,inv,ovf -traceback" \
 	"FFLAGS_OMP = -mp" \
 	"CFLAGS_OMP = -mp" \
-	"FFLAGS_ACC = -Mnofma -acc -ta=tesla:cc70,cc80 -Minfo=accel" \
+	"FFLAGS_ACC = -Mnofma -acc -Minfo=accel" \
 	"CFLAGS_ACC =" \
 	"PICFLAG = -fpic" \
 	"BUILD_TARGET = $(@)" \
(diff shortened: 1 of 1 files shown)


gdicker1:
The force push from 91928e5 to faed816 references a squashed commit. On a failed make attempt, files created by the openacc_test target (openacc.c, openacc.f90, etc) used to remain. Now, these files are removed before issuing the "exit 1" command.

mgduda (changes requested):
In general I think this looks good, with just a few minor issues to consider.

Makefile
```diff
@@ -951,7 +951,64 @@ ifeq "$(OPENMP)" "true"
...
+	    echo "************ ERROR ************"; exit 1) \
+	else \
+	   echo "=> FFLAGS_ACC or CFLAGS_ACC defined"; \
+	fi
+
+	@#
+	@# Create test C and FORTRAN programs that look for OpenACC header file and parallelize a loop
+	@#
+	@printf '#include <openacc.h>\nint main(){\n int CORES=acc_get_num_cores(); int i,n=0;\n #pragma acc kernels\n for (i=0; i<10; i++); n=n+i;\n return 0;}\n' > openacc.c
+	@printf 'program openacc1\n use openacc; external acc_get_num_cores\n integer :: i,n=0,CORES=0\n CORES=acc_get_num_cores()\n !$$acc kernels\n DO i=0,10; n=n+i; END DO\n !$$acc end kernels\nend program\n' > openacc.f90
```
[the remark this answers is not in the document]
gdicker1: I did the one-liner to try and limit the number of lines added to the Makefile, but this version would let someone inspect the test programs without having to remove the `rm` instructions. My preference leans towards the one-liner since I'm still new to `sed`.

Makefile
```diff
@@ -951,7 +951,64 @@ ifeq "$(OPENMP)" "true"
...
+	  echo "Simple OpenACC program could not be compiled by $(SFC)."; echo "Build output:"; cat openacc_bld.txt; \
+	  echo "************ ERROR ************"; \
+	  rm -rf openacc.* openacc_[cf].out openacc_bld.txt; exit 1)
+	@($(FC) -c openacc.f90 $(FCINCLUDES) $(FFLAGS) $(LDFLAGS) $(LIBS) -o openacc_f.out &> openacc_bld.txt && echo "=> $(FC) can compile simple OpenACC") || \
+	 (echo "************ ERROR ************"; \
+	  echo "Simple OpenACC program could not be compiled by $(FC)."; echo "Build output:"; cat openacc_bld.txt; \
+	  echo "************ ERROR ************"; \
+	  rm -rf openacc.* openacc_[cf].out openacc_bld.txt; exit 1)
+
+	@rm -rf openacc.* openacc_[cf].out openacc_bld.txt
```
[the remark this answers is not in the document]
mgduda: I think the `rm` command you've given would be just fine (and probably what I would do myself).

Makefile
```diff
@@ -951,7 +951,64 @@ ifeq "$(OPENMP)" "true"
...
+	@# Create test C and FORTRAN programs that look for OpenACC header file and parallelize a loop
+	@#
+	@printf '#include <openacc.h>\nint main(){\n int CORES=acc_get_num_cores(); int i,n=0;\n #pragma acc kernels\n for (i=0; i<10; i++); n=n+i;\n return 0;}\n' > openacc.c
+	@printf 'program openacc1\n use openacc; external acc_get_num_cores\n integer :: i,n=0,CORES=0\n CORES=acc_get_num_cores()\n !$$acc kernels\n DO i=0,10; n=n+i; END DO\n !$$acc end kernels\nend program\n' > openacc.f90
+
+	@#
+	@# See whether the test programs can be compiled
+	@#
+	@echo "Checking [$(BUILD_TARGET)] compilers for OpenACC support..."
+	@($(SCC) -c openacc.c $(CPPINCLUDES) $(CFLAGS) $(LDFLAGS) $(LIBS) -o openacc_c.out &> openacc_bld.txt && echo "=> $(SCC) can compile simple OpenACC") || \
```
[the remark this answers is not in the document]
mgduda: I actually have a PR in the works that rewrites the PIO tests, so I'll make sure to do redirects in a POSIX-compliant way in that PR. No need to worry about that for this PR, then.

Makefile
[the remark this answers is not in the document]
gdicker1: Push to a22bee0 should resolve this with multi-line test programs

gdicker1:
@mgduda see if the changes here work for you. I'm particularly interested if you think the error messages in the compilation test section are sufficient.

mgduda:
@gdicker1 If I add `-fopenacc` to the `gfortran` target, I get the following error from the build of the `openacc.f90` test program:
```
openacc.f90:4:11:

    4 |     n_devs=acc_get_num_devices( acc_device_current )
      |           1
Error: There is no specific function for the generic ‘acc_get_num_devices’ at (1)
```
From the OpenACC 2.7 standard:
```
Many of the routines accept or return a value corresponding to the type of device. In C and C++, the
datatype used for device type values is acc_device_t; in Fortran, the corresponding datatype
is integer(kind=acc_device_kind). The possible values for device type are implemen-
tation specific, and are defined in the C or C++ include file openacc.h and the Fortran module
openacc. Four values are always supported: acc_device_none, acc_device_default,
acc_device_host and acc_device_not_host. For other values, look at the appropriate
files included with the implementation, or read the documentation for the implementation. The
value acc_device_default will never be returned by any function; its use as an argument will
tell the runtime library to use the default device type for that implementation.
```

gdicker1:
@mgduda, thank you for checking that, when I change from `acc_device_current` to `acc_device_default` the code now builds as expected. Plus this led me to check the GCC implementation of OpenACC and the GCC installations on Casper.

Before I make the push with the fix, should I keep the compilation commands so the output goes to a `openacc_[cf].log` file or do you think it would be better to align this with PR#980 and send compilation output to `/dev/null`? I would argue it's nice to have command output present so a potential user doesn't have to re-run the command (and set up their modules/environment to re-run the command) so they can check how the compilation fails.

mgduda:
@gdicker1 I don't have a strong preference either way for leaving the log of the failed compilation or just sending all build output to `/dev/null`. My guess would be that -- if users are building the code interactively -- there wouldn't be a need to reload any module environment, and they should be able to simply copy-and-paste the build commands from the error message if it wasn't obvious to them why OpenACC might not be working on their system.

Makefile
```diff
@@ -951,7 +951,111 @@ ifeq "$(OPENMP)" "true"
 endif
 
 
-pio_test:
+openacc_test :
```
mgduda, line 954: It looks like there's a stray space at the end of this line. Also, for consistency with other make targets, should we remove the space before the `:`.

Makefile
```diff
@@ -951,7 +951,111 @@ ifeq "$(OPENMP)" "true"
...
+	    echo "ERROR: OPENACC=true was specified, but [$(BUILD_TARGET)] build target does not seem to support OpenACC:"; \
+	    echo "    FFLAGS_ACC and CFLAGS_ACC are both undefined for [$(BUILD_TARGET)] in the top-level Makefile."; \
+	    echo "Please set these variables to appropriate OpenACC compilation flags in the [$(BUILD_TARGET)] target to enable OpenACC support."; \
+	    echo "*********************************************************"; exit 1)\
+	else \
+	   echo "=> FFLAGS_ACC or CFLAGS_ACC defined"; \
+	fi
+
+	@#
+	@# Create test C and FORTRAN programs that look for OpenACC header file and parallelize a loop
```
mgduda, line 972: It's a minor stylistic issue, but these days I think Fortran is most often written with just a capital `F` (see, e.g., the Fortran 2003 draft standard), while all-caps seems to have been the prevalant style in decades past.

Makefile
```diff
@@ -951,7 +951,111 @@ ifeq "$(OPENMP)" "true"
...
+	@printf "program openacc\n \
+		&    use openacc\n \
+		&    integer :: i,n=0,n_devs=0\n \
+		&    n_devs=acc_get_num_devices( acc_device_current )\n \
+		&    !\$$acc kernels\n \
+		&    do i=0,10\n \
+		&      n=n+i\n \
+		&    end do\n \
+		&    !\$$acc end kernels\n \
+		&end program\n" | sed 's/^[ \t]*&//' > openacc.f90
```
mgduda, line 992: For the C test program, indentation uses a tab followed by spaces, while for the Fortran test program, it's two tabs. It might be nice to make these consistent (and my preference would lean towards the former).

Makefile
```diff
@@ -951,7 +951,111 @@ ifeq "$(OPENMP)" "true"
...
+	  && echo "=> $(SCC) can compile test OpenACC program" || \
+	 (echo "*********************************************************"; \
+	  echo "ERROR: Test OpenACC C program could not be compiled by $(SCC)."; \
+	  echo "Following compilation command failed with errors:" ; \
+	  echo "$(SCC) openacc.c $(CPPINCLUDES) $(CFLAGS) $(LDFLAGS) $(LIBS) -o openacc_c.out"; \
+	  echo ""; \
+	  echo "Test program openacc.c and output openacc_c.log have been left"; \
+	  echo "in the top-level MPAS directory for further debugging"; \
+	  echo "*********************************************************"; \
+	  rm -rf openacc.f90 openacc_f.out openacc_f.log; exit 1)
```
mgduda, line 1008: `openacc_f.out` and `openacc_f.log` should never exist at this point, right? Or is the idea to remove any failed Fortran test output from a previous build of MPAS?
Also, I think the `r` option may not be needed in the `rm` command.

Makefile
gdicker1: Thanks for the correction!

Makefile
gdicker1: Thanks for catching that. I agree that the test programs should be consistent and also should be indented by one tab and then spaces.

Makefile
gdicker1: Yes to the second question. My idea with these `rm` commands after the error message was that only the relevant files will be left after a failed build. So on a first build, these files shouldn't exist but they may be there if someone had a failed build and adjusted their inputs to make. That said, I do think I need to change such `rm` commands a little so they remove any openacc_[cf].out that might exist. Those would also be possible extraneous files from the build.

I agree to the `r` option being unneeded. That will be removed.

Makefile
```diff
@@ -938,7 +938,111 @@ ifeq "$(OPENMP)" "true"
...
+	@#
+	@printf "#include <openacc.h>\n\
+	        &int main(){\n\
+	        &    int n_devs=acc_get_num_devices( acc_device_default );\n\
+	        &    int i,n=0;\n\
+	        &    #pragma acc kernels\n\
+	        &    for (i=0; i<10; i++)\n\
+	        &      n=n+i;\n\
+	        &    return 0;\n\
+	        &}\n" | sed 's/^[ \t]*&//' > openacc.c
```
mgduda, line 969: When indenting test programs with tab followed by spaces, I think we can omit the `\t` in the `sed` commands.
From the GNU Make manual on [Splitting Recipe Lines](https://www.gnu.org/software/make/manual/make.html#Splitting-Recipe-Lines):

> However, in contrast to how they are treated in other places in a makefile (see [Splitting Long Lines](https://www.gnu.org/software/make/manual/make.html#Splitting-Lines)), backslash/newline pairs are not removed from the recipe. Both the backslash and the newline characters are preserved and passed to the shell. How the backslash/newline is interpreted depends on your shell. If the first character of the next line after the backslash/newline is the recipe prefix character (a tab by default; see [Special Variables](https://www.gnu.org/software/make/manual/make.html#Special-Variables)), then that character (and only that character) is removed. Whitespace is never added to the recipe.

Makefile
```diff
@@ -938,7 +938,111 @@ ifeq "$(OPENMP)" "true"
...
+	        &      n=n+i\n\
+	        &    end do\n\
+	        &    !\$$acc end kernels\n\
+	        &end program\n" | sed 's/^[ \t]*&//' > openacc.f90
+
+	@#
+	@# See whether the test programs can be compiled
+	@#
+	@echo "Checking [$(BUILD_TARGET)] compilers for OpenACC support..."
+	@$(SCC) openacc.c $(CPPINCLUDES) $(CFLAGS) $(LDFLAGS) $(LIBS) -o openacc_c.out > openacc_c.log 2>&1 \
```
mgduda, line 985: It's probably a matter of taste, but I find that the use of more than one of either `&&` or `||`, especially when spanning multiple lines, leads to shell that is harder to follow. As an alternative, something like the following could be a bit more readable:
```
        @( $(SCC) openacc.c $(CPPINCLUDES) $(CFLAGS) $(LDFLAGS) $(LIBS) -o openacc_c.out > openacc_c.log 2>&1; \
           if [ $$? -eq 0 ]; then \
               echo "=> $(SCC) can compile test OpenACC program"; \
           else \
               echo "*********************************************************"; \
               echo "ERROR: Test OpenACC C program could not be compiled by $(SCC)."; \
               echo "Following compilation command failed with errors:" ; \
               echo "$(SCC) openacc.c $(CPPINCLUDES) $(CFLAGS) $(LDFLAGS) $(LIBS) -o openacc_c.out"; \
               echo ""; \
               echo "Test program openacc.c and output openacc_c.log have been left"; \
               echo "in the top-level MPAS directory for further debugging"; \
               echo "*********************************************************"; \
               rm -f openacc.f90 openacc_[cf].out openacc_f.log; exit 1; \
           fi )
```

mgduda (changes requested):
Just a few final points to consider; otherwise, this looks good.

Makefile
gdicker1: I agree, the `\t` can be ommited. When I was playing around with this sed command I figured out what the documentation said through trial and error.

Makefile
gdicker1: The version with `&&` and `||` definitely has some more "arcane" shell qualities as opposed to using if/else. I agree this should be changed.

gdicker1:
@mgduda, I edited the sed command and changed my tests to use if-else commands instead of `&&` and `||` in the push to be5f4de. Let me know if there's anything else you want me to look at!

mgduda (approved):
Looks good to me!

PR #980, opened 2022-07-07
Automatically detect the PIO library version in the top-level Makefile

mgduda:
This PR enables the top-level Makefile to automatically detect the correct PIO library
version.

Previously, the MPAS build system assumed that a PIO 1.x library version was to
be used unless the `USE_PIO2` option was set to `true` in the compilation command.
Additionally, the top-level Makefile ran a check to verify that the available
PIO version matched what was assumed (1.x) or specified (2.x), and in case of a
mismatch, the build stopped with an error.

Given that we are able to reliably detect whether a valid PIO 1.x or 2.x library
is available through the compilation of test programs, there is no need for the
user to specify which PIO version they have. Instead, we can just automatically
detect this at build time in the top-level Makefile.

In case no PIO library is detected, an error message is printed. Included in
this message are the compilation commands that were tried, along with a note
that the test programs to be compiled by those commands may be found in the
top-level MPAS directory.

Also included in the changes to the Makefile is a print statement to tell the user
that the USE_PIO2 option has been deprecated in case it was specified.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -573,13 +573,6 @@ FCINCLUDES =
 LIBS =
 
 #
-# If user has indicated a PIO2 library, define USE_PIO2 pre-processor macro
-#
-ifeq "$(USE_PIO2)" "true"
-	override CPPFLAGS += -DUSE_PIO2
-endif
-
-#
 # Regardless of PIO library version, look for a lib subdirectory of PIO path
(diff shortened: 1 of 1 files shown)


gdicker1 (approved):
Looks good to me and the expected output was present when building. It might be better to merge this before merging those in PR#979

PR #981, opened 2022-07-11
Report runtime settings for MPI and OpenMP in [init_]atmosphere log files

mgduda:
This PR adds code to report runtime settings for MPI and OpenMP near the beginning
of log files for the init_atmosphere and atmosphere cores.

The setup_log routines for the init_atmosphere and atmosphere cores now write
messages about the runtime values of MPI task count and maximum OpenMP thread
count to log files. For example:
```
 Run-time settings:
   MPI task count: 6
   OpenMP max threads: 4
```
This information can be helpful in determining the parallel runtime settings
that were used for a run when only log files are available.

Note that, if OPENMP=true was not specified in the build options, then no
reporting of the maximum OpenMP thread count is performed. This allows for
disambiguation of cases where the model was run with a single OpenMP thread or
with no OpenMP support at all.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core_interface.F
+++ b/src/core_atmosphere/mpas_atm_core_interface.F
@@ -240,6 +240,9 @@ module atm_core_interface
 
       use mpas_derived_types, only : mpas_log_type, domain_type
       use mpas_log, only : mpas_log_init, mpas_log_open
+#ifdef MPAS_OPENMP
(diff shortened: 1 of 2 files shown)


gdicker1 (approved):
Simple change that only affects logging and uses information already captured in the domain. I think this could be added whenever.

I tested on Cheyenne using GNU compilers with and without OPENMP=true and got the expected outputs.

PR #983, opened 2022-07-14
Remove block loops from mpas_atm_time_integration.F

gdicker1:
A MPAS domain encompasses the entire problem space being simulated and is decomposed across blocks when running in parallel. To maintain generality it was not initially assumed that one thread would be associated with one block (i.e. one thread could have multiple blocks or blocks could be shared across threads). This means that any operation that would modify a variable in a subpool (e.g. horizontal wind speed `u`), would need to start at the head of the blocklist, fetch the necessary subpools from the block, fetch arrays/fields/dimensions from subpools, perform computation, and then repeat for each block in the blocklist. These "block loops" and are illustrated in the example below. Since there should now only be one block per thread, the block loops, the whitespace they consumed, and repeated fetching of pointers is unnecessary.

This pull request removes the block loops from `mpas_atm_time_integration.F`, reclaims the blank lines and columns used for the loops, and moves the calls to `mpas_pool_get_subpool` and `mpas_pool_get_dimensions` within `atm_srk3` to the top of the subroutine where feasible. 

General form of a block loop:
```
block => domain % blocklist
do while (associated(block))
   ! Get the subpools that contain arrays, fields, or dimensions that are needed
   call mpas_pool_get_subpool(block % structs, 'state', state)
   call mpas_pool_get_subpool(block % structs, 'mesh', mesh)

   ! Get the dimensions and arrays needed
   call mpas_pool_get_dimension(state, 'nThreads', nThreads)
   call mpas_pool_get_dimension(mesh, 'nCells', nCells)
   call mpas_pool_get_array(state, 'u', u)

    ...
   ! Code that updates u for this block
   ...
   ! Advance to the next block in the blocklist
   block => block % next
end do
```

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -203,8 +203,8 @@ module atm_time_integration
       clock => domain % clock
       block => domain % blocklist
 
-      call mpas_pool_get_config(domain % blocklist % configs, 'config_time_integration', config_time_integration)
-      call mpas_pool_get_config(domain % blocklist % configs, 'config_apply_lbcs', config_apply_lbcs)
+      call mpas_pool_get_config(block % configs, 'config_time_integration', config_time_integration)
+      call mpas_pool_get_config(block % configs, 'config_apply_lbcs', config_apply_lbcs)
 
       if (trim(config_time_integration) == 'SRK3') then
          call atm_srk3(domain, dt, itimestep)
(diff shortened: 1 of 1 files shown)


mgduda:
I've tested the following dynamics configurations to verify that we get bit-identical results before and after the changes in this PR:

- [ ] config_time_integration_order=2, config_split_dynamics_transport=true, config_number_of_sub_steps=2, config_dynamics_split_steps=3

mgduda (approved):
This looks great to me. It's always nice when we remove more lines in a PR than we add!

skamaroc (approved):

PR #984, opened 2022-07-20
Remove whitespace in Makefile

stephenmsachs:
This seems trivial but I have actually written an sed parser that trips over
this here:
https://github.com/spack/spack/blob/43673fee808f9e02efcb4330c6a7fa2c9b80c14c/var/spack/repos/builtin/packages/mpas-model/package.py#L74

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ xlf:
 	"USE_PAPI = $(USE_PAPI)" \
 	"OPENMP = $(OPENMP)" \
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI" )
- 
+
 ftn:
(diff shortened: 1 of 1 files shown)


mgduda:
@stephenmsachs Thanks for opening this PR. The change looks good to me, of course. One small request, though: could you update the commit message and PR description to describe in a sentence or two what the change accomplishes without referencing a link that may disappear in future? Something to the effect of "Remove a space on an otherwise blank line separating two build targets in the top-level Makefile for consistency."? If you don't mind, I can also just force-push an update to your branch.

stephenmsachs:
Closing in favor of https://github.com/MPAS-Dev/MPAS-Model/pull/987

PR #986, opened 2022-08-10
Fix PV registry descriptions

theweathermanda:
In the Registry, the pv_vertex, pv_edge, and pv_cell variables are described as "absolute vorticity/rho_zz", but the density component was previously removed from the model code. This request updates the descriptions and units of these variables to be consistent with the absolute vertical vorticity. 

In Line 4816 of mpas_atm_time_integration.F (v7.3): "the original definition of pv_edge had a factor of 1/density. We have removed that factor given that it was not integral to any conservation property of the system"

Relative vertical vorticity is calculated beginning at Line 5606 (in v7.3) as:

```
      do iVertex=vertexStart,vertexEnd
         vorticity(1:nVertLevels,iVertex) = 0.0
         do i=1,vertexDegree
            iEdge = edgesOnVertex(i,iVertex)
            s = edgesOnVertex_sign(i,iVertex) * dcEdge(iEdge)
!DIR$ IVDEP
            do k=1,nVertLevels
               vorticity(k,iVertex) = vorticity(k,iVertex) + s * u(k,iEdge)
            end do
         end do
!DIR$ IVDEP
         do k=1,nVertLevels
            vorticity(k,iVertex) = vorticity(k,iVertex) * invAreaTriangle(iVertex)
         end do
      end do

```
(with 'u' as the uncoupled normal component of horizontal velocity), and planetary vorticity is added to get absolute vorticity at Line 5754:

`            pv_vertex(k,iVertex) = (fVertex(iVertex) + vorticity(k,iVertex))`

Thus, pv_vertex is the absolute vertical vorticity at a vertex.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -1565,8 +1565,8 @@
                 <var name="vorticity" type="real" dimensions="nVertLevels nVertices Time" units="s^{-1}"
                      description="Relative vorticity at vertices"/>
 
-                <var name="pv_edge" type="real" dimensions="nVertLevels nEdges Time" units="s^{-1} kg^{-1} m^3"
-                     description="absolute vorticity/rho_zz averaged to the cell edge from the vertices"/>
+                <var name="pv_edge" type="real" dimensions="nVertLevels nEdges Time" units="s^{-1}"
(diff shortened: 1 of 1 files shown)


mgduda (approved):
Looks good to me!

PR #988, opened 2022-08-16
Eliminate the use of hard-coded Fortran unit numbers in atmosphere code

gdicker1:
Some files contained hard-coded numbers for file units instead of searching for unused units. This PR changes those in the atmosphere core to instead use `mpas_io_units` via the `mpas_new_unit` and `mpas_release_unit` subroutines. This ensures that MPAS can be used in more contexts where it isn't certain whether a particular unit (e.g. 27) is already in use.

This PR copies and extends the work done in JCSDA-internal/MPAS-Model PR#12

### DIFF
--- a/src/core_atmosphere/diagnostics/soundings.F
+++ b/src/core_atmosphere/diagnostics/soundings.F
@@ -232,6 +232,7 @@ module soundings
         use mpas_derived_types, only : MPAS_Time_type, MPAS_NOW
         use mpas_timekeeping, only : MPAS_is_alarm_ringing, MPAS_reset_clock_alarm, MPAS_get_clock_time, MPAS_get_time
         use mpas_constants, only : rvord
+        use mpas_io_units, only: mpas_new_unit, mpas_release_unit
 
         implicit none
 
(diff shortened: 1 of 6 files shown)


gdicker1:
This was tested with a default run of the 120km benchmark. I was hoping this would pass through visual inspection. If necessary I can do further tests to be sure the changes (e.g. to Thompson microphysics or soundings diagnostic) are correct.

mgduda (changes requested):
Since the `module_mp_thompson.F` code is shared with WRF (though the versions may have drifted apart slightly), it may be good to try our best at keeping our version compatible with WRF. As was done in `module_ra_cam_support.F`, it may be good to surround calls to `mpas_new_unit` and `mpas_release_unit` with `#if defined(mpas)` guards, and for non-MPAS builds, to simply set `mp_unit` to 11. We can also move the use of the `mpas_io_units` module up to the top of the file near the use of other MPAS-specific modules (although those aren't guarded by an `#if defined(mpas)` directive, at least they'll all be in the same place).

gdicker1:
@mgduda let me know if you have anything else! Commit [f6f5662](https://github.com/MPAS-Dev/MPAS-Model/pull/988/commits/f6f5662a9541db2b5577a90df6826cc29d6ca4b7) should fix what you asked for so far!

mgduda (approved):
This PR looks great to me. Thanks!

PR #989, opened 2022-08-18
Fix memory leak in deallocate_microphysics for recloud_p, reice_p, and resnow_p

mgduda:
This PR fixes a memory leak in the deallocate_microphysics routine for the recloud_p,
reice_p, and resnow_p arrays.

The logic for deallocating the recloud_p, reice_p, and resnow_p fields was incorrect,
leading to a memory leak at the end of a simulation. Rather than checking if these arrays
are allocated and allocating them if they are not, the code now checks whether these
arrays are allocated and deallocates them if they are.


ldfowler58:
I reviewed the PR and totally agree with the memory leak fix. I have not idea what was going on in my head when I did this several years ago. Thanks.

ldfowler58 (approved):

PR #990, opened 2022-08-18
Add optional arguments for namelist and streams files to mpas_init

gdicker1:
These changes allows for those that create their own driver script to override the default namelist and streams files by adding optional arguments `namelistFileParam` and `streamsFileParam` to the `mpas_init` routine. Since the default files can also be overriden by supplying the `-n` and `-s` command-line arguments, if the optional `FileParam` arguments to `mpas_init` are provided the command-line arguments aren't parsed by `mpas_init`.

This PR copies and extends the work from JCSDA-internal/MPAS-Model PR#10

### DIFF
--- a/src/driver/mpas_subdriver.F
+++ b/src/driver/mpas_subdriver.F
@@ -116,7 +116,8 @@ module mpas_subdriver
          if (len_trim(namelistFileParam) == 0) then
             write (0,*) 'WARNING: mpas_init argument namelistFileParam has 0 length and will be ignored'
          else if (len_trim(namelistFileParam) > len(namelistFile)) then
-            write (0,*) 'CRITICAL ERROR: mpas_init argument namelistFileParam is too long for namelistFile variable.'
+            write(0,'(A,I5,A,I5,A)') 'CRITICAL ERROR: mpas_init argument ''namelistFileParam'' has length ',&
+                                     len_trim(namelistFileParam), ', but the maximum allowed is ', len(namelistFile), ' characters'
             stop
          else
             readNamelistArg = .true.
@@ -128,7 +129,8 @@ module mpas_subdriver
          if (len_trim(streamsFileParam) == 0) then
             write (0,*) 'WARNING: mpas_init argument streamsFileParam has 0 length and will be ignored'
          else if (len_trim(streamsFileParam) > len(streamsFile)) then
-            write(0,*) 'CRITICAL ERROR: mpas_init argument streamsFileParam is too long for streamsFile variable.'
+            write(0,'(A,I5,A,I5,A)') 'CRITICAL ERROR: mpas_init argument ''streamsFileParam'' has length ',&
+                                     len_trim(streamsFileParam), ', but the maximum allowed is ', len(streamsFile), ' characters'
             stop
          else
             readStreamsArg = .true.
--- a/src/driver/mpas_subdriver.F
+++ b/src/driver/mpas_subdriver.F
@@ -39,7 +39,7 @@ module mpas_subdriver
    contains
 
 
-   subroutine mpas_init(corelist, domain_ptr, mpi_comm)
+   subroutine mpas_init(corelist, domain_ptr, mpi_comm, namelistFileParam, streamsFileParam)
 
       use mpas_stream_manager, only : MPAS_stream_mgr_init, MPAS_build_stream_filename, MPAS_stream_mgr_validate_streams
       use iso_c_binding, only : c_char, c_loc, c_ptr, c_int
@@ -53,6 +53,8 @@ module mpas_subdriver
       type (core_type), intent(inout), pointer :: corelist
       type (domain_type), intent(inout), pointer :: domain_ptr
       integer, intent(in), optional :: mpi_comm
+      character(len=*), intent(in), optional :: namelistFileParam
+      character(len=*), intent(in), optional :: streamsFileParam
 
       integer :: iArg, nArgs
       logical :: readNamelistArg, readStreamsArg
@@ -108,6 +110,15 @@ module mpas_subdriver
 
       readNamelistArg = .false.
       readStreamsArg = .false.
+
+      if (present(namelistFileParam) .and. len_trim(namelistFileParam) > 0 ) then
+         readNamelistArg = .true.
+         namelistFile = trim(namelistFileParam)
+      endif
+      if (present(streamsFileParam)  .and. len_trim(streamsFileParam) > 0 ) then
+         readStreamsArg = .true.
+         streamsFile = trim(streamsFileParam)
+      endif
 
       nArgs = command_argument_count()
       iArg = 1
--- a/src/driver/mpas_subdriver.F
+++ b/src/driver/mpas_subdriver.F
@@ -111,47 +111,66 @@ module mpas_subdriver
       readNamelistArg = .false.
       readStreamsArg = .false.
 
-      if (present(namelistFileParam) .and. len_trim(namelistFileParam) > 0 ) then
-         readNamelistArg = .true.
-         namelistFile = trim(namelistFileParam)
-      endif
-      if (present(streamsFileParam)  .and. len_trim(streamsFileParam) > 0 ) then
-         readStreamsArg = .true.
-         streamsFile = trim(streamsFileParam)
-      endif
-
-      nArgs = command_argument_count()
-      iArg = 1
-      do while (iArg < nArgs)
-         call get_command_argument(iArg, argument)
-         if (len_trim(argument) == 0) exit
-
-         if ( trim(argument) == '-n' ) then
-            iArg = iArg + 1
+      ! If provided, error check the namelistFileParam and copy it to namelistFile to override default
+      if (present(namelistFileParam)) then
+         if (len_trim(namelistFileParam) == 0) then
+            write (0,*) 'WARNING: mpas_init argument namelistFileParam has 0 length and will be ignored'
+         else if (len_trim(namelistFileParam) > len(namelistFile)) then
+            write (0,*) 'CRITICAL ERROR: mpas_init argument namelistFileParam is too long for namelistFile variable.'
+            stop
+         else
             readNamelistArg = .true.
-            call get_command_argument(iArg, namelistFile)
-            if ( len_trim(namelistFile) == 0 ) then
-                write(0,*) 'ERROR: The -n argument requires a namelist file argument.'
-                stop
-            else if ( trim(namelistFile) == '-s' ) then
-                write(0,*) 'ERROR: The -n argument requires a namelist file argument.'
-                stop
-            end if
-         else if ( trim(argument) == '-s' ) then
-            iArg = iArg + 1
-            readStreamsArg = .true.
-            call get_command_argument(iArg, streamsFile)
-            if ( len_trim(streamsFile) == 0 ) then
-                write(0,*) 'ERROR: The -s argument requires a streams file argument.'
-                stop
-            else if ( trim(streamsFile) == '-n' ) then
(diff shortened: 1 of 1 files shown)


gdicker1:
This was tested with the 120km benchmark atmosphere core testcase with multiple tests outlined below.

- Provide empty strings for `namelistFileParam` and `streamsFileParam`
     - Expected behavior resulted: warnings about the `namelistFileParam` and `streamsFileParams` having 0 length were printed and the default `namelist.atmosphere` and `streams.atmosphere` were used instead. Note: because of the implementation, command-line arguments would be ignored.
- Providing strings in `namelistFileParam` and `streamsFileParam` that are longer than the size of `namelistFile` and `streamsFile`.
     - Expected behavior resulted: error messages about the size mismatch were printed and the model was stopped.
- Pointing `namelistFileParam` and `streamsFileParam` to files with no content (empty files).
     - In this case the model tried to load default values for content that would have been in the namelist file. However, once the model progressed to reading the empty streams file a critical error caused the simulation to end. This seems like appropriate behavior to me.
- Providing both `FileParam` optional arguments and `-n` and `-s` command-line arguments.
     - Expected behavior resulted: the command-line arguments were ignored and the simulation carried out normally.

src/driver/mpas_subdriver.F
```diff
@@ -109,38 +111,66 @@ end subroutine xml_stream_get_attributes
...
-         if (len_trim(argument) == 0) exit
-
-         if ( trim(argument) == '-n' ) then
-            iArg = iArg + 1
+      ! If provided, error check the namelistFileParam and copy it to namelistFile to override default
+      if (present(namelistFileParam)) then
+         if (len_trim(namelistFileParam) <= 0) then
+            write (0,*) 'WARNING: mpas_init argument namelistFileParam has 0 length and will be ignored'
+         else if (len_trim(namelistFileParam) > len(namelistFile)) then
+            write (0,*) 'CRITICAL ERROR: mpas_init argument namelistFileParam is too long for namelistFile variable.'
```
mgduda, line 119: Here and for the streamsFile handling, we could print the length of the argument as well as the maximum acceptable length (i.e., the length of the local variable to which the argument will be assigned). This could help reduce debugging time for cases, e.g., where a long string with corrupted contents or other garbage is accidentally passed to the function: being told that the namelist filename is, say, 1024 characters long could lead to a quicker diagnosis of the issue.

gdicker1: @mgduda, how does it look now after I pushed? I pretty much wrote what I would put in a debug print statement for my use, so there's probably a better format.

src/driver/mpas_subdriver.F
```diff
@@ -109,38 +111,72 @@ end subroutine xml_stream_get_attributes
...
-                stop
-            end if
-         else if ( trim(argument) == '-s' ) then
-            iArg = iArg + 1
+            namelistFile = trim(namelistFileParam)
+         endif
+      endif
+      ! If provided, error check the streamsFileParam and copy it to streamsFile to override default
+!      if (present(streamsFileParam)) then
+      if (.true.) then
```
mgduda, line 130: It looks like some debugging changes may have been pushed to the PR branch on GitHub?

src/driver/mpas_subdriver.F
```diff
@@ -109,38 +111,72 @@ end subroutine xml_stream_get_attributes
...
-
-         if ( trim(argument) == '-n' ) then
-            iArg = iArg + 1
+      ! If provided, error check the namelistFileParam and copy it to namelistFile to override default
+      if (present(namelistFileParam)) then
+         if (len_trim(namelistFileParam) <= 0) then
+            write(0,"(A)") 'WARNING: mpas_init argument namelistFileParam has 0 or negative length and will be ignored'
+         else if (len_trim(namelistFileParam) > len(namelistFile)) then
+            write(0,"(A)") 'CRITICAL ERROR: mpas_init argument namelistFileParam is too long for namelistFile variable.'
+            write(0,"(A,I5,A,I5)") 'len_trim(namelistFileParam)=', len_trim(namelistFileParam), &
```
mgduda, line 120: It's a minor issue, but we've generally tended to use single-quotes whenever possible (one exception is when writing strings that contain single quotes, where it can sometimes be clearer to enclose the entire string in double-quotes).

src/driver/mpas_subdriver.F
```diff
@@ -109,38 +111,72 @@ end subroutine xml_stream_get_attributes
...
+      ! If provided, error check the streamsFileParam and copy it to streamsFile to override default
+!      if (present(streamsFileParam)) then
+      if (.true.) then
+!         if (len_trim(streamsFileParam) <= 0) then
+!            write(0,"(A)") 'WARNING: mpas_init argument streamsFileParam has 0 or negative length and will be ignored'
+!         else if (len_trim(streamsFileParam) > len(streamsFile)) then
+          if (.true.) then
+            write(0,"(A)") 'CRITICAL ERROR: mpas_init argument streamsFileParam is too long for streamsFile variable.'
+            write(0,"(A,I5,A,I5)") 'len_trim(streamsFileParam)=', len_trim(streamsFileParam), &
+                                   ' > len(streamsFile)=', len(streamsFile)
```
mgduda, line 137: In my opinion, including internal Fortran variable names and intrinsics (`LEN` and `LEN_TRIM`) in the message here may not be the clearest way to communicate the error. I'm of course open to other ideas, but one suggestion for this message would be something like:
```
'The ''streamsFileParam'' argument to mpas_init() has length ', len_trim(streamsFileParam), ', which exceeds'
'the maximum allowable length of ', len(streamsFile), ' characters.'
```

mgduda (changes requested):

src/driver/mpas_subdriver.F
mgduda: Since the message is meant for human consumption, I've offered an alternative in a review comment that (in my opinion) may state the issue a little more clearly.

src/driver/mpas_subdriver.F
gdicker1: Yep! That's what I get for editing in a separate window.

src/driver/mpas_subdriver.F
gdicker1: That's a standard I can stick to. Thanks!

src/driver/mpas_subdriver.F
gdicker1: Because I'm just rephrasing the print statements and fixing the above things, these have been addressed by a fixup commit and force push (from f26a276 to 47acce4). What do you think now?

src/driver/mpas_subdriver.F
```diff
@@ -109,38 +111,68 @@ end subroutine xml_stream_get_attributes
...
-            end if
-         else if ( trim(argument) == '-s' ) then
-            iArg = iArg + 1
+            namelistFile = trim(namelistFileParam)
+         endif
+      endif
+      ! If provided, error check the streamsFileParam and copy it to streamsFile to override default
+      if (present(streamsFileParam)) then
+         if (len_trim(streamsFileParam) <= 0) then
+            write(0,'(A)') 'WARNING: mpas_init argument ''streamsFileParam'' has 0 or negative length and will be ignored'
```
mgduda, line 130: I forgot to add this to my previous review, but I don't think `len_trim` will ever return a negative number, so perhaps mentioning (and testing for) a negative length here may be unnecessarily confusing?
Here's what the Fortran 2003 standard has to say about the return value of `len_trim`:
```
The result has a value equal to the number of characters remaining after any
trailing blanks in STRING are removed. If the argument contains no nonblank characters, the
result is zero.
```

src/driver/mpas_subdriver.F
```diff
@@ -109,38 +111,68 @@ end subroutine xml_stream_get_attributes
...
-         if (len_trim(argument) == 0) exit
-
-         if ( trim(argument) == '-n' ) then
-            iArg = iArg + 1
+      ! If provided, error check the namelistFileParam and copy it to namelistFile to override default
+      if (present(namelistFileParam)) then
+         if (len_trim(namelistFileParam) <= 0) then
+            write(0,'(A)') 'WARNING: mpas_init argument ''namelistFileParam'' has 0 or negative length and will be ignored'
+         else if (len_trim(namelistFileParam) > len(namelistFile)) then
+            write(0,'(A,I5,A,I5,A)') 'CRITICAL ERROR: mpas_init argument ''namelistFileParam'' has length ', len_trim(namelistFileParam),&
```
mgduda, line 119: Let's wrap this line to ensure it's at most 132 columns (and similarly for the other `CRITICAL ERROR` write statement. (I do realize that there are other lines in this source file that exceed 132 column, but they got in before we decided that standards compliance was a good thing; we can fix those other lines another time...).

Thus spake the Fortran 2003 standard:
```
In free source form there are no restrictions on where a statement (or portion of a statement) may
appear within a line. A line may contain zero characters. If a line consists entirely of characters of
default kind (4.4.4), it may contain at most 132 characters. If a line contains any character that is not
of default kind, the maximum number of characters allowed on the line is processor dependent.
```

src/driver/mpas_subdriver.F
```diff
@@ -109,38 +111,68 @@ end subroutine xml_stream_get_attributes
...
-                write(0,*) 'ERROR: The -n argument requires a namelist file argument.'
-                stop
-            else if ( trim(namelistFile) == '-s' ) then
-                write(0,*) 'ERROR: The -n argument requires a namelist file argument.'
-                stop
-            end if
-         else if ( trim(argument) == '-s' ) then
-            iArg = iArg + 1
+            namelistFile = trim(namelistFileParam)
+         endif
```
mgduda, line 125: Another minor stylistic issue: for the same reason we write `else if` with a space, I think `end if` is preferable to `endif`. Also, one can `end` many things in Fortran: `do`, `subroutine`, `program`, etc., so why not `if` as well?

src/driver/mpas_subdriver.F
gdicker1: To avoid too many force pushes, would the below be an acceptable re-write? In case the formatting doesn't show correctly, I'm trying to align `len_trim` with the first quote mark before `CRITICAL`.

>              write(0,'(A,I5,A,I5,A)') 'CRITICAL ERROR: mpas_init argument ''namelistFileParam'' has length ',&
>                                       len_trim(namelistFileParam), ', but the maximum allowed is ', len(namelistFile), ' characters'

mgduda: That seems fine to me. (I'm not sure if GitHub changed the indentation or if I copied the text incorrectly, but as long as we're indented three columns within the `else if`, that should work.)

src/driver/mpas_subdriver.F
gdicker1: This just came from me writing my if tests to be really exact. I agree it would be better without the negative length check/print confusing the issue.

src/driver/mpas_subdriver.F
gdicker1: Gotcha, that can be easily fixed.

src/driver/mpas_subdriver.F
mgduda: In general I agree that writing conditional statements so that they completely cover all possibilities is the correct thing to do. I suppose (not that we need to make yet another amendment for this PR) we could have tested whether `len_trim(streamsFileParam) <= 0`, but have the write statement say something less algebraic, e.g., "streamsFileParam is empty and will be ignored.".

gdicker1:
@mgduda, could you check again? (I rebased to get some of the last changes in, thus the force push).

mgduda (approved):
Thanks for the changes. This looks good to me!

PR #992, opened 2022-09-16
Remove 'hx' and 'advCells' from init_atmosphere input and output streams

mgduda:
This PR removes the `hx` and `advCells` fields from the input and output streams in the init_atmosphere core.
These two fields are not needed in the MPAS-Atmosphere model initial conditions, and furthermore they are
each only used in a single processing stage within the init_atmosphere core.


gdicker1 (approved):
Seems reasonable. Passes visual inspection for me.

PR #997, opened 2022-10-24
Pre processing: scripts for downloading ERA5 data for running MPAS-A

daniloceano:
GetERA5{suffix} scripts download ERA5 data for running the MPAS-A model. GetERA5_MPAS-RUN.sh call the subroutines that download initial conditions and surface fields for creating update files through the model integration.


daniloceano:
Wrong repo.

PR #999, opened 2022-10-27
Start file names in atmosphere diagnostics with "mpas_"

gdicker1:
This is a follow-on PR to PR#993 and seeks to prevent file name conflicts when MPAS-A is compiled as part of a larger system.

NOTE: this is not a concern in stand-alone MPAS-A since name collisions would prevent builds from succeeding.


gdicker1:
Let me know what would be the best way to reference the [other PR](https://github.com/MPAS-Dev/MPAS-Model/pull/993) or if this description needs more work.

mgduda (approved):

mgduda:
> Let me know what would be the best way to reference the [other PR](https://github.com/MPAS-Dev/MPAS-Model/pull/993) or if this description needs more work. I didn't think a Markdown-style link would be good for the eventual merge-commit message.

It may not be entirely necessary to reference PR #993 in the description/commit message for this PR. Here's what I was planning to use for the merge commit message:
```
    This merge modifies the filenames of MPAS-A diagnostics modules so that each
    filename begins with the prefix "mpas_". While this change has no practical
    impacts for stand-alone MPAS-A, it may be useful when MPAS-A source files are
    included in a larger system (e.g., CESM) to prevent filename collisions or
    other build-system issues.
```

gdicker1:
That looks like a better commit message than the path I was on. Looks good to me!

PR #1000, opened 2022-10-28
Variable grid

daniloceano:
Changed the spherical_grid.py for allowing the user to specify grid parameters: grid spacing of the high resolution area, radius of the high resolution area, radius of the transition zone and global grid spacing (low resolution)


daniloceano:
Wrong repo again. Sorry.

PR #1005, opened 2022-11-08
Improve performance of parallel static field remapping by bounding k-d tree searches

mgduda:
This PR significantly improves the performance of parallel static field 
remapping in the init_atmosphere core by bounding searches for the nearest cell 
to each pixel of a static field. Aside from the positive impact on runtime
performance, the changes in this merge have no impact on the resulting fields.

Previously, searches for the nearest cell to a static field pixel that was 
sufficiently far from any point in the k-d tree of cells could effectively
trigger a search through all or nearly all branches in the tree. Since we
ultimately only want to consider searches that result in an MPAS mesh cell that
actually contains the pixel, and not just the nearest cell -- which may in fact be
geographically far from the pixel -- a reasonable upper bound for the search
distance is based on the maximum diameter of any cell in the block of cells
assigned to an MPI task.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -99,6 +99,8 @@
  integer, pointer :: category_min
  integer, pointer :: category_max
 
+ real(kind=RKIND) :: max_kdtree_distance2
+
  contains
 
 !==================================================================================================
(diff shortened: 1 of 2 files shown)


gdicker1 (approved):
Nothing stands out and this should help with the amount of time the init_atmosphere core takes. Looks good to me!

PR #1011, opened 2022-12-14
Optimized kernels in cuascn subroutine in Ntiedtke

Pranay-Reddy-Kommera:
Optimized OpenACC kernels in the cuascn subroutine in cu_ntiedtke


Pranay-Reddy-Kommera:
New PR will be submitted with additional optimizations.

PR #1012, opened 2022-12-18
Optimized ntiedtke

Pranay-Reddy-Kommera:
The OpenACC kernels in the 'cu_ntiedtke' subroutine has been optimized.


mgduda (commented):
This PR seems to have no negative impact on CPU-only runs with the GNU compilers. Both before and after locally merging this PR for testing, I'm seeing failures in 24-hour simulations on the x1.10242 mesh using OpenACC with 4 MPI tasks (2 for radiation, 2 for integration); so, I can't comment on whether the changes here are correct for GPUs.
@Pranay-Reddy-Kommera @raghuraj19 If you'd like me to do so, I can just merge this PR as-is. But, if you'd rather wait until we can debug the failures with OpenACC in the base branch, that would be fine as well.

raghuraj19:
@mgduda : I reran this PR on internal nodes. The error originates from RRTMG spcvmc code and is independent of this code. So, you can go ahead and accept the PR. I'll create a new PR for the bug. Thanks.

mgduda (approved):

PR #1013, opened 2022-12-19
Updated MPAS_to_physics_gpu subroutine

Pranay-Reddy-Kommera:
In this PR, the data transfers to host in the `MPAS_to_physics_gpu` subroutine are commented out as they are not required if all the physics scheme are on GPU. In addition, the loop which checks if the pressure in the layer above the surface is greater than that in the layer above it is placed in-between 'MPAS_OPENACC' preprocessor directives as we are not updating the variables on host.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_interface.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_interface.F
@@ -707,8 +707,8 @@
  enddo
 !$acc end parallel
  enddo
-!$acc update host(qv_p,qc_p,qr_p,qi_p,qs_p,qg_p,u_p,v_p, &
-!$acc zz_p,rho_p,th_p,t_p,pi_p,pres_p,zmid_p,dz_p)
+!!!$acc update host(qv_p,qc_p,qr_p,qi_p,qs_p,qg_p,u_p,v_p, &
+!!!$acc zz_p,rho_p,th_p,t_p,pi_p,pres_p,zmid_p,dz_p)
 
  pbl_select: select case (trim(pbl_scheme))
     case("bl_mynn")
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/physics/mpas_atmphys_interface.F
```diff
@@ -766,28 +766,28 @@ subroutine MPAS_to_physics_gpu(configs,mesh,state,time_lev,diag,diag_physics,mpa
...
+!!!       call mpas_log_write('lonCell=$r', realArgs=(/lonCell(i)/degrad/))
+!!!       do k = kts,kte
+!!!          call mpas_log_write('$i $i $i $r $r $r $r $r $r $r $r', intArgs=(/j,i,k/),&
+!!!             realArgs=(/dz_p(i,k,j),pressure_b(k,i),pressure_p(k,i),pres_p(i,k,j), &
+!!!                        rho_p(i,k,j),th_p(i,k,j),t_p(i,k,j),qv_p(i,k,j)/))
+!!!       enddo
+!!!!      call mpas_log_write('pressure increasing with height', messageType=MPAS_LOG_CRIT)
+!!!    endif
+!!! enddo
+!!! enddo
```
mgduda, line 790: Rather than commenting-out this loop, perhaps it would be better to deactivate it through preprocessing macros only if we're using OpenACC?
E.g.,
```
#ifndef MPAS_OPENACC
...
#endif
```

mgduda (changes requested):

raghuraj19:
@mgduda : Pranay has turned in the correction.

mgduda:
@Pranay-Reddy-Kommera The description of this PR says "Please do not merge this PR.". If you'd like this PR to be merged, could you please update the description?

PR #1014, opened 2022-12-28
Remove 'east' and 'north' fields from atmosphere restart stream

mgduda:
This PR removes the `east` and `north` fields from the atmosphere core's restart stream.

The `east` and `north` fields are always computed when the model starts up, and
do not need to be written to and read from the restart stream.

Note that the `init_dirs_forphys` subroutine that computes the `east` and
`north` fields is only active if the model was compiled with DO_PHYSICS defined;
but, these fields are currently only used in physics-related code (specifically,
`physics_get_tend`) that is active if the model was compiled with DO_PHYSICS.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -599,8 +599,6 @@
 			<var name="defc_a"/>
 			<var name="defc_b"/>
 			<var name="coeffs_reconstruct"/>
-			<var name="east"/>
-			<var name="north"/>
 			<var_array name="scalars"/>
 			<var name="rainprod" packages="mp_thompson_in"/>
 			<var name="evapprod" packages="mp_thompson_in"/>


ldfowler58:
I agree that we can safely remove the east and north fields from the atmosphere core's restart stream: The east and north fields are initialized in subroutine init_dirs_forphys in mpas_atmphys_init.F when the model starts up, including when config_do_restart is true and DO_PHYSICS is defined. Therefore, both fields can be safely removed from the restart stream in Registry.xml.

gdicker1 (approved):
This looks good to me. However, I think changing the line from "Note that the code that computes..." to "Note that the `init_dirs_forphys` subroutine that computes..." in the PR message makes tracing/verifying this change easier

mgduda:
> This looks good to me. However, I think changing the line from "Note that the code that computes..." to "Note that the `init_dirs_forphys` subroutine that computes..." in the PR message makes tracing/verifying this change easier

Thanks for the suggestion! I've updated the commit message as well as the PR description. If the updates look good to you, could you re-approve?

gdicker1 (approved):
Yep that looks better. It seems good to go now.

PR #1015, opened 2022-12-28
Remove 'meshScaling*' fields from the atmosphere core's restart stream

mgduda:
This PR removes the `meshScaling*` fields from the atmosphere core's restart stream.

The `meshScalingDel2`, `meshScalingDel4`, `meshScalingRegionalCell`, and `meshScalingRegionalEdge` fields are always computed by the `atm_mpas_mesh_scaling` routine when the model starts up, and they therefore do not need to be written and read from the restart stream by the atmosphere core.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -570,10 +570,6 @@
 			<var name="fEdge"/>
 			<var name="fVertex"/>
 			<var name="meshDensity"/>
-			<var name="meshScalingDel2"/>
-			<var name="meshScalingDel4"/>
-			<var name="meshScalingRegionalCell"/>
-			<var name="meshScalingRegionalEdge"/>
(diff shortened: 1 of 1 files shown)


gdicker1 (approved):
This looks good. Not required, but to me it would improve the PR message if it mentioned that these meshScaling* variables are initialized in the `atm_mpas_mesh_scaling` subroutine.

mgduda:
> This looks good. Not required, but to me it would improve the PR message if it mentioned that these meshScaling* variables are initialized in the `atm_mpas_mesh_scaling` subroutine.

@gdicker1 Thanks for the suggestion. I've updated the commit message and PR description. Could you re-approve if the new text looks alright to you?

gdicker1 (approved):
That looks better and I approve!

skamaroc (approved):
The changes look fine and make sense.

PR #1016, opened 2022-12-28
Remove 'uReconstruct{Zonal,Meridional}' fields from atmosphere restart stream

mgduda:
This PR removes the `uReconstruct{Zonal,Meridional}` fields from the atmosphere core's
restart stream.

The `uReconstructZonal` and `uReconstructMeridional` fields are always computed
via the call to `mpas_reconstruct` in the `atm_mpas_init_block` routine when the
model starts up and do not need to be written to and read from the atmosphere
core's restart stream.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -619,8 +619,6 @@
 			<var name="rho"/>
 			<var name="theta"/>
 			<var name="relhum"/>
-			<var name="uReconstructZonal"/>
-			<var name="uReconstructMeridional"/>
 			<var name="circulation"/>
 			<var name="exner"/>
 			<var name="exner_base"/>


gdicker1 (changes requested):
I can't convince myself that "... fields are always computed when the model starts up" is true so far from looking at the code. I might need some extra guidance either personally or from the PR message.

So far I've looked through most of the initialization execution path and don't see where any uReconstruct variables would have been initialized. Near as I can tell those variables are updated whenever `mpas_reconstruct` is called in [`atm_srk3` near line 1098](https://github.com/MPAS-Dev/MPAS-Model/blob/153aebb1c0171c654a0a7ab76e6751186606c7b5/src/core_atmosphere/dynamics/mpas_atm_time_integration.F#L1098), but I think I'm wrong.

mgduda:
@gdicker1 I've updated the commit message and PR description to call out where the `uReconstructZonal` and `uReconstructMeridional` fields are computed at model start-up. Does the updated text seem sufficient?

gdicker1:

gdicker1 (approved):
Apologies, my last comment was meant to be an approval. Looks good to me!

PR #1017, opened 2022-12-29
Initialize climatological ozone profiles for both cold-start and restart runs

mgduda:
This PR modifies the atmosphere core so that climatological ozone profiles are initialized for both cold-start
and restart runs. Previously, this initialization was handled by the `init_o3climatology` routine only for cold-start
runs, with the expectation that the climatological ozone field on the MPAS mesh was available in restart files.

By interpolating the monthly-mean ozone profiles from `OZONE_PLEV.TBL`, `OZONE_LAT.TBL`, and `OZONE_DAT.TBL` when `config_o3climatology` is `.true.` regardless of whether the run is a cold-start or a restart run, it is no longer necessary to write the `pin` and `ozmixm` fields to restart files in the atmosphere core.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -801,8 +801,6 @@
 			<var name="rniblten"/>
 			<var name="rthratensw"/>
 			<var name="rthratenlw"/>
-			<var name="pin"/>
-			<var name="ozmixm"/>
 			<var name="isltyp"/>
 			<var name="ivgtyp"/>
 			<var name="mminlu"/>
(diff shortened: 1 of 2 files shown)


ldfowler58:
Initially, I had decided to add the pin and ozmixm fields to the restart stream in Registry.xml to avoid redoing the horizontal interpolation of the input latitude-longitude monthly mean climatological ozone to the MPAS mesh. This had the impact of increasing the size of restart files, particularly for high-resolution meshes. I agree that it is more important to reduce the size of restart files and recalculate ozmixm for cold starts and restart runs. Let's go ahead with this PR.

gdicker1 (approved):
Looks good to me from code inspection.

gdicker1:
I ran an experiment with PR #1014, PR #1015, PR #1016, and PR #1017 merged into the develop branch to check the restarts. I used the 120km benchmark test case for input data and completed two runs for comparison: the default 5 day run, and a 3 day run that restarted using the restart file from day 2 of the default run. I then saw exactly what I wanted and expected: the global min/max w and u had no differences from the restart point onward between the runs.

Extra reason to say approved!

PR #1023, opened 2023-01-24
Fix for coriolis parameter calculation in OML call

thomasuhoh:
This merge corrects the calculation of the Coriolis parameter used in the ocean mixed layer model which depends on the sin(latitude) and not on cos(latitude).


ldfowler58:
The bug fix on the calculation of the coriolis parameter looks right to me.

skamaroc (commented):
This change appears to be correct.  I'm guessing that the error was my fault from way way back.

mgduda (approved):

mgduda:
@thomasuhoh Thanks for spotting this and for providing a PR!

PR #1029, opened 2023-01-27
Fix zero values of LWDNT and LWDNTC in RRTMG LW

thomasuhoh:
This bug fix prevents that LWDNT and LWDNTC show zero values in RRTMG LW. 

In the original version, the indices of dflx and dflxc were running from (1,nlayers+1) but this leads to zero values for LWDNT and LWDNTC. The index has to be limited to (1,nlayers).


ldfowler58:
LWDNT and LWDNTC are the all-sky and clear-sky downward long wave radiation fluxes at the top-of-the-atmosphere (TOA) and defined at the interface between layers. nlayers is the total number of layers and includes the layers added between the top-of-the-model and the top-of-the-atmosphere. It makes sense that they are equal to zero.

PR #1033, opened 2023-03-23
Add build target for the NVIDIA HPC SDK

mgduda:
This PR adds a build target (`nvhpc`) for the NVIDIA HPC SDK.

The `nvhpc` build target supports OpenACC as well as OpenMP builds in addition to the default MPI-only build.


mgduda:
@supreethms1809 Based on your experience with the NVHPC compilers, do you have any suggestions for changes to the compiler flags in the new build target?

@cponder From an NVIDIA perspective, are there any changes you'd recommend?

cponder:
This looks like what we want, thanks.

cponder (approved):
This looks like what we want, thanks.

PR #1034, opened 2023-03-23
Add build target, 'cray', for use with the Cray Compiling Environment

mgduda:
This PR adds a build target (`cray`) for use with the Cray Compiling Environment.

The `cray` build target supports OpenMP builds in addition to the default MPI-only build.


mgduda:
@gdicker1 I've tested the new `cray` build target on Gust, but it would be helpful to get independent verification that this allows MPAS-A to build and run successfully.

gdicker1 (approved):
I was able to successfully build and run a default 120km benchmark on gust with this target. 

Note: the first build on Gust can't be on a batch node since they don't have internet and MPAS-A needs to fetch the DBL and TBL files for physics_wrf. Common for batch systems.

PR #1037, opened 2023-04-05
speed up call to subroutine get_pblh

ldfowler58:
When using the intel compiler with the -g option, calling subroutine get_pblh is extremely slow to run. This is because the call to get_pblh is inside an "its to ite" loop, and arrays in its argument list contain the "i" index. Copying these arrays to temporary 1D dummy arrays that only depend on the vertical dimension solves this issue.

### DIFF
--- a/src/core_atmosphere/physics/physics_wrf/module_bl_ysu.F
+++ b/src/core_atmosphere/physics/physics_wrf/module_bl_ysu.F
@@ -537,6 +537,8 @@ contains
                                                                      buoy_ysu
    real,    dimension( ims:ime )             ::                       pblh_ysu,&
                                                                       vconvfx
+  real, dimension( kts:kte  ) :: thvx_1d,tke_1d,dzq_1d
+  real, dimension( kts:kte+1) :: zq_1d
 !
 !-------------------------------------------------------------------------------
 !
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/physics/physics_wrf/module_bl_ysu.F
```diff
@@ -537,6 +537,8 @@ subroutine ysu2d(j,ux,vx,tx,qx,p2d,p2di,pi2d,                               &
                                                                      buoy_ysu
    real,    dimension( ims:ime )             ::                       pblh_ysu,&
                                                                       vconvfx
+   real,    dimension( kts:kte )   :: dummy1,dummy2,dummy4
+   real,    dimension( kms:kme )   :: dummy3
```
mgduda, line 541: Since these `dummyN` variables each have a specific use, we could consider giving them names like `thvx_1d`, `tke_ysu_1d`, `zq_1d`, and `dzq_1d`.

ldfowler58: I totally agree that it will add more clarity to our modifications to the original sourcecode.

mgduda (approved):

PR #1038, opened 2023-04-06
Add a new build target, 'gnu', for the GNU Fortran, C, and C++ compilers

mgduda:
This PR adds a new build target, `gnu`, for the GNU Fortran, C, and C++ compilers.

The only difference between the new `gnu` target and the existing `gfortran` target is the addition of `-std=f2008` to the `FFLAGS` variable to check for compliance with the Fortran 2008 standard.

With the new `gnu` target, builds with the GNU 7.4.0 (and presumably, earlier) compilers fail with the error
```
   mpas_stream_manager.F:5844:32:

        call c_f_pointer(manager_c, manager)
                                1
   Error: TS 29113/TS 18508: Noninteroperable array FPTR at (1) to C_F_POINTER: Expression is a noninteroperable derived type
```
However, builds with the GNU 8.3.0 and later compilers are successful.

In future, the `gfortran` target will likely be deprecated and superseded by the `gnu` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,36 @@ endif
 dummy:
 	( $(MAKE) error )
 
+gnu:   # BUILDTARGET GNU Fortran, C, and C++ compilers
+	( $(MAKE) all \
+	"FC_PARALLEL = mpif90" \
+	"CC_PARALLEL = mpicc" \
+	"CXX_PARALLEL = mpicxx" \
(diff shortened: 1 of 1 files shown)


gdicker1 (approved):
Looks good! I was able to build this target with GNU 10.1.0 on Cheyenne and run the 120km benchmark case.

PR #1043, opened 2023-04-10
V8 add 1.7 gradient coefficients

skamaroc:
This PR adds fields of coefficients for computing the x- and y-derivative of cell-centered fields, `cell_gradient_coef_x` and `cell_gradient_coef_y`. The computation of these coefficients takes place in the init_atmosphere core, and the coefficients are written to static and initial conditions files.

Also included in this PR are changes to read the new coefficient fields into the atmosphere core, and to read/write these from/to MPAS-A restart files.

At present, the new `cell_gradient_coef_x` and `cell_gradient_coef_y` fields are not used in stand-alone MPAS-A.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -466,6 +466,10 @@
 			<var name="deriv_two"/>
 			<var name="defc_a"/>
 			<var name="defc_b"/>
+#ifdef MPAS_CAM_DYCORE
+			<var name="cell_gradient_coef_x"/>
+			<var name="cell_gradient_coef_y"/>
+#endif
 			<var name="coeffs_reconstruct"/>
 			<var_array name="scalars"/>
 			<var name="initial_time"/>
@@ -598,6 +602,10 @@
 			<var name="deriv_two"/>
 			<var name="defc_a"/>
 			<var name="defc_b"/>
+#ifdef MPAS_CAM_DYCORE
+			<var name="cell_gradient_coef_x"/>
+			<var name="cell_gradient_coef_y"/>
+#endif
 			<var name="coeffs_reconstruct"/>
 			<var name="east"/>
 			<var name="north"/>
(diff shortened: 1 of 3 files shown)


src/core_init_atmosphere/mpas_atm_advection.F
```diff
@@ -894,28 +901,21 @@ subroutine atm_initialize_deformation_weights( mesh, nCells, on_a_sphere, sphere
...
             if (ip1 == n) ip1 = 1
-            dl = sqrt((xp(ip1)-xp(i))**2 + (yp(ip1)-yp(i))**2)
+            dx = xp(ip1)-xp(i)
+            dy = yp(ip1)-yp(i)
             area_cell = area_cell + 0.25*(xp(i)+xp(ip1))*(yp(ip1)-yp(i)) - 0.25*(yp(i)+yp(ip1))*(xp(ip1)-xp(i))
-            area_cellt = area_cellt + (0.25*(xp(i)+xp(ip1))*cos(thetat(i)) + 0.25*(yp(i)+yp(ip1))*sin(thetat(i)))*dl
+            thetat(i) = atan2(dy,dx)-pii/2.
          end do
 
+         ! coefficients - see documentation for the formulas.
```
mgduda, line 917: Does this documentation have a URL?

src/core_atmosphere/Registry.xml
```diff
@@ -1370,6 +1374,12 @@
                 <var name="defc_b" type="real" dimensions="maxEdges nCells" units="unitless"
                      description="Coefficients for computing the diagonal components of the horizontal deformation"/>
 
+                <var name="cell_gradient_coef_x" type="real" dimensions="maxEdges nCells" units="m^-1"
+                     description="Coefficients for computing the x derivative of a cell-centered variable"/>
+
+                <var name="cell_gradient_coef_y" type="real" dimensions="maxEdges nCells" units="m^-1"
+                     description="Coefficients for computing the y derivative of a cell-centered variable"/>
```
mgduda, line 1381: Are these actually zonal and meridional gradients? If not, how are the x- and y-directions defined?

src/core_atmosphere/Registry.xml
```diff
@@ -466,6 +466,8 @@
 			<var name="deriv_two"/>
 			<var name="defc_a"/>
 			<var name="defc_b"/>
+                        <var name="cell_gradient_coef_x"/>
+                        <var name="cell_gradient_coef_y"/>
```
mgduda, line 470: We could enclose occurrences of these new fields in
```
#ifdef MPAS_CAM_DYCORE
```
```
#endif
```
directives to prevent them from being allocated, read, and written in stand-alone MPAS-A.

src/core_atmosphere/Registry.xml
mgduda: Actually, I just read the commit message for ddb57a8f, and it seems that these are for zonal and meridional derivatives. Should we change the descriptions (or even the names) to make this clear?

src/core_atmosphere/Registry.xml
mgduda: When the MPAS LES capability is available, we could remove the preprocessing directives or define a package so these would only be allocated, read, and written if we're operating as a dycore in CAM or running with LES options enabled.

src/core_init_atmosphere/mpas_atm_advection.F
skamaroc: No, it doesn't have a URL, and I am the only person with the "documentation".  My intent is to put it in the technote.

mgduda (approved):

PR #1048, opened 2023-04-27
Copy xml_stream_parser.c from master to maint-1.0

ndkeen:
Copying this file from E3SM master (as of April 27 2023) to allow compilation on maint-1.0 with Intel compiler.

Avoids:
```
xml_stream_parser.c:1732:6: error: call to undeclared function 'stream_mgr_add_field_c'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                                        stream_mgr_add_field_c(manager, streamID, (const char *)fieldname, packages_local, &err);
```
[bfb]

### DIFF
--- a/src/framework/xml_stream_parser.c
+++ b/src/framework/xml_stream_parser.c
@@ -3,7 +3,7 @@
 //
 // Unless noted otherwise source code is licensed under the BSD license.
 // Additional copyright and license information can be found in the LICENSE file
-// distributed with this code, or at http://mpas-dev.github.com/license.html
+// distributed with this code, or at http://mpas-dev.github.io/license.html
 //
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


jonbob (approved):
Looks fine from a visual inspection

xylar (approved):
I'm fine with bringing in the full changes with the caveat that I can't say for sure whether the masking changes will interface properly with this version of the code.  It looks like the mask parameter might just be `NULL` but again, I'm not completely sure.

A potentially safer fix would be just to cherry-pick this commit:
https://github.com/MPAS-Dev/MPAS-Model/pull/198/commits/9bf0ed3afc3e001ec210caeef91fa4225fc945b9

ndkeen:
I don't know how to cherry-pick.  Would it make more sense for someone with knowledge of what this is doing to make PR?

jonbob:
verified that @xylar's recommended cherry-pick allows maint-1.0 to successfully complete all-active tests. Working now on getting the fix committed

PR #1051, opened 2023-05-12
Introduce new I/O layer option in MPAS: the Simple MPAS I/O Layer (SMIOL)

mgduda:
This PR introduces a new I/O layer into MPAS: the Simple MPAS I/O Layer (SMIOL).

SMIOL provides a subset of the parallel I/O functionality of PIO, but with a simplified design that minimizes external dependencies to just the Parallel-NetCDF library. All SMIOL code is included directly in the MPAS-Model source tree in `src/external/SMIOL`, and the use of SMIOL in place of PIO is activated at build time by *not* setting the `PIO` environment variable.

If MPAS the use of SMIOL was activated at build time (by *not* setting the `PIO` environment variable), the message
```
Using the SMIOL library.
```
will appear in the build summary (rather than `Using the PIO 2.x library.` or `Using the PIO 1.x library.`).

At present, all output files written by SMIOL are in 64-bit data (CDF-5) format, and so the `io_type` attribute in streams is effectively ignored.

Although a snapshot of the SMIOL code is included under `src/external/SMIOL` in this PR, SMIOL itself is developed in a separate repository (https://github.com/MPAS-Dev/SMIOL), and in future it may be possible to pull the SMIOL code directly from its repository into MPAS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -628,6 +628,7 @@ CPPINCLUDES =
 FCINCLUDES =
 LIBS =
 
+ifneq "$(PIO)" ""
 #
 # Regardless of PIO library version, look for a lib subdirectory of PIO path
 # NB: PIO_LIB is used later, so we don't just set LIBS directly
@@ -671,6 +672,11 @@ ifneq ($(wildcard $(PIO_LIB)/libpioc\.*), )
 endif
 ifneq ($(wildcard $(PIO_LIB)/libgptl\.*), )
 	LIBS += -lgptl
+endif
+
+else # Not using PIO, using SMIOL
+	LIBS += -L$(PWD)/src/external/SMIOL -lsmiolf -lsmiol
+	FCINCLUDES += -I$(PWD)/src/external/SMIOL
 endif
 
 ifneq "$(NETCDF)" ""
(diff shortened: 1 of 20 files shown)


mgduda:
@gdicker1 I just pushed an additional commit ([208c8708](https://github.com/MPAS-Dev/MPAS-Model/pull/1051/commits/208c8708)) that improves the error messages when calls to SMIOL return non-success error codes. If you haven't already started reviewing this PR, I can just squash this commit onto the previous HEAD of the PR branch; but, in case you've already started, I thought I'd keep the new changes separate. Ultimately, I'll squash the commit onto its parent before merging this PR just to keep the commit history a bit cleaner.

gdicker1:
@mgduda I think you can go ahead and squash that commit. I haven't done more than glance through this PR yet. Thanks for the extra messages!

mgduda:
@gdicker1 I've just squashed commit 208c8708, and I've also added a new commit (ba694cb5) to indicate in log files that SMIOL was used. I think this PR should be ready for review whenever you're ready!

gdicker1 (approved):
Thanks for the work on this! It looks good and works for the tests I ran.

The tests were the same as in PR #1057 which I describe in [this comment](https://github.com/MPAS-Dev/MPAS-Model/pull/1057#issuecomment-1574343472).

PR #1052, opened 2023-05-17
Initial implementation of the CCPP-compliant parameterization of the gravity wave drag over orography

ldfowler58:
1. In this PR, we added the directory physics_mmm. In its initial implementation, physics_mmm contains parameterizations of moist physics (surface layer scheme, planetary boundary layer scheme, convection scheme, and cloud microphysics scheme) for the mesoscale reference suite.
2. Parameterizations in physics_mmm are CCPP-compliant (initial versions) and are shared between MPAS, WRF, and CM1.
3. The initial version of module_bl_gwdo.F is split in two separate files: bl_gwdo.F in physics_mmm contains the actual parameterization whereas the modified version of module_bl_gwdo.F reduces to a buffer between the parameterization and the MPAS physics driver.

### DIFF
--- a/src/core_atmosphere/physics/Makefile
+++ b/src/core_atmosphere/physics/Makefile
@@ -198,6 +198,7 @@ mpas_atmphys_update.o: \
 clean:
 	$(RM) *.o *.mod *.f90 libphys.a
 	( cd physics_wrf; $(MAKE) clean )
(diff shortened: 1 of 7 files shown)


weiwangncar (approved):
Code changes are as expected, and compiles as it is.

PR #1053, opened 2023-05-17
Initial implementation of the CCPP-compliant parameterization of the YSU planetary boundary layer.

ldfowler58:
1. In this PR, we added the directory physics_mmm. In its initial implementation, physics_mmm contains parameterizations of moist physics (surface layer scheme, planetary boundary layer scheme, convection scheme, and cloud microphysics scheme) for the mesoscale reference suite.
2.  Parameterizations in physics_mmm are CCPP-compliant (initial versions) and are shared between MPAS, WRF, and CM1.
3. The initial version of module_bl_ysu.F is split in two separate files: bl_ysu.F in physics_mmm contains the actual parameterization of the YSU scheme whereas the modified version of module_bl_ysu.F reduces to a buffer between the parameterization and the MPAS physics driver.

### DIFF
--- a/src/core_atmosphere/physics/physics_wrf/module_bl_ysu.F
+++ b/src/core_atmosphere/physics/physics_wrf/module_bl_ysu.F
@@ -1,23 +1,22 @@
+#define NEED_B4B_DURING_CCPP_TESTING 1
 !=================================================================================================================
-!module_bl_ysu.F was modified to accomodate both the WRF and MPAS models / 2018-12-7 
+ module module_bl_ysu
+ use mpas_log
+ use mpas_kind_types,only: kind_phys => RKIND
+ use bl_ysu
+
+ implicit none
+ private
+ public:: ysu
+
+
+ contains
+
+
 !=================================================================================================================
-!WRF:model_layer:physics
-!
-!
-!
-!
-!
-!
-!
-module module_bl_ysu
-contains
-!
-!
-!-------------------------------------------------------------------------------
-!
-   subroutine ysu(u3d,v3d,th3d,t3d,qv3d,qc3d,qi3d,p3d,p3di,pi3d,               &
+   subroutine ysu(u3d,v3d,t3d,qv3d,qc3d,qi3d,p3d,p3di,pi3d,                    &
                   rublten,rvblten,rthblten,                                    &
-                  rqvblten,rqcblten,rqiblten,flag_qi,                          &
+                  rqvblten,rqcblten,rqiblten,flag_qc,flag_qi,                  &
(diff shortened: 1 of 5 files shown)


src/core_atmosphere/physics/physics_wrf/module_bl_ysu.F
```diff
@@ -110,1655 +132,343 @@ subroutine ysu(u3d,v3d,th3d,t3d,qv3d,qc3d,qi3d,p3d,p3di,pi3d,               &
...
-              & MIN((thetav1D(k)-(minthv + delt_thv))/MAX(thetav1D(k)-thetav1D(k-1),1E-6),1.0)
-        ENDIF
-        k = k+1
-         IF (k .EQ. kte-1) zi = zw1D(kts+1) !EXIT SAFEGUARD
-      ENDDO
+      if(present(a_u_bep) .and. present(a_v_bep) .and. present(a_t_bep) .and.  &
+         present(a_q_bep) .and. present(a_e_bep) .and. present(b_u_bep) .and.  &
+         present(b_v_bep) .and. present(b_t_bep) .and. present(b_q_bep) .and.  &
+         present(b_e_bep) .and. present(dlg_bep) .and. present(dl_u_bep) .and. &
+         present(sf_bep)  .and. present(vl_bep)  .and. present(frc_urb2d)) then
```
weiwangncar, line 351: @ldfowler58 Because this group of variables are together, either they are all available, or not, would just testing one of them, or even just checking on flag_bep be sufficient?

ldfowler58: I would prefer to keep it as is to highlight that those arrays are actually optional. To simply flag_bep may not be sufficient because we may want to have flag_bep be false but the arrays to be present to ensure that results remain the same.

mgduda: @weiwangncar I agree that performance could be improved if we checked for the presence of just a single optional argument. For now, perhaps we could just let the code be as is, and work out an optimization for a future PR.

ldfowler58: I like Michael's suggestion for now. Thanks.

weiwangncar (approved):
You could say this code matches with WRF v4.5 YSU.

mgduda:
> You could say this code matches with WRF v4.5 YSU.

@weiwangncar I can add that info to the final merge commit message. Thanks for the suggestion.

mgduda (approved):

PR #1057, opened 2023-05-31
Convert halo exchanges in MPAS-A to use group exchange routines from mpas_dmpar

mgduda:
This PR converts halo exchanges in MPAS-A to use group halo exchange routines from the mpas_dmpar module.

Previously, MPAS-A used older halo exchange routines that handled a single field per call. Now, halos are exchanged using newer (c. 2016) routines in the mpas_dmpar module that can communicate halos for a group of fields in a single call.

As part of the changes in this PR, the use of temporary, ad hoc local fields for exchanging halos of fields in the monotonic scalar transport has been reworked to make use of scratch arrays.

In cases where only a single field needs to have its halo exchanged at a specific point in the code, that halo exchange has still been updated to use the group halo exchange routines by creating an exchange group with a single member.

Note: In this PR, the halo exchanges in diagnostics (specifically, isobaric diagnostics and PV diagnostics) have not been modified.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -1873,6 +1873,12 @@
                 </var_array>
         </var_struct>
 
+        <var_struct name="halo_scratch" time_levs="1">
+                <var name="scale" type="real" dimensions="nVertLevels TWO nCells" units="-"
+                     description="Scaling coefficients for monotonic-limited horizontal fluxes"
+                     persistence="scratch" />
+        </var_struct>
+
 
 <!-- ================================================================================================== -->
 <!--  DECLARATIONS OF ALL PHYSICS VARIABLES (will need to be moved to a Physics Registry shared by the  -->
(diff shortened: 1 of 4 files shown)


mgduda:
@gdicker1 For a bit more background, in working towards the implementation of the OpenACC-friendly group halo exchanges, an intermediate step was to make use of the existing group halo exchange routines from the `mpas_dmpar` module. Since the overall code changes were extensive, I thought it might make it easier to review (and for anyone in future to follow what's being done) if I broke the work into several smaller PRs. This would be the first of those PRs.

gdicker1 (approved):
Looks good! I had no differences in the restart files for the few tests I did with models compiled with [MPAS-Dev:develop](https://github.com/MPAS-Dev/MPAS-Model/tree/develop) and [mgduda:atmosphere/group_halos](https://github.com/mgduda/MPAS-Model/tree/atmosphere/group_halos).

gdicker1:
**Tests:**
- 1 day simulation
- 12 + 12 simulation
    - The model ran for 12 sim hours, stopped, then restarted for another 12 hours.


**Common details:**
- Took place on NCAR's Cheyenne machine
- Used the GNU 10.1.0 compiler software stack
- Were initialized with the v7.0 benchmark test case on a 120km mesh
- Restart files were written every 12 hours

PR #1062, opened 2023-06-07
Enable the use of mpas_halo routines for communicating halo groups in MPAS-A

mgduda:
This PR adds the ability to use the `mpas_halo` routines for communicating halo groups in MPAS-A.

As an alternative to the group halo routines from the `mpas_dmpar` module, MPAS-A can now make
use of the routines from the `mpas_halo` module. Which module to use is runtime configurable
through a new namelist option, `config_halo_exch_method` in a new `&development` namelist group.
The `&development` group is hidden by default, and the `config_halo_exch_method` defaults to `'mpas_halo'`.
To switch to the `mpas_dmpar` halo routines, the `config_halo_exch_method` namelist option may be set
to `'mpas_dmpar'`.

The specific choice of halo exchange module (`mpas_dmpar` or `mpas_halo`) should have no impact on
simulation results, all else being equal.

In order to make the setup and use of different halo exchange methods cleaner and more modular, this PR
includes refactoring of the calls to create and destroy halo exchange groups as well as the use of procedure
arguments to supply specific halo exchange routines to the dynamics and physics-dynamics coupling code.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -382,7 +382,7 @@
                 <nml_option name="config_halo_exch_method" type="character" default_value="mpas_dmpar"
                      units="-"
                      description="Method to use for exchanging halos"
-                     possible_values="`mpas_dmpar'"/>
+                     possible_values="`mpas_dmpar', `mpas_halo'"/>
         </nml_record>
 
 <!-- **************************************************************************************** -->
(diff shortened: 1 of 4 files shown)


mgduda:
@dimomatt @gdicker1 The commit messages provide quite a bit more detail than the PR description. For bd97fed7 in particular, the diff can seem daunting; but, that's mostly because it makes changes to the indentation for a lot of code. If you run `git show -b bd97fed7`, the commit is much easier to understand.

dimomatt (approved):
Looks good to me- everything builds and runs, and the code looks clean!

gdicker1 (commented):
Looks good to me too!

PR #1069, opened 2023-06-09
Add cloud top temperature and pressure estimates from NCL

thomasuhoh:
This PR add estimates for cloud top temperature and pressure based on the NCL "CTT" routine from https://github.com/NCAR/wrf-python/blob/develop/fortran/wrf_fctt.f90 to the diagnostics output. If no cloud is present, then cloud top temperature is set to -9999. as missing value.

### DIFF
--- a/src/core_atmosphere/diagnostics/mpas_convective_diagnostics.F
+++ b/src/core_atmosphere/diagnostics/mpas_convective_diagnostics.F
@@ -10,6 +10,7 @@ module mpas_convective_diagnostics
     use mpas_derived_types, only : MPAS_pool_type, MPAS_clock_type, MPAS_LOG_ERR, MPAS_LOG_CRIT
     use mpas_kind_types, only : RKIND, R8KIND
     use mpas_log, only : mpas_log_write
+    use calculate_ctt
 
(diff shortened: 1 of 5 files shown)


src/core_atmosphere/diagnostics/mpas_convective_diagnostics.F
```diff
@@ -397,6 +417,25 @@ subroutine convective_diagnostics_compute()
                   umeridional_6km(iCell) = column_height_value(umeridional(1:nVertLevels,iCell), zp, 6000.0_RKIND, nVertLevels)
               end if
 
+            !  calculate ctt or HCF
+            if (need_ctt .OR. need_ctpres) then
+!           from NCL https://github.com/NCAR/wrf-python/blob/develop/fortran/wrf_fctt.f90
+                do iCell=1, nCellsSolve
```
CharlesZheZhang, line 423: Please see this line of code, this do loop is redundant, since it is already included in the do loop above

CharlesZheZhang:
Thank you so much for coded up the ctt function from ncl/wrf-python into MPAS-A diagnostic. 
I did try to compile it and had the above issue, see my reply above. 
Thank you so much!

PR #1078, opened 2023-06-12
Bug fix in call to the parameterization of the gravity wave drag over orography

ldfowler58:
This PR corrects a bug discovered by Wei Wang in the call to the subroutine gwdo in mpas_atmphys_driver_gwdo.F.  In the earlier version, the variable z was erroneously defined as the height at the interface between layers, or z = z_p where
z_p is equal to zgrid. In the corrected version, the variable z is now defined as the height defined at the middle of layers, or z = zmid_p, as needed by subroutine gwdo.


mgduda:
@ldfowler58 Is it only the last of the commits (da30ff91) in this PR branch that is actually needed? If so, I can rebase this branch to avoid the inclusion of the first five commits.

weiwangncar:
@ldfowler58 @mgduda It looks like most of the changed files do not belong to this PR.
@ldfowler58 The PR message is good.

mgduda (approved):

weiwangncar (approved):

PR #1079, opened 2023-06-12
Eliminate the need to specify config_len_disp through the introduction of a new variable, nominalMinDc

mgduda:
This PR eliminates the need for users to specify `config_len_disp` in their `namelist.atmosphere` files
by introducing a new scalar variable, `nominalMinDc`, which specifies the nominal minimum
grid distance where meshDensity == 1.0. The `nominalMinDc` variable is read from the "input" stream by
the init_atmosphere core and scaled along with other mesh fields like `dcEdge` and `dvEdge`; the `nominalMinDc`
variable is then written to the "output" stream by the init_atmosphere core, making it available to the atmoshere core.

At start-up, the atmosphere core attempts to read both the `config_len_disp` namelist variable and the `nominalMinDc`
variable, both of which default to 0.0.

1) If both `config_len_disp` and `nominalMinDc` are provided (in the namelist and input file, respectively) as positive values, the `config_len_disp` value is used as the horizontal length scale for horizontal diffusion and 3-d divergence damping; `nominalMinDc` is set to this value as well. Additionally, if `config_len_disp` and `nominalMinDc` differ
(by more than ~1.0e-6 relative difference), a warning is printed to the log file; e.g.:
```
WARNING: nominalMinDc was read from input file as a positive value (120000.) that differs
WARNING: from the specified config_len_disp value (125000.)
```

2) If only one of `config_len_disp` and `nominalMinDc` is provided as a positive value, that positive value is used in the model (and in any case, `nominalMinDc` takes that value when written to model restart files).

3) If neither `config_len_disp` nor `nominalMinDc` are provided, the atmosphere core stops with an error:
```
      ERROR: Both config_len_disp and nominalMinDc are <= 0.0.
      ERROR: Please either specify config_len_disp in the &nhyd_model namelist group,
      ERROR: or use an input file that provides a valid value for the nominalMinDc variable.
```

The changes in this PR are backwards-compatible, in the sense that old mesh files that lack the `nominalMinDc` variable can still be used, in which case the user must supply the `config_len_disp` option in their `namelist.atmosphere` file.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -142,10 +142,10 @@
                      description="Formulation of horizontal mixing"
                      possible_values="`2d_fixed' or `2d_smagorinsky'"/>
 
-                <nml_option name="config_len_disp" type="real" default_value="120000.0"
+                <nml_option name="config_len_disp" type="real" default_value="0.0" in_defaults="false"
                      units="m"
                      description="Horizontal length scale, used by the Smagorinsky formulation of horizontal diffusion and by 3-d divergence damping"
-                     possible_values="Positive real values"/>
+                     possible_values="Positive real values. A zero value implies that the length scale is prescribed by the nominalMinDc value in the input file."/>
 
                 <nml_option name="config_visc4_2dsmag" type="real" default_value="0.05"
                      units="-"
@@ -457,6 +457,7 @@
 			<var name="fEdge"/>
 			<var name="fVertex"/>
 			<var name="meshDensity"/>
+			<var name="nominalMinDc"/>
 			<var name="bdyMaskCell"/>
 			<var name="bdyMaskEdge"/>
 			<var name="bdyMaskVertex"/>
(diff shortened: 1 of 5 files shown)


gdicker1 (approved):
Code looks good and (correctly setup) tests did what I expected.

My only note is that in the PR description the following line isn't clear to me:

> Additionally, if `config_len_disp` and `nominalMinDc` differ by more than 1 part in 10e6, a warning is printed ...

It might be clearer to say:

> Additionally, if the difference between `config_len_disp` and `nominalMinDc` is greater than 1E-6, a warning is printed ...

skamaroc (approved):

mgduda:
> Code looks good and (correctly setup) tests did what I expected.
> 
> My only note is that in the PR description the following line isn't clear to me:
> 
> > Additionally, if `config_len_disp` and `nominalMinDc` differ by more than 1 part in 10e6, a warning is printed ...
> 
> It might be clearer to say:
> 
> > Additionally, if the difference between `config_len_disp` and `nominalMinDc` is greater than 1E-6, a warning is printed ...

Thanks for the suggestion. I've updated the PR description. The tolerance is actually 1e-6 scaled by `config_len_disp`, which I've tried to convey in the updated text.

PR #1080, opened 2023-06-14
Add missing dependency on core_physics_mmm in core_atmosphere/physics/Makefile

mgduda:
This PR corrects parallel build failures in the atmosphere core by adding a missing dependency on `core_physics_mmm`
in `src/core_atmosphere/physics/Makefile`. The `core_physics_wrf` build target should depend on `core_physics_mmm`, since compilation of modules in the `physics_wrf` directory depends on successful compilation of modules in the `physics_mmm` directory.

### DIFF
--- a/src/core_atmosphere/physics/Makefile
+++ b/src/core_atmosphere/physics/Makefile
@@ -52,7 +52,7 @@ lookup_tables:
 core_physics_mmm: core_physics_init
 	(cd physics_mmm; $(MAKE) all)
 
-core_physics_wrf: core_physics_init
+core_physics_wrf: core_physics_init core_physics_mmm
 	(cd physics_wrf; $(MAKE) all COREDEF="$(COREDEF)")
 
 core_physics_init: $(OBJS_init)


ldfowler58:
In MPAS-Dev:develop:

in ./physics/Makefile, we added -I./physics_mmm at the bottom of the file.
in ./physics/physics_wrf/Makefile, we added -I../physics_mmm at the bottom of the file.

so I do not understand why we need to modify line 55 to core_physics_wrf: core_physics_init core_physics_mmm

Thanks.

gdicker1:
@ldfowler58 this has to do with GNU Make and the compilation process, instead of compilation flags. So far, builds have succeeded because `core_physics_mmm` just happens to build before `core_physics_wrf`. My recent parallel build attempts showed me that this isn't guaranteed. Adding this prerequisite would guarantee the compilation order.

gdicker1 (approved):
I confirmed this fixes parallel builds. Thanks!

PR #1085, opened 2023-06-15
Always use geog_data_path with trailing slash in mpas_init_atm_static

mgduda:
This PR corrects an issue in which static field processing would fail when the `config_geog_data_path` option
was specified in a namelist without a trailing '/' character. In such cases, trying to build paths to geographical data tiles at various points in the `mpas_init_atm_static` module would fail to produce a valid path. A local variable, `geog_data_path`, that is the same as `config_geog_data_path` but guaranteed to have a trailing '/' was already being set near the beginning of the `init_atm_static()` routine, and this PR simply uses that variable in place of the user-provided `config_geog_data_path`.

With the changes in this PR, static field processing can work with a geographical data path from the namelist that either does or does not have a trailing '/'.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -374,7 +374,7 @@
  end select
 
  call mpas_log_write('--- start interpolate TER')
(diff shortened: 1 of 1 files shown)


dimomatt (approved):
Everything seems to be in order here

PR #1086, opened 2023-06-15
Correct several Fortran 2008 compliance issues in physics_mmm

mgduda:
This PR corrects Fortran 2008 compliance issues in physics modules that were caught by the GNU Fortran compiler when using the `-std=f2008` flag. Most of these issues are related to the use of non-standard kind/type specifications.

### DIFF
--- a/src/core_atmosphere/physics/physics_mmm/cu_ntiedtke.F
+++ b/src/core_atmosphere/physics/physics_mmm/cu_ntiedtke.F
@@ -2527,9 +2527,10 @@
       implicit none
 
 !--- input arguments:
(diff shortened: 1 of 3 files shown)


ldfowler58:
I completely agree with your updates proposed in cu_ntiedtke.F and mp_radar.F. Regarding module_libmassv.F, it may be cleaner to add the definitions of R4KIND and R8KIND in ccpp_kinds.F instead of defining them as private at the top of the modules.

ldfowler58:
I can work on this, of course, if you agree.

ldfowler58 (approved):

PR #1090, opened 2023-06-28
Fix memory leak in deallocate_microphysics for recloud_p, reice_p, and resnow_p (v8.0.1)

mgduda:
This PR fixes a memory leak in the deallocate_microphysics routine for the recloud_p,
reice_p, and resnow_p arrays.

The logic for deallocating the recloud_p, reice_p, and resnow_p fields was incorrect,
leading to a memory leak at the end of a simulation. Rather than checking if these arrays
are allocated and allocating them if they are not, the code now checks whether these
arrays are allocated and deallocates them if they are.

This PR mirrors PR #989 but targets the `hotfix-v8.0.1` branch.


mgduda:
@gdicker1 The changes in this PR were already approved for inclusion in the MPAS v7.4 release, which we decided to skip since the v8.0.0 release was near. I think it should be sufficient to just verify that this PR introduces the same changes as in PR #989.

PR #1091, opened 2023-06-28
Allow real-valued attributes to be read with different precision using SMIOL

mgduda:
This PR allows real-valued attributes to be read with a different precision from
MPAS's native precision when using SMIOL.

Previously, the `mpas_io_get_att_real0d()` routine relied on its 'precision'
optional argument to determine whether to attempt to read real-valued attributes
in a precision different from MPAS's native precision. However, in some parts of
the MPAS framework (specifically, in `mpas_bootstrap_framework_phase1()`), the 
`MPAS_io_get_att()` routine is called without the 'precision' argument, leading to
problems in reading, e.g., 'sphere_radius' when the precision of the input file
is not the MPAS native precision.
    
Now, the `mpas_io_get_att_real0d()` routine first tries to read an attribute in
MPAS's native precision; if that fails, a second attempt is made to read the 
attribute using the opposite precision (single vs. double). If both attempts
fail, the attribute argument is set to a fill value and an error code is
returned. This allows MPAS to read real-valued attributes using SMIOL during the 
bootstrapping process regardless of the precision of the input file.

### DIFF
--- a/src/framework/mpas_io.F
+++ b/src/framework/mpas_io.F
@@ -4603,77 +4603,45 @@ module mpas_io
          local_precision = MPAS_IO_NATIVE_PRECISION
       end if
 
-      ! Query attribute value
 #ifdef MPAS_PIO_SUPPORT
+      ! Query attribute value
       pio_ierr = PIO_inq_att(handle % pio_file, varid, attName, xtype, len)
       if (pio_ierr /= PIO_noerr) then
          if (present(ierr)) ierr = MPAS_IO_ERR_PIO
          return
       end if
-#endif
 
       if ((local_precision == MPAS_IO_SINGLE_PRECISION) .and. &
           (MPAS_IO_NATIVE_PRECISION /= MPAS_IO_SINGLE_PRECISION)) then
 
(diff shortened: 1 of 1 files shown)


mgduda:
For an example of the problem that this PR addresses, here are the warnings we get in a log file when the init_atmosphere core is compiled in double-precision (the default) and when we read a single-precision static file (which definitely has a `sphere_radius` attribute):
```
WARNING: Attribute sphere_radius not found in x1.10242.static.nc
WARNING:    Setting sphere_radius to 1.0
```

mgduda:
@gdicker1 The diff may look a bit complicated, since the calls to read attributes using PIO and SMIOL were previously interleaved, necessitating multiple sets of preprocessing directives. I thought it would be cleaner to just separate the PIO and SMIOL sections of the `mpas_io_get_att_real0d` routine so that each I/O layer has a contiguous chunk of code.

gdicker1 (approved):
Looks good to me!

Generating initial conditions with a single-precision init_atmosphere core and then running with double-precision atmosphere core was successful. There were no warnings/messages like @mgduda mentioned above.

PR #1092, opened 2023-06-28
Fix units and description for rt_diabatic_tend in atmosphere core Registry.xml

mgduda:
This PR corrects the units and description for the `rt_diabatic_tend` field in the atmosphere `Registry.xml`.

The `rt_diabatic_tend` field represents the tendency of the modified potential temperature due to cloud microphysics in K/s; accordingly, this commit corrects the units and description attributes for `rt_diabatic_tend` in the atmosphere core's `Registry.xml` file.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -1791,8 +1791,8 @@
                 <var name="tend_theta" name_in_code="theta_m" type="real" dimensions="nVertLevels nCells Time" units="kg K m^{-3} s^{-1}"
                      description="tendency of coupled potential temperature rho*theta_m/zz from dynamics and physics, updated each RK step"/>
 
-                <var name="rt_diabatic_tend" type="real" dimensions="nVertLevels nCells Time" units="kg K s^{-1}"
-                     description="Tendency of coupled potential temperature from physics"/>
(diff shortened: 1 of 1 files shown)


smileMchen (approved):
This change is correct.

ldfowler58:
I am wondering if it would make sense to move rt_diabatic_tend from "tend" to "tend_physics" in Registry.xml. My reasoning is twofold:

1) rt_diabatic_tend is actually a physics-related tendency from cloud microphysics processes, and it is an uncoupled variable as the other physics tendencies in "tend_physics".

2) All the variables in "tend" are coupled tendencies but rt_diabatic_tend is not.

Considering that rt_diabatic_tend is only needed in subroutine atm_recover_large_step_variables, it may not be too much work to do this.

Michael, if you are OK to consider it, I am happy to add a PR to include this change. Thanks.

mgduda:
@ldfowler58 I don't have a strong opinion on where `rt_diabatic_tend` is stored, but I think you make a good argument for moving it to the `tend_physics` pool/var_struct. As you said, that would be a separate PR, which would target the `develop` branch.

For this PR specifically, do the changes to the `units` and `description` attributes look good to you?

PR #1094, opened 2023-06-28
Add mpas_modify_att() subroutines to mpas_attlist

dimomatt:
CF compliance requires certain attributes to be updated periodically throughout a run, or at the beginning of a run after compilation. The current attlist module only supports creation and retrieval of attributes. This commit adds the functionality to modify attributes and some basic tests for those routines. The subroutines work by traversing the linked list until they find a node who's attName matches the attName to be changes. It then checks the attType of the node to make sure that the correct type of value is being passed in to the attribute. The error code is then set to 0 and the subroutine exits.

It is also noted that these procedures are not currently threadsafe. Future work may include threadsafe attribute modification and a more comprehensive test suite for all of the attlist procedures.

### DIFF
--- a/src/core_test/mpas_test_core_field_tests.F
+++ b/src/core_test/mpas_test_core_field_tests.F
@@ -82,11 +82,12 @@ module test_core_field_tests
       integer, intent(out) :: ierr
 
       type ( att_list_type ), pointer :: srcList, destList
-      integer :: srcInt, destInt
-      integer, dimension(:), pointer :: srcIntA, destIntA
-      real (kind=RKIND) :: srcReal, destReal
+      integer :: srcInt, destInt, modifyInt
+      integer, dimension(:), pointer :: srcIntA, destIntA, modifyIntA
+      real (kind=RKIND) :: srcReal, destReal, modifyReal
       real (kind=RKIND), dimension(:), pointer :: srcRealA, destRealA
-      character (len=StrKIND) :: srcText, destText
+      real (kind=RKIND), dimension(:), pointer :: modifyRealA
+      character (len=StrKIND) :: srcText, destText, modifyText
 
       integer :: threadNum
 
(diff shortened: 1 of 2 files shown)


src/framework/mpas_attlist.F
```diff
@@ -565,7 +819,6 @@ subroutine mpas_remove_att(attList, attName, ierr)!{{{
                deallocate(cursor % attValueIntA)
             end if
             deallocate(cursor)
-
```
mgduda, line 568: Since we aren't otherwise modifying the `mpas_remove_att` routine, I'd suggest just keeping this blank line as it was.

src/core_test/mpas_test_core.F
```diff
@@ -138,7 +138,7 @@ function test_core_run(domain) result(iErr)!{{{
       end if
 
       !$omp parallel default(firstprivate) shared(domain, threadErrs)
-      call test_core_test_fields(domain, threadErrs, ierr)
+      call test_core_test_fields(domain, threadErrs, iErr)
```
mgduda, line 141: I'd generally agree that consistency (of capitalization, in this case) is great, but since this is the only modification to this file, I'd vote for not making this change and having one fewer file in the list of modified files.

src/framework/mpas_attlist.F
```diff
@@ -246,13 +254,259 @@ subroutine mpas_add_att_text(attList, attName, attValue, ierr)!{{{
...
+    character (len=*), intent(in) :: attValue !< Input: Updated Att. value
+    integer, intent(out), optional :: ierr !< Output: Error flag
+
+    type (att_list_type), pointer :: cursor
+
+    if (present(ierr)) ierr = 0
+
+    ! Return an error if there is somehow no attList
+    if (.not. associated(attList)) then
+      ierr = 1
```
mgduda, line 292: Since `ierr` is an optional argument, we should verify that it's `present` before setting it here.

mgduda (changes requested):

mgduda:
@dimomatt No need to adopt any of the following, but it's just an idea that came to mind.

Currently, the logic requires setting `ierr` in multiple places and it has multiple `return` statements:
```fortran
    if (present(ierr)) ierr = 0

    ! Return an error if there is somehow no attList
    if (.not. associated(attList)) then
      ierr = 1
      return
    else
      ! Traverese list looking for attName
      cursor => attlist
      do while (associated(cursor))
        if (trim(cursor % attName) == trim(attName)) then
          if (cursor % attType /= MPAS_ATT_TEXT) then
            if (present(ierr)) ierr = 1
          else
            write(cursor % attValueText,'(a)') trim(attValue)
          end if
          return
        end if
        cursor => cursor % next
      end do
      if (present(ierr)) ierr = 1
    end if
```
As an alternative, the following structure seems simpler to my mind:
```fortran
    if (present(ierr)) ierr = 1

    ! Traverese list looking for attName
    cursor => attlist
    do while (associated(cursor))
      if (trim(cursor % attName) == trim(attName)) then
        if (cursor % attType == MPAS_ATT_TEXT) then
          if (present(ierr)) ierr = 0
          write(cursor % attValueText,'(a)') trim(attValue)
        end if
        return
      end if
      cursor => cursor % next
    end do
```
*I haven't actually tested the above...* but the idea is to set `ierr` initially to a bad return code, and to only change it in case we succeed in modifying the requested attribute. There is only one `return` statement, and we make use of the fact that setting a pointer (`cursor`) to an unassociated pointer (potentially `attlist`) results in that pointer also being unassociated.

src/core_test/mpas_test_core.F
dimomatt: I would agree with you that a pedantic fix like that normally wouldn't be something to add to a commit like this, but as far as I can tell, if I leave it as "ierr", lines 142-146 won't actually report test failures caused by the field tests, as they check the "iErr" variable, not "ierr".

mgduda: Fortran isn't case sensitive, so I'm surprised that changing between `ierr` and `iErr` makes a difference.

src/framework/mpas_attlist.F
```diff
@@ -253,6 +261,213 @@ subroutine mpas_add_att_text(attList, attName, attValue, ierr)!{{{
...
+    type (att_list_type), pointer :: attList !< Input/Output: Attribute List
+    character (len=*), intent(in) :: attName !< Input: Att. name to modify
+    character (len=*), intent(in) :: attValue !< Input: Updated Att. value
+    integer, intent(out), optional :: ierr !< Output: Error flag
+
+    type (att_list_type), pointer :: cursor
+
+    if (present(ierr)) ierr = 1
+
+      ! Traverese list looking for attName
```
mgduda, line 290: It's probably worth correcting the indentation here to match that of the code block below.

Also, here and in other places, "Traverese" -> "Traverse".

src/framework/mpas_attlist.F
```diff
@@ -253,6 +261,213 @@ subroutine mpas_add_att_text(attList, attName, attValue, ierr)!{{{
...
+        if (cursor % attType == MPAS_ATT_INT) then
+          if (present(ierr)) ierr = 0
+          cursor % attValueInt = attValue
+        end if
+        return
+      end if
+      cursor => cursor % next
+    end do
+
+
```
mgduda, line 345: For consistency, I'd suggest just one blank line here.

src/framework/mpas_attlist.F
```diff
@@ -253,6 +261,213 @@ subroutine mpas_add_att_text(attList, attName, attValue, ierr)!{{{
...
+    if (present(ierr)) ierr = 1
+
+    ! Traverese list looking for attName
+    cursor => attlist
+    do while (associated(cursor))
+      if (trim(cursor % attName) == trim(attName)) then
+        if (cursor % attType == MPAS_ATT_REALA) then
+          if (present(ierr)) ierr = 0
+          cursor % attValueRealA(:) = attValue(:)
```
mgduda, line 462: Here and also in the `int1d` routine, it may be good to also verify that the size of the `attValue` array matches the size of the `attValueRealA` array.

src/core_test/mpas_test_core_field_tests.F
```diff
@@ -154,6 +155,52 @@ subroutine test_core_attribute_list_test(domain, threadErrs, ierr)!{{{
...
+         call mpas_modify_att(srcList, 'testInt', modifyInt)
+         call mpas_modify_att(srcList, 'testIntA', modifyIntA)
+         call mpas_modify_att(srcList, 'testReal', modifyReal)
+         call mpas_modify_att(srcList, 'testRealA', modifyRealA)
+         call mpas_modify_att(srcList, 'testText', modifyText)
+
+         call mpas_get_att(srcList, 'testInt', destInt)
+         call mpas_get_att(srcList, 'testIntA', destIntA)
+         call mpas_get_att(srcList, 'testReal', destReal)
+         call mpas_get_att(srcList, 'testRealA', destRealA)
```
mgduda, line 176: It looks like the `mpas_get_att` routine allocates memory for the returned array in the case of a 1-d real or integer attribute, so after the previous set of tests (and after these tests), I think it may be necessary to deallocate `destIntA` and `destRealA` to avoid a memory leak.

Since this is just a testing core, a few small memory leaks may in practice be inconsequential, but working towards a situation in which we can run the entire test core with valgrind and not turn up any memory leaks would be nice, I think.

mgduda (changes requested):

mgduda (approved):

PR #1095, opened 2023-06-28
tend_w_pgf and tend_w_buoy never explicitly used

theweathermanda:
The variables tend_w_pgf and tend_w_buoy are never explicitly used in the calculation of these terms. I've commented them out here since they aren't needed.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -3913,7 +3913,7 @@ module atm_time_integration
 
       real (kind=RKIND), dimension(:,:), pointer :: defc_a, defc_b
 
-      real(kind=RKIND), dimension(:,:), pointer :: tend_w_pgf, tend_w_buoy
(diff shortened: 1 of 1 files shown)


mgduda:
@theweathermanda I agree that `tend_w_pgf` and `tend_w_buoy` aren't actually used, and I think it would actually be fine to just delete these entirely (rather than commenting them out). Besides the code in `mpas_atm_time_integration.F`, we could remove the `Registry.xml` entries for these two fields. Also, in the current `develop` branch, there are a few more code lines that reference `tend_w_pgf` and `tend_w_buoy` in the context of OpenACC data movement.

If you'd like, I can help out in rebasing this branch onto the current `develop` branch and picking up a few of the additional lines related to `tend_w_pgf` and `tend_w_buoy`. Just let me know how you'd prefer to proceed.

Thanks in any case for spotting these unused variables, and apologies that it's taken so long to get to this PR!

theweathermanda:
@mgduda Thanks, Michael! I'd greatly appreciate any help with rebasing as I haven't kept up with the latest MPAS updates.

PR #1099, opened 2023-07-05
Fix OpenMP error in deallocation of rthdynten without certain cumulus schemes

mgduda:
This PR fixes an OpenMP error in the deallocation of `rthdynten` when certain cumulus schemes are not used.

When neither the Grell-Freitas nor the Tiedtke/nTiedtke cumulus schemes were used, the `rthdynten` array was allocated locally in the `compute_dyn_tend(...)` routine, since it is a required argument to the `compute_dyn_tend_work(...)` routine. The local `rthdynten` pointer was initialized and therefore implicitly had the `SAVE` attribute, which, according to 2.15.1.2 of the OpenMP 4.5 standard, made the `rthdynten` pointer shared:
```
  Local variables declared in called routines in the region and that have the
  save attribute, or that are data initialized, are shared unless they appear in
  a threadprivate directive.
```
The allocation and deallocation of `rthdynten` happened inside of a threaded region, leading to potential model failures when multiple threads attempted to deallocate the same shared `rthdynten` pointer.

The simple solution is to remove packages from the `rthdynten` field in the atmosphere core's `Registry.xml` file, and to remove the additional code that handled selective local allocation of this array in the `compute_dyn_tend(...)` routine.

Since `rthdynten` is allocated by the `compute_dyn_tend(...)` routine if it is not otherwise allocated due to packages, simply removing packages from the `rthdynten` field in the atmosphere core's `Registry.xml` file should have little impact on the memory usage of the dynamics, only increasing memory use outside of the dynamics in cases where neither
the Grell-Freitas nor the Tiedtke/nTiedtke schemes are used. However, this potentially larger memory use comes with the benefit of significantly simplified logic in the dynamics.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -2962,6 +2962,9 @@
                      description="Total dry air tendency from physics"
                      persistence="scratch" />
 
+                <var name="rthdynten" type="real" dimensions="nVertLevels nCells Time" units="K s^{-1}"
+                     description="tendency of temperature due to horizontal and vertical advections"/>
+
                 <!-- Scratch variables used when propagating cell-centered winds to edges -->
                 <var name="tend_uzonal" type="real" dimensions="nVertLevels nCells Time" units="m s^{-1} s^{-1}"
                      description="Total cell-centered zonal wind tendency from physics"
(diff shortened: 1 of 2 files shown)


weiwangncar (approved):
A user tested the change and it solved the problem he encountered when OpenMP is used.

PR #1101, opened 2023-07-11
Add arbitrary registry attributes

dimomatt:
This PR allows the addition of arbitrary attributes to variables defined in Registry.xml.

In pursuit of CF compliance, certain attributes will need to be added to variables. To make this easier, this pull request allows for the addition of arbitrary attributes by looping over all attributes in a variable, as opposed to hardcoding them in. Some support functions were also created to filter out attributes that aren't used in output files and to allow for the modification of attribute names if they have a different name in the code (namely, description vs. long_name).

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -18,6 +18,9 @@
 #define STR(s) #s
 #define MACRO_TO_STR(s) STR(s)
 
+#define NUM_MODIFIED_ATTRS 1
+#define NUM_IGNORED_ATTRS 6
+
 void write_model_variables(ezxml_t registry){/*{{{*/
 	const char * suffix = MACRO_TO_STR(MPAS_NAMELIST_SUFFIX);
 	const char * exe_name = MACRO_TO_STR(MPAS_EXE_NAME);
@@ -83,6 +86,79 @@ int write_field_pointer_arrays(FILE* fd){/*{{{*/
 	return 0;
 }/*}}}*/
 
+// Checks for a string in a list of strings.
+// Returns the index of the string if it does exist in the array,
+// and -1 if it does not appear in the array.
+int find_string_in_array(char *input_string, char *array[]){
+  int i;
+  for (i = 0; array[i] != NULL; i++ ){
+    if (strcmp(input_string, array[i]) == 0){
+      return i;
+    }
+  }
+  return -1;
+}
+
+// Add an attribute to the generated fortran code.
+// This handles the fortprint and all associated things
+void fortprint_add_attribute(FILE *fd, char* index, char *att_name, char *pointer_name_arr, char *temp_str){
+  fortprintf(fd, "      call mpas_add_att(%s %% attLists(%s) %% attList, '%s', '%s')\n", pointer_name_arr,
+                                                                                         index,
+                                                                                         att_name,
+                                                                                         temp_str);
+  return;
+}
+
+// Helper function to change attribute names in accordance with
+// "attrs_to_modify" within the "add_attribute_if_not_ignored" function
+void modify_attr(char* attr, const char* array[NUM_MODIFIED_ATTRS][2], size_t rows) {
+  for (size_t i = 0; i < rows; i++) {
+    if (strcmp(attr, array[i][0]) == 0) {
+        strcpy(attr, array[i][1]);
+        return;
+        }
+    }
+}
+
+void add_attribute_if_not_ignored(FILE *fd, char* index, char *att_name, char *pointer_name_arr, char *temp_str){
+  // Will certainly be faster to pass a pointer to this array than reconstruct
+  // it
+  const char *attrs_to_ignore[NUM_IGNORED_ATTRS][1024] = {
+    "name",
+    "type",
+    "dimensions",
+    "units",
+    "persistence",
+    "packages"
+  };
+
+  const char *attrs_to_modify[NUM_MODIFIED_ATTRS][2] = {
+    {"description", "long_name"}
+  };
+  
+  modify_attr(att_name, attrs_to_modify, NUM_MODIFIED_ATTRS);
+
+  if (find_string_in_array(att_name, attrs_to_ignore) == -1){
+      char *string, *tofree, *token;
+			string = strdup(temp_str);
+			tofree = string;
+			token = strsep(&string, "'");
+
+			sprintf(temp_str, "%s", token);
+
+			while ( ( token = strsep(&string, "'") ) != NULL ) {
+				sprintf(temp_str, "%s''%s", temp_str, token);
+			}
+      free(tofree);
+
+    fortprint_add_attribute(fd, index, att_name, pointer_name_arr, temp_str);
+  }
+  else{
+    printf("Ignoring variable with name %s\n", att_name);
+  }
+
+  return;
+}
 
 int set_pointer_name(int type, int ndims, char *pointer_name, int time_levs){/*{{{*/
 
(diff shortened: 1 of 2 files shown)


mgduda (changes requested):
Before any further review, I think it would be good to correct the whitespace in the new code. For C code, we generally use tabs for indentation. Also, I think we've gravitated towards placing the `*` for pointer variable declarations next to the name, e.g.,
```
const char *foo;
```

dimomatt:
Testing has been done to compare output in structs_and_variables.inc on develop vs this branch. Changes appear to be limited to the replacement of single quotes with double quotes, and the addition of attributes that were not hard coded to be included on develop.

mgduda:
> Testing has been done to compare output in structs_and_variables.inc on develop vs this branch. Changes appear to be limited to the replacement of single quotes with double quotes, and the addition of attributes that were not hard coded to be included on develop.

Which additional attributes are being included? We can add those to the list of ignored attributes. I think we should aim to get identical (with the exception of quotes v. apostrophes) generated code.

mgduda (changes requested):
I'm not yet sure if I'm just missing something, but I think the treatment of attributes for `var_array` XML elements may be different after applying the changes in this PR to the `develop` branch.

To demonstrate this, you can compile the `test` core on the `develop` branch, save the `src/core_test/inc/structs_and_variables.inc` file, then merge this PR, recompile, and check the attributes being added to the `test1` and `test2` variables in the `structs_and_variables.inc` code.

For example, for the `test1` variable, with changes in this PR we get:
```
      call mpas_add_att(r2Ptr % attLists(const_index) % attList, "units", "units1")
      call mpas_add_att(r2Ptr % attLists(const_index) % attList, "long_name", "Test var 1 in a var_array")
```
whereas before we got the following generated code:
```
         call mpas_add_att(r2Ptr % attLists(const_index) % attList, 'long_name', 'Test var 1 in a var_array')
         call mpas_add_att(r2Ptr % attLists(const_index) % attList, 'units', 'units1')
         call mpas_add_att(r2Ptr % attLists(const_index) % attList, '_FillValue', MPAS_REAL_FILLVAL)
```

mgduda:
I think there's just one final issue. If we add the `missing_value` attribute to both a `var_array` and one of its constituent `var`s, e.g.,
```
<var_array name="arrTest" type="real" dimensions="nCells" missing_value="FILLVAL">
        <var name="test1" units="units1" description="Test var 1 in a var_array" array_group="arrTest" missing_value="-1.0"/>
        <var name="test2" units="units2" description="Test var 2 in a var_array" array_group="arrTest"/>             
</var_array>
```
the current `develop` branch generates the following in `structs_and_variables.inc`:
```
         call mpas_add_att(r2Ptr % attLists(const_index) % attList, 'long_name', 'Test var 1 in a var_array')
         call mpas_add_att(r2Ptr % attLists(const_index) % attList, 'units', 'units1')
         call mpas_add_att(r2Ptr % attLists(const_index) % attList, '_FillValue', MPAS_REAL_FILLVAL)
```
So, the `missing_value` attribute on constituents is ignored, and a `missing_value` can only be applied to an entire `var_array`.

However, with the changes in this PR, the following code is generated in `structs_and_variables.inc`:
```
      call mpas_add_att(r2Ptr % attLists(const_index) % attList, "units", "units1")
      call mpas_add_att(r2Ptr % attLists(const_index) % attList, "long_name", "Test var 1 in a var_array")
      call mpas_add_att(r2Ptr % attLists(const_index) % attList, "_FillValue", MPAS_REAL_FILLVAL)
      call mpas_add_att(r2Ptr % attLists(const_index) % attList, "_FillValue", MPAS_REAL_FILLVAL)
```

src/tools/registry/gen_inc.c
```diff
@@ -1377,40 +1447,18 @@ int parse_var_array(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t var
...
-				token = strsep(&string, "'");
-				sprintf(temp_str, "%s", token);
 
-				while ( ( token = strsep(&string, "'") ) != NULL ) {
-					sprintf(temp_str, "%s''%s", temp_str, token);
+			for (attr = var_xml->attr; attr && *attr; attr+=2) {
+				// If the attr is "missing_value", ignore it and later on take
+				// the value from the var array.
+				if (strcmp(attr[0], "missing_value") == 0) {
+					printf("WARNING: Ignoring missing_value definition in variable defined under %s\n", vararrname);
```
mgduda, line 1455: Just a suggestion, but we could add the name of the variable in question to this warning message, e.g.,
```
printf("WARNING: Ignoring missing_value attribute for var %s defined in var_array %s\n", varname, vararrname);
```
(I also modified the terminology a bit.)

dimomatt: Agreed, change made

src/tools/registry/gen_inc.c
```diff
@@ -83,6 +109,51 @@ int write_field_pointer_arrays(FILE* fd){/*{{{*/
...
+		return;
+	}
+
+	// check if the attribute is numeric
+	if (find_string_in_array(att_name, NUMERIC_ATTRS, NUM_NUMERIC_ATTRS) >= 0){
+		format_string = "      call mpas_add_att(%s %% attLists(%s) %% attList, \"%s\", %s)\n";
+	} 
+	// If it isn't numeric, make sure to wrap  att_value in quotes
+	else {
+ 		format_string = "      call mpas_add_att(%s %% attLists(%s) %% attList, \"%s\", \"%s\")\n";
```
mgduda, line 151: It looks like we're getting some compilation issues with the `init_atmosphere` core when attributes are long enough that the line splitting in the `fortprintf` should be acting upon a long code line.

Here's an example warning message:
```
./inc/structs_and_variables.inc:576:0:

  576 | velocity for an edge")
      | 
Warning: Missing ‘&’ in continued character constant at (1) [-Wampersand]
```

I think these all stem from the use of `"` characters as the string delimiters in the generated `mpas_add_att` calls in the `add_attribute_if_not_ignored` function.

src/tools/registry/gen_inc.c
```diff
@@ -83,6 +109,122 @@ int write_field_pointer_arrays(FILE* fd){/*{{{*/
...
+		if ( stringIn[i] == '\'' ) {
+			if (resultIndex + 1 > bufferSize) return 1;
+			result[resultIndex++] = '\'';
+		}
+		if (resultIndex + 1 > bufferSize) return 1;
+		result[resultIndex++] = stringIn[i];
+	}
+
+	if (resultIndex + 1 > bufferSize) return 1;
+	result[resultIndex++] = '\0';
```
mgduda, line 154: Just a suggestion for simplification: I think the code on lines 153-154 can be removed if the for-loop on line 144 goes from 0 to `strlen(stringIn) + 1`.

src/tools/registry/gen_inc.c
```diff
@@ -83,6 +109,122 @@ int write_field_pointer_arrays(FILE* fd){/*{{{*/
...
+
+// Doubles single quotes in stringIn, and places the results in the buffer
+// stringOut. stringOut should be large enough to store (len(stringIn) * 2 + 1)
+// characters. Returns 1 if the buffer is too small for the result. 
+int escape_quotes(const char * stringIn, char * result, size_t bufferSize){
+	size_t resultIndex = 0;
+	
+	for (size_t i = 0; i < strlen(stringIn); i++) {
+		if ( stringIn[i] == '\'' ) {
+			if (resultIndex + 1 > bufferSize) return 1;
```
mgduda, line 146: This is probably a matter of personal preference, but if you change the condition to `resultIndex >= bufferSize`, then there's no need for arithmetic.

mgduda (changes requested):
Can you please clean up the whitespace? For example, compare lines 171 - 174 with lines 177 - 179.

Generally, we're looking for only tabs to indent to the beginning of a level of indentation, and then spaces if further indentation within a level is needed (to align with code above).

src/tools/registry/gen_inc.c
```diff
@@ -83,6 +109,122 @@ int write_field_pointer_arrays(FILE* fd){/*{{{*/
...
+		format_string = "      call mpas_add_att(%s %% attLists(%s) %% attList, '%s', %s)\n";
+	} 
+	// If it isn't numeric, make sure to wrap  att_value in quotes
+	else {
+ 		format_string = "      call mpas_add_att(%s %% attLists(%s) %% attList, '%s', '%s')\n";
+	}
+	
+	// Escape the quotes
+	err = escape_quotes(att_value, escaped_value, value_buffer_size);
+	if ( err == 1 ){
```
mgduda, line 200: We could save having to declare the local variable `err` if the conditional just used the call to `escape_quotes` on the left of the comparison (and similarly for the if-test on line 211).

mgduda (approved):

PR #1104, opened 2023-07-19
Update test Registry.xml to have numeric missing_values

dimomatt:
PR #1101 created a new array of names of numeric attributes, one of which is the "missing_value" attribute. In the test core, however, the only uses of "missing_value" assign it to the string "FILLVAL". This is nonnumeric, and causes compiler errors on that branch. This PR updates the two uses of "FILLVAL" to numeric values- 0.0 in the case of the one real valued field, and -999 in the case of the integer valued field to match other variables in the registry.

The test core compiles and runs properly with these modifications.

### DIFF
--- a/src/core_test/Registry.xml
+++ b/src/core_test/Registry.xml
@@ -76,7 +76,7 @@
 		<var name="xtime"                             type="text"     dimensions="Time"/>
 	</var_struct>
 	<var_struct name="mesh" time_levs="1">
-		<var_array name="arrTest" type="real" dimensions="nCells" missing_value="FILLVAL">
+		<var_array name="arrTest" type="real" dimensions="nCells" missing_value="0.0">
 			<var name="test1" units="units1" description="Test var 1 in a var_array" array_group="arrTest"/>
 			<var name="test2" units="units2" description="Test var 2 in a var_array" array_group="arrTest"/>
 		</var_array>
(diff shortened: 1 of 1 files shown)


mgduda:
Apologies for not doing a better job of checking the current behavior of the code earlier. It does appear that the registry code generator does translate the string `"FILLVAL"` into a numeric constant that is the specified fill value for real and integer values.

In the `test` core, we have, for example,
```
<var name="indexToCellID"                     type="integer"  dimensions="nCells" missing_value="FILLVAL"/>
```
in the `Registry.xml` file.

The logic beginning around [line 454 in gen_inc.c](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.0.1/src/tools/registry/gen_inc.c#L454-L475) translates `"FILLVAL"` to a numeric constant in the `missing_value` string for a variable.

Then, `missing_value` is used in the `fortprintf` call that writes a numeric attribute for a variable around [line 1640 of gen_inc.c](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.0.1/src/tools/registry/gen_inc.c#L1639-L1641) (and elsewhere for var_arrays).

It might be worth taking a look at PR #1101 to see if we can handle `"FILLVAL"` specifications for numerical attributes in an elegant way.

PR #1105, opened 2023-07-26
Remove the azure-pipelines.yml file, as it is no longer used

mgduda:
This PR removes the `azure-pipelines.yml` file from the top-level MPAS-Model directory,
as the file is no longer used.


gdicker1 (approved):
Looks good to me!

PR #1106, opened 2023-07-26
Start mpas_string_utils

dimomatt:
Create mpas_string_utils file

Throughout the MPAS code, there are a number of utility functions and subroutines that deal with the manipulation of strings. As more procedures are added, it begins to make sense to have a common space for these things to be contained in. To this effect, we have added the "mpas_string_utils" module, and moved the "mpas_split_string" subroutine from mpas_timekeeping.F to this module. Future updates will include more functions and subroutines that allow further manipulation of strings.

### DIFF
--- a/src/framework/Makefile
+++ b/src/framework/Makefile
@@ -34,7 +34,8 @@ OBJS = mpas_kind_types.o \
        xml_stream_parser.o \
        regex_matching.o \
        mpas_log.o \
-       mpas_halo.o
+       mpas_halo.o \
+       mpas_string_utils.o
 
 all: framework $(DEPS)
 
@@ -78,7 +79,7 @@ mpas_dmpar.o: mpas_sort.o mpas_kind_types.o mpas_derived_types.o mpas_hash.o mpa
 
 mpas_sort.o: mpas_kind_types.o mpas_log.o
 
-mpas_timekeeping.o: mpas_kind_types.o mpas_derived_types.o mpas_dmpar.o mpas_threading.o mpas_log.o
+mpas_timekeeping.o: mpas_string_utils.o mpas_kind_types.o mpas_derived_types.o mpas_dmpar.o mpas_threading.o mpas_log.o
 
 mpas_timer.o: mpas_kind_types.o mpas_dmpar.o mpas_threading.o mpas_log.o
 
--- a/src/framework/mpas_string_utils.F
+++ b/src/framework/mpas_string_utils.F
@@ -1,0 +1,66 @@
+! Copyright (c) 2023 The University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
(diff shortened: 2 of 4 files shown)


src/framework/Makefile
```diff
@@ -52,7 +53,7 @@ mpas_framework.o: mpas_dmpar.o \
                   mpas_io_units.o \
                   mpas_block_decomp.o \
                   mpas_stream_manager.o \
-                  mpas_c_interfacing.o
+                  mpas_c_interfacing.o
```
mgduda, line 56: Can you remove the space that was added here at the end of the line?

src/framework/mpas_string_utils.F
```diff
@@ -0,0 +1,54 @@
+! Copyright (c) 2021-2023 The University Corporation for Atmospheric Research (UCAR).
```
mgduda, line 1: Apologies for the misleading suggestion to follow the copyright statement from `mpas_halo.F` file. In this case, I think the copyright year can be 2023, since this is a new file only introduced this year.

src/framework/mpas_timekeeping.F
```diff
@@ -12,6 +12,7 @@ module mpas_timekeeping
    use mpas_dmpar
    use mpas_threading
    use mpas_log
+   use mpas_string_utils
```
mgduda, line 15: I mentioned that we're slowly improving, and here may be an example of where the existing code doesn't follow our current best practice of using the `only` clause with `use` statements. There's no need to modify the other, pre-existing `use` statements here, but for the `mpas_string_utils` module, we could update this to be:
```
   use mpas_string_utils, only : mpas_split_string
```

src/framework/mpas_string_utils.F
```diff
@@ -0,0 +1,54 @@
...
+!> \details
+!>  This module provides functions and subroutines used for string
+!>  manipulations and utilities.
+!
+!-----------------------------------------------------------------------
+module mpas_string_utils
+
+  contains
+
+  subroutine mpas_split_string(string, delimiter, subStrings)
```
mgduda, line 22: I suppose we might as well add a documentation block for this subroutine now that we're migrating it into a new module. If you'd like to chat at all about how we describe this routine, just let me know.

src/framework/mpas_string_utils.F
```diff
@@ -0,0 +1,54 @@
...
+    index = 1
+    do i = 1, len(string)
+      if (string(i:i) == delimiter) then
+          subStrings(index) = string(start:i-1)
+          index = index + 1
+          start = i + 1
+      end if
+    end do
+    subStrings(index) = string(start:len(string))
+  end subroutine mpas_split_string
```
mgduda, line 51: It's a minor detail, but leaving a blank line before the `end subroutine` line would follow existing style.

mgduda (changes requested):

mgduda (approved):

PR #1107, opened 2023-07-28
Add basic string utils tests

dimomatt:
This commit adds the infrastructure required to test the new mpas_string_utils function. This includes adding a test file, adding some minimal tests, and adding these files to the Makefile for the test core. The tests as they stand simply test basic string splitting capability.

### DIFF
--- a/src/core_test/mpas_test_core.F
+++ b/src/core_test/mpas_test_core.F
@@ -93,7 +93,8 @@ module test_core
       use test_core_streams, only : test_core_streams_test
       use test_core_sorting, only : test_core_test_sorting
       use mpas_halo_testing, only : mpas_halo_tests
-   
+      use test_core_string_utils, only : mpas_test_string_utils
+
       implicit none
    
       type (domain_type), intent(inout) :: domain
@@ -167,6 +168,10 @@ module test_core
          call mpas_log_write('Stream I/O tests: FAILURE', MPAS_LOG_ERR)
       end if
 
+      ! Run string util tests
+      call mpas_log_write('')
+      call mpas_test_string_utils(iErr)
+      call mpas_log_write('')
 
       call test_core_test_intervals(domain, threadErrs, iErr)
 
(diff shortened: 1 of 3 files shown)


src/core_test/mpas_test_core_string_utils.F
```diff
@@ -0,0 +1,70 @@
...
+        character :: delimiter
+        character(len=StrKIND), pointer, dimension(:) :: splitStrings
+        integer, intent(out) :: err
+        integer :: i
+
+        ! Test a basic case
+        delimiter = ' '
+        testString = 'This is a basic test'
+        call mpas_split_string(testString, delimiter, splitStrings)
+        if (trim(splitStrings(1)) /= 'This' .or. &
```
mgduda, line 35: If the returned `splitStrings` array contains too few elements, these checks may just result in a segfault. To guard against this possible failure mode, we could first check that the returned `splitStrings` is of the right size.

src/core_test/mpas_test_core.F
```diff
@@ -167,6 +168,14 @@ function test_core_run(domain) result(iErr)!{{{
          call mpas_log_write('Stream I/O tests: FAILURE', MPAS_LOG_ERR)
       end if
 
+      iErr = 0
+      ! Test string split function
+      call mpas_test_split_string(iErr)
+      if ( iErr == 0 ) then
+          call mpas_log_write('Split Strings tests: SUCCESS')
```
mgduda, line 175: Admitting that the current output from most of the `test` core tests is barely readable, we could nonetheless make the logged message for the string tests stand out a bit more. Right now, we get something like this in the log file:
```
 Stream I/O tests: SUCCESS
 Split Strings tests: SUCCESS
  Performing time interval tests
```
What do you think about modifying the print statements here so that the result in the log file looks more like this:
```
 Stream I/O tests: SUCCESS
           
 String utils tests: 
   mpas_split_string: SUCCESS
   
  Performing time interval tests
```
There could be a blank line, a generic line indicating that string utils will be tested, and then a single line for each function that we test. This seems like it might be a bit more extensible as we add more tests for the string utilities in future.

mgduda (changes requested):
Just a few small changes, then I think we'll be good to merge.

You may want to proofread the commit message, too. I think `function` should be plural in the first sentence, and `splititng` should be `splitting`.

src/core_test/mpas_test_core.F
dimomatt: yep, I like that change.

mgduda (approved):

PR #1108, opened 2023-07-31
Trim strings before splitting

dimomatt:
This commit adds a step to the mpas_split_string subroutine. In the process of adding tests, it was discovered that splitting a string while using a space as a delimiter causes the string to be split on every trailing space, leading to a splitStrings array that is equal to the size of the untrimmed input string, minus the number of non-space characters, plus one. This behavior is unwanted.

This has been solved this by trimming off trailing spaces before splitting the string. This has been tested with the new tests in PR #1107 and confirmed to pass these tests as expected. This PR should be merged prior to #1107, as the first test splits based on a space character delimiter, and will fail without these changes to the mpas_split_string routine.

### DIFF
--- a/src/framework/mpas_string_utils.F
+++ b/src/framework/mpas_string_utils.F
@@ -17,50 +17,56 @@
 !-----------------------------------------------------------------------
 module mpas_string_utils
 
-  contains
+    contains
 
-   !-----------------------------------------------------------------------
-   !  routine mpas_split_string
-   !
-   !> \brief This routine splits a string on a specified delimiting character
-   !> \author Michael Duda, Doug Jacobsen
-   !> \date   07/23/2014
-   !> \details This routine splits the given "string" on the delimiter
-   !>          character, and returns an array of pointers to the substrings
-   !>          between the delimiting characters. 
-   !
-   !-----------------------------------------------------------------------
-  subroutine mpas_split_string(string, delimiter, subStrings)
+    !-----------------------------------------------------------------------
+    !  routine mpas_split_string
+    !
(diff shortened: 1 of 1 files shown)


src/framework/mpas_string_utils.F
```diff
@@ -35,14 +35,16 @@ subroutine mpas_split_string(string, delimiter, subStrings)
     implicit none
 
     character(len=*), intent(in) :: string
+    character(len=len(trim(string))) :: trimString
```
mgduda, line 38: I'd propose moving the declaration of this local variable down to around line 42 with other local variables. Generally (and we should add this to a coding guidelines document!), I think we'd like to have dummy argument declarations first, then some separation, then local variable declarations. In newer code (e.g., in [mpas_halo.F](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.0.0/src/framework/mpas_halo.F#L118-L124)) we've gone so far as to add comments `! Arguments` and `! Local variables` to call out which declarations are which.

mgduda: Also, rather than `len(trim(string))` you can use `len_trim(string)`. Section 13 of the Fortran 2003 standard, which covers intrinsic procedures, is good reading if you're ever waiting on something to compile!

mgduda (changes requested):
In addition to the comments I've left in the code, it may be worth adding something to the documentation block (under `\details`) to explain that the input string will be trimmed before being split. (Imagine how useful it would have been had the original authors mentioned that strings are not trimmed, and that any trailing whitespace was subject to splitting on the delimiter character!)

mgduda (approved):

PR #1109, opened 2023-08-01
Remove original CMake build files

mgduda:
This PR removes CMake build files from the MPAS-Model repository to make way for a new CMake build for MPAS-Atmosphere.

Existing CMake build files were originally added in late 2019 to support E3SM. These build files didn't work for the `atmosphere` and `init_atmosphere` cores, and to make way for CMake build files that will support the `atmosphere` and `init_atmosphere` cores, this commit removes the previous `CMakeLists.txt` and `*.cmake` files.


gdicker1 (approved):
Looks good to me!

PR #1111, opened 2023-08-07
Add Cf compliant time to core_atmosphere

dimomatt:
CF compliance in time axis requires a unit of '<time unit> since YYYY-MM-DD hh:mm:ss'. This commit creates the 'Time' variable in the core_atmosphere Registry.xml, initializes it to 0, and changes the 'units' attribute to 'seconds since <start_timestamp>'. The start timestamp is modified from the initial xtime format to remove the underscore and replace it with a space to match the CF compliant time string format.

We can then start to update the Time variable at the same time as the xtime variable by adding 'dt' to the Time at time level 1, and saving it to time level 2. This produces a CF-compliant time variable that tracks xtime at each step.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -495,6 +495,7 @@
 			<var_array name="scalars"/>
 			<var name="initial_time"/>
 			<var name="xtime"/>
+			<var name="Time"/>
 			<var name="u"/>
 			<var name="w"/>
 			<var name="rho"/>
@@ -631,6 +632,7 @@
 			<var_array name="scalars"/>
 			<var name="initial_time"/>
 			<var name="xtime"/>
+			<var name="Time"/>
 			<var name="u"/>
 			<var name="w"/>
 			<var name="rho_zz"/>
@@ -922,6 +924,7 @@
 			<var name="zz"/>
 			<var name="initial_time"/>
 			<var name="xtime"/>
+			<var name="Time"/>
 			<var name="u"/>
 			<var name="w"/>
 			<var name="pressure"/>
@@ -1006,6 +1009,7 @@
 
 			<var name="initial_time"/>
 			<var name="xtime"/>
+			<var name="Time"/>
 
                         <!-- Begin isobaric diagnostics defined in diagnostics/Registry_isobaric.xml -->
 			<var name="mslp"/>
@@ -1477,6 +1481,9 @@
 
                 <var name="xtime" type="text" dimensions="Time" units="YYYY-MM-DD_hh:mm:ss"
                      description="Model valid time"/>
+
+                <var name="Time" type="real" dimensions="Time" units="seconds since YYYY-MM-DD hh:mm:ss"
+                    description="CF-compliant valid time" standard_name="time"/>
 
                 <!-- Prognostic variables: read from input, saved in restart, and written to output -->
                 <var name="u" type="real" dimensions="nVertLevels nEdges Time" units="m s^{-1}"
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -274,6 +274,8 @@ module atm_time_integration
       type (MPAS_TimeInterval_type) :: dtInterval
       character (len=StrKIND), pointer :: xtime
       character (len=StrKIND) :: xtime_new
+      real (kind=RKIND), pointer :: Time
+      real (kind=RKIND) :: Time_new
       type (mpas_pool_type), pointer :: state
       character (len=StrKIND), pointer :: config_time_integration
 
@@ -298,6 +300,14 @@ module atm_time_integration
       call mpas_pool_get_subpool(block % structs, 'state', state)
       call mpas_pool_get_array(state, 'xtime', xtime, 2)
       xtime = xtime_new
+
+      ! Get CF-compliant time at current timestep
+      call mpas_pool_get_array(state, 'Time', Time, 1)
+      Time_new = Time + dt
+
+      ! Write CF-compliant time for advanced timestep
+      call mpas_pool_get_array(state, 'Time', Time, 2)
+      Time = Time_new
 
    end subroutine atm_timestep
 
--- a/src/core_atmosphere/mpas_atm_core.F
+++ b/src/core_atmosphere/mpas_atm_core.F
@@ -51,6 +51,8 @@ module atm_core
       use mpas_atm_threading, only : mpas_atm_threading_init
       use atm_time_integration, only : mpas_atm_dynamics_init
       use mpas_timer, only : mpas_timer_start, mpas_timer_stop
+      use mpas_attlist, only : mpas_modify_att
+      use mpas_string_utils, only : mpas_string_replace
 
       implicit none
 
@@ -67,12 +69,14 @@ module atm_core
       type (mpas_pool_type), pointer :: mesh
       type (mpas_pool_type), pointer :: diag
       type (field2DReal), pointer :: u_field, pv_edge_field, ru_field, rw_field
+      type (field0DReal), pointer :: Time_field
       character (len=StrKIND), pointer :: xtime
       character (len=StrKIND), pointer :: initial_time1, initial_time2
       type (MPAS_Time_Type) :: startTime
 
       real (kind=RKIND), pointer :: nominalMinDc
       real (kind=RKIND), pointer :: config_len_disp
+      real (kind=RKIND), pointer :: Time
 
       integer, pointer :: nVertLevels, maxEdges, maxEdges2, num_scalars
       character (len=ShortStrKIND) :: init_stream_name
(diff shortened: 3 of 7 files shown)


src/core_atmosphere/mpas_atm_core.F
```diff
@@ -262,6 +266,15 @@ function atm_core_init(domain, startTimeStamp) result(ierr)
          call mpas_pool_get_array(state, 'xtime', xtime, 1)
          xtime = startTimeStamp
 
+         ! Set the units to be cf compliant 'seconds since <cf-timestamp>'
+         call mpas_pool_get_field(state, 'Time', Time_field) 
+         call mpas_modify_att(Time_field % attLists(1) % attList, 'units', &
+              'seconds since ' // mpas_string_replace(startTimeStamp, '_', ' '))
```
mgduda, line 272: So that we use a consistent reference time for restart runs, I think we'll need to use the `initial_time` variable rather than the start time. For example, if I do a cold-start run and simulate for two days, I get
```
	float Time(Time) ;
		Time:units = "seconds since 2010-10-23 00:00:00" ;
		Time:long_name = "CF convention compliant model valid time" ;
```
with values
```
 Time = 0, 21600, 43200, 64800, 86400, 108000, 129600, 151200, 172800 ;
```
However, if I restart from T+24h, I get
```
	float Time(Time) ;
		Time:units = "seconds since 2010-10-24 00:00:00" ;
		Time:long_name = "CF convention compliant model valid time" ;
```
with times
```
 Time = 21600, 43200, 64800, 86400 ;
```
in the model output. (Note that for a restart, the first history file record isn't written, hence no value of 0 for `Time`).

src/core_atmosphere/mpas_atm_core.F
```diff
@@ -262,6 +266,15 @@ function atm_core_init(domain, startTimeStamp) result(ierr)
...
+
+         ! Initialize Time to 0
+         call mpas_pool_get_array(state, 'Time', Time, 1)
+         Time = 0.
```
mgduda, line 276: Since `Time` is read from the restart file, I don't think we want to always reset it to zero here.

mgduda: Following up on this, all real-valued variables declared in the `Registry.xml` file are set to zero after being allocated. But, I think we may want to consider changes to the `init_atmosphere` core (to be included as additional commits in this PR branch) that add a CF-compliant `Time` variable to the ICs, LBCs, and surface update files.

src/core_init_atmosphere/Registry.xml
```diff
@@ -599,9 +601,10 @@
                         filename_interval="output_interval"
                         packages="lbcs"
                         immutable="true">
-
+
```
mgduda, line 604: Please remove the added space here.

src/core_init_atmosphere/Registry.xml
```diff
@@ -599,9 +601,10 @@
                         filename_interval="output_interval"
                         packages="lbcs"
                         immutable="true">
-
+ 
 			<var name="xtime"/>
-			<var_struct name="lbc_state"/>
+			<var name="Time"/>
+            <var_struct name="lbc_state"/>
```
mgduda, line 607: I think the indentation here also needs fixing.

src/core_init_atmosphere/Registry.xml
```diff
@@ -940,6 +943,9 @@
                 <var name="xtime" type="text" dimensions="Time" units="YYYY-MM-DD_hh:mm:ss"
                      description="Model valid time"/>
 
+                <var name="Time" type="real" dimensions="Time" units="Seconds since YYYY-MM-DD hh:mm:ss"
+                     desciption="CF convention compliant model valid time"/>
```
mgduda, line 947: `desciption` should be `description`.

mgduda (changes requested):
In addition to the comments, I think there should be some additional code in the init_atmosphere core to set the description for the `Time` variable to use the actual time of the initial conditions. Right now, the ICs file that's produced has the following `units`:
```
	float Time(Time) ;
		Time:units = "Seconds since YYYY-MM-DD hh:mm:ss" ;
```

mgduda:
After more testing, I think there are a few additions that we may like to include in this PR:

- [ ] Add the CF-compliant `Time` variable to all default output streams in the `atmosphere` core
- [ ] Add the `standard_name` attribute with a value of `time` to the `Time` variable
- [ ] Ensure that the value of the `Time` variable is correct for each of the time slices in surface update files produced by the `init_atmosphere` core
- [ ] Ensure that the value of the `Time` variable is correct for each of the time slices in lateral boundary files produced by the `init_atmosphere` core

src/core_init_atmosphere/mpas_init_atm_surface.F
```diff
@@ -44,29 +44,38 @@ subroutine init_atm_case_sfc(domain, dminfo, stream_manager, mesh, fg, state, di
...
+ start_time = mpas_get_clock_time(domain % clock, MPAS_START_TIME)
 
  do while (curr_time <= stop_time)
     call mpas_get_time(curr_time, dateTimeString=timeString)
     xtime = timeString
 
+    time_since_start = curr_time - start_time
+    call mpas_get_timeInterval(time_since_start, dt=dt)
+    Time = dt
+!
```
mgduda, line 78: I think it would be cleaner to remove the `!` that was added on this line -- just a blank line will suffice.

src/core_atmosphere/Registry.xml
```diff
@@ -1478,6 +1482,9 @@
                 <var name="xtime" type="text" dimensions="Time" units="YYYY-MM-DD_hh:mm:ss"
                      description="Model valid time"/>
 
+                <var name="Time" type="real" dimensions="Time" units="Seconds since YYYY-MM-DD hh:mm:ss"
```
mgduda, line 1485: Since the attribute that's ultimately written to files uses `seconds since ...`, would it be more consistent to un-capitalize `Seconds` here?

src/core_atmosphere/Registry.xml
```diff
@@ -1478,6 +1482,9 @@
...
+                    description="CF convention compliant model valid time" standard_name="time"/>
```
mgduda, line 1486: We could shorten this to something like "CF-compliant valid time".

src/core_init_atmosphere/Registry.xml
```diff
@@ -940,6 +943,9 @@
                 <var name="xtime" type="text" dimensions="Time" units="YYYY-MM-DD_hh:mm:ss"
                      description="Model valid time"/>
 
+                <var name="Time" type="real" dimensions="Time" units="Seconds since YYYY-MM-DD hh:mm:ss"
+                     description="CF convention compliant model valid time" standard_name="time"/>
```
mgduda, line 947: Similar comment as in `core_atmosphere/Registry.xml` regarding the capitalization of `Seconds` and the description.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -299,6 +301,14 @@ subroutine atm_timestep(domain, dt, nowTime, itimestep, exchange_halo_group)
       call mpas_pool_get_array(state, 'xtime', xtime, 2)
       xtime = xtime_new
 
+      ! Get CF-compliant time at current timestep
+      call mpas_pool_get_array(state, 'Time', Time, 1)
+      Time_new = Time + dt
+
+      ! Write CF-compliant time for advanced timestep
+      call mpas_pool_get_array(state, 'Time', Time, 2)
+      Time = Time_new
```
mgduda, line 310: An alternative to consider for this new code might be the following:
```
      ! Update CF-compliant time at end of timestep
      call mpas_pool_get_array(state, 'Time', Time, 1)
      call mpas_pool_get_array(state, 'Time', Time_new, 2)
      Time_new = Time + dt
```
There's one less line of code, and it seems clearer to me. But I recognize that this is purely subjective, so I wouldn't object to the code as it is in the PR. Just a different way to consider. (Note that the variant code here requires `Time_new` to be declared with the `pointer` attribute.)

mgduda (changes requested):
Other than a few minor comments, I think this PR looks good from a correctness and functionality perspective.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
dimomatt: I think I'm going to leave it as is for now, but this is certainly a more concise way to do this. I tried to follow the style of the xtime update in the lines above, which maybe helps with clarity if you're transitioning from using the xtime variable to the Time variable?

mgduda: In the case of the `xtime` update, the code is necessarily more complicated because we're updating a character string from a `real` and an `MPAS_Time_type`; on the other hand, in the case of `Time`, we're simply adding two `real` variables together. In my opinion, it's unnecessarily confusing to have the `Time` variable point to two different quantities only a few lines apart and to assign a value to `Time_new` only to then assign `Time_new` to `Time` (but not the same `Time` as a few lines earlier).

mgduda (approved):

PR #1112, opened 2023-08-08
Add ability for `clean` target to check for CORE existence

gdicker1:
Previously MPAS would run the clean recipe without regard for the core actually existing. This change ensures that the clean recipe will only run if a valid value for CORE is provided.


mgduda (approved):
That's an elegant fix!

PR #1113, opened 2023-08-23
Add NSSL 2/3-moment cloud microphysics

MicroTed:
Add NSSL 2/3-moment cloud microphysics

Adds option to use the NSSL 2-moment cloud microphysics scheme with 6 hydrometeor species (droplets, rain, cloud ice, snow, graupel, and hail). It includes the option for 3-moment rain/graupel/hail (adds reflectivity moment) and an option to turn off hail (for non-severe convection). Details on using the scheme are in docs/README.NSSLmp. Also adds a 3D reflectivity field for optional output (refl10cm). The parameterization code is (or will be) at parity with the current WRF PR for the 3-moment scheme https://github.com/wrf-model/WRF/pull/1876 (and eventual CCPP update).

Tested on daily 3-km regional CONUS forecasts (with v7.3), so similar performance is expected with v8.0

Outstanding issue: At least in the physics_mmm directory, the .mod files are not updated by recompiling newer .F files. They have to be deleted in order to be replaced. I don't know if this is general or a gfortran quirk.


MicroTed:
Hi, @ldfowler58 would you have time to look at this at some point? Let me know if this is the approach you want for physics_mmm or if there is more to shift from the model code to the interface. The physics module is an updated version from CCPP, so it has all the extra comments for documentation. The interface 'run' subroutine is from CCPP (for SCM/FV3), but isn't actually used since MPAS can call the microphysics directly with the "_p" arrays.
Thanks!

mgduda:
I think it would be good to separate the changes for NetCDF compression from the changes to add the NSSL microphysics scheme, as these (if I understand correctly) are unrelated changes.

MicroTed:
I agree that it would be cleaner to split it off the netcdf part, and it is only 2 files and I think no overlapping changes. It's a bare-bones implementation that assumes one would only choose the netcdf4 option because of compression being available. Otherwise the speed of pnetcdf for uncompressed writes really can't be beat.

MicroTed:
I moved the netcdf changes to a separate branch:  https://github.com/MicroTed/MPAS-Model/tree/dev-nc4deflate
I could make a draft PR or just leave it on ice for now -- whatever you want to do given your priorities and limited time. Ideally I would add a netcdf version check and an extra flag for deflate level etc.

ldfowler58:
I started to look at the changes and add-ons made by Ted in order to include the NSSL cloud microphysics scheme. Here my first suggestions for the top files that I have been modified:

1) In ./src/core_atmosphere/Registry.xml, I suggest to keep the "mp_" reference in front of the packages named nssl_hail_in, nssl_ccn_in, nssl3m_in, and nssl3m_hail_in. Users may appreciate that those packages point to cloud microphysics options. nssl3m_in can be changed to mp_nssl3m_in since you defined the NSSL 2-moment package as mp_nssl2m_in. Thanks.

2) In ./src/core_atmosphere/physics/mpas_atmphys_packages.F, can you please keep everything "lower case" and remove extra blank spaces to ensure "continuity" in how the sourcecode has been originally written (if not much to ask). For instance, change "
    if(config_nssl_hail_on) then
      nssl_hail_in = .true.
   endif

3) In ./src/core_atmosphere/physics/mpas_atmphys_control.F, change lines 
           config_microp_scheme .eq. 'mp_thompson' .or. &
           config_microp_scheme .eq. 'mp_nssl2m'  .or. &
to
           config_microp_scheme .eq. 'mp_nssl2m'      .or. &
           config_microp_scheme .eq. 'mp_thompson' .or. &

4) In ./src/core_atmosphere/physics/mpas_atmphys_vars.F (lines 197-203), do you mind putting one variable per line as for lines 188-195 so that I can add the definition of the different variables later. Thanks.


5) In ./src/core_atmosphere/physics/mpas_atmphys_manager.F, same comments as in 2) regarding lower case and space. Wouldn't it be simpler to use 75._RKIND instead of 1.25001*60._RKIND? Thanks.

Will look at the actual microphysics next. Thanks.

MicroTed:
Thanks for the comments! My thinking for leaving off "mp_" from nssl3m_in and nssl_hail_in etc. is that they are package options rather than full packages. They are not seen by the user, but rather activated by config_ options. So maybe it doesn't really matter? The package is still the mp_nssl2m_in, and the others are options (subpackages?) to turn variables on or off. 

Perhaps I could just rename mp_nssl2m_in to just mp_nssl_in? Or something like that? I also have a version with an option to run single-moment, but I'm not sure that's needed for a release code.

I can change the other formatting things to match the rest of the code. (Although I'm firmly in the camp of having a space after 'if' :) considering there is an unbalanced space in front of 'then', haha.)

MicroTed:
Updated (merged) to 8.2.1-hotfix. Also added function in init_atmosphere to generate LBCs for hydrometeor number concentrations when the LBC data only have mixing ratio. I still need to run some more tests and update the NSSL microphysics code. (Also some enhancements to the ideal initialization will be added.)

MicroTed:
I ran into an odd compiler issue with nvfortran/pgf90 at -O3 with mpas_atmphys_interface.F: If a pointer assignment is inside an IF structure, it may not work correctly. For example, the optional hail variables here:

```
             IF ( config_nssl_hail_on ) THEN
              call mpas_pool_get_dimension(state,'index_qh'  ,index_qh  )
              call mpas_pool_get_dimension(state,'index_nh'  ,index_nh  )
              call mpas_pool_get_dimension(state,'index_volh' ,index_volh  )
               qh   => scalars(index_qh,:,:)
               nh   => scalars(index_nh,:,:)
               volh => scalars(index_volh,:,:)
             ENDIF
```

At -O0 it works fine, but -O2 or 3, the pointer for qh seems to go randomly somewhere, maybe within the scalar structure or maybe not. I have only looked at the min/max values of qh. Obviously this is bad compiler behavior, but I found two work-arounds for it:
 1) Remove the IF clause and just always assign qh etc. even if the index is -1, but still have an IF clause around accessing the array. I don't know if this will cause debug problems, however (e.g., bounds check, although the array isn't being accessed)
 2) Just replace 
                  qh_p(i,k,j)   = qh(k,i)
      with
                 qh_p(i,k,j)   = scalars(index_qh,k,i)

I currently have option 2 in the PR code, but can switch it to 1 or a better suggestion.

Also for this reason I have not yet included mp_nssl with the mass/number tendencies (e.g., rqcmpten). Those seem to be diagnostic only, but maybe there will be a need for those.

ldfowler58:
Hi Ted: Thanks for pointing this out. I do not use nvfortran or pgf90 at all but it is definitely something that we should look into. Is it a "feature" that only occurs with the NSSL cloud microphysics or a "feature" that could also occur with, for instance, the Thompson cloud microphysics scheme? I'll investigate with a different cloud microphysics scheme separately.
Laura

MicroTed:
Hi, Laura,
It seems to only be a compiler bug with the IF-ENDIF sections, not with the SELECT-CASE at all. So I think the mp-thompson sections are OK. I did test the regular mp_thompson and it ran fine. We have some AMD epyc computers locally, and the nvfortran/pgf runs pretty well for them. I haven't tried gfortran yet for comparison, but will try to do that.

That said, I don't yet fully understand the structure of the 'scalar' array and how it works to reference the first index. (Basically how it converts a s(numscalar)%vals2d(nlevs,ncells) kind of record into something that looks like a 3D array -- too fancy!)

weiwangncar:
@MicroTed Apparently not updating .mod file is a gfortran feature as I just learnt. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=31587

MicroTed:
Updated to use the manage externals functionality to clone the microphysics code to physics_noaa

MicroTed:
Question for anybody: How much work would be involved to get the namelist filename from within mpas_atmphys_driver_microphysics.F?  (In the case where somebody is using the '-n' command line option to change it from the default.) The init subroutine takes the file name as an input, and it is currently hardwired to the default in the call. The reason is so that it can check for its internal namelist with rarely-used options. Most users would never use it, but it would be nice to correct this at some point.

EDIT: code has been updated to pass the namelist filename to the nssl init routine

jihyeonjang:
Hi Ted, 

Here are my suggestions based on my tests (before dev-nssl3m was merged into the current develop at [840ef2d](https://github.com/MPAS-Dev/MPAS-Model/pull/1113/commits/840ef2dc709065cd86c16d9c6bd15df2771d83d4)). 

 1. Radar reflectivity for Thompson scheme
There was an issue with the radar reflectivity output for the Thompson scheme after merging the NSSL-MP changes. To address this, please consider the following changes: 
 
**src/core_atmosphere/physics/mpas_atmphys_interface.F**
- Remove L1022 (no longer needed; related to the PR [#1329](https://github.com/MPAS-Dev/MPAS-Model/pull/1329)) 
```
L1022:        call mpas_pool_get_array(diag_physics,'refl10cm',refl10cm)
```
- Remove L1044 and L1053 for the Thompson scheme: 
```
L1044:              call mpas_pool_get_array(diag_physics,'refl10cm',refl10cm)
L1053:                 refl10cm(k,i) = refl10cm_p(i,k,j)
```

**src/core_atmosphere/physics/mpas_atmphys_driver_microphysics.F**
- Remove L163 and L274 for the Thompson scheme:
```
L163:           if(.not.allocated(refl10cm_p)) allocate(refl10cm_p(ims:ime,kms:kme,jms:jme))
L274:          if(allocated(refl10cm_p)) deallocate(refl10cm_p)
```

Note: The physics-intermediate refl10cm_p array was removed in the PR [#1329](https://github.com/MPAS-Dev/MPAS-Model/pull/1329), but it is still needed for NSSL-MP, since 10-cm radar reflectivity is computed in nssl_2mom_driver. 

&nbsp;

2. In src/framework/mpas_dmpar.F, the following lines are no longer needed after the commit [a5fc16f](https://github.com/MPAS-Dev/MPAS-Model/pull/1113/commits/a5fc16fbf5b42f09355b747982375a475fa31a1d). 
```
L87:   integer, public :: numprocs = 1
L313:       numprocs = mpi_size
```

&nbsp;

3. In src/core_atmosphere/physics/mpas_atmphys_packages.F, remove the following duplicated line. 
```
L127:    call mpas_log_write('    config_nssl_hail_on            = $l', logicArgs=(/config_nssl_hail_on/))
```

jihyeonjang:
I would also like to confirm the preferred namelist name for NSSL-MP: "mp_nssl2m" vs "mp_nssl". When config_nssl_3moment = .true., the 3-moment option becomes available, but still the microphysics name remains mp_nssl2m. Also, your branch name for this PR is dev-nssl3m. However, the subroutine names (nssl_2mom_init_const, nssl_2mom_driver, nssl_2mom_gs) are consistent with "mp_nssl2m", so mp_nssl2m could be more consistent. 

Could you decide which namelist name you prefer for NSSL-MP?

MicroTed:
> I would also like to confirm the preferred namelist name for NSSL-MP: "mp_nssl2m" vs "mp_nssl". When 

I agree it is awkward and was I just keeping the '2m' for legacy. It would make sense to simplify it to 'mp_nssl'. It can even be run in single-moment still, but I didn't see the need to add the flags to enable that option. I can change those.

I'll look at the changes for the reflectivity issue, too. Thanks for finding and pointing that out!

MicroTed:
@jihyeonjang OK, I have fixed the Thompson issue and updated tokens from mp_nssl2m to mp_nssl (with a backward compatibility to allow 'mp_nssl2m' for config input). 
mpas_atmphys_init.F has a flag f_qg for graupel, and although it doesn't seem to do anything, I also added one for hail in case it is ever needed.

MicroTed:
One more commit fixes an oversight with the MYNN scheme where the cloud droplet concentration tendencies were not being applied but the mass tendency was. It would have some effect, but it's unclear how large.

src/core_atmosphere/CMakeLists.txt
```diff
@@ -355,6 +377,7 @@ add_library(core_atmosphere ${ATMOSPHERE_CORE_SOURCES}
             ${ATMOSPHERE_CORE_PHYSICS_MMM_SOURCES}
             ${ATMOSPHERE_CORE_PHYSICS_WRF_SOURCES}
             ${ATMOSPHERE_CORE_PHYSICS_NOAA_SOURCES}
+            ${ATMOSPHERE_CORE_PHYSICS_NSSL-MP_SOURCES}
```
MicroTed, line 380: I don't know cmake very well, but I see that other packages have explicit source file lists. I can add that for NSSLmp if that's needed.

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -6081,6 +6101,70 @@ subroutine init_atm_case_lbc(timestamp, block, mesh, nCells, nEdges, nVertLevels
             rho_zz(k,iCell) = pressure(k,iCell) / rgas / (p(k,iCell) * t(k,iCell))
             rho_zz(k,iCell) = rho_zz(k,iCell) / (1.0 + scalars(index_qv,k,iCell))
          end do
+
+         ! check for missing moments that are needed
+         do k=1,nVertLevels
+            IF ( index_qs > 0 .and. index_ns > 0 ) THEN
```
mgduda, line 6107: Since `qs`, `ns`, etc. aren't defined as scalar constituents in the `Registry.xml` file, I think this code will lead to the use of unassociated pointers (`index_qs`, etc.).

Should there be additional changes in the `Registry.xml` file for the `init_atmosphere` core for scalar constituents beyond `qv`, `qc`, `qr`, `nifa`, and `nwfa`?

MicroTed: Yes, there is work to be done here! We recently ran into an issue where gfortran thought a pointer was associated because it wasn't explicitly initialized as null. (in the GSL version) One question here is whether to add the microphysics scheme to the init namelist as in core_atmosphere and set the microphysics variables accordingly (i.e., have to know what scheme will be used). That would limit file sizes by not including unneeded variables. Or do we just provide all possible variables to have a generic IC/BC?

mgduda: Given the number of new scalars, my inclination would be to only include these in LBC files if they will be needed in the model (i.e., if the NSSL scheme will be used). If we were to attach packages to each of the scalars, we could activate packages based on a specification in the `namelist.init_atmosphere` file of the microphysics scheme(s) that will be used in the model.

In future, though, it might be possible to scan through input files and activate packages for each scalar individually based on whether that scalar contains initial data in the input files. My thinking here is that, if we have no initial data for, say, `qs`, there would be no value in including it in the LBC files (as it would simply be zero everywhere, which is what the model will assume for a missing scalar anyway).

mgduda:
@MicroTed Regarding this part of the PR description,

>Outstanding issue: At least in the physics_mmm directory, the .mod files are not updated by recompiling newer .F files. They have to be deleted in order to be replaced. I don't know if this is general or a gfortran quirk.

I think this should be fixed with the merge of https://github.com/NCAR/MMM-physics/pull/6 .

MicroTed:
> @MicroTed Regarding this part of the PR description,
> 
> > Outstanding issue: At least in the physics_mmm directory, the .mod files are not updated by recompiling newer .F files. They have to be deleted in order to be replaced. I don't know if this is general or a gfortran quirk.
> 
> I think this should be fixed with the merge of [NCAR/MMM-physics#6](https://github.com/NCAR/MMM-physics/pull/6) .

Oh, it has been so long... I recall somebody saying that this behavior was a 'feature' and that gfortran only generated a new module if the interface or variables had changed -- or I'm misremembering. I haven't had any problems with gnu-12.x (or 15.x)

src/core_init_atmosphere/mpas_init_atm_cases.F
MicroTed: Also, perhaps the diagnosed scalars (e.g., number concentration from a single-moment species) could be done in 'atmosphere' if the LBC is not found? For example, if there is snow mixing ratio, but the scheme needs the number concentration, it might be better to create that when mpas is reading lbcs and finds qs but not ns. It would be basically the same code that I have now on the init side, and would save having to write 'bogus' LBCs. That combined with your idea of checking what is even available would help limit the number of vars in the LBC file.

mgduda: Apologies for the long delay in following up. I've finally spent a bit of time looking into the possibility of setting up packages for individual species based on either (1) the contents of the initial intermediate file or (2) a namelist-provided list of species. I've pushed a new branch, [demo/hydrometeor_packages](https://github.com/mgduda/MPAS-Model/commits/demo/hydrometeor_packages/), to my fork of the MPAS-Model repository with what I think might be a workable solution. In particular, the [last commit](https://github.com/mgduda/MPAS-Model/commit/4c81e2a4) on that branch adds the new package setup logic. Furthermore, I like your suggestion of diagnosing number concentrations and other scalars in the `atmosphere` core, as this would save on LBC file size.

If the approach take in the `demo/hydrometeor_packages` branch looks like it would work from your perspective, I think we could get that capability into the `develop` branch fairly quickly, and we could then rebase this PR branch and add packages for the new species in this PR (`qs`, etc.). Does that sound like a reasonable approach?

MicroTed: Looks like a pretty clever approach to peek at the intermediate file. Sounds good to me.

As far as diagnosing LBC values, I guess that would need to be done in mpas_atm_update_bdy_tend?

mgduda: I agree that `mpas_atm_update_bdy_tend` is probably the best place to diagnose number concentrations, etc. Around line 169 (in the v8.4.0 code) might be ideal, since we will have just read in the new boundary state into `lbc_*` time level 2 but not yet computed the time tendencies for the LBC fields: https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/dynamics/mpas_atm_boundaries.F#L169-L171 .

MicroTed: Yunheng Wang wrote a subroutine to calculate reflectivity in the init phase for the initial output time, and that routine could be used as a template (or expanded) for diagnosing missing number concentrations, etc.  
https://github.com/NOAA-National-Severe-Storms-Laboratory/frdd-MPAS-Model/blob/da78534402b341adc4a921d062f1c2785ae35907/src/core_atmosphere/physics/mpas_atmphys_driver_microphysics.F#L1465 (corrected link)
It would be cleaner to call a similar routine than trying to handle various microphysics here. For NSSL-MP, it is ultimately calling the nssl_column_dbz and calcnfromq routines from the module.

mgduda:
@MicroTed We've just merged PR #1467 into the `develop` branch, so if you're able to merge `develop` to this PR branch, or to rebase this PR branch onto the HEAD of the `develop` branch, that should provide the groundwork needed for adding packages to the new scalars in ICs and LBCs that are introduced by this PR. I'm glad to help with any of this -- just let me know if you'd like me to do so (though I may need to be added as a GitHub 'collaborator' on your fork?).

MicroTed:
@mgduda  I was out of the office last week, so I just got to this. It should be merged up to the current develop branch. I also found double-counting issue in the NSSL-MP that was causing excessive hail reflectivity in the 2-moment option. Hopefully good to go at this point.
I'm happy to add you as a collaborator, too.

PR #1116, opened 2023-08-31
Improve error message in MPAS I/O layer

mgduda:
This PR modifies the `mpas_io` and `mpas_io_streams` modules to provide more helpful error messages
in cases where errors occur in the PIO or SMIOL libraries. Prior to the changes in this PR, an error in
the PIO or SMIOL library would generally result in the following less-than-helpful message in the MPAS
log file:
```
MPAS IO Error: Bad return value from PIO
```
Now, for example, the error messages give further details about the nature of the error in PIO or SMIOL:
```
MPAS IO Error: PIO error -62: NetCDF: One or more variable sizes violate format constraints
```

Also included in this PR are changes to clean up several parts of the `mpas_io` module, for example,
replacing the PIO-specific error code `MPAS_IO_ERR_PIO` with a more generic name (`MPAS_IO_ERR_BACKEND`)
and adding code to properly set the error code in various places when the SMIOL library is used.

### DIFF
--- a/src/framework/mpas_io.F
+++ b/src/framework/mpas_io.F
@@ -6375,10 +6375,11 @@ module mpas_io
    end function MPAS_io_handle_dminfo
 
 
-   subroutine MPAS_io_err_mesg(ierr, fatal)
+   subroutine MPAS_io_err_mesg(ioContext, ierr, fatal)
 
       implicit none
 
+      type (mpas_io_context_type), intent(inout) :: ioContext
       integer, intent(in) :: ierr
       logical, intent(in) :: fatal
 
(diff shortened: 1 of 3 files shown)


gdicker1:
I'm still looking through this PR, but just noticed that this is to the master branch. Is that correct?

mgduda:
> I'm still looking through this PR, but just noticed that this is to the master branch. Is that correct?

Good catch -- it should be a PR to the `develop` branch. I've just updated the PR.

gdicker1 (approved):
Looks good to me!

PR #1118, opened 2023-09-27
Check for active packages for variables in prewrite_reindex

mgduda:
This PR modifies the `prewrite_reindex` routine to check for active packages
before reindexing fields in an output stream.

The `MPAS_stream_mgr_write` routine contains calls to the routines
`prewrite_reindex` and `postwrite_reindex` to translate between local and global
indices for any mesh indexing fields that are being written by a stream. In
cases where an indexing field is present in a stream but contains no active
packages, that field will not actually be written, and it is therefore not 
necessary to convert its indices to global indices and back to local indices.
    
This PR adds two new pool arguments to the `prewrite_reindex` routine so that
it can determine when an indexing field is not active and avoid any reindexing.
   
No package checks are needed in the `postwrite_reindex` routine, since only
indexing fields that have associated `*_save` pointers (e.g., `cellsOnCell_save`)
undergo global-to-local index translation, and these `*_save` pointers are all 
nullified in the `prewrite_reindex` routine and then set only if their associated
indexing fields are active.
    
Note that converting the indices for inactive indexing fields from local to
global and then back to local indices is not a problem from a correctness
perspective, but doing so results in unnecessary computation and data
movement. In that sense, the changes in this PR are an optimization only.

### DIFF
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -3363,7 +3363,7 @@ module mpas_stream_manager
            !
            if ( .not. stream % blockWrite ) then
               STREAM_DEBUG_WRITE(' -- Prewrite reindex for stream ' // trim(stream % name))
-              call prewrite_reindex(manager % allFields, stream % field_pool)
+              call prewrite_reindex(manager % allFields, manager % allPackages, stream % field_pool, stream % field_pkg_pool)
            end if
 
            ! 
@@ -4837,7 +4837,7 @@ module mpas_stream_manager
     !>                  the top of this module where this routine is made public.
     !
     !-----------------------------------------------------------------------
-    subroutine prewrite_reindex(allFields, streamFields) !{{{
+    subroutine prewrite_reindex(allFields, allPackages, streamFields, fieldPkgPool) !{{{
 
         implicit none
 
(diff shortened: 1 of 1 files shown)


dimomatt (approved):

src/framework/mpas_stream_manager.F
```diff
@@ -5159,6 +5197,11 @@ subroutine postwrite_reindex(allFields, streamFields) !{{{
 
         integer :: i, j, threadNum
 
+        character (len=StrKIND), pointer :: packages
+        logical :: active_field
+        integer :: err_level
```
mgduda, line 5202: Looking through these code changes again, I'm not sure why I've added these variable declarations here.

src/framework/mpas_stream_manager.F
```diff
@@ -4869,6 +4871,11 @@ subroutine prewrite_reindex(allFields, streamFields) !{{{
...
```
mgduda, line 4876: For consistency with planned changes to PR #1117, I should probably switch to snakeCase here, too.

mgduda: As in the comment discussion on PR #1118, I'm proposing to keep the snake_case here to make it easier to find and factor-out common package checking code in future.

mgduda:
@dimomatt After some thinking, I realized that we don't actually need package checks in the `postwrite_reindex` routine, and so I've force-pushed changes to this PR branch to remove code changes to that routine.

dimomatt (approved):
looks good to me!

PR #1119, opened 2023-10-07
Set single-precision as the default build

gdicker1:
This PR switches the logic for the PRECISION Makefile argument so single-precision variables are now the default and double-precision builds have to be requested. A check is also added to try to prevent any values besides "double" or "single" from being valid settings for PRECISION.

While double-precision variables allows for more accuracy, many users would be better served by the reduced data storage and throughput increase from single-precision variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -831,14 +831,15 @@ ifeq "$(OPENMP_OFFLOAD)" "true"
 	LDFLAGS += $(LDFLAGS_GPU)
 endif #OPENMP_OFFLOAD IF
 
-ifeq "$(PRECISION)" "single"
+ifeq "$(PRECISION)" "double"
+	FFLAGS += $(FFLAGS_PROMOTION)
+	PRECISION_MESSAGE="MPAS was built with default double-precision reals."
+else
+$(if $(PRECISION),$(info NOTE: PRECISION=single is unnecessary, single is the default))
 	CFLAGS += "-DSINGLE_PRECISION"
 	CXXFLAGS += "-DSINGLE_PRECISION"
 	override CPPFLAGS += "-DSINGLE_PRECISION"
 	PRECISION_MESSAGE="MPAS was built with default single-precision reals."
-else
-	FFLAGS += $(FFLAGS_PROMOTION)
-	PRECISION_MESSAGE="MPAS was built with default double-precision reals."
 endif #PRECISION IF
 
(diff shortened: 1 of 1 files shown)


gdicker1:
Force push from [7c4f828 ](https://github.com/MPAS-Dev/MPAS-Model/commit/7c4f828d1007379cdd2b65c63cc96dd33c3548e1)to [f4ef6af ](https://github.com/MPAS-Dev/MPAS-Model/commit/f4ef6afb5119446458c5c61dadb159d31f61d136)
[now](https://github.com/MPAS-Dev/MPAS-Model/pull/1119#event-10580674357) fixes typo in `filter-out` statement.

Makefile
```diff
@@ -831,14 +831,18 @@ ifeq "$(OPENMP_OFFLOAD)" "true"
 	LDFLAGS += $(LDFLAGS_GPU)
 endif #OPENMP_OFFLOAD IF
 
-ifeq "$(PRECISION)" "single"
+ifneq (,$(filter-out double single,$(PRECISION)))
+$(error PRECISION should be "", "single", or "double"; recieved value "$(PRECISION)")
```
mgduda, line 835: `recieved` -> `received`

mgduda (changes requested):
Although the HEAD of the PR branch does compile, `a261cdbc` doesn't compile if `PRECISION=single` or `PRECISION` is not specified. Generally, I think it's good to have every commit leave the code in a state that compiles (and ideally, runs).

Makefile
gdicker1: > Although the HEAD of the PR branch does compile, a261cdbc doesn't compile if PRECISION=single or PRECISION is not specified. Generally, I think it's good to have every commit leave the code in a state that compiles (and ideally, runs).

Whoops! It seems that I rebased poorly and added that fix to 3f25e68 instead of where I meant to. If you don't mind, I can do another force push to fix that and the mispelling of `received`.

gdicker1:
Force-push [3f25e68 ](https://github.com/MPAS-Dev/MPAS-Model/commit/3f25e689c391191e601194734e80aa5fa7018109) to [68c49d5 ](https://github.com/MPAS-Dev/MPAS-Model/commit/68c49d52f4cde00403cc2adce85bc06dc6a6f7bb) fixes commits so each builds and fixes typo in error message.

mgduda (approved):

PR #1120, opened 2023-10-09
Enable supersampling of 30" terrestrial fields: ter, ivgtyp, isltyp, greenfrac

mgduda:
This PR enables the supersampling of the 30" terrain elevation, land use category, soil category, and green vegetation fraction datasets through the introduction of a new namelist option, `config_30s_supersample_factor`, in the `&data_sources` namelist group in the `namelist.init_atmosphere` file.

The `config_30s_supersample_factor` option defaults to 1 (i.e., to no supersampling), which appears to be reasonable for MPAS meshes with >= 3 km cells. Experimentally, a supersampling factor of 2 was found to be sufficient for a 1-km quasi-uniform regional mesh centered over Colorado.

The `init_atm_map_static_data` routine, which is used by the `interp_terrain`, `interp_landuse`, and `interp_soilcat` routines in the init_atmosphere core, already supported supersampling, and all that was required for these fields was to pass the namelist supersampling value down the subroutine call stack. The code to remap the MODIS FPAR 30-arc-second greenness fraction data did not support supersampling, and so was modified to use the same approach as for terrain, land use, and soil category.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -140,6 +140,7 @@
  type(c_ptr) :: rarray_ptr
 
  integer, pointer :: supersample_fac
+ integer, pointer :: supersample_fac_30s
 
  real(kind=RKIND):: lat,lon,x,y
  real(kind=RKIND):: lat_pt,lon_pt
@@ -214,6 +215,7 @@
  call mpas_pool_get_config(configs, 'config_albedo_data', config_albedo_data)
  call mpas_pool_get_config(configs, 'config_maxsnowalbedo_data', config_maxsnowalbedo_data)
  call mpas_pool_get_config(configs, 'config_supersample_factor', supersample_fac)
+ call mpas_pool_get_config(configs, 'config_30s_supersample_factor', supersample_fac_30s)
 
  write(geog_data_path, '(a)') config_geog_data_path
  i = len_trim(geog_data_path)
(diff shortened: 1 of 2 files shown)


src/core_init_atmosphere/mpas_init_atm_static.F
```diff
@@ -750,10 +759,13 @@ subroutine init_atm_static(mesh, dims, configs)
 
             all_pixels_mapped_to_halo_cells = .true.
 
-            do j = tile_bdr + 1, tile_ny + tile_bdr, 1
-                do i = tile_bdr + 1, tile_nx + tile_bdr, 1
+            do j = supersample_fac_30s * tile_bdr + 1, supersample_fac_30s * (tile_ny + tile_bdr), 1
+                do i = supersample_fac_30s * tile_bdr + 1, supersample_fac_30s * (tile_nx + tile_bdr), 1
+
+                    ii = (i - 1) / supersample_fac_30s + 1
```
dimomatt, line 765: Would it be reasonable to change these variables from "ii" and "jj" to something more descriptive? "tile_x" and "tile_y" or something like that? I have a vendetta against ii and jj as indexing variables, but maybe it's more trouble to change here than it's worth.

mgduda: In my opinion, the index variable names are probably fine as they are. The code you've referenced mirrors what is done in a more general routine, [init_atm_map_static_data](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.0.1/src/core_init_atmosphere/mpas_init_atm_static.F#L1200), and so I think there's value in keeping the variable names consistent. Not that I have anything against descriptive variable names, but do we then insist that the loop variables `i` and `j` be given better names? `ii` and `jj` are the counterparts of `i` and `j`, but in non-supersampled indexing space.

dimomatt (approved):

PR #1125, opened 2023-11-03
Fix sprintf usage and memleaks in gen_inc.c

gdicker1:
Previous to this PR, gen_inc.c used sprintf in a standards non-compliant way to append to a string. This could cause certain compilers and software stacks could fail to build any MPAS core because the auto-generated *.inc files would be malformed. The uses of sprintf to append to a string were changed to strcat in this PR.

This PR also includes fixes for memory leaks in gen_inc.c.


gdicker1:
This PR recreates and extends the work in PR #715.

I checked this PR by comparing different builds of the atmosphere and init_atmosphere cores. The builds succeeded and no differences were found in the *.inc files.

mgduda (approved):
Looks good. Thanks for putting this PR together!

PR #1127, opened 2023-11-09
Add mpas_dmpar_sum_int8 to mpas_dmpar

dimomatt:
This PR adds a function to sum the values of an 8-byte integer across all MPI ranks. It is functionally identical to mpas_dmpar_sum_int, but returns an integer(kind=I8KIND) instead of a standard 4-byte integer.

### DIFF
--- a/src/core_test/Makefile
+++ b/src/core_test/Makefile
@@ -8,7 +8,8 @@ OBJS = mpas_test_core.o \
        mpas_test_core_timekeeping_tests.o \
        mpas_test_core_sorting.o \
        mpas_halo_testing.o \
-       mpas_test_core_string_utils.o
+       mpas_test_core_string_utils.o \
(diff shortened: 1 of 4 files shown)


mgduda:
I ended up writing a few tests for the new `mpas_dmpar_sum_int8` routine that I turned into tests in the test core. Would it be fine if I pushed a additional commit to this PR branch to include those tests?

dimomatt:
yes, that sounds good!

mgduda:
> yes, that sounds good!

I've just pushed another commit with a couple of tests. If all looks good to you, I think we're ready to merge.

mgduda (approved):

dimomatt:
Tests all look good and pass with the intel compiler- I am happy to have this merged

PR #1128, opened 2023-11-20
Get the size of the largest field in a stream before writing

dimomatt:
Different types of output file allow for different sizes of output variables. This commit checks the size of all of the fields within a stream in preparation for selecting file type with SMIOL in a future commit. A new function is added to mpas_stream_manager.F to call mpas_field_list_get_largest_field_size on all field lists for a given stream and all of its substreams.

mpas_field_list_get_largest_field_size returns an integer(kind=I8KIND) value of the size in bytes of the largest array associated with each field in a given field list. This function sums the sizes of the arrays on each MPI task before returning the value.

### DIFF
--- a/src/framework/mpas_io_streams.F
+++ b/src/framework/mpas_io_streams.F
@@ -4362,4 +4362,56 @@ module mpas_io_streams
 
    end subroutine put_get_field_atts
 
+   function mpas_field_list_get_largest_field_size(fieldList, dmInfo) result(maxSize)
+      implicit none
+      type (field_list_type), pointer :: fieldList
+      type (dm_info) :: dmInfo
+
+      integer(kind=I8KIND) :: maxSize
+      integer(kind=I8KIND) :: fieldSize
+      character (len=strKIND) :: fieldNameOut
+      type (field_list_type), pointer :: fieldCursor
+
+      maxSize = 0
+
+      if (.not. associated(fieldList)) return
+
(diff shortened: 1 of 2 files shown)


src/framework/mpas_io_streams.F
```diff
@@ -4362,4 +4362,56 @@ subroutine put_get_field_atts(handle, ioDirection, fieldname, attList)
...
+      maxSize = 0
+
+      if (.not. associated(fieldList)) return
+
+      fieldCursor => fieldList
+
+      do while (associated(fieldCursor))
+         select case (fieldCursor % field_type)
+            case (FIELD_0D_INT)
+               call mpas_dmpar_sum_int8(dmInfo, sizeof(fieldCursor % int0dField % scalar), fieldSize)
```
mgduda, line 4384: With the GNU compilers and `-std=f2008`, this code will fail to compile:
```
mpas_io_streams.F:4384:48:

 4384 |                call mpas_dmpar_sum_int8(dmInfo, sizeof(fieldCursor % int0dField % scalar), fieldSize)
      |                                                1
Error: Function ‘sizeof’ at (1) has no IMPLICIT type
```
The `sizeof` function is a GNU extension: https://gcc.gnu.org/onlinedocs/gfortran/SIZEOF.html

mgduda (changes requested):

PR #1129, opened 2023-11-29
Add build target 'intel' for the Intel oneAPI compilers

mgduda:
This PR adds a new build target (`intel`) for the Intel oneAPI compilers. The `ifx`, `icx`, and `icpx` compilers are used for serial code, while the `mpifort`, `mpicc`, and `mpic++` wrappers are used for MPI code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -654,6 +654,33 @@ cray:   # BUILDTARGET Cray Programming Environment
 	"OPENMP = $(OPENMP)" \
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI" )
 
+intel:   # BUILDTARGET Intel oneAPI Fortran, C, and C++ compiler suite
+	( $(MAKE) all \
+	"FC_PARALLEL = mpifort" \
+	"CC_PARALLEL = mpicc" \
+	"CXX_PARALLEL = mpic++" \
+	"FC_SERIAL = ifx" \
+	"CC_SERIAL = icx" \
+	"CXX_SERIAL = icpx" \
+	"FFLAGS_PROMOTION = -real-size 64" \
+	"FFLAGS_OPT = -O3 -convert big_endian -free -align array64byte" \
+	"CFLAGS_OPT = -O3" \
+	"CXXFLAGS_OPT = -O3" \
+	"LDFLAGS_OPT = -O3" \
+	"FFLAGS_DEBUG = -g -convert big_endian -free -check all -fpe0 -traceback" \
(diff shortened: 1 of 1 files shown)


mgduda:
On Derecho, after loading the `intel` module, setting the environment variable `INTEL_COMPILER_TYPE=ONEAPI` seems to cause the oneAPI compilers to be invoked in favor of the Intel classic compilers.

mgduda:
@gdicker1 It looks like builds also fail when using the PIO library with the existing `ifort` target. So the failures don't seem to be specific to the proposed new `intel` target.

gdicker1:
To test this PR I did builds on Derecho using the same software stack (`intel/2023.0.0`, `cray-mpich/8.1.25`,...) to try builds for:

- `ifort` and `intel` targets
- SMIOL and PIO I/O libraries
- test and atmosphere cores

Only the combination of `intel` and PIO failed to build. 

Other times I've seen this kind of error has been with the OneAPI compilers when I was mistakenly using them with the `ifort` target or when the OneAPI compilers have been used to compile the other mpas-* cores in EarthWorks [^1]. To me this points to an issue with how our `core_interface` code interacts with the OneAPI+PIO combo.

[^1]: NOTE: the mpas-framework is based more on the v7.3+ MPAS-Dev/MPAS-Model than whatever E3SM uses

mgduda:
That's true that I actually used the Intel oneAPI compilers with the `ifort` build target when noting a failure there with PIO as the I/O option. I think my point is that this demonstrates that it's not the new build target itself that's wrong, but rather, that there's a problem with the compiler (or its interaction with our code, to be fair). I'd propose that fixing builds with the oneAPI compilers should be a separate PR, and this PR shouldn't mix code changes with Makefile changes, as the two goals are fundamentally different.

gdicker1 (approved):
Thanks for that perspective, I now agree 100%. Code changes look good and resolving the build issue can be done another time.

HathewayWill:
https://community.intel.com/t5/Intel-oneAPI-Base-Toolkit/Intel-LLVM-MPI-commands/m-p/1560761/emcs_t/S2h8ZW1haWx8dG9waWNfc3Vic2NyaXB0aW9ufExSNE4zUEFJNTgzVFZJfDE1NjA3NjF8U1VCU0NSSVBUSU9OU3xoSw#M3358

HathewayWill:
Link for some information on the intel compilers:

https://community.intel.com/t5/Intel-oneAPI-Base-Toolkit/Intel-LLVM-MPI-commands/m-p/1560215

PR #1132, opened 2023-12-26
Add new mpas_stream_inquiry infrastructure module for querying contents of streams files

mgduda:
This PR adds a new infrastructure module, `mpas_stream_inquiry`, for querying 
the contents of streams XML files. The new module operates on instances of a new
derived type, `MPAS_streamInfo_type`, to determine whether a streams XML file
contains a definition of a given stream, whether a stream provides a
specification for a given attribute, and the values for any attributes present 
in the XML file.

The `mpas_stream_inquiry` module can only return information that was explicitly 
provided in a streams XML file, but it cannot, for example, return the
Registry-specified default value for an attribute unless that default value was
expressed in the XML file. However, being only dependent on the contents of a
streams XML file, the `mpas_stream_inquiry` module can be employed even before
the MPAS stream manager has been initialized.

The domain type now includes a pointer to an `MPAS_streamInfo_type` type, which 
is allocated and initialized with a core's streams file from within the
`mpas_init` routine, after `mpas_framework_init_phase2` but before defining
packages. This will allow cores to make queries concerning the stream
definitions provided by a user during the MPAS start-up process.

Also included in this PR are unit tests for the stream inquiry functionality.

### DIFF
--- a/src/driver/mpas_subdriver.F
+++ b/src/driver/mpas_subdriver.F
@@ -413,6 +413,7 @@ module mpas_subdriver
 
       use mpas_stream_manager, only : MPAS_stream_mgr_finalize
       use mpas_log, only : mpas_log_finalize, mpas_log_info
+      use mpas_derived_types, only : MPAS_streamInfo_type
 
       implicit none
 
@@ -420,6 +421,7 @@ module mpas_subdriver
       type (domain_type), intent(inout), pointer :: domain_ptr
 
       integer :: iErr
+      type (MPAS_streamInfo_type), pointer :: streamInfo
 
 
       !
@@ -440,7 +442,8 @@ module mpas_subdriver
       !
       call MPAS_stream_mgr_finalize(domain_ptr % streamManager)
 
-      if (domain_ptr % streamInfo % finalize() /= 0) then
+      streamInfo => domain_ptr % streamInfo
+      if (streamInfo % finalize() /= 0) then
          call mpas_log_write('Finalization of streamInfo object failed for core '//trim(domain_ptr % core % coreName), &
                              messageType=MPAS_LOG_ERR)
       end if
--- a/src/driver/mpas_subdriver.F
+++ b/src/driver/mpas_subdriver.F
@@ -47,6 +47,7 @@ module mpas_subdriver
       use mpas_timekeeping, only : mpas_get_clock_time, mpas_get_time
       use mpas_bootstrapping, only : mpas_bootstrap_framework_phase1, mpas_bootstrap_framework_phase2
       use mpas_log
+      use mpas_stream_inquiry, only : MPAS_stream_inquiry_new_streaminfo
  
       implicit none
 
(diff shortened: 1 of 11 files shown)


gdicker1 (commented):
I haven't checked runs yet, but builds had some comments/issues.

I was able to build `atmosphere` and `test` cores successfully for the `ifort` (intel-classic) and `gnu`[^2] targets on Derecho. The `nvhpc` target failed to build either due to an Internal Compiler Error.

[^2]: after adding the `-fallow-argument-mismatch` for the mpi library.

src/driver/mpas_subdriver.F
```diff
@@ -426,6 +440,12 @@ subroutine mpas_finalize(corelist, domain_ptr)
       !
       call MPAS_stream_mgr_finalize(domain_ptr % streamManager)
 
+      if (domain_ptr % streamInfo % finalize() /= 0) then
```
gdicker1, line 443: The `nvhpc` build points to this line as a source of the Internal Compiler error.

Part of my build log:
```
mpifort -D_MPI -DCPRPGI -DCORE_ATMOSPHERE -DMPAS_NAMELIST_SUFFIX=atmosphere -DMPAS_EXE_NAME=atmosphere_model - DMPAS_DEBUG -DSINGLE_PRECISION -DMPAS_NATIVE_TIMERS -DMPAS_GIT_VERSION=v8.0.1-58-gf642c4ea6-dirty -            DMPAS_BUILD_TARGET=nvhpc -DMPAS_SMIOL_SUPPORT -O0 -g -Mbounds -Mchkptr -byteswapio -Mfree -Ktrap=divz,fp,inv,  ovf -traceback -c mpas_subdriver.F -I/glade/u/apps/derecho/23.06/spack/opt/spack/netcdf/4.9.2/cray-mpich/8.1.  25/nvhpc/23.1/o74c/include -I/glade/u/apps/derecho/23.06/spack/opt/spack/parallel-netcdf/1.12.3/cray-mpich/8.  1.25/nvhpc/23.1/qzqm/include -I/glade/work/gdicker/mpas-work/PullRequestTesting/2023Dec31_MPASAPR1132/MPAS-    Model/src/external/SMIOL -I/glade/u/apps/derecho/23.06/spack/opt/spack/netcdf/4.9.2/cray-mpich/8.1.25/nvhpc/   23.1/o74c/include -I/glade/u/apps/derecho/23.06/spack/opt/spack/parallel-netcdf/1.12.3/cray-mpich/8.1.25/      nvhpc/23.1/qzqm/include -I../framework -I../operators -I../core_atmosphere -I../external/esmf_time_f90
nvfortran-Warning-CUDA Fortran or OpenACC GPU targets disables -Mbounds
NVFORTRAN-S-0000-Internal compiler error. check_member: member arrived with wrong derived type    5621         (mpas_subdriver.F: 443)
 0inform,   0 warnings,   1 severes, 0 fatal for mpas_finalize
make[4]: *** [Makefile:24: mpas_subdriver.o] Error 2
```

gdicker1 (changes requested):
I haven't checked runs yet, but builds had some comments/issues.

I was able to build `atmosphere` and `test` cores successfully for the `ifort` (intel-classic) and `gnu`[^2] targets on Derecho. The `nvhpc` target failed to build either due to an Internal Compiler Error.

[^2]: after adding the `-fallow-argument-mismatch` for the mpi library.

src/driver/mpas_subdriver.F
mgduda: I've just added a commit to this PR (f41bedae) to work around what at first glance appears to be a compiler bug. Just as a sanity check, I also verified that the code compiled (before the additional commit) with the Cray compilers, in addition to the Intel and GNU compilers.

gdicker1 (approved):
This looks good to go now!

PR #1133, opened 2023-12-27
Add `streamInfo` argument to `setup_packages` and `get_mesh_stream` interfaces for all cores

mgduda:
This PR adds the `domain % streamInfo` instance of an `MPAS_streamInfo_type` type
as an argument to the `setup_packages` and `get_mesh_stream` core interface routines for
all cores. Before the calls to these two core interface routines in `mpas_init`, the `streamInfo`
instance has been initialized with the core's streams XML file, enabling cores to make decisions
based on the streams XML file regarding packages and the stream to provide the mesh
information for a core.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core_interface.F
+++ b/src/core_atmosphere/mpas_atm_core_interface.F
@@ -337,15 +337,16 @@ module atm_core_interface
    !>  are available.
    !
    !-----------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


mgduda:
The changes in this PR depend on PR #1132.

gdicker1 (approved):
This builds for me and everything looks good!

PR #1137, opened 2024-01-24
Introduce "invariant stream" capability to MPAS-Atmosphere

mgduda:
This PR introduces the capability to store time-invariant fields in a separate file, thereby avoiding the need to store copies of these fields in every time-varying initial conditions or restart file.

Without any changes in workflow, MPAS-A will appear to a user to work as before. However, if a definition of the `"invariant"` input stream is added to the `streams.atmosphere` file at runtime, and if the `input_interval` is set to `"initial_only"`, e.g.,
```xml
<immutable_stream name="invariant"
                  type="input"
                  filename_template="x1.40962.L32.invariant.nc"
                  input_interval="initial_only" />
```
, the model will read all time-invariant fields from the designated file when starting up -- either from a cold-start or a restart; additionally, when restart files are periodically written during a simulation, time-invariant fields will be omitted from those restart output streams.

A file containing time-invariant fields can be produced while running the `init_atmosphere_model` program by setting `config_vertical_grid = true` while also setting `config_met_interp = false` in the `namelist.init_atmosphere` file for the real-data initialization case (init case 7). The invariant file can be used as input to the `init_atmosphere_model` program when interpolating real-data ICs with `config_met_interp = true`, and this same invariant file is suitable for use in the "invariant" stream in the `streams.atmosphere` file.

### DIFF
--- a/src/core_atmosphere/Makefile
+++ b/src/core_atmosphere/Makefile
@@ -26,7 +26,7 @@ core_reg:
 core_input_gen:
 	if [ ! -e default_inputs ]; then  mkdir default_inputs; fi
 	( cd default_inputs; $(NL_GEN) ../Registry_processed.xml namelist.atmosphere in_defaults=true )
-	( cd default_inputs; $(ST_GEN) ../Registry_processed.xml streams.atmosphere stream_list.atmosphere. listed )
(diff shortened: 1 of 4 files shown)


gdicker1 (approved):
I was able to test this and everything looks good.

120km restarts that don't contain invariant variables are about 73% the size of the default restarts in my test. Nice!

PR #1148, opened 2024-03-21
Generalize mpas_sphere_angle function to work for non-unit-radius spheres

mgduda:
This PR generalizes the `mpas_sphere_angle` function so that provides correct results for non-unit-radius spheres.

The methods previously employed in the `mpas_sphere_angle` function for computing the side lengths of a spherical triangle were only applicable to unit-radius spheres. This PR generalizes the computation of these side lengths so that the `mpas_sphere_angle` function can be used to compute spherical angles on spheres of arbitrary radius.


mgduda:
Although the changes in this PR generalize a geometry function, that generalization leads immediately to a fix for Issue #1149.

PR #1151, opened 2024-03-21
Added separate driver to call the parameterization of seaice

ldfowler58:
In preparation to adding the Noah-MP land surface scheme (LSM), and because the Noah and Noah-MP LSMs share the same parameterization of sea-ice, I moved all the sourcecode related to sea-ice that currently sits in the Noah LSM driver to its own driver (mpas_atmphys_driver_seaice.F). The call to the sea-ice driver is located in the main driver after the call to the Noah LSM. All outputs are bit for bit the same relative to the original parameterization.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver.F
@@ -17,6 +17,7 @@
  use mpas_atmphys_driver_pbl
  use mpas_atmphys_driver_radiation_lw
  use mpas_atmphys_driver_radiation_sw 
+ use mpas_atmphys_driver_seaice,only: allocate_seaice,deallocate_seaice,driver_seaice
  use mpas_atmphys_driver_sfclayer
  use mpas_atmphys_driver_oml
  use mpas_atmphys_constants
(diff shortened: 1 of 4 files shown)


src/core_atmosphere/physics/mpas_atmphys_driver.F
```diff
@@ -270,14 +271,21 @@ subroutine physics_driver(domain,itimestep,xtime_s)
...
        do thread=1,nThreads
           call driver_lsm(itimestep,block%configs,mesh,diag_physics,sfc_input, &
                           cellSolveThreadStart(thread), cellSolveThreadEnd(thread))
        end do
 !$OMP END PARALLEL DO
-       call deallocate_lsm(config_frac_seaice)
+       call deallocate_lsm
+
+       call allocate_seaice
+       do thread=1,nThreads
```
mgduda, line 284: So that the `driver_seaice` call can be parallelized with OpenMP, I think we should add directives around this `do` loop so that the code looks like:
```fortran
!$OMP PARALLEL DO
       do thread=1,nThreads
          call driver_seaice(block%configs,diag_physics,sfc_input, &
                             cellSolveThreadStart(thread),cellSolveThreadEnd(thread))
       enddo
!$OMP END PARALLEL DO
```
(essentially following what was done for the thread loop around the call to `driver_lsm`.

mgduda (changes requested):
As noted, I think it would be good to add OpenMP directives for loop around the `driver_seaice` call.

src/core_atmosphere/physics/mpas_atmphys_driver.F
ldfowler58: As you suggested, I added OpenMP directives around the loop that calls the main seaice driver. Thanks for the comment.

src/core_atmosphere/physics/mpas_atmphys_driver_seaice.F
```diff
@@ -0,0 +1,493 @@
+! Copyright (c) 2013,  Los Alamos National Security, LLC (LANS)
+! and the University Corporation for Atmospheric Research (UCAR).
```
mgduda, line 2: Sorry for not catching this in my first review. Since this is an entirely new code module (and file), I think we could update this copyright statement to:
```
Copyright (c) 2024 The University Corporation for Atmospheric Research (UCAR).
```

mgduda (changes requested):
Just one more small change regarding a copyright statement.

src/core_atmosphere/physics/mpas_atmphys_driver_seaice.F
ldfowler58: corrected as you suggested.

mgduda (approved):

Issue #1150, opened 2024-03-21
SIGFPE: Floating-point exception in Thompson microphysics v8.0.1 when DEBUG=true


mgduda:
When MPAS v8.0.1 is compiled with `DEBUG=true` and when running with the Thompson microphysics scheme (part of the `convection_permitting` suite), the model stops with a floating-point exception on [line 1860 of module_mp_thompson.F](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.0.1/src/core_atmosphere/physics/physics_wrf/module_mp_thompson.F#L1860).

mgduda:
This issue was pointed out in this forum post: https://forum.mmm.ucar.edu/threads/floating-point-exception-erroneous-arithmetic-operation-when-running-with-thompson-microphysics-and-debug-true.16159/#post-40111 .

As in the forum thread, it appears that the `mvd_r` variable is used without first being initialized at some k levels.

mgduda:
Closed by PR #1152.

PR #1152, opened 2024-03-21
Initialize `mvd_r` and `mvd_c` before first use in Thompson microphysics scheme

mgduda:
This PR initializes two local variables, `mvd_r` and `mvd_c`, in the `mp_thompson` routine to avoid the potential use of uninitialized memory.

Note: Identical changes to those included in this PR have separately been introduced into newer versions of the Thompson microphysics scheme, and so the changes here can be considered a back-port of these fixes.


mgduda:
The changes in this PR fix Issue #1150.

ldfowler58:
I completely agree with the description of the added initializations and approve this PR.

PR #1154, opened 2024-03-25
Remove LIBS from openacc_test in top-level Makefile

gdicker1:
Since these OpenACC compilation tests don't depend on NetCDF, GPTL, or other libraries, problems can be prevented by removing the linking commands that come from the LIBS variable. For example, the openacc_test rule will fail and the build will abort if a library the compiler expects to link against hasn't built yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -982,70 +982,70 @@ ifeq "$(OPENACC)" "true"
 	@# See whether the test programs can be compiled
 	@#
 	@echo "Checking [$(BUILD_TARGET)] compilers for OpenACC support..."
-	@( $(SCC) openacc.c $(CPPINCLUDES) $(CFLAGS) $(LDFLAGS) $(LIBS) -o openacc_c.out > openacc_c.log 2>&1; \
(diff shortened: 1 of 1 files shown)


gdicker1:
This fixes an issue that occurs when builds are attempted with SMIOL and `OPENACC=true` (assuming a target with valid `*_ACC` flags in the top-level Makefile). Since the SMIOL library hasn't been built at the early point in the build that the `openacc_test` rule is attempted, all attempts of these OpenACC compilation tests will fail without some modification to the LIBS variable (or removal of LIBS as this PR does).

E.g. this can be tested on NCAR's Derecho machine with module commands similar to below (note the lack of a module that provides PIO):
```
 module load ncarenv/23.06
 module load cuda
 module load nvhpc
 module load cray-mpich
 module load cray-libsci
 module load netcdf-mpi
 module load parallel-netcdf
 module load ncarcompilers
```
and the build command:
```
make nvhpc CORE=test OPENACC=true
```

mgduda (approved):
With the changes in this PR, I'm now able to build with the NVHPC compilers with SMIOL and with `OPENACC=true`.

PR #1157, opened 2024-04-01
Merge v8.0.2 from MPAS-Dev

favba:
Merge v8.0.2 from MPAS-Dev with the following modifications:

Merge branch 'hotfix-v8.0.2'

This merge corrects several issues in the MPAS-Atmosphere model, specifically:

* Generalize the mpas_sphere_angle function to provide correct results for
  non-unit-radius spheres. This generalization fixes problems in remapping
  static fields to the outermost layer of specified-zone cells in limited-area
  meshes.

* Initialize mvd_r and mvd_c before first use in the Thompson microphysics
  scheme to prevent the potential use of uninitialized memory, which could lead
  to floating-point exceptions.

* Correct the units for the grid scale input field to the Thompson cloud
  fraction scheme.


favba:
Sorry, meant to create a PR to our fork.

PR #1159, opened 2024-04-03
added initialization of local variable we in bl_ysu.F

ldfowler58:
In ./src/core_atmosphere/physics/physics_mmm/bl_ysu.F, added the initialization of the local variable we.

### DIFF
--- a/src/core_atmosphere/physics/physics_mmm/bl_ysu.F
+++ b/src/core_atmosphere/physics/physics_mmm/bl_ysu.F
@@ -561,6 +561,7 @@
      hgamt(i)  = 0.
      hgamq(i)  = 0.
      wscale(i) = 0.
+     we(i)     = 0.
      kpbl(i)   = 1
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/physics/physics_mmm/bl_ysu.F
```diff
@@ -892,6 +893,7 @@ subroutine bl_ysu_run(ux,vx,tx,qvx,qcx,qix,nmix,qmix,p2d,p2di,pi2d,     &
        endif
      enddo
    enddo
+
```
mgduda, line 896: Would it be preferable to remove this blank line that was added?

ldfowler58: Done. I must have added this blank line when debugging the sourcecode and adding print statements. This extra line has been removed as suggested.

mgduda (approved):

PR #1163, opened 2024-05-02
Introduce framework routine `mpas_framework_report_settings` for reporting compile-time options and run-time settings

mgduda:
This PR introduces a new framework routine, `mpas_framework_report_settings`, that reports compile-time options and run-time settings to the log file. For example, the following is written by the routine when MPAS is compiled with the `nvhpc` build target with `OPENACC=true`:
```
 Output from 'git describe --dirty': v8.1.0-dirty

 Compile-time options:
   Build target: nvhpc
   OpenMP support: no
   OpenACC support: yes
   Default real precision: single
   Compiler flags: optimize
   I/O layer: SMIOL

 Run-time settings:
   MPI task count: 4
```
The output is core-independent, and any core can therefore invoke the `mpas_framework_report_settings`, for example, in its `setup_log` routine.

As part of this PR, a call to `mpas_framework_report_settings` has been added to the test core's `test_setup_log` routine, and copy-and-paste code for writing the same compile-time option and run-time setting information in the init_atmosphere and atmosphere cores has been replaced with calls to `mpas_framework_report_settings`.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core_interface.F
+++ b/src/core_atmosphere/mpas_atm_core_interface.F
@@ -262,9 +262,7 @@ module atm_core_interface
 
       use mpas_derived_types, only : mpas_log_type, domain_type
       use mpas_log, only : mpas_log_init, mpas_log_open
-#ifdef MPAS_OPENMP
-      use mpas_threading, only : mpas_threading_get_num_threads
-#endif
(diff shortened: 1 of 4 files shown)


gdicker1 (approved):
Looks good to me!

PR #1166, opened 2024-05-16
Add stream_function and velocity_potential fields to the diag struct.

jim-p-w:
This PR adds two new fields to the diag struct, stream_function and velocity_potential.


mgduda (approved):

1JunGu:
is there code added for calculating stream function in /dianostic ？

mgduda:
> is there code added for calculating stream function in /dianostic ？

I don't think so. As far as I'm aware, MPAS-JEDI just writes these two fields to some (non-default) output streams to serve as placeholders.

PR #1167, opened 2024-05-16
Add da_state input/output stream

jim-p-w:
This PR adds a "da_state" inout/output stream.
This stream is only used by MPAS-JEDI.
This stream is associated with the "jedi_da" package, and so it will only be active if config_jedi_da = true.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -1135,6 +1135,73 @@
                         <var_array name="lbc_scalars" />
 
                 </stream>
+
+		<stream name="da_state"
+                        type="input;output"
+                        filename_template="mpasout.$Y-$M-$D_$h.$m.$s.nc"
+                        input_interval="initial_only"
+                        output_interval="0_06:00:00"
+                        immutable="true"
+                        packages="jedi_da">
+
+			<var_array name="scalars"/>
+			<var name="initial_time"/>
+			<var name="xtime"/>
+			<var name="cldfrac"/>
+			<var name="re_cloud" packages="mp_thompson_in;mp_wsm6_in"/>
+			<var name="re_ice" packages="mp_thompson_in;mp_wsm6_in"/>
+			<var name="re_snow" packages="mp_thompson_in;mp_wsm6_in"/>
+			<var name="u"/>
+			<var name="w"/>
+			<var name="rho"/>
+			<var name="pressure_p"/>
+			<var name="pressure_base"/>
+			<var name="theta"/>
+			<var name="relhum"/>
+			<var name="rho_base"/>
+			<var name="theta_base"/>
+			<var name="uReconstructZonal"/>
+			<var name="uReconstructMeridional"/>
+			<var name="surface_pressure"/>
+			<var name="refl10cm_max"/>
+			<var name="rainc"/>
+			<var name="rainnc"/>
+			<var name="lai"/>
+			<var name="isice_lu"/>
+			<var name="iswater_lu"/>
+			<var name="sfc_albbck"/>
+			<var name="sfc_albedo"/>
+			<var name="sfc_emibck"/>
+			<var name="mavail"/>
+			<var name="sfc_emiss"/>
+			<var name="thc"/>
+			<var name="ust"/>
+			<var name="xicem"/>
+			<var name="z0"/>
+			<var name="znt"/>
+			<var name="skintemp"/>
+			<var name="snow"/>
+			<var name="snowc"/>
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/Registry.xml
```diff
@@ -1135,6 +1135,72 @@
                         <var_array name="lbc_scalars" />
 
                 </stream>
+		<stream name="da_state"
```
mgduda, line 1138: Could you add a blank line between the end of the "lbc_in" stream and the beginning of the "da_state" stream?

src/core_atmosphere/Registry.xml
```diff
@@ -1135,6 +1135,72 @@
...
+                        type="input;output"
+                        filename_template="mpasout.$Y-$M-$D_$h.$m.$s.nc"
+                        input_interval="initial_only"
+                        output_interval="0_06:00:00"
+                        immutable="true">
```
mgduda, line 1143: Since the "da_state" stream is not conditionally compiled into the model, I think we should attach the "jedi_da" package to the stream. You can check the "lbc_in" stream for an example of how to attach a package to a stream.

mgduda (changes requested):

src/core_atmosphere/Registry.xml
jim-p-w: Yes.

src/core_atmosphere/Registry.xml
jim-p-w: Yes I will. FMI What are the semantics of adding a "packages" attribute to a stream?

src/core_atmosphere/Registry.xml
jim-p-w: Done.

src/core_atmosphere/Registry.xml
jim-p-w: Done.

mgduda: A stream with packages will only be read or written (effectively, it will only be active) if at least one of the attached packages is active. So in this case, even though the "da_state" stream has an `input_interval` and and `output_interval`, it will only actually be read or written if `config_jedi_da` is `true` in the namelist (which activates the "jedi_da" package.

mgduda: I think you need to remove the `>` character from the line with the `immutable` attribute.

mgduda (changes requested):

src/core_atmosphere/Registry.xml
jim-p-w: Right, done. My bad for not running the nit tests.

src/core_atmosphere/Registry.xml
```diff
@@ -1135,6 +1135,74 @@
...
+			<var name="surface_pressure"/>
+			<var name="refl10cm_max"/>
+			<var name="rainc"/>
+			<var name="rainnc"/>
+			<var name="lai"/>
+			<var name="isice_lu"/>
+			<var name="iswater_lu"/>
+			<var name="sfc_albbck"/>
+			<var name="sfc_albedo"/>
+			<var name="sfc_albedo_seaice"/>
```
mgduda, line 1174: This field has been removed from MPAS-Atmosphere, and trying to include it in this stream leads to the following build error:
```
ERROR: Trying to add undefined variable sfc_albedo_seaice to stream da_state.
```
Based on the email discussion with @liujake, it should be safe to remove `sfc_albedo_seaice` from the "da_state" stream.

mgduda (changes requested):
Besides removing sfc_albedo_seaice, I would suggest including at least a couple of additional points in the PR description: (1) this stream is only used by MPAS-JEDI, and (2) the stream is associated with the "jedi_da" package, and so it will only be active if `config_jedi_da = true`.

jim-p-w:
> Besides removing sfc_albedo_seaice, I would suggest including at least a couple of additional points in the PR description: (1) this stream is only used by MPAS-JEDI, and (2) the stream is associated with the "jedi_da" package, and so it will only be active if `config_jedi_da = true`.

Should I add points (1) and (2) to the commit message too?

mgduda:
> Should I add points (1) and (2) to the commit message too?

Sure, I think that would be good in case someone happens upon the commit itself and not the merge commit (for which we'll use the PR description).

src/core_atmosphere/Registry.xml
jim-p-w: The field is removed.

mgduda (approved):

PR #1168, opened 2024-05-21
CMake build for MPAS v8.1.0 atmosphere core

amstokely:
This PR adds support for building the atmosphere and init atmosphere MPAS cores with CMake vs GNU Make. CMake allows the support MPAS cores to be built outside the repository source directory, simplifies platform/compiler specific linking/compilation flags, and also allows MPAS to be integrated into other CMake managed projects (eg. mpas-jedi).
This PR was specifically created so that MPAS v8.1.0 can be used by mpas-jedi.

The CMake build successfully runs a small calculation (240km test on derecho) in single and double precision modes when compiled with GNU 11.4.1 and GNU 12.2.0 compilers. The results are comparable to those obtained when running the same calculation using the current develop branch.


1JunGu:
why close？ a good job

amstokely:
> why close？ a good job

I'm cleaning up the commit history and then I'll reopen it!

PR #1172, opened 2024-05-28
Add equation of time correction to local time for sw radiation calculation

weiwangncar:
This PR adds equation of time correction to the local time for computing shortwave radiation. The EOT adds a max of 16 min in Nov, and a min of -14 min in Feb. This correction has been added to WRF since 2014.

The PR improves accuracy of shortwave radiation, but relatively small overall effects are expected.


mgduda:
@weiwangncar It seems like there are quite a few magic numbers in the new code:
```fortran
       da=6.2831853071795862*(julian-1.)/365.
       eot=(0.000075+0.001868*cos(da)-0.032077*sin(da) &
            -0.014615*cos(2*da)-0.04089*sin(2*da))*(229.18)
       xt24 = mod(xtime+radt*0.5,1440.)+eot
```
Could some of these be defined as parameters with descriptive names? For example, it seems like `6.2831853071795862` could be `twopi` or `2.0 * pi`. I have no idea what `0.032077`, `0.001868`, etc. represent.

dudhia:
Yes, that is 2*pi. The other numbers are a fit to an annual cycle.

mgduda:
> Yes, that is 2*pi. The other numbers are a fit to an annual cycle. They describe the phase.

If we can't create descriptively named parameters for the numbers, would it at least be possible to lay out the equation with an explanation of the parameters in a comment block?

dudhia:
It's a simple harmonic function fit based on the day of the year. The numbers are empirical fit coefficients.

weiwangncar:
Will revisit at a later time.

weiwangncar:
Here is a [reference](https://gml.noaa.gov/grad/solcalc/solareqns.PDF) for the equation used, the second equation on the first page. Would this help with this PR?

jihyeonjang:
@weiwangncar. Could you rebase this PR onto the current develop branch? It has been a while since this PR was opened, so there are now conflicts between this branch and develop. Thank you!

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
```diff
@@ -10035,9 +10024,14 @@ subroutine rrtmg_swrad( &
 
        !--- calculate the cosine of the solar zenith angle at the current time step to determine if the sun is
        !    above or below the horizon (xt24 is the fractional part of simulation days plus half of radt in 
-       !    units of minutes, julian is in days, and radt is in minutes). do not call rrtmg_sw is night-time:
+       !    units of minutes, julian is in days, and radt is in minutes). eot (equation of time) is a correction
+       !    of mean solar time due to the difference between that time and the apparent solar time.
```
jihyeonjang: Could you add this link for the reference of this modification here?  (link: https://gml.noaa.gov/grad/solcalc/solareqns.PDF)

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
```diff
@@ -10035,9 +10024,14 @@ subroutine rrtmg_swrad( &
 
        !--- calculate the cosine of the solar zenith angle at the current time step to determine if the sun is
        !    above or below the horizon (xt24 is the fractional part of simulation days plus half of radt in 
-       !    units of minutes, julian is in days, and radt is in minutes). do not call rrtmg_sw is night-time:
+       !    units of minutes, julian is in days, and radt is in minutes). eot (equation of time) is a correction
+       !    of mean solar time due to the difference between that time and the apparent solar time. 
+       !    do not call rrtmg_sw is night-time:
        dorrsw = .true.
-       xt24 = mod(xtime+radt*0.5,1440.)
+       da=6.2831853071795862*(julian-1.)/365.
```
jihyeonjang: It seems that 6.2831853071795862 can be replaced with `2*pi`. Could you modify this? For example, we can get `pii` from `mpas_atmphys_constants`.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
```diff
@@ -10035,9 +10024,14 @@ subroutine rrtmg_swrad( &
...
+       eot=(0.000075+0.001868*cos(da)-0.032077*sin(da) &
+            -0.014615*cos(2*da)-0.04089*sin(2*da))*(229.18)
```
jihyeonjang: Do you think we can replace 229.18 in this equation for `eot` with `4.*180./pi`, which converts units from radians to minutes of solar time. As I mentioned above, I think `pi` can be obtained as `pii` from `mpas_atmphys_constants`.

jihyeonjang: The coefficient 0.04089 in the fifth term could be replaced with 0.0408**4**9, consistent with the reference equation.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
jihyeonjang: Could you confirm that `(julian-1.)` can be replaced with `(julian-0.5)`? 

In the reference, the equation for da is:
da = 2*pi/365 * (**day_of_year - 1 + (hour-12)/24**). 

Since `julian` is a REAL value that comes from `curr_julday` in `physics_timetracker` as:
`curr_julday = real(julday-1) + utc_h / 24.0`,

I think `julian-0.5` would be consistent with the equation in the reference.

jihyeonjang (changes requested):

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
weiwangncar: But this would make the code look different from the reference.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
weiwangncar: Will do so.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
weiwangncar: Will add.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
weiwangncar: Will make the change.

dudhia: Good catch. WRF also needs this fix.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
dudhia: I agree with this change.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
jihyeonjang: I agree. To keep the consistency with the reference, let's leave 229.18 as-is. Thanks!

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
dudhia: Now I wonder. I think we can leave it as the code does not use hours for da. It would be accurate for hour=12.

dudhia: OK, now I see why we need the 0.5 version from Jihyeon. The julday is a real number accounting for hours.

weiwangncar: Corrected.

weiwangncar:
@jihyeonjang Thank you very much for your careful review. @dudhia Can you both review and see if the changes are correct. We will need to do a correction in WRF as well.

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
```diff
@@ -10197,9 +10200,15 @@ subroutine rrtmg_swrad( &
...
        !--- calculate the cosine of the solar zenith angle at the current time step to determine if the sun is
        !    above or below the horizon (xt24 is the fractional part of simulation days plus half of radt in 
-       !    units of minutes, julian is in days, and radt is in minutes). do not call rrtmg_sw is night-time:
+       !    units of minutes, julian is in days, and radt is in minutes). eot (equation of time) is a correction
+       !    of mean solar time due to the difference between that time and the apparent solar time
+       !    (https://gml.noaa.gov/grad/solcalc/solareqns.PDF).
+       !    do not call rrtmg_sw at night-time:
        dorrsw = .true.
-       xt24 = mod(xtime+radt*0.5,1440.)
+       da=2.*pii*julian/365.-0.5
```
jihyeonjang: it should be changed to: 
da = 2. * pii * (julian-0.5) / 365.

jihyeonjang (changes requested):

src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F
weiwangncar: My bad... It's corrected.

jihyeonjang:
@weiwangncar Could you make the following change for `pii`? I think this should be the last modification I'll ask for. Thank you!

 `src/core_atmosphere/physics/physics_wrf/module_ra_rrtmg_sw.F`

```
@@ -10008,7 +10008,7 @@

 !------------------------------------------------------------------
 MODULE module_ra_rrtmg_sw
-use mpas_atmphys_constants,only : cp,g=>gravity
+use mpas_atmphys_constants,only : cp,g=>gravity,pii
 use module_ra_rrtmg_vinterp,only: vinterp_ozn

 use parrrsw, only : nbndsw, ngptsw, naerec
@@ -10042,7 +10042,6 @@ CONTAINS
                        its,ite, jts,jte, kts,kte                       &
                        )

- use mpas_constants, only : pii
  implicit none
```

weiwangncar:
@jihyeonjang Done.

jihyeonjang (approved):

dudhia (approved):

PR #1071, opened 2023-06-09
Updated and user defined GHG concentrations for RRTMG

thomasuhoh:
This PR updates the initial GHG concentrations in RRTMG SW+LW to more recent values. In addition, the user can specifiy the values by setting co2_content, n2o_content, and ch4_content to the desired values in the physics section of namelist.atmosphere.
New default values are 410 ppmv for CO2, 330 ppbv for N2O, and 1850 ppbv for CH4.

PR #1173, opened 2024-05-29
Make CO2 a function of year instead of a single constant

weiwangncar:
This PR makes CO2 a function of the year following what's done in WRF since v4.2, instead of using a single constant of 379, which is valid for 2005. The value for the year is approximately the same as what's observed. For example, for 2020, the observed and computed value is about 414.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_lw.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_lw.F
@@ -870,6 +870,7 @@
             rre_snow   = rresnow_p     , lwupt     = lwupt_p     , lwuptc   = lwuptc_p , &
             lwdnt      = lwdnt_p       , lwdntc    = lwdntc_p    , lwupb    = lwupb_p  , &
             lwupbc     = lwupbc_p      , lwdnb     = lwdnb_p     , lwdnbc   = lwdnbc_p , &
+            yr         = year          ,                                                 &
             ids = ids , ide = ide , jds = jds , jde = jde , kds = kds , kde = kde ,      &
             ims = ims , ime = ime , jms = jms , jme = jme , kms = kms , kme = kme ,      &
             its = its , ite = ite , jts = jts , jte = jte , kts = kts , kte = kte        &
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_lw.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_radiation_lw.F
@@ -17,6 +17,7 @@
  use mpas_atmphys_camrad_init
  use mpas_atmphys_rrtmg_lwinit
  use mpas_atmphys_vars
+ use mpas_atmphys_functions, only: co2_estimate
 
 !wrf physics:
  use module_ra_cam            
@@ -835,6 +836,7 @@
 !local variables:
  integer:: o3input
  real(kind=RKIND):: radt,xtime_m
+ real(kind=RKIND):: co2
 
 !-----------------------------------------------------------------------------------------------------------------
 !call mpas_log_write(' --- enter subroutine driver_radiation_lw: ')
@@ -851,6 +853,7 @@
     case ("rrtmg_lw")
        o3input = 0
        if(config_o3climatology) o3input = 2
+       co2 = co2_estimate(year,curr_julday)
 
        call mpas_timer_start('RRTMG_lw')
        call rrtmg_lwrad( &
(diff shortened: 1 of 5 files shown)


mgduda:
@weiwangncar I appreciate that we may be trying to mirror changes from the WRF versions of `module_ra_rrtmg_lw.F` and `module_ra_rrtmg_sw.F`, but just in case any possibilities come to mind, I think it would be better if we didn't need to define the CO2 function identically in two separate places. Is there any possibility of a shared function, `co2_concentration(year)` or similar that would define:
```
co2 = (280. + 90.*exp(0.02*(year-2000)))*1.e-6
```

weiwangncar:
@mgduda This can certainly be done. Does it matter which module the function will reside in?

mgduda:
> @mgduda This can certainly be done. Does it matter which module the function will reside in?

I suppose it would be ideal if neither SW nor LW "owned" the function, since it would be used by both and there doesn't seem to be a precedent for one of the modules (SW or LW) depending on the other. Would it make sense to have a new "radiation utils" module? That would eliminate significant build dependencies, too. If not, then maybe the function could go in the LW module (assuming CO2 is more relevant there)?

In any case, we should probably get @ldfowler58's input.

ldfowler58:
Is there a reference for this equation. I find it a little curious that CO2 changes on January 1st every year? As Michael and I talked about last week, is it just a step function? Thanks.

weiwangncar:
@ldfowler58 Jimy came up with this equation. Yes, it would be changed every year. We could do a linear interpolation like what CAM and ghg_input = 1 in WRF does so that CO2 would vary smoothly in time.
@dudhia Comment?

dudhia:
The annual increments are small (0.5%), and the steps would hardly be noticeable in my view. We have not needed interpolation in WRF.

ldfowler58:
PR 1071 does exactly what you are planning to do, but in a more complete manner. I think that we should go that route after I review the PR.

weiwangncar:
@mgduda @ldfowler58 Based on some discussion I had with Michael, the co2 equation is now added to the model via a function call. Because the function can deal with fractional years, that capability is added to the function. This will smooth out the change of co2 from one day to the next, and from one year to the next.

ldfowler58:
I still think that we should consider PR 1071 since it also includes N2O and CH4 and is user driven.

weiwangncar:
@ldfowler58 While solution proposed in this PR isn't perfect, it offers a way to update CO2 for long simulations. PR1071 isn't general either. It lets user set values for the four gases in the namelist, but it won't change during the simulations. Again it is probably not ideal for long simulations.

dudhia:
At some point the GHG table option which is now default in WRF could be
added.

ldfowler58:
In all fairness to users, I also think that we should review "related" PRs in the order that they have been submitted. I really suggest to review the GHG PR first and then to review your PR.

jihyeonjang:
I am reviewing this PR along with PR #1071. 

Using a year-dependent function for CO2 instead of a single constant seems reasonable to me, especially for long-term simulations. Aligning this change with WRF also seems beneficial. Thank you, @weiwangncar, for creating a function in the `mpas_atmphys_functions` module. 

Related to this PR, I have a couple of questions for @dudhia and @weiwangncar:

- Jimy mentioned that 

> At some point the GHG table option which is now default in WRF could be added.

Does this refer to `module_ra_clWRF_support.F`, which reads GHG concentrations from `CAMtr_volume_mixing_ratio*` files, with `ghg_input=1`? 

- Do you think the GHG table option (`ghg_input=1`, which has been the default since WRF4.4) is needed for MPAS? I think the year-dependent CO2 function in this PR is sufficient for MPAS now, because the GHG table option is more relevant for long-term simulations, which may not be the most common use case for MPAS. 

- The function in this PR and the GHG table in WRF both update values on a yearly basis and do not vary within a year. Is that correct? 

- I would also like to allow users to specify the GHG concentrations (constant), following PR 1071 . Do you have any thoughts on that? 

Thank you!

dudhia:
The table year-dependent CO2 is based on observed values for past years and official climate scenarios for future values. The function form is a close fit for past CO2 growth and extrapolates it for the future. Fixed values would not be advised unless the user is knowledgable of what value is best for the year simulated. In regional climate simulations of multiple decades, a scenario-based changing value would be best, but in WRF we have also used a fixed higher future value.

weiwangncar:
@jihyeonjang Here are my replies to your questions: 1. Yes, the GHG table option will need additional code to read and interpolate the GHG data and ghg_input controls it. 2. I'd agree with you that a time-varying CO2 is ok for the time being, though the limitation is that the other GHG won't get updated. 3. The CO2 function in this PR will only change by the year, while the GHG in the table option is interpolated in time. Though one may argue that the year to year change of the value isn't too huge. 4. What is your argument for allowing users to edit the GHG values? If there is a consensus we can certainly do it. It will at least allow users to change other GHG values. Though it is not for everyone, it can be useful for someone who uses it with a good intention.

jihyeonjang:
Thank you, @dudhia and @weiwangncar!  

It seems the GHG table option would be better for users. I will take a look at the GHG table option in WRF and see how much work it will take to add it to MPAS. Thanks!

jihyeonjang (commented):
A follow-up PR (#1457) has been opened based on this work. This PR can be closed.

PR #1174, opened 2024-05-29
Remove Fortran only flag from LDFLAGS_DEBUG for NVIDIA compiler targets

gdicker1:
The -Mchkptr flag is only valid for Fortran compilation. If included in a C build with NVHPC compilers (like the check with openacc.c does), the build fails.

Before this change, builds using the nvhpc target and OPENACC=true and DEBUG=true would fail. Removing -Mchkptr allows successful debug openacc builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,7 @@ nvhpc:   # BUILDTARGET NVIDIA HPC SDK
 	"FFLAGS_DEBUG = -O0 -g -Mbounds -Mchkptr -byteswapio -Mfree -Ktrap=divz,fp,inv,ovf -traceback" \
 	"CFLAGS_DEBUG = -O0 -g -traceback" \
 	"CXXFLAGS_DEBUG = -O0 -g -traceback" \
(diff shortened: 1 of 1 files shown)


gdicker1:
The build of openacc.c occurs as part of the `openacc_test` target in the top-level Makefile. Without this change, builds fail around [line 1123 of the top-level Makefile](https://github.com/MPAS-Dev/MPAS-Model/blob/369341a8ee85c9204471b83c3392eef00aa7f77b/Makefile#L1123) (before any actual MPAS files are touched).

User's would then have to compile the openacc.c program (as directed by the build) to see the error message related to `-Mchkptr`.

mgduda:
@gdicker1 Just a note for similar PRs in future, since the changes are in the top-level `Makefile` and not specific to MPAS-Atmosphere a branch name with `framework/` rather than `atmosphere/` may be preferable. But no worries for this PR!

gdicker1:
> @gdicker1 Just a note for similar PRs in future, since the changes are in the top-level `Makefile` and not specific to MPAS-Atmosphere a branch name with `framework/` rather than `atmosphere/` may be preferable. But no worries for this PR!

Thanks for the reminder!

mgduda (approved):

PR #1177, opened 2024-05-30
Implementation of the aerosol-aware option in the Thompson MP

ldfowler58:
This PR updates the Thompson cloud microphysics parameterization to include the aerosol-aware option based on the WRF model release 4.1.4.

* In ./src/core_init_atmosphere, we added the initialization of the "water-friendly" and "ice-friendly" aerosols as a function of a monthly-mean based climatology from the GOddard Chemistry Aerosol and Transport Model (GOCART). The variables nwfa and nifa are included in the scalars array.

* In ./src/core_atmosphere/physics, the aerosol-aware option is triggered when config_microp_scheme is set to "mp_thompson_aerosols", and used to activate CCN and IN to form cloud liquid water droplets and ice crystals. In addition to nwfa and nifa, the number concentration of cloud liquid water droplets (nc) is an additional prognostic variable, also in the scalars array.

* In ./src/core_atmosphere/physics, the MYNN PBL scheme is used for mixing of aerosols in the PBL and free-troposphere. The radiative effect of aerosols is included in the RRTMG shortwave radiation code.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -3073,11 +3073,24 @@
 
 
                 <!-- ================================================================================================== -->
-                <!-- ... PARAMERIZATION OF CLOUDINESS:                                                                  -->
+                <!-- ... PARAMETERIZATION OF CLOUDINESS:                                                                -->
                 <!-- ================================================================================================== -->
 
                 <var name="cldfrac" type="real" dimensions="nVertLevels nCells Time" units="unitless"
                      description="horizontal cloud fraction"/>
+
+
+                <!-- ================================================================================================== -->
+                <!-- ... PARAMERIZATION OF AEROSOL OPTICAL PROPERTIES:                                                  -->
+                <!-- ================================================================================================== -->
+
+                <var name="taod5502d" type="real" dimensions="nCells Time" units="unitless"
+                     description="total aerosol optical depth at 550 nm from the Thompson cloud microphysics"
+                     packages="mp_thompson_aers_in"/>
+
+                <var name="taod5503d" type="real" dimensions="nVertLevels nCells Time" units="unitless"
+                     description="aerosol optical depth at 550 nm from the Thompson cloud microphysics"
+                     packages="mp_thompson_aers_in"/>
 
 
                 <!-- ================================================================================================== -->
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -397,6 +397,7 @@
         <packages>
                 <package name="mp_kessler_in" description="parameterization of kessler microphysics."/>
                 <package name="mp_thompson_in" description="parameterization of Thompson cloud microphysics."/>
+                <package name="mp_thompson_aers_in" description="parameterization of scale-aware Thompson cloud microphysics."/>
                 <package name="mp_wsm6_in" description="parameterization of WSM6 cloud microphysics."/>
                 <package name="cu_grell_freitas_in" description="parameterization of Grell-Freitas convection."/>
                 <package name="cu_kain_fritsch_in" description="parameterization of Kain-Fritsch convection."/>
@@ -608,13 +609,15 @@
 			<var name="surface_pressure"/>
 
 #ifdef DO_PHYSICS
-			<var name="rainprod" packages="mp_thompson_in"/>
-			<var name="evapprod" packages="mp_thompson_in"/>
-			<var name="re_cloud" packages="mp_thompson_in;mp_wsm6_in"/>
-			<var name="re_ice" packages="mp_thompson_in;mp_wsm6_in"/>
-			<var name="re_snow" packages="mp_thompson_in;mp_wsm6_in"/>
-			<var name="nt_c" packages="mp_thompson_in"/>
-			<var name="mu_c" packages="mp_thompson_in"/>
+			<var name="rainprod"/>
+			<var name="evapprod"/>
+			<var name="re_cloud"/>
+			<var name="re_ice"/>
+			<var name="re_snow"/>
+			<var name="nwfa2d"/>
+			<var name="nifa2d"/>
+			<var name="nt_c"/>
+			<var name="mu_c"/>
                         <var name="tend_sfc_pressure"/>
 			<var name="rt_diabatic_tend"/>
 			<var name="nsteps_accum"/>
@@ -627,7 +630,9 @@
 			<var name="i_rainnc"/>
 			<var name="rainncv"/>
 			<var name="rainnc"/>
+			<var name="snowncv"/>
 			<var name="snownc"/>
+			<var name="graupelncv"/>
 			<var name="graupelnc"/>
 			<var name="sr"/>
 			<var name="i_rainc"/>
@@ -775,6 +780,18 @@
 			<var name="xmb_shallow"/>
 			<var name="qc_cu"/>
 			<var name="qi_cu"/>
+			<var name="rthmpten"/>
+			<var name="rqvmpten"/>
+			<var name="rqcmpten"/>
+			<var name="rqrmpten"/>
+			<var name="rqimpten"/>
+			<var name="rqsmpten"/>
+			<var name="rqgmpten"/>
+			<var name="rncmpten"/>
+			<var name="rnimpten"/>
+			<var name="rnrmpten"/>
+			<var name="rnifampten"/>
+			<var name="rnwfampten"/>
 			<var name="rthcuten"/>
 			<var name="rqvcuten"/>
 			<var name="rqccuten"/>
@@ -1521,31 +1538,43 @@
 
                         <var name="qc" array_group="moist" units="kg kg^{-1}"
                              description="Cloud water mixing ratio"
-                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_kessler_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_kessler_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="qr" array_group="moist" units="kg kg^{-1}"
                              description="Rain water mixing ratio"
-                             packages="mp_kessler_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="mp_kessler_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="qi" array_group="moist" units="kg kg^{-1}"
                              description="Ice mixing ratio"
-                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="qs" array_group="moist" units="kg kg^{-1}"
                              description="Snow mixing ratio"
-                             packages="bl_mynn_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="bl_mynn_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="qg" array_group="moist" units="kg kg^{-1}"
                              description="Graupel mixing ratio"
-                             packages="mp_thompson_in;mp_wsm6_in"/>
+                             packages="mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="ni" array_group="number" units="nb kg^{-1}"
                              description="Cloud ice number concentration"
-                             packages="bl_mynn_in;mp_thompson_in"/>
+                             packages="bl_mynn_in;mp_thompson_in;mp_thompson_aers_in"/>
 
                         <var name="nr" array_group="number" units="nb kg^{-1}"
                              description="Rain number concentration"
-                             packages="mp_thompson_in"/>
+                             packages="mp_thompson_in;mp_thompson_aers_in"/>
+
+                        <var name="nc" array_group="number" units="nb kg^{-1}"
+                             description="Cloud water number concentration"
+                             packages="mp_thompson_aers_in"/>
+
+                        <var name="nifa" array_group="number" units="nb kg^{-1}"
+                             description="Ice-friendly aerosol number concentration"
+                             packages="mp_thompson_aers_in"/>
+
+                        <var name="nwfa" array_group="number" units="nb kg^{-1}"
+                             description="Water-friendly aerosol number concentration"
+                             packages="mp_thompson_aers_in"/>
                 </var_array>
 #endif
 
@@ -1851,31 +1880,43 @@
 
                         <var name="tend_qc" name_in_code="qc" array_group="moist" units="kg m^{-3} s^{-1}"
                              description="Tendency of cloud water mass per unit volume divided by d(zeta)/dz"
-                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_kessler_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_kessler_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="tend_qr" name_in_code="qr" array_group="moist" units="kg m^{-3} s^{-1}"
                              description="Tendency of rain water mass per unit volume divided by d(zeta)/dz"
-                             packages="mp_kessler_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="mp_kessler_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="tend_qi" name_in_code="qi" array_group="moist" units="kg m^{-3} s^{-1}"
                              description="Tendency of ice mass per unit volume divided by d(zeta)/dz"
-                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="bl_mynn_in;bl_ysu_in;cu_ntiedtke_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="tend_qs" name_in_code="qs" array_group="moist" units="kg m^{-3} s^{-1}"
                              description="Tendency of snow mass per unit volume divided by d(zeta)/dz"
-                             packages="bl_mynn_in;mp_thompson_in;mp_wsm6_in"/>
+                             packages="bl_mynn_in;mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="tend_qg" name_in_code="qg" array_group="moist" units="kg m^{-3} s^{-1}"
                              description="Tendency of graupel mass per unit volume divided by d(zeta)/dz"
-                             packages="mp_thompson_in;mp_wsm6_in"/>
+                             packages="mp_thompson_in;mp_thompson_aers_in;mp_wsm6_in"/>
 
                         <var name="tend_ni" name_in_code="ni" array_group="number" units="nb m^{-3} s^{-1}"
                              description="Tendency of cloud ice number concentration multiplied by dry air density divided by d(zeta)/dz"
-                             packages="bl_mynn_in;mp_thompson_in"/>
+                             packages="bl_mynn_in;mp_thompson_in;mp_thompson_aers_in"/>
 
                         <var name="tend_nr" name_in_code="nr" array_group="number" units="nb m^{-3} s^{-1}" 
                              description="Tendency of rain number concentration multiplied by dry air density divided by d(zeta)/dz"
-                             packages="mp_thompson_in"/>
+                             packages="mp_thompson_in;mp_thompson_aers_in"/>
+
+                        <var name="tend_nc" name_in_code="nc" array_group="number" units="nb m^{-3} s^{-1}"
+                             description="Tendency of cloud water number concentration multiplied by dry air density divided by d(zeta)/dz"
+                             packages="mp_thompson_aers_in"/>
+
+                        <var name="tend_nifa" name_in_code="nifa" array_group="number" units="nb m^{-3} s^{-1}"
+                             description="Tendency of ice-friendly aerosol number concentration multiplied by dry air density divided by d(zeta)/dz"
+                             packages="mp_thompson_aers_in"/>
+
+                        <var name="tend_nwfa" name_in_code="nwfa" array_group="number" units="nb m^{-3} s^{-1}"
+                             description="Tendency of water-friendly aerosol number concentration multiplied by dry air density divided by d(zeta)/dz"
+                             packages="mp_thompson_aers_in"/>
                 </var_array>
 #endif
         </var_struct>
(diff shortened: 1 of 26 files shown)


weiwangncar:
@mgduda should the base of the PR be changed?

mgduda (changes requested):
With the GNU 12.2.0 compilers, I'm seeing compilation failures for `module_ra_rrtmg_sw_aerosols.F`:
```
module_ra_rrtmg_sw_aerosols.F:87:20:
   87 |     real, dimension(ims:ime, kms:kme, jms:jme), intent(in) :: p,     & ! pressure (Pa)
      |                    1
Error: GNU Extension: Symbol 'ims' is used before it is typed at (1)
module_ra_rrtmg_sw_aerosols.F:70:40:

   70 | subroutine calc_aerosol_rrtmg_sw(ht,dz8w,p,t3d,qv3d,aer_type,                               &
      |                                        1
Error: Symbol 'dz8w' at (1) has no IMPLICIT type
module_ra_rrtmg_sw_aerosols.F:70:42:

   70 | subroutine calc_aerosol_rrtmg_sw(ht,dz8w,p,t3d,qv3d,aer_type,                               &
      |                                          1
Error: Symbol 'p' at (1) has no IMPLICIT type
module_ra_rrtmg_sw_aerosols.F:70:46:

   70 | subroutine calc_aerosol_rrtmg_sw(ht,dz8w,p,t3d,qv3d,aer_type,                               &
      |                                              1
Error: Symbol 't3d' at (1) has no IMPLICIT type
module_ra_rrtmg_sw_aerosols.F:70:51:

   70 | subroutine calc_aerosol_rrtmg_sw(ht,dz8w,p,t3d,qv3d,aer_type,                               &
      |                                                   1
Error: Symbol 'qv3d' at (1) has no IMPLICIT type
```
I think these can be resolved by ensuring that dummy arguments are declared such that none of them are referenced before they're declared.

mgduda:
@ldfowler58 Regarding the compilation failure, I think all that's needed is to move this block:
```
    integer, intent(in) :: ims,ime,jms,jme,kms,kme, &
                           its,ite,jts,jte,kts,kte
```
above this block of declarations in module_ra_rrtmg_sw_aerosols.F:
```
    real, dimension(ims:ime, kms:kme, jms:jme), intent(in) :: p,     & ! pressure (Pa)
                                                              t3d,   & ! temperature (K)
                                                              dz8w,  & ! dz between full levels (m)
                                                              qv3d     ! water vapor mixing ratio (kg/kg)
```

src/core_init_atmosphere/mpas_init_atm_thompson_aerosols.F
```diff
@@ -0,0 +1,753 @@
...
+ type(mpas_pool_type),intent(inout)   :: state
+!local variables and pointers:
+ character (len=StrKIND),pointer:: config_start_time
+ character(len=StrKIND):: filename_gocart
+ character(len=StrKIND):: initial_date,mess
+
+ logical:: lexist
+
+!-----------------------------------------------------------------------------------------------------------------
+ call mpas_log_write('--- enter subroutine init_atm_gocart:')
```
mgduda, line 56: I'd propose removing this print statement as well as the `--- end subroutine init_atm_gocart.` print statement at the end of this subroutine.

ldfowler58: As suggested, commented out extra print statements.

ldfowler58: As suggested, moved the declaration of dimensions above that of variables.

ldfowler58:
commented out the extra print statements related to how the AOPs are computed.

src/core_atmosphere/physics/mpas_atmphys_init.F
```diff
@@ -294,47 +290,47 @@ subroutine physics_init(dminfo,clock,configs,mesh,diag,tend,state,time_lev,diag_
...
+          do iCell = 1, nCellsSolve
+             h_oml(iCell) = h_oml_initial(iCell)
+             ! WRF COMMENT:
+             ! fill in near coast area with SST: 200 K was set as missing value in ocean pre-processing code
+             if( (t_oml_200m_initial(iCell) > 200.) .and. (t_oml_200m_initial(iCell) <= 200.) )  &
+                  t_oml_200m_initial(iCell) = sst(iCell)
+          enddo
+       endif
+    endif
+ endif
```
weiwangncar, line 333: Is it ok to include this in this subject commit?

weiwangncar:
It seems with this code, even if the data file, QNWFA_QNIFA_Monthly_Sigma.dat, isn't present in the working directory, the relevant fields are in the IC and likely IBC files with values set to zero. It seems it would be better when the data is not used, they don't appear in IC/LBC files.
Also is it necessary to change the input file name from QNWFA_QNIFA_MONTHLY_SIGMA.dat to QNWFA_QNIFA_Monthly_Sigma.dat? That tricked me at least.

src/core_atmosphere/physics/mpas_atmphys_init.F
ldfowler58: Can you please clarify the question. I do not understand what you mean. Thanks.

weiwangncar: @ldfowler58 I understand that part of this code modification is clean up. But since this is a PR for adding aerosol Thompson scheme, the code change for ocean initialization seems out of place.

ldfowler58:
I do agree with your comments with these extra fields being zero out on the init side (Note that they are not read from the input file if mp_thompson_aerosols = false on the core_atmosphere side). To not have these fields in the IC and IBC files would require to add a namelist parameter related to the aerosol-aware option and define a "package" attached to it. I am not sure that I want to start adding specify physics options on the core_init_atmosphere.   

When I started to work on the aerosol-aware version of the Thompson microphysics, I think that "SIGMA" was actually "Sigma", including in the original f90 sourcecode that read the GOCART netCDF file that Greg Thompson used to output the binary file. I can certainly changed that the filename, once we decide where that file will reside for download.

weiwangncar:
@ldfowler58 I would generally advocate for simpler coding. But the file sizes with the aerosols are 55% larger, which is pretty significant. Also is it necessary to write nifa/nwfa/nwif_gocart_clim fields into initial condition file?

weiwangncar:
@ldfowler58 Perhaps one can use detection of the file, QNWFA_QNIFA_Monthly_Sigma.dat, as a cue for not processing these fields?

src/core_atmosphere/Registry.xml
```diff
@@ -397,6 +397,7 @@
         <packages>
                 <package name="mp_kessler_in" description="parameterization of kessler microphysics."/>
                 <package name="mp_thompson_in" description="parameterization of Thompson cloud microphysics."/>
+                <package name="mp_thompson_aers_in" description="parameterization of scale-aware Thompson cloud microphysics."/>
```
weiwangncar, line 400: Should this be called "parameterization of **aerosol**-aware Thompson cloud microphysics"?

mgduda:
> @ldfowler58 Perhaps one can use detection of the file, QNWFA_QNIFA_Monthly_Sigma.dat, as a cue for not processing these fields?

I think that would definitely be possible: to set up a package that is activated by the existence of the `QNWFA_QNIFA_Monthly_Sigma.dat` file in the run directory. I'd propose we take this as a "cleanup" item once we've created the `release-v8.2.0` branch (hopefully in the next couple of days?). It would be a small enough change, and easily testable.

src/core_init_atmosphere/mpas_init_atm_thompson_aerosols.F
```diff
@@ -0,0 +1,753 @@
...
+             if(x < 0.5) then
+                lon = lon + 360.0
+                call latlon_to_ij(proj,lat,lon,x,y)
+             elseif(x > real(field%nx,kind=RKIND)+ 0.5) then
+                lon = lon - 360.0
+                call latlon_to_ij(proj,lat,lon,x,y)
+             endif
+
+             if(maskval /= -1.0) then
+                destField3d(nmonths,k,iCell) = interp_sequence(x,y,1,rslab,-2,field%nx+3,1,field%ny,1,1,msgval, \
```
mgduda, line 697: Compiling with the 'nvhpc' build target gives these errors:
```
NVFORTRAN-S-0023-Syntax error - unbalanced parentheses (mpas_init_atm_thompson_aerosols.F: 697)
NVFORTRAN-S-0023-Syntax error - unbalanced parentheses (mpas_init_atm_thompson_aerosols.F: 698)
NVFORTRAN-S-0023-Syntax error - unbalanced parentheses (mpas_init_atm_thompson_aerosols.F: 700)
NVFORTRAN-S-0023-Syntax error - unbalanced parentheses (mpas_init_atm_thompson_aerosols.F: 701)
```
I think converting the `\` to a `&` on these lines would resolve the issue.

ldfowler58:
typos have been corrected.

ldfowler58:
This would be similar to what I do to detect if we do or not do the interpolation of nwfa and nifa which is also based on the presence of that file or not. I will start working on it after we finish the other items, as you suggested.

ldfowler58:
of course. corrected the description of package mp_thompson_aers_in. thanks.

ldfowler58:
These latest commits aim to resolve the issue related to nwfa and nifa being set to zero when the input GOCART climatological file is not present in the run directory.

weiwangncar:
> > @ldfowler58 Perhaps one can use detection of the file, QNWFA_QNIFA_Monthly_Sigma.dat, as a cue for not processing these fields?
> 
> I think that would definitely be possible: to set up a package that is activated by the existence of the `QNWFA_QNIFA_Monthly_Sigma.dat` file in the run directory. I'd propose we take this as a "cleanup" item once we've created the `release-v8.2.0` branch (hopefully in the next couple of days?). It would be a small enough change, and easily testable.

After removing the climatological fields of NIFA/NWFA from the IC file, the size is now only 5% larger. One can argue that it would be clean if one is not choosing the aerosal option, the data isn't in the input. But it's up to you. I also see lbc_qc and lbc_qr in the LBC file with zero values.

ldfowler58:
**"One can argue that it would be clean if one is not choosing the aerosal option, the data isn't in the input. But it's up to you."**  I do not understand what you mean. 

**"I also see lbc_qc and lbc_qr in the LBC file with zero values."** I do not understand what you mean either. Using the develop branch at hash 369341a8ee85c (i.e. without any aerosol-aware sourcecode) with the GFS files, that you pointed me to this morning, the lbc files contain the field lbc_qc and lbc_qr and they are both zero. That would not change with the aerosol-aware version?

weiwangncar:
@ldfowler58 The following comment "I also see lbc_qc and lbc_qr in the LBC file with zero values." simply points out that these two fields are in the LBC file even there are not input to them from GFS, as in this case. I understand this has nothing to do with your change.

ldfowler58:
OK, then. It looks like I replied to all the comments that have been posted regarding the initialization of nwfa and nifa in core_init_atmosphere.

ldfowler58:
I revised the initialization of the lateral boundary conditions for nc, nifa, and nwfa. The LBCs are computed using the monthly-mean fields nifa_gocart_clim, nwfa_goart_clim, and pwif_gocart_clim, and pressure interpolated from the first-guess fields for each timestamp.

In mpas_atm_time_integration.F added calls to subroutine scalars_driving for nc, nifa, and nwfa when using the mp_thompson_aerosols cloud microphysics option.

weiwangncar:
@ldfowler58 Thanks! The Thompson+aerosol with convection_permitting suite works now in my regional case. Nice! But it still fails right away with mesoscale_reference physics. It fails with segmentation fault.

ldfowler58:
I know. One thing and one commit at a time!

ldfowler58:
I modified the sourcecode so that the option mp_thompson_aerosols can be used with the mesoscale_reference suite. I tested the code changes to ensure that short forecasts with the convection_permitting suite remained the same as the original forecasts. I also tested the sourcecode for bit for bit restartibility.

I would appreciate it if Wei could re-run her regional forecasts with the mesoscale_reference suite.

weiwangncar:
@ldfowler58 I've rerun my regional test case with both mesoscale_reference and convection_permitting suites using the latest code in this branch and both are working well now with a 3-h forecast tests. Great!

ldfowler58:
Great. That's what I got as well. Thanks.

weiwangncar:
As a general comment, since the IC/LBC file sizes are now back to the larger ones (55% larger in IC file), I would again suggest not to include the data if the option is off or the input file isn't in the directory. If you agree with the idea, it can be changed later.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -471,6 +471,11 @@ subroutine atm_srk3(domain, dt, itimestep, exchange_halo_group)
          call mpas_pool_get_dimension(state, 'index_qg', index_qg)  
          call mpas_pool_get_dimension(state, 'index_nr', index_nr)
          call mpas_pool_get_dimension(state, 'index_ni', index_ni)
+         if(config_microp_scheme == 'mp_thompson_aerosols') then
```
mgduda, line 474: I don't think we need these conditional checks on `config_microp_scheme` here and elsewhere in `mpas_atm_time_integration.F`. If the aerosol-aware Thompson scheme isn't being used, these indices will be returned as -1 (the same as, e.g., `index_nr` and `index_ni`).

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -927,7 +932,17 @@ subroutine atm_srk3(domain, dt, itimestep, exchange_halo_group)
...
-                  
+                  if(config_microp_scheme == 'mp_thompson_aerosols') then
+                     if (index_nc > 0) then
+                        scalars_driving(index_nc,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', rk_timestep(rk_step) )
+                     end if
+                     if (index_nifa > 0) then
+                        scalars_driving(index_nifa,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', rk_timestep(rk_step) )
+                     end if
+                     if (index_nwfa > 0) then
+                        scalars_driving(index_nwfa,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', rk_timestep(rk_step) )
```
mgduda, line 943: I think the second-to-last arguments aren't correct, and they should be `nc`, `nifa`, and `nwfa`.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1094,7 +1109,17 @@ subroutine atm_srk3(domain, dt, itimestep, exchange_halo_group)
...
-               
+               if (config_microp_scheme == 'mp_thompson_aerosols') then
+                  if(index_nc > 0) then
+                     scalars_driving(index_nc,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', rk_timestep(rk_step) )
+                  end if
+                  if(index_nifa > 0) then
+                     scalars_driving(index_nifa,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', rk_timestep(rk_step) )
+                  end if
+                  if(index_nwfa > 0) then
+                     scalars_driving(index_nwfa,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', rk_timestep(rk_step) )
```
mgduda, line 1120: Similarly here, the penultimate argument to `scalars_driving` looks incorrect.

mgduda (changes requested):

src/core_init_atmosphere/Registry.xml
```diff
@@ -583,6 +594,9 @@
                         <var name="q2" packages="met_stage_out"/>
                         <var name="rh2" packages="met_stage_out"/>
                         <var name="t2m" packages="met_stage_out"/>
+                        <var name="nifa_gocart_clim" packages="met_stage_out;mp_thompson_aers_in"/>
+                        <var name="nwfa_gocart_clim" packages="met_stage_out;mp_thompson_aers_in"/>
+                        <var name="pwif_gocart_clim" packages="met_stage_out;mp_thompson_aers_in"/>
```
mgduda, line 599: I think we might have discussed this, but could you remind me why we write `nifa_gocart_clim`, `nwfa_gocart_clim`, and `pwif_gocart_clim` to the initial conditions file?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
ldfowler58: you are right. I removed lines 474 and 478.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
ldfowler58: you are right. corrected.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
ldfowler58: you are right. corrected.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1254,7 +1279,17 @@ subroutine atm_srk3(domain, dt, itimestep, exchange_halo_group)
...
-      
+         if (config_microp_scheme == 'mp_thompson_aerosols') then
+            if (index_nc > 0) then
+               scalars_driving(index_nc,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', dt )
+            end if
+            if (index_nifa > 0) then
+               scalars_driving(index_nifa,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', dt )
+            end if
+            if (index_nwfa > 0) then
+               scalars_driving(index_nwfa,1:nVertLevels,1:nCells+1) = mpas_atm_get_bdy_state( clock, block, nVertLevels, nCells, 'ni', dt )
```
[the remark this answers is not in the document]
ldfowler58: you are right. corrected.

ldfowler58:
I suppose that commit cfb4df8902b8 can be squashed (reordered?) with commit a10e416d5507 to simplify the log file. Not sure if it is needed or how to proceed. Thanks.

src/core_atmosphere/physics/mpas_atmphys_control.F
```diff
@@ -166,9 +168,10 @@ subroutine physics_namelist_check(configs)
...
- if(.not. (config_microp_scheme .eq. 'off'         .or. &
-           config_microp_scheme .eq. 'mp_kessler'  .or. &
-           config_microp_scheme .eq. 'mp_thompson' .or. &
+ if(.not. (config_microp_scheme .eq. 'off'                  .or. &
+           config_microp_scheme .eq. 'mp_kessler'           .or. &
+           config_microp_scheme .eq. 'mp_thompson'          .or. &
+           config_microp_scheme .eq. 'mp_thompson_aerosols' .or. &
            config_microp_scheme .eq. 'mp_wsm6')) then
 
     write(mpas_err_message,'(A,A10)') 'illegal value for config_microp_scheme:', &
```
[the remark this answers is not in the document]
ldfowler58: you are right. corrected. note that I should probably change the A10 format to A20 format for convection, radiation, others as well. I will submit a PR in a version 8.2 bug-fix release?

src/core_init_atmosphere/mpas_init_atm_thompson_aerosols.F
```diff
@@ -0,0 +1,889 @@
...
+ call mpas_pool_get_array(mesh,'latCell' ,latCell )
+ call mpas_pool_get_array(mesh,'lonCell' ,lonCell )
+
+ call mpas_pool_get_array(mesh,'nifa_gocart_clim',nifa_clim)
+ call mpas_pool_get_array(mesh,'nwfa_gocart_clim',nwfa_clim)
+ call mpas_pool_get_array(mesh,'pwif_gocart_clim',pwif_clim)
+
+!open intermediate file:
+ istatus = 0
+ call read_gocart_init(trim(filename_gocart),istatus)
```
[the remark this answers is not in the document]
ldfowler58: I do not recall  what I was exactly thinking when I added subroutine read_gocart_init. Maybe 1) the filename does not have a time stamp and therefore "simpler" to open? and 2) wanting to keep everything in one module to keep the thompson/gocart sourcecode compact.

src/core_init_atmosphere/Registry.xml
ldfowler58: Yes, we did discuss this earlier last week. Keeping those three fields in the init file was an easy way to use them to build the lateral boundary conditions for nifa and nwfa. We discussed that it may not be the ideal way for DA, but that we could review that issue after the 8.2 release. One option would be to add them to the invariant stream instead.

With these proposed changes, I was able to create the LBC and run regional MPAS with and without the gocart aerosols.

mgduda: Ok, thanks for the (re-)explanation!

src/core_init_atmosphere/mpas_init_atm_thompson_aerosols.F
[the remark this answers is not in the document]
mgduda: I think it would be better to use the `read_met_init` routine. You can see in that routine that if the `source_is_constant` arugment is `.true.`, we use the filename as specified rather than adding the date suffix:
https://github.com/MPAS-Dev/MPAS-Model/blob/v8.1.0/src/core_init_atmosphere/mpas_init_atm_read_met.F#L72-L76

src/core_atmosphere/physics/mpas_atmphys_control.F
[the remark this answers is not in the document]
mgduda: I think it's fine to adjust the format as part of this PR.

src/core_init_atmosphere/mpas_init_atm_thompson_aerosols.F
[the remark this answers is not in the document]
ldfowler58: replaced calls to subroutine read_gocart_init with calls to subroutine read_met_init, as suggested.

mgduda:
@ldfowler58 I can rebase the branch to squash cfb4df89 into a10e416d.

Also, I would propose removing some of these "enter subroutine" and "end subroutine" log messages:
```
 --- enter subroutine init_microphysics:
 --- enter subroutine init_thompson_aerosols_forMPAS:
 --- end read table_ccnAct:
 --- global_nwfa_min = 0.104950E+09
 --- global_nwfa_max = 0.258438E+10
 --- initialize nwfa using the climatological GOCART data.
 --- global_nifa_min = 2.62297
 --- global_nifa_max = 0.129468E+07
 --- initialize nifa using the climatological GOCART data.
 --- end subroutine init_thompson_aerosols_forMPAS.
 --- end subroutine init_microphysics:
```
(i.e., the first two lines and the last two lines from this block).

If you're alright with the rebase and with removing the enter/exit messages, just let me know and I can force-push an update to the PR branch.

ldfowler58:
corrected the build of the lateral boundary conditions when mp_thompson_aerosols is set to true.

ldfowler58:
commented extra print statements, as suggested. Although I acted on your suggestions, it seems to me that it is easier to find the print statement "end read table_ccnAct" in the sourcecode when we know where that print statement comes from. Since it is only written once to the logfile at the init timestep, I do not think that it burdens the logfile that much. It actually helps me (and may help users) to follow what actually happens when mp_thompson_aerosols is true.

mgduda:
> commented extra print statements, as suggested. Although I acted on your suggestions, it seems to me that it is easier to find the print statement "end read table_ccnAct" in the sourcecode when we know where that print statement comes from. Since it is only written once to the logfile at the init timestep, I do not think that it burdens the logfile that much. It actually helps me (and may help users) to follow what actually happens when mp_thompson_aerosols is true.

Although I understand the more general point, in the case of "end read table_ccnAct" specifically, a simple `grep -r "end read table_ccnAct" *` in the `src/core_atmosphere` directory returns the file `physics/mpas_atmphys_init_microphysics.F` in just a few seconds, and it's easy to search for "end read table_ccnAct" within that file.

We are considering the addition of a "log/debug level" for log messages that should make it easier to keep debug and/or development printouts in the code and to selectively activate them at runtime. Hopefully sometime by the end of summer this capability may exist.

Anyway, I won't object to leaving the `--- end read table_ccnAct:` message as it is.

mgduda (approved):
Thanks for reworking the LBC code so that `nifa_gocart_clim`, `nwfa_gocart_clim`, and `pwif_gocart_clim` are no longer needed in the ICs file!

ldfowler58:
no need to reintroduce the log statement. Regarding introducing "log/debug" level print statements, it would be very useful if it could be designed in an general way for sue in the physics_mmm directory (may using #include ?)

ldfowler58:
About the LBC code, I was not have the time to change how the mp_thompson_aerosols is activated along with the other packages. I can review that for a future release. But this does not affect anything anyway.

PR #1180, opened 2024-06-13
Make the registry 'parse' tool core-independent

mgduda:
This PR introduces changes to allow the registry `parse` tool to be core-independent, with any core-dependent
output that it produces depending only on run-time arguments.

The `parse` tool now requires its first command-line argument to be the name of a `Registry.xml` file, and it requires
subsequent arguments that specify definitions of at least the following macros in the form `-D<macro>=<value>`:

 - `MPAS_EXE_NAME`
 - `MPAS_GIT_VERSION`
 - `MPAS_BUILD_TARGET`
 - `MPAS_NAMELIST_SUFFIX`

Additionally, this PR updates the main `Makefile` for each core so that the processed `Registry.xml` file is passed as the first command-line argument to `parse` and the contents of `$(CPPFLAGS)`, which contains definitions of at least the aforementioned macros, as subsequent command-line arguments.

### DIFF
--- a/src/core_atmosphere/Makefile
+++ b/src/core_atmosphere/Makefile
@@ -30,7 +30,7 @@ core_input_gen:
 
 gen_includes: core_reg
 	(if [ ! -d inc ]; then mkdir -p inc; fi) # To generate *.inc files
-	(cd inc; $(REG_PARSE) < ../Registry_processed.xml )
+	(cd inc; $(REG_PARSE) ../Registry_processed.xml )
 
 post_build:
 	if [ ! -e $(ROOT_DIR)/default_inputs ]; then mkdir $(ROOT_DIR)/default_inputs; fi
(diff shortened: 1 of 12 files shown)


mgduda:
@gdicker1 With the changes in this PR, I think PR #1179 will actually work as intended.

mgduda:
@gdicker1 I've just pushed three new commits to fix a couple of minor issues. Before this PR is ultimately merged, I'll squash those commits.

gdicker1 (commented):
Thanks for doing the fixups as separate commits; you were right that it made reviewing these changes easier.

gdicker1 (approved):
These changes do what is expected and it is nice that the callback+variadic implementation of `parse_macros` can be extended in the future.

I also took the liberty of trying these with #1179 - combining these changes and the changes in #1179 allow for core-independent builds.

mgduda:
@gdicker1 Thanks! I just force-pushed the rebased branch, and I confirmed that no changes were inadvertently made during the rebase with `git diff a2df6eed b8f077cc`.

PR #1181, opened 2024-06-13
Change URL for MMM-physics external to use https rather than git

mgduda:
This PR changes the URL for the `MMM-physics` external to use `https` rather than `git` so that the checkout
of the `MMM-physics` repository will work even for those who don't have SSH keys set up on GitHub.

### DIFF
--- a/src/core_atmosphere/Externals.cfg
+++ b/src/core_atmosphere/Externals.cfg
@@ -1,7 +1,7 @@
 [MMM-physics]
 local_path = ./physics_mmm
 protocol = git
-repo_url = git@github.com:NCAR/MMM-physics.git
+repo_url = https://github.com/NCAR/MMM-physics.git
 tag = 20240525-MPASv8.2
 
 required = True


gdicker1:
@mgduda I wonder if it would also be good to add more to the src/core_atmosphere/physics/.gitignore here too. Right now `git status` (in a MPAS-Model dir) shows changes in the ".../physics_mmm" directory when really the MPAS-Model repo shouldn't care about those files.

Something like the below would match what was done in CESM. This way the dir that contains the external is ignored. (This could be in addition to the specific files already mentioned.)
```
physics_mmm/
```

mgduda:
@gdicker1 Agreed that we should update the `.gitignore` file; but I can make a separate PR for that to keep this one focused on fixing the URL for the `MMM-physics` repository.

ldfowler58:
Whatever the two of you would like to proceed is totally OK with me, of course. You definitely are the software engineers with ruling privileges on those matters, and I simply appreciate your help and suggestions. Thanks,

gdicker1 (approved):
Looks good, thanks!

PR #1182, opened 2024-06-14
Parallelize the remapping of Noah-MP static fields in init_atmosphere core

mgduda:
This PR parallelizes the remapping of Noah-MP static fields in the init_atmosphere core.

Following the strategy employed for the parallel remapping of other static fields in the init_atmosphere core, this PR reworks the code for remapping `soilcomp`, `soilcl1`, `soilcl2`, `soilcl3`, and `soilcl4` so that no special graph partition files are needed in order to get bit-wise identical values for these fields when using any number of MPI tasks.

The `soilcomp` field is handled as a multi-layer, continuous field, while `soilcl{1,2,3,4}` are handled as categorical fields (although they are declared in the Registry as `real` fields).

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_static.F
+++ b/src/core_init_atmosphere/mpas_init_atm_static.F
@@ -89,11 +89,14 @@
  ! use module level variables for now...
  !
  integer (kind=I8KIND), dimension(:), pointer :: ter_integer
+ integer (kind=I8KIND), dimension(:,:), pointer :: soilcomp_int
+ real (kind=RKIND) :: soilcomp_msgval = 255.0_RKIND   ! Modified later based on index file for soilcomp
  integer, dimension(:), pointer :: lu_index
  integer, dimension(:), pointer :: soilcat_top
  integer, dimension(:), pointer :: nhs
  integer, dimension(:,:), allocatable:: ncat
- ! Landmask is used by the accumulation function for maxsnoalb so it needs to be a global variable
+ ! Landmask is used by the accumulation function for maxsnoalb and soilcomp,
+ ! so it needs to be a global variable
  integer, dimension(:), pointer :: landmask
 
  integer, pointer :: category_min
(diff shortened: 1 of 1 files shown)


ldfowler58:
Hi Michael:
Thanks for working on the parallel remapping of those 5 fields. Looking at the updated sourcecode, it looks like a lot of the original sourcecode has been replaced with subroutine calls. Unfortunately, I am very unfamiliar on the parallelization of the static fields. So it is a little difficult for me to comment on theses changes. Would it be useful for you if I were to try using your updated sourcecode separately? Just let me know, and I can redo the initialization of those static fields later this afternoon.

mgduda:
@ldfowler58 I think just giving the code a try would be sufficient for the purposes of a review.

ldfowler58:
Hi Michael:
I used your updated sourcecode to build the static fields and compared the different soil* fields against the ones without parallelization. While soilcl1, soilcl2, soilcl3, and soilcl4 are the same, I found very large differences for the soilcomp fields.

I out my original plots in /glade/u/home/laura/forMichael/ and plots using your sourcecode in /glade/u/home/laura/forMichael/parallel_noahmp_static/parallel.

Can you please look at the difference between the soilcomp fields. Thanks.

ldfowler58:
Maybe, we are missing the scale factor? Also, in subroutine interp_soilcomp, I did not find where soilcomp_int was computed? Thanks.

mgduda:
@ldfowler58 Thanks for checking the fields. It looks like I did forget to apply the `scale_factor` to `soilcomp`. I've just pushed another commit (that I'll squash before merging this PR) with two fixes: (1) apply `scale_factor` to the `soilcomp` field, and (2) only remap the `soilcomp` field to land points as defined by the MPAS `landmask` field.

mgduda:
@ldfowler58 To answer your other question, `soilcomp_int` is updated in the `soilcomp_interp_accumulation` routine, which is called from `init_atm_map_static_data` (which in turn is called from `interp_soilcomp`).

ldfowler58:
I updated my plots of soilcomp using the ncl script, making sure that I was not doing an extra rescaling. For all soilcomp fields, the parallel and non-parallel sourcecode produce similar interpolated fields. I agree to replace the non-parallel sourcecode with the one that parallelizes the remapping of the Noah-MP static fields.

ldfowler58 (approved):

mgduda:
@ldfowler58 Thanks for the review! I've just force-pushed an update to squash the last fixup commit for the scale factor into the previous commit, but I've verified that the rebase introduced no changes.

PR #1187, opened 2024-06-20
Add "-check all" to LDFLAGS_DEBUG for 'intel' build target

mgduda:
This PR adds `-check all` to `LDFLAGS_DEBUG` for the `intel` build target to resolve linking errors.

When building MPAS with the Intel oneAPI compilers and `DEBUG=true`, the linking step fails with undefined
references to various `__msan` symbols. According to the Intel compiler documentation, if `-check all` is used
to compile source files, that option must also be used during linking. From the Intel 2024.1 manual:
```
     Caution
     Files compiled with option check all should also be
     linked with this same option, or the link step may fail.
```

PR #1188, opened 2024-06-20
Add ONLY clause to USE statements for the 'mpi_f08' module

mgduda:
This PR adds `ONLY` clauses to all `USE` statements for the `mpi_f08` module in MPAS.

Besides being a generally good practice, the use of only those entities from the 'mpi_f08' module that are actually needed eliminates link-time issues with debug builds using the Intel oneAPI compilers (with the Cray MPICH library).

### DIFF
--- a/src/framework/mpas_abort.F
+++ b/src/framework/mpas_abort.F
@@ -33,7 +33,7 @@ module mpas_abort
 #ifdef _MPI
 #ifndef NOMPIMOD
 #ifdef MPAS_USE_MPI_F08
-      use mpi_f08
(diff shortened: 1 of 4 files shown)


mgduda:
When combined with PR #1187, `DEBUG=true` builds with the `intel/2024.0.2` module on Derecho should be successful. However, runs fail almost immediately due to what look like issues with uninitialized variables. So there may be another PR needed to turn off some checks (for example, changing `-check all` to something like `-check pointers,bounds`).

gdicker1 (approved):
As you say, these changes are best practice and make sense. Approved by inspection.

PR #1190, opened 2024-06-21
Fix intermittent parallel build failures in MPAS-A physics

mgduda:
This PR fixes intermittent parallel build failures in MPAS-A physics.

Firstly, this PR adds a missing dependency in `src/core_atmosphere/physics/Makefile`. The `core_physics` target should depend not only on `core_physics_wrf` but also on `core_physics_noahmp`. Note that `core_physics_wrf` already has a dependency on `core_physics_mmm`, and so it isn't necessary to also list `core_physics_mmm` as a dependency for `core_physics`.

Secondly, this PR splits the compilation of source files in MPAS-A physics. Now, Makefiles have been modified so that source files in `physics_wrf`, `physics_noahmp`, and `physics_mmm` can be compiled in parallel first, and as a subsequent step, object files from these directories are added to `libphys.a` in a serial fashion to avoid race conditions during library updates.

### DIFF
--- a/src/core_atmosphere/Externals.cfg
+++ b/src/core_atmosphere/Externals.cfg
@@ -2,7 +2,7 @@
 local_path = ./physics_mmm
 protocol = git
 repo_url = git@github.com:NCAR/MMM-physics.git
-tag = 20240525-MPASv8.2
+tag = 20240626-MPASv8.2
 
 required = True
 
--- a/src/core_atmosphere/physics/Makefile
+++ b/src/core_atmosphere/physics/Makefile
@@ -70,7 +70,7 @@ core_physics_noahmp:
 core_physics_init: $(OBJS_init)
 	ar -ru libphys.a $(OBJS_init)
 
-core_physics: core_physics_wrf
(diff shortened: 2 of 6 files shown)


ldfowler58:
I should have checked that parallel builds work after adding the directory physics_noahmp, but did not (I do not use parallel builds very often). After compiling the sourcecode with the updated Makefile, parallel builds (-j 4) worked with the intel and gnu compilers. I approve this PR. Thanks for fixing the Makefile.

mgduda:
Upon further testing, it looks like there are other issues with parallel builds related to the creating or updating of `libphys.a`. I'll see if it's possible to address those issues as part of this PR.

mgduda:
@ldfowler58 Can you re-review this PR? I've added another commit with the Makefile changes that allow us to build source directories in parallel, but to serialize the updating of `libphys.a`.

ldfowler58:
I reviewed this PR and now understand the issue regarding the parallel build. After reviewing changes to the Makefile, I agree with the PR. Thanks.

PR #1096, opened 2023-06-29
Add missing physics_mmm module search path in atmosphere Makefiles

mgduda:
This PR corrects compilation issues with certain Fortran compilers by adding the 
`physics_mmm` directory to the module search path in two atmosphere core
Makefiles. Some compilers (e.g., nvfortran) require the paths of modules that
are used indirectly to be present in the module search path provided with the
-I... option. Accordingly, this PR adds `-I./physics/physics_mmm` to the set of
compiler flags in `src/core_atmosphere/Makefile` and `-I../physics/physics_mmm` to
the set of compiler flags in `src/core_atmosphere/dynamics/Makefile` to fix
compilation errors.

PR #1191, opened 2024-06-21
Add module search paths to support compilation of Noah-MP code with nvfortran

mgduda:
This PR adds module search paths to support compilation of MPAS-A with Noah-MP using the nvfortran compiler.

The nvfortran compiler requires paths of modules that are used indirectly to be present in the module search path provided with the `-I...` option. To support compilation of Noah-MP code in MPAS-Atmosphere with the nvfortran compiler, this commit (1) adds Noah-MP source directories to `FCINCLUDES`, and (2) adds the path to the `esmf_time_f90` code when compiling Noah-MP code.


mgduda:
@gdicker1 This sort of change may seem familiar: it's along the lines of what we did in PR #1096.

gdicker1 (approved):
This does fix the nvhpc target for the release-v8.2.0 branch. Approved!

PR #1197, opened 2024-06-27
Don't include FCINCLUDES in build options checked for build compatibility

mgduda:
This PR removes `FCINCLUDES` from the set of build variables that are checked for build compatibility.

With commit c7401502, the `FCINCLUDES` variable is modified with core-specific options (for the atmosphere core specifically) and so including `FCINCLUDES` in the set of build options that are checked for compatibility in the top-level `Makefile` leads to a failed check, forcing a clean between the compilation of the atmosphere core and other cores.

The `FCINCLUDES` variable contains very similar information to the `CPPINCLUDES` variable, which is included in compatibility checks, and in order to enable the atmosphere core to be compiled after compilation of another core, this PR removes `FCINCLUDES` from the variables that are checked for build compatibility.


gdicker1 (approved):
Works as advertised, approved!

PR #1202, opened 2024-06-28
Use more mpi_f08 features in mpi_f08_test for better detection of valid mpi_f08 module

islas:
Use more mpi_f08 features in mpi_f08_test for odd cases where certain MPI implementations not compiled with mpi_f08 enabled still pass the current check.

Using OpenMPI 4.0.0 and gfortran 11.4.1 compiling without mpi_f08 features the MPI wrapper will still link and supply mpi_f08 (-lmpi_usempif08) despite not being valid or usable. This causes the `use mpi_f08` and some features to compile fine, but other mpi_f08 features to not exist and fail compilation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1267,9 +1267,10 @@ mpi_f08_test:
 	$(info Checking for mpi_f08 support...)
 	$(eval MPAS_MPI_F08 := $(shell $\
 		printf "program main\n$\
-		        &   use mpi_f08, only : MPI_Init, MPI_Comm\n$\
+		        &   use mpi_f08, only : MPI_Init, MPI_Comm, MPI_INTEGER, MPI_Datatype\n$\
 		        &   integer :: ierr\n$\
 		        &   type (MPI_Comm) :: comm\n$\
+		        &   type (MPI_Datatype), parameter :: MPI_INTEGERKIND = MPI_INTEGER\n$\
 		        &   call MPI_Init(ierr)\n$\
 		        end program main\n" | sed 's/&/ /' > mpi_f08.f90; $\
 		$\


mgduda:
@islas When you configured OpenMPI, did you add a flag to explicitly disable `mpi_f08`?

islas:
@mgduda Nope, I used `--enable-mpi-fortran`. From OpenMPI 4.0.0 :
```
  --enable-mpi-fortran    specify which Fortran MPI bindings to build: yes,
                          none (or no), best-effort, mpifh (build only mpif.h
                          support), usempi (build mpif.h and the mpi module),
                          or usempif08 (or all, build mpifh, the mpi module,
                          and the mpi_f08 module) (default: "yes" if Fortran
                          compiler found)
```

Though the mpi_f08 module exists in lib/ it fails to compile with : 
```
gfortran -I/home/aislas/dependencies//openmpi/include -I/home/aislas/dependencies//openmpi/lib -Wl,-rpath -Wl,/home/aislas/dependencies//openmpi/lib -Wl,--enable-new-dtags -L/home/aislas/dependencies//openmpi/lib -lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh -lmpi -o test_f08 mpi_f08_test.f90 -std=f2008
mpi_f08_test.f90:5:53:

    5 |   type (MPI_Datatype), parameter :: MPI_INTEGERKIND = MPI_INTEGER
      |                                                     1
Error: Parameter ‘mpi_integer’ at (1) has not been declared or is a variable, which does not reduce to a constant expression
```
Wrapper expanded with `-showme` and `-std=f2008` added to manual compilation of the test program. Given the older version of OpenMPI, it wouldn't surprise me if this is a bug. 

Also re-configuring OpenMPI with `--enable-mpi-fortran=usempif08` yields the same results.

mgduda (approved):

PR #1203, opened 2024-07-02
Enhance lib / lib64 library detection for pnetcdf and netcdf

islas:
The original additions of checking lib64/ for pnetcdf/netcdf (#592) made assumptions that when the variable supplied for the respective library location, if a lib64/ folder exists to use that. This leads to problems when existence of the actual library is not checked and if a lib64/ folder does exist but the library is not installed there. As this configuration layout is not unusual, the fix is to keep the precedence of lib64/, but check that the library exists within the directory before assigning values.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -549,10 +549,10 @@ endif
 
 
 ifneq "$(PNETCDF)" ""
-ifneq ($(wildcard $(PNETCDF)/lib), )
+ifneq ($(wildcard $(PNETCDF)/lib/libpnetcdf.*), )
 	PNETCDFLIBLOC = lib
 endif
-ifneq ($(wildcard $(PNETCDF)/lib64), )
+ifneq ($(wildcard $(PNETCDF)/lib64/libpnetcdf.*), )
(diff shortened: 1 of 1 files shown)


islas:
Future work should consider using `pkg-config` or any of the `pnetcdf-config`  / `nc-config` / `nf-config` to supply the appropriate link / compile flags. This would generally reduce the maintenance burden from rewriting logic that is already provided by those projects. Likewise, this would lean towards more standard package / dependency management rather than project-specific variables.

mgduda:
@islas The commit message for ef3c4b38 references a commit (3dab782b) that doesn't seem to (no longer?) exists. Could you update the commit message?

islas:
Ah! I cleaned up with a rebase before updating the remote, but didn't update the appropriate descriptions. Incoming fix

islas:
islas/MPAS-Model@3dab782 now changed to islas/MPAS-Model@fe46d4f to reference islas/MPAS-Model@0ca4208

mgduda (approved):

PR #1161, opened 2024-04-14
Implemented the Noah-MP Land Surface Model

ldfowler58:
This PR adds the implementation of the Noah-MP land surface model (LSM):

* The sourcecode of the Noah-MP LSM was downloaded from the Noah-MP github repository and corresponds to the release-v5.0.1. It is located in the directory ./src/core_atmosphere/physics/physics_noahmp.

* ./src/core_atmosphere/physics/Registry_noahmp.xml contains all the run options and variables needed to link the Noah-MP sourcecode to the MPAS physics framework.

* ./src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F includes the initialization of the Noah-MP LSM.

* ./src/core_atmosphere/physics/mpas_atmphys_driver_lsm_noahmp.F is the main driver for the Noah-MP LSM.

PR #1204, opened 2024-07-03
Add core_physics dependencies for mpas_atmphys_lsm_noahmpfinalize module

islas:
Module `mpas_atmphys_lsm_noahmpfinalize` added in #1161 does not have the appropriate dependencies in the respective makefile.

This adds the necessary dependencies to the `mpas_atmphys_lsm_noahmpfinalize.o`

### DIFF
--- a/src/core_atmosphere/physics/Makefile
+++ b/src/core_atmosphere/physics/Makefile
@@ -194,6 +194,9 @@ mpas_atmphys_lsm_noahmpinit.o: \
 	mpas_atmphys_utilities.o \
 	mpas_atmphys_vars.o
 
+mpas_atmphys_lsm_noahmpfinalize.o : \
+	mpas_atmphys_vars.o
+
 mpas_atmphys_manager.o: \
 	mpas_atmphys_constants.o \
 	mpas_atmphys_o3climatology.o \


src/core_atmosphere/physics/Makefile
```diff
@@ -165,6 +165,10 @@ mpas_atmphys_driver_sfclayer.o: \
 mpas_atmphys_finalize.o: \
 	mpas_atmphys_lsm_noahmpfinalize.o
 
+mpas_atmphys_lsm_noahmpfinalize.o : \
+	mpas_atmphys_vars.o \
+	core_physics_noahmp
```
mgduda, line 170: I'd suggest moving this dependency block below that for `mpas_atmphys_lsm_noahmpinit.o`. Also, we're guaranteed that `core_physics_noahmp` will be built before any of the object files in `$(OBJS)`, so I would suggest removing `core_physics_noahmp` for consistency with other targets in this directory that also depend on Noah-MP.

mgduda (changes requested):

src/core_atmosphere/physics/Makefile
islas: I missed that `core_physics_noahmp` is added as a dependency in later commits. 
Making bug fix changes based on the the originating commit leaves it open to rediscovering or fighting against issues solved in later commits.

mgduda:
@islas Can you add some detail to the commit message (and use capitalization where appropriate)? A couple of questions to consider for the commit message:
 - what defines "minimum dependencies"?
 - how have we determined that "mpas_atmphys_vars.o" is the minimum dependency?

islas:
Is this roughly what you are looking for from the commit message?

> Add the minimum object dependencies in the makefile for mpas_atmphys_lsm_noahmpfinalize.o. This corresponds to all unresolved modules used yet to be compiled by the time this rule is evaluated.
Modules originating from NoahMP and framework are ignored as they should be handled separately before the core_physics target is started.

I'm refraining from saying "NoahMP and framework _are_ already compiled by the time the core_physics target is started" since the fix is based on a commit that does not have those fix in it, thus would be referring to future patches that have yet to occur if one were to try and take this individual commit.

mgduda (approved):

PR #1205, opened 2024-07-05
Fixed CMake build for atmosphere core

amstokely:
Prior to releasing MPAS v8.2.0, several changes necessitated updates to the GNU Makefiles. These changes included:

- Inclusion of the noamp library
- Renaming of several existing files
- Removing the `physics_mmm` directory from the source code and moving it to its own GitHub repository
- Adding a source file to the init atmosphere core
- Modifying how the registry is processed

Unfortunately, the CMake build files were not updated accordingly, resulting in the loss of CMake support in MPAS v8.2.

This PR includes the necessary updates to the CMake files, enabling MPAS to be configured and built with CMake. These updates ensure that the latest version of MPAS can be used with `mpas-jedi` and allow MPAS-Model to be compiled using both GNU and Intel compilation suites.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ src/core_*/*.f90
 src/framework/*.f90
 src/driver/*.f90
 src/operators/*.f90
+src/core_atmosphere/physics/physics_mmm
 
 # All object files
 *.o
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ src/core_*/*.f90
 src/framework/*.f90
 src/driver/*.f90
 src/operators/*.f90
-src/core_atmosphere/physics/physics_mmm
 
 # All object files
 *.o
--- a/src/core_atmosphere/CMakeLists.txt
+++ b/src/core_atmosphere/CMakeLists.txt
@@ -2,165 +2,354 @@
 ## Source files
 # physics/
 set(ATMOSPHERE_CORE_PHYSICS_SOURCES
-        ccpp_kinds.F
-        mpas_atmphys_camrad_init.F
-        mpas_atmphys_constants.F
-        mpas_atmphys_control.F
-        mpas_atmphys_date_time.F
-        mpas_atmphys_driver_cloudiness.F
-        mpas_atmphys_driver_microphysics.F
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@ src/core_*/*.f90
 src/framework/*.f90
 src/driver/*.f90
 src/operators/*.f90
+src/core_atmosphere/physics/physics_mmm
```
mgduda, line 9: The `physics_mmm` directory is already ignored in the `hotfix-v8.2.1` branch in the `src/core_atmosphere/physics/.gitignore` file.

src/framework/mpas_dmpar.F
```diff
@@ -551,6 +551,7 @@ subroutine mpas_dmpar_bcast_reals(dminfo, n, rarray, proc)!{{{
 
    end subroutine mpas_dmpar_bcast_reals!}}}
 
+
```
mgduda, line 554: Can you remove this whitespace change?

mgduda (changes requested):

src/core_atmosphere/CMakeLists.txt
```diff
@@ -2,165 +2,354 @@
...
+    module_cam_shr_kind_mod.F
+    module_cam_support.F
+    module_cu_gf.mpas.F
+    module_mp_kessler.F
+    module_mp_radar.F
+    module_mp_thompson.F
+    module_mp_thompson_cldfra3.F
+    module_mp_thompson_aerosols.F
+    module_mp_wsm6.F
+    module_ra_rrtmg_sw_aerosols
```
mgduda, line 65: I think this filename is missing a suffix. (Also, in my opinion, this would have been easier to spot if so many whitespace changes were not introduced.)

mgduda (changes requested):

src/core_atmosphere/CMakeLists.txt
amstokely: Good catch! Also you can hide white space changes. I always do this and find it to be really helpful.

mgduda (approved):
Thanks for these fixes -- all looks good to me!

PR #1206, opened 2024-07-05
Remove debugging prints around calls to sfclayrev in driver_sfclayer

mgduda:
This PR removes some debugging log writes around calls to `sfclayrev` in the `driver_sfclayer` subroutine. These writes led to messages like the following appearing in the log file for every model timestep:
```
 --- enter subroutine sfclayrev:
 --- end subroutine sfclayrev:
 --- enter subroutine sfclayrev seaice:
 --- end subroutine sfclayrev seaice:
```


mgduda:
@ldfowler58 If you'd prefer, I can comment-out the log writes rather than remove them. But for consistency with calls to the old surface-layer scheme, I thought to simply remove the writes instead.

ldfowler58:
No problems. It makes more sense to remove those write statements. I approve this PR. Thanks.

PR #1207, opened 2024-07-09
Add mpas_dmpar_bcast_real4s routine for broadcasting real(kind=R4KIND) arrays

mgduda:
This PR adds a new `mpas_dmpar_bcast_real4s` routine for broadcasting `real(kind=R4KIND)` arrays.

The `mpas_dmpar_bcast_real4s` routine mirrors the `mpas_dmpar_bcast_reals` routine, with the dummy array argument being of `R4KIND` rather than `RKIND` kind.

### DIFF
--- a/src/framework/mpas_dmpar.F
+++ b/src/framework/mpas_dmpar.F
@@ -98,6 +98,7 @@ include 'mpif.h'
    public :: mpas_dmpar_bcast_ints
    public :: mpas_dmpar_bcast_real
    public :: mpas_dmpar_bcast_reals
+   public :: mpas_dmpar_bcast_real4s
    public :: mpas_dmpar_bcast_double
    public :: mpas_dmpar_bcast_doubles
    public :: mpas_dmpar_bcast_logical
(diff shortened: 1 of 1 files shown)


mgduda:
This PR targets the `hotfix-v8.2.1` branch since the `mpas_dmpar_bcast_real4s` routine added in this PR is required by PR #1208.

amstokely (approved):
I think this is a good temporary solution for fixing the double precision compilation problem. However, I think we should come up with a more robust way to broadcast real arrays in the future, as this subroutine would not be supported in a generic interface.

mgduda:
> I think this is a good temporary solution for fixing the double precision compilation problem. However, I think we should come up with a more robust way to broadcast real arrays in the future, as this subroutine would not be supported in a generic interface.

Just to clarify, the `mpas_dmpar_bcast_real4s` routine definitely can be used as a specific implementation of a generic interface. If we were to do so, as I think you're getting at, there would be issues in calling the generic interface with multi-dimensional arrays, but that's no fault of the `mpas_dmpar_bcast_real4s` routine.

PR #1208, opened 2024-07-09
Fix double-precision builds of MPAS-A v8.2.0

mgduda:
This PR corrects double-precision build failures of `mpas_atmphys_init_microphysics.F` in MPAS-A v8.2.0.

The `mpas_atmphys_init_microphysics` module contains a routine, `table_ccnAct`, that
broadcasts an `R4KIND` real array, `tnccn_act`, from the Thompson microphysics
scheme. This broadcast was previously performed by a call to
`mpas_dmpar_bcast_reals` through the macro `DM_BCAST_MACRO`. However,
because the `mpas_dmpar_bcast_reals` routine broadcasts arrays of kind `RKIND`, when
MPAS-A is compiled with double-precision reals, the `mpas_dmpar_bcast_reals`
routine can no longer be used with the `tnccn_act` array.
    
With this PR, the `DM_BCAST_MACRO` in `mpas_atmphys_init_microphysics.F` has 
been defined to `mpas_dmpar_bcast_real4s`, which broadcasts arrays of `R4KIND`
reals. This works for both single- and double-precision builds of MPAS-A, since
the kind type of the `tnccn_act` array is independent of the MPAS-A build
precision.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_init_microphysics.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_init_microphysics.F
@@ -5,7 +5,7 @@
 ! Additional copyright and license information can be found in the LICENSE file
 ! distributed with this code, or at http://mpas-dev.github.com/license.html
 !
-#define DM_BCAST_MACRO(A) call mpas_dmpar_bcast_reals(dminfo,size(A),A)
+#define DM_BCAST_MACRO(A) call mpas_dmpar_bcast_real4s(dminfo,size(A),A)
 
 !=================================================================================================================
  module mpas_atmphys_init_microphysics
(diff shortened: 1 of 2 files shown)


mgduda:
This bugfix PR depends on PR #1207 to add the necessary broadcast routine for `R4KIND` real arrays.

ldfowler58:
This is quite interesting. I had actually done this in the past (adding a mpas_dmpar_bcast_real4s) for exactly that reason but never made a PR since we now mostly use single precision. Totally forgot to test compiling MPAS using double precision. I, of course, approve this PR. Thanks.

amstokely (approved):
Thanks for fixing this! I was able to compile MPAS-Model with double precision enabled on both gnu and intel platforms with GNU Make and CMake.

PR #1220, opened 2024-07-30
Enable GPU execution of atm_recover_large_step_variables via OpenACC

gdicker1:
This PR enables GPU calculation of post-acoustic step variable reconstitution by adding OpenACC directives to the `atm_recover_large_step_variables_work` routine.

Timing information for the OpenACC data transfers in this routine is captured in the log file by a new timer: `atm_recover_large_step_variables [ACC_data_xfer]`

Invariant fields used in the work routine are copied in during `mpas_atm_dynamics_init` and deleted in `mpas_atm_dynamics_finalize` by building off the fields already handled in previous PRs.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -2816,13 +2816,21 @@ module atm_time_integration
       integer :: i, iCell, iEdge, k, cell1, cell2
       real (kind=RKIND) :: invNs, rcv, p0, flux
 
+      MPAS_ACC_TIMER_START('atm_recover_large_step_variables [ACC_data_xfer]')
+      !$acc enter data copyin(rho_p_save,rho_pp,rho_base,rw_save, &
+      !$acc                  rw_p,rtheta_base,rtheta_p_save,rtheta_pp, &
+      !$acc                  rt_diabatic_tend,exner_base,ru_save,ru_p,ruAvg)
+      !$acc enter data create(rho_zz,rho_p,rw,w,rtheta_p, &
+      !$acc                  theta_m,exner,pressure_p,ru,u)
+      !$acc enter data copyin(wwAvg,ruAvg)
+      MPAS_ACC_TIMER_STOP('atm_recover_large_step_variables [ACC_data_xfer]')
 
       rcv = rgas/(cp-rgas)
       p0 = 1.0e+05  ! this should come from somewhere else...
 
       ! Avoid FP errors caused by a potential division by zero below by 
       ! initializing the "garbage cell" of rho_zz to a non-zero value
-      !$acc parallel
+      !$acc parallel default(present)
       !$acc loop vector
       do k=1,nVertLevels
          rho_zz(k,nCells+1) = 1.0
@@ -2834,7 +2842,7 @@ module atm_time_integration
 
       invNs = 1 / real(ns,RKIND)
 
-      !$acc parallel
+      !$acc parallel default(present)
       !$acc loop gang worker
       do iCell=cellStart,cellEnd
 
@@ -2864,7 +2872,7 @@ module atm_time_integration
       !$acc end parallel
 
       if (rk_step == 3) then
-         !$acc parallel
+         !$acc parallel default(present)
          !$acc loop collapse(2)
          do iCell=cellStart,cellEnd
 !DIR$ IVDEP
@@ -2880,7 +2888,7 @@ module atm_time_integration
          end do
          !$acc end parallel
       else
-         !$acc parallel
+         !$acc parallel default(present)
          !$acc loop collapse(2)
          do iCell=cellStart,cellEnd
 !DIR$ IVDEP
@@ -2899,7 +2907,7 @@ module atm_time_integration
 
 !$OMP BARRIER
 
-      !$acc parallel
+      !$acc parallel default(present)
       !$acc loop gang worker
       do iEdge=edgeStart,edgeEnd
 
@@ -2959,6 +2967,15 @@ module atm_time_integration
 
       end do
       !$acc end parallel
+
+      MPAS_ACC_TIMER_START('atm_recover_large_step_variables [ACC_data_xfer]')
+      !$acc exit data delete(rho_p_save,rho_pp,rho_base,rw_save, &
+      !$acc                  rw_p,rtheta_base,rtheta_p_save,rtheta_pp, &
+      !$acc                  rt_diabatic_tend,ru_save,ru_p,ruAvg)
+      !$acc exit data copyout(rho_zz,rho_p,rw,w,rtheta_p, &
+      !$acc                   theta_m,exner,pressure_p,ru,u)
+      !$acc exit data copyout(wwAvg,ruAvg)
+      MPAS_ACC_TIMER_STOP('atm_recover_large_step_variables [ACC_data_xfer]')
 
    end subroutine atm_recover_large_step_variables_work
 
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -212,6 +212,13 @@ module atm_time_integration
       real (kind=RKIND), dimension(:), pointer :: invAreaCell
       integer, dimension(:), pointer :: bdyMaskCell
       integer, dimension(:), pointer :: bdyMaskEdge
+      real (kind=RKIND), dimension(:,:), pointer :: zz
+      real (kind=RKIND), dimension(:,:,:), pointer :: zb
+      real (kind=RKIND), dimension(:,:,:), pointer :: zb3
+      real (kind=RKIND), dimension(:,:,:), pointer :: zb_cell
+      real (kind=RKIND), dimension(:,:,:), pointer :: zb3_cell
+      real (kind=RKIND), dimension(:), pointer :: fzm
+      real (kind=RKIND), dimension(:), pointer :: fzp
 #endif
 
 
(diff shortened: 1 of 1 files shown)


gdicker1:
NOTE: the changes in this PR slightly change answers. The changes seem to come from a couple of loops. In my regional test case, if the calculations of `rho_p`, `rho_zz`, `rtheta_p`, `exner`, and `pressure_p` are done on the CPU instead, no differences occur.

First this loop
```
         do k = 1, nVertLevels
            rho_p(k,iCell) = rho_p_save(k,iCell) + rho_pp(k,iCell)
            rho_zz(k,iCell) = rho_p(k,iCell) + rho_base(k,iCell)
         end do
```

And in this loop
```
            do k = 1, nVertLevels
               rtheta_p(k,iCell) = rtheta_p_save(k,iCell) + rtheta_pp(k,iCell) &
                                 - dt * rho_zz(k,iCell) * rt_diabatic_tend(k,iCell)
               theta_m(k,iCell) = (rtheta_p(k,iCell) + rtheta_base(k,iCell))/rho_zz(k,iCell)
               exner(k,iCell) = (zz(k,iCell)*(rgas/p0)*(rtheta_p(k,iCell)+rtheta_base(k,iCell)))**rcv
               ! pressure_p is perturbation pressure
               pressure_p(k,iCell) = zz(k,iCell) * rgas * (exner(k,iCell)*rtheta_p(k,iCell)+rtheta_base(k,iCell)  &
                                                          * (exner(k,iCell)-exner_base(k,iCell)))
            end do
```

mgduda:
Although there isn't anything in particular that looks suspect to me, it does seem like the differences in results after the first timestep of a 12-km regional simulation comparing with the current `develop` branch are a bit larger than I would have expected (based on the differences we see before and after the merge of #1216).

I also get a core dump at the end of the simulation on Derecho when using the following modules:
```
module --force purge
ml ncarenv/23.09    
ml craype/2.7.23    
ml nvhpc/24.3
ml ncarcompilers/1.0.0
ml cray-mpich/8.1.27
ml parallel-netcdf/1.12.3
ml cuda/12.2.1
```

It may be worth taking a second look at the changes in this PR to see if we can find anything that's subtly off.

gdicker1:
@mgduda, this is ready for another round of review!

There are slight answer differences, but isolated just to `exner` and `pressure_p` from what I can tell. No answer differences occur if the code below is added after the loop to calculate `rtheta_p`, `theta_m`, `exner`, and `pressure_p` (near line 2892 in this PR).

```
         ! Calculate exner and pressure_p on CPU
         !$acc update host(rtheta_p)
         do iCell=cellStart,cellEnd
            do k = 1, nVertLevels
               exner(k,iCell) = (zz(k,iCell)*(rgas/p0)*(rtheta_p(k,iCell)+rtheta_base(k,iCell)))**rcv
               ! pressure_p is perturbation pressure
               pressure_p(k,iCell) = zz(k,iCell) * rgas * (exner(k,iCell)*rtheta_p(k,iCell)+rtheta_base(k,iCell)  &
                                                          * (exner(k,iCell)-exner_base(k,iCell)))
            end do
         end do
         !$acc update device(exner,pressure_p)
```

mgduda:
@gdicker1 If I use the current HEAD of the `develop` branch (ce4d6709) as a baseline, I find that in addition to the code you mention above to compute `exner` and `pressure_p` on the host, I also need to add
```
         rw(1,iCell) = 0.0
```
and
```
         rw(nVertLevels+1,iCell) = 0.0
```
around lines 2860 and 2873 (in this PR branch) in order to get identical results with the baseline.
It makes sense to me that this might be needed, since we `create` rather than `copyin` the `rw` field; but what surprises me is that this isn't needed in the PR branch (if we use 63fdcf70 as the baseline, and don't merge onto the HEAD of the `develop` branch). Perhaps we are just lucky in getting some zeros in the memory that's given to us on the device?

If you agree that we should be initializing the `rw` field at the first and last level, can you add those changes to the PR branch?

gdicker1:
> @gdicker1 If I use the current HEAD of the develop branch (https://github.com/MPAS-Dev/MPAS-Model/commit/ce4d6709342b796dab5d08ea1c175ca9fa3c5534) as a baseline, I find that in addition to the code you mention above to compute exner and pressure_p on the host, I also need to add ...

I agree with those changes, and will get them pushed. I would assume we were just getting lucky before this point!

gdicker1:
@mgduda this is ready for re-review. Apologies for not pinging after I pushed [cde6447](https://github.com/MPAS-Dev/MPAS-Model/pull/1220/commits/cde6447d5b529a2210510d1c1626f670c1edfdf4).

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -2759,45 +2815,71 @@ subroutine atm_recover_large_step_variables_work(nCells, nEdges, nCellsSolve, nE
       integer :: i, iCell, iEdge, k, cell1, cell2
       real (kind=RKIND) :: invNs, rcv, p0, flux
 
+      MPAS_ACC_TIMER_START('atm_recover_large_step_variables [ACC_data_xfer]')
+      !$acc enter data copyin(rho_p_save,rho_pp,rho_base,rw_save,rw_p, &
+      !$acc                   rtheta_p_save,rtheta_pp,rtheta_base, &
+      !$acc                   ru_save,ru_p) &
+      !$acc            create(rho_zz,rho_p,rw,w,rtheta_p,theta_m, &
+      !$acc                   ru,u) &
+      !$acc            copyin(wwAvg,ruAvg)
```
mgduda, line 2824: Out of curiosity, is there a reason for placing `wwAvg` and `ruAvg` in their own `copyin` clause rather than including them in the earlier `copyin` clause?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -2759,45 +2815,71 @@ subroutine atm_recover_large_step_variables_work(nCells, nEdges, nCellsSolve, nE
...
+      end if
+      MPAS_ACC_TIMER_STOP('atm_recover_large_step_variables [ACC_data_xfer]')
 
       rcv = rgas/(cp-rgas)
       p0 = 1.0e+05  ! this should come from somewhere else...
 
-      ! Avoid FP errors caused by a potential division by zero below by 
+      ! Avoid FP errors caused by a potential division by zero below by
       ! initializing the "garbage cell" of rho_zz to a non-zero value
+      !$acc parallel
```
mgduda, line 2836: This might have been something that we talked about as a potential "best practice" after this PR was started, but would it be worth adding `default(present)` clauses to `parallel` directives?

gdicker1: I'll get them added. I also think this PR pre-dates that discussion.

I think the `default(present)` directives are useful right now as we're porting and when we get to the point where we want to fuse data regions. Once the port is "finished" I think these clauses may cause friction whenever new code gets added. (Even that might be a benefit - e.g. if a physics change makes the `scalars` array partially present because it only modifies moisture species, that should be addressed before the PR is accepted.)

gdicker1: Addressed in [cc7e7b7](https://github.com/MPAS-Dev/MPAS-Model/pull/1220/commits/cc7e7b7d9ec1887d3b9860ad71398f2dedadaa5b).

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: Nothing I remember now. This is addressed in fixup commit [f631856](https://github.com/MPAS-Dev/MPAS-Model/pull/1220/commits/f631856105c78079ced1b347f95e5415cb270592)

gdicker1:
@mgduda, ready for another review!

mgduda:
@gdicker1 Even after applying the patch to compute `exner` and `pressure_p` on the host, I'm seeing answer differences with the changes in this PR. Can you verify that you are able to get bitwise-identical results when correcting the computation of just `exner` and `pressure_p`?

mgduda:
@gdicker1 It looks like there's a possible bug in your commit f6318561: `wwAvg` and `ruAvg` are now in a `delete` clause rather than a `copyout` clause.

gdicker1:
Force-push [cc7e7b7](https://github.com/MPAS-Dev/MPAS-Model/commit/cc7e7b7d9ec1887d3b9860ad71398f2dedadaa5b) to [adfdc2f](https://github.com/MPAS-Dev/MPAS-Model/commit/adfdc2f4b96d331943c5537e8342b0a9eb65e510) to rebase this branch on develop. (Currently 4ea6abbd, the merge of #1223 onto develop).

gdicker1:
This should be ready for another review after pushing fixup [c29fcb5](https://github.com/MPAS-Dev/MPAS-Model/pull/1220/commits/c29fcb5392d97ccb7a74e02e94c5c07160c9328d).

@mgduda, does this pass as bitwise identical with this output from `compare_netcdf.py`?
```
Comparing files: f1 - f2
        f1=baseline_acc_1736985688/restart.2019-09-01_00.06.00.nc
        f2=merge-dev-1223_1737050011/restart.2019-09-01_00.06.00.nc

            Variable  Min       Max       
              relhum -10.240860  106.321869
        refl10cm_max -35.000000  2.681815
```

Above, the baseline results compared with this PR with:
1. The PR branch modified so the calculations for `exner` and `pressure_p` are conducted on the host
2. Both the baseline and test compiled with `OPENACC=true`

gdicker1:
> @mgduda, does this pass as bitwise identical with this output from `compare_netcdf.py`?
> ...

I found the difference, I have different output requests in my streams.atmosphere for the baseline and the test run. This means some diagnostics are computed on different schedules. I'll ping again once I've made sure the runs are exactly the same.

gdicker1:
@mgduda, using the same streams.atmosphere settings, I now get the same results. This is ready for review!

mgduda (changes requested):

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -344,6 +346,9 @@ subroutine mpas_atm_dynamics_init(domain)
       call mpas_pool_get_array(mesh, 'zz', zz)
       !$acc enter data copyin(zz)
 
+      call mpas_pool_get_array(mesh, 'zz', zz)
+      !$acc enter data copyin(zz)
```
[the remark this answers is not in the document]
gdicker1: Crud, thanks for the catch!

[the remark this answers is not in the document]
gdicker1: Addressed by fixup [baa8b40](https://github.com/MPAS-Dev/MPAS-Model/pull/1220/commits/baa8b401ce5bc5032b6ba17bc3630379b4a16d51)

gdicker1:
Force-push [baa8b40](https://github.com/MPAS-Dev/MPAS-Model/commit/baa8b401ce5bc5032b6ba17bc3630379b4a16d51) to [be338c9](https://github.com/MPAS-Dev/MPAS-Model/commit/be338c9590088b5598ed860c9667a51d71d47082) squashes this into one commit.

mgduda (approved):

PR #1221, opened 2024-08-02
Fix MPAS-A dycore-only build in version 8.2.0

kuanchihwang:
When building MPAS-A as a dycore, all physics-related components are disabled. An example scenario of this is for use with CAM/CAM-SIMA.

However, the MPAS-A registry file in version 8.2.0 contains the following regressions, which cause the dycore-only build to break:

* The MPAS-A registry file introduced a new stream, `da_state`, which unconditionally includes variables that exist only when physics are enabled.
* The MPAS-A registry file unconditionally includes a new registry file from Noah-MP, which is applicable only when physics are enabled.

This PR fixes MPAS-A dycore-only build by guarding the above regions with the `DO_PHYSICS` macro.

For stand-alone MPAS-A, this PR is an NFC (No Functional Change).

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -1155,10 +1155,6 @@
 			<var_array name="scalars"/>
 			<var name="initial_time"/>
 			<var name="xtime"/>
-			<var name="cldfrac"/>
-			<var name="re_cloud" packages="mp_thompson_in;mp_wsm6_in"/>
-			<var name="re_ice" packages="mp_thompson_in;mp_wsm6_in"/>
-			<var name="re_snow" packages="mp_thompson_in;mp_wsm6_in"/>
 			<var name="u"/>
 			<var name="w"/>
 			<var name="rho"/>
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/Makefile
```diff
@@ -4,8 +4,11 @@
 # To build a dycore-only MPAS-Atmosphere model, comment-out or delete
 # the definition of PHYSICS, below
 #
-PHYSICS=-DDO_PHYSICS
-
+# If MPAS_CAM_DYCORE is found in CPPFLAGS, PHYSICS will become undefined automatically
+#
+ifeq ($(findstring MPAS_CAM_DYCORE,$(CPPFLAGS)),)
```
mgduda, line 9: Although I agree that this is a reasonable change in principle, I think it may be outside the scope what we want to include in the v8.2.1 release. My understanding is that CAM's build already ensured that `DO_PHYSICS` was not defined, and the need for this change may be due to the reuse of makefiles in ongoing work on CAM-SIMA?

mgduda (changes requested):

mgduda (approved):
I've verified that the changes in this PR fix dynamics-only builds of stand-alone MPAS-A.

src/core_atmosphere/Makefile
mgduda: I fully support creating a separate PR to the `develop` branch with this `MPAS_CAM_DYCORE` logic.

PR #1227, opened 2024-09-03
Remove unused arguments to atm_set_smlstep_pert_variables_work

gdicker1:
This PR removes unused arguments from the atm_set_smlstep_pert_variables_work and atm_set_smlstep_pert_variables routines.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -2224,19 +2224,19 @@ module atm_time_integration
       do iCell=cellSolveStart,cellSolveEnd
 
          if (bdyMaskCell(iCell) <= nRelaxZone) then  !  no conversion in specified zone, regional_MPAS
-         do i=1,nEdgesOnCell(iCell)
-            iEdge = edgesOnCell(i,iCell)
+            do i=1,nEdgesOnCell(iCell)
+               iEdge = edgesOnCell(i,iCell)
+!DIR$ IVDEP
+               do k = 2, nVertLevels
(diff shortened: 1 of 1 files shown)


gdicker1:
Note, I structured this PR so the final two commits (that deal with arguments to the main routine) can be dropped if those changes aren't desired.

mgduda:
In my opinion, having five commits to remove unused arguments and variables may be unnecessary. I'd suggest we consider revising this PR so that it contains just two commits: one commit to make whitespace-only changes to executable code, and a second commit to remove unused arguments and variables (but without any reordering).

gdicker1:
> In my opinion, having five commits to remove unused arguments and variables may be unnecessary...

Agreed. I will rebase to collapse these commits and update the PR description.

gdicker1:
@mgduda, please take another look. Despite the double force-push, the only changes are to the number of commits - no code changes.

mgduda (changes requested):
The changes look good to me. I'd only suggest adding a bit more detail in the message for the whitespace-only commit. Taken on its own, the message doesn't give a good idea of what part of the code is affected in my opinion.

gdicker1:
@mgduda, the force push now uses the commit message

> Whitespace, improve if-statement readability in atm_set_smlstep_pert_variables_work

Please let me know if there's any other adjustments needed!

mgduda (approved):

gdicker1:
Sub-PRs:
- [ ] https://github.com/EarthWorksOrg/mpas-seaice/pull/19
- [ ] https://github.com/EarthWorksOrg/mpas-ocean/pull/17

PR #1229, opened 2024-09-09
Fix portability issue in parse_macros function used by the MPAS registry

mgduda:
This PR fixes a portability issue in the `parse_macros` function used by the MPAS registry.

The `parse_macros` function that is used by the MPAS registry `parse` tool previously made use of the state/saveptr/lasts argument to `strtok_r` when parsing command-line macro definitions. This was not portable and resulted in incorrectly generated output on some systems.

This PR reworks the `parse_macros` function so that it no longer uses the state/saveptr/lasts argument to `strtok_r` and instead relies only on the return value of `strtok_r` to provide portability across systems with different `strtok_r` implementations.


gdicker1 (approved):
Approved!

PR #1232, opened 2024-09-16
add `brtemp` and `cldmask` under `diag` var_struct of `src/core_atmosphere/Registry.xml

jim-p-w:
Add brtemp and cldmask fields to atmosphere Registry.xml file for MPAS-JEDI.

These fields are required by MPAS-JEDI and are associated with the `jedi_da` package,
which is active only if `config_jedi_da = true`.
Without setting `config_jedi_da = true` this PR should have no impact on
memory usage for stand-alone MPAS-A applications.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -1839,6 +1839,15 @@
                 <var name="surface_pressure" type="real" dimensions="nCells Time" units="Pa"
                      description="Diagnosed surface pressure"/>
 
+                <!-- observation-related variables interfacing to Non-Variational SAtellite-based Cloud Analysis (SACA)  -->
+                <var name="cldmask" type="real" dimensions="nCells Time" units="unitless"
+                     description="cloud mask (1=cloudy ; 0=clear)"
+                     packages="jedi_da"/>
+
+                <var name="brtemp" type="real" dimensions="nCells Time" units="K"
+                     description="brightness temperature to calculate cloud top height; usually from geostationary IR window channel"
+                     packages="jedi_da"/>
+
         </var_struct>
 
         <var_struct name="tend" time_levs="1">


mgduda:
@jim-p-w Can you update the PR description and also the commit message to explain why we need to add variables to the MPAS-A Registry that are not used anywhere in the MPAS-A model? Also, it would be worth noting that these variables should have no impact on memory usage for stand-alone MPAS-A applications due to the package that's attached to the variables. As it is, the description is probably not useful for MPAS-A developers, and I think an important function of PR descriptions is to convince maintainers of a repository that code changes should be accepted into the repository.

jim-p-w:
@mgduda The PR description and commit message have been updated.

mgduda:
@jim-p-w The commit message begins with "This merge adds brtemp and cldmask under diag var_struct of src/core_atmosphere/Registry.xml (MPAS-Dev#43)". The commit isn't a merge, and MPAS-Dev#43 isn't anything related to these changes in the MPAS-Dev/MPAS-Model repository. Also, note that we try to keep the one-line summary of commit messages to 80 characters or less.

I would also suggest removing the bit about madwrf and SACA from the PR description, removing the "DESCRIPTION OF CHANGES:" and "LIST OF MODIFIED FILES:" lines, and also removing the part about cherry-picking.

jim-p-w:
How about
For the commit message: 
```
Add brtemp and cldmask under diag var_struct of src/core_atmosphere/Registry.xml

These fields are used by MPAS-JEDI and are associated with the `jedi_da` package,
which is active only if `config_jedi_da = true`.
Without setting `config_jedi_da = true` this commit should have no impact on
memory usage for stand-alone MPAS-A applications.
```

For the PR message use the same verbiage, but change `commit` to `PR`

mgduda:
> How about For the commit message:
> 
> ```
> Add brtemp and cldmask under diag var_struct of src/core_atmosphere/Registry.xml
> 
> These fields are used by MPAS-JEDI and are associated with the `jedi_da` package,
> which is active only if `config_jedi_da = true`.
> Without setting `config_jedi_da = true` this commit should have no impact on
> memory usage for stand-alone MPAS-A applications.
> ```
> 
> For the PR message use the same verbiage, but change `commit` to `PR`

I think it would be nice if there were a way to include "MPAS-JEDI" in the one-line summary, which we'd see with, e.g., `git log --oneline`. How about something like "Add brtemp and cldmask fields to atmosphere Registry.xml file for MPAS-JEDI" for the first line?

Also, perhaps rather than "used by MPAS-JEDI", would it be fair to say "required by MPAS-JEDI"?

jim-p-w:
@mgduda I've done another update to the PR and commit verbiage .

mgduda (approved):
I can confirm that I see no increase in memory usage when running stand-alone MPAS-A with `config_jedi_da = false`.

PR #1234, opened 2024-09-17
Enable PIO to Add Attributes to Existing NetCDF Files

amstokely:
# Add Ability to Write New Attributes to Existing NetCDF Files in PIO

## Current Problem:
The Parallel I/O (PIO) system in MPAS currently lacks the ability to add new attributes to existing NetCDF files. This limitation arises because when PIO attempts to write a new attribute, the file remains in data mode, which does not allow for the addition of attributes. Instead, attributes can only be added when the file is in define mode.

Unfortunately, the public APIs of both PIO and NetCDF do not provide a way to inquire about a file’s current mode. Switching to define mode each time a new attribute is needed is expensive, as transitioning between data and define modes introduces significant I/O costs.

## Solution:
This PR implements a try-fail approach to improve performance by reducing unnecessary mode switches:

1. **Try-Fail Method**: The system first attempts to write the new attribute while in data mode. If it fails, the system switches to define mode, writes the attribute, and then returns to data mode. This minimizes the number of transitions to define mode, thus improving performance.
   
2. **New Generic Interface**: A new interface, `put_att_pio`, is introduced, which encapsulates the try-fail logic. This interface is designed to handle various data types and dimensions:
   - Supports scalar and 1D attributes.
   - Handles data types such as `int`, `real`, `double`, and `string`.

3. **Code Efficiency**: 
   - `select type` statements are used to avoid code duplication and bloat.
   - Extensive logging is added at major steps in the newly introduced logic, enhancing traceability and debugging capabilities.

## Impact:
This PR enables MPAS users to add new attributes to NetCDF files that were generated by earlier versions of MPAS when using the PIO platform, providing better backward compatibility and flexibility in managing metadata.

## Additional Notes:
- This change ensures that users can manage NetCDF files more efficiently without sacrificing performance by unnecessarily switching between file modes.
- The approach is designed to maintain backward compatibility while enhancing PIO's functionality within MPAS.


mgduda:
@amstokely Can you fix up the commit history? As of now it looks like there are 23 files changed.

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+      type(MPAS_IO_Handle_type), intent(inout) :: handle
+      integer :: pio_ierr
+
+      call mpas_log_write('Calling PIO_redef')
+      pio_ierr = PIO_redef(handle % pio_file)
+      if (pio_ierr /= PIO_noerr) then
+         io_global_err = pio_ierr
+         return
+      end if
+      call mpas_log_write('Successfully called PIO_redef')
```
mgduda, line 5051: Let's either remove or comment-out these logging statements.

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+   function handle_put_att_pio_redef(handle) result (pio_ierr)
+      implicit none
+      type(MPAS_IO_Handle_type), intent(inout) :: handle
+      integer :: pio_ierr
+
+      call mpas_log_write('Calling PIO_redef')
+      pio_ierr = PIO_redef(handle % pio_file)
+      if (pio_ierr /= PIO_noerr) then
+         io_global_err = pio_ierr
+         return
```
mgduda, line 5049: I don't think this return statement is needed (except to avoid printing out the log message below).

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+   function handle_put_att_pio_enddef(handle) result (pio_ierr)
+      implicit none
+      type(MPAS_IO_Handle_type), intent(inout) :: handle
+      integer :: pio_ierr
+
+      call mpas_log_write('Calling PIO_enddef')
+      pio_ierr = PIO_enddef(handle % pio_file)
+      if (pio_ierr /= PIO_noerr) then
+         io_global_err = pio_ierr
+         return
```
mgduda, line 5064: Similar to the comment earlier, if we remove the print statements, then this `return` statement can probably be removed as well.

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+      if (pio_ierr /= PIO_noerr) then
+         io_global_err = pio_ierr
+         if (present(ierr)) ierr = MPAS_IO_ERR_BACKEND
+
+         if (handle % preexisting_file .and. .not. handle % data_mode) then
+            if (handle_put_att_pio_redef(handle) /= PIO_noerr) return
+
+            select type(attValue)
+               type is (integer)
+                  call mpas_log_write('Calling PIO_put_att for integer attribute '//trim(attname))
```
mgduda, line 5104: Let's remove these logging statements as well.

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+            if (pio_ierr /= PIO_noerr) then
+               io_global_err = pio_ierr
+               return
+            end if
+
+            if (handle_put_att_pio_enddef(handle) /= PIO_noerr) return
+
+            if (present(ierr)) ierr = MPAS_IO_NOERR
+         end if
+         return
```
mgduda, line 5126: It doesn't look like this `return` statement is necessary.

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+      integer, intent(in) :: varid
+      character(len=*), intent(in) :: attName
+      class(*), dimension(:), intent(in) :: attValue
+      integer, optional :: ierr
+      integer :: pio_ierr
+      character(len=*), parameter :: log_message_prefix = 'Calling PIO_put_att for'
+
+      select type(attValue)
+         type is (integer)
+            call mpas_log_write(log_message_prefix//' integer 1D-array attribute '//trim(attname))
```
mgduda, line 5142: Let's remove these logging statements as well.

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+
+            if (pio_ierr /= PIO_noerr) then
+               io_global_err = pio_ierr
+               return
+            end if
+
+            if (handle_put_att_pio_enddef(handle) /= PIO_noerr) return
+            if (present(ierr)) ierr = MPAS_IO_NOERR
+         end if
+         return
```
mgduda, line 5178: We could remove this `return` statement.

src/framework/mpas_io.F
```diff
@@ -5733,6 +5880,14 @@ subroutine MPAS_io_put_att_real0d(handle, attName, attValue, fieldname, syncVal,
...
          return
       end if
+
+!      if (handle % preexisting_file) then
+!         pio_ierr = PIO_enddef(handle % pio_file)
+!         if (pio_ierr /= PIO_noerr) then
+!            io_global_err = pio_ierr
+!            return
+!         end if
+!      end if
```
mgduda, line 5890: Perhaps it's better to just remove this block of code rather than leaving it commented-out?

src/framework/mpas_io.F
```diff
@@ -5919,20 +6074,20 @@ subroutine MPAS_io_put_att_real1d(handle, attName, attValue, fieldname, syncVal,
          allocate(singleVal(size(attValueLocal)))
          singleVal(:) = real(attValueLocal(:),R4KIND)
 #ifdef MPAS_PIO_SUPPORT
-         pio_ierr = PIO_put_att(handle % pio_file, varid, attName, singleVal) 
+        pio_ierr = put_att_pio(handle, varid, attName, singleVal, ierr=ierr)
```
mgduda, line 6077: I think there may be another space of indentation that's needed here to align with other code.

src/framework/mpas_io.F
```diff
@@ -5950,6 +6105,9 @@ subroutine MPAS_io_put_att_real1d(handle, attName, attValue, fieldname, syncVal,
    end subroutine MPAS_io_put_att_real1d
 
 
+
+
+
```
mgduda, line 6110: It would be preferable to not add these three blank lines.

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
 
    end subroutine MPAS_io_get_att_real1d
 
+   function handle_put_att_pio_redef(handle) result (pio_ierr)
```
mgduda, line 5040: If these new routines are only used by PIO, can we enclose them in an `MPAS_PIO_SUPPORT` pre-processing directive?

src/framework/mpas_io.F
jim-p-w, line 5051: Should this function clear `handle%data_mode`? on success?

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+      type(MPAS_IO_Handle_type), intent(inout) :: handle
+      integer :: pio_ierr
+
+      call mpas_log_write('Calling PIO_enddef')
+      pio_ierr = PIO_enddef(handle % pio_file)
+      if (pio_ierr /= PIO_noerr) then
+         io_global_err = pio_ierr
+         return
+      end if
+      call mpas_log_write('Successfully called PIO_enddef')
```
jim-p-w, line 5066: should this function set `handle%data_mode` on success?

src/framework/mpas_io.F
```diff
@@ -5033,6 +5037,149 @@ subroutine MPAS_io_get_att_real1d(handle, attName, attValue, fieldname, precisio
...
+            call mpas_log_write(log_message_prefix//' text attribute '//trim(attname))
+            pio_ierr = PIO_put_att(handle % pio_file, varid, attName, attValue)
+      end select
+
+      if (pio_ierr /= PIO_noerr) then
+         io_global_err = pio_ierr
+         if (present(ierr)) ierr = MPAS_IO_ERR_BACKEND
+
+         if (handle % preexisting_file .and. .not. handle % data_mode) then
+            if (handle_put_att_pio_redef(handle) /= PIO_noerr) return
```
jim-p-w, line 5100: I think I don't understand what `data_mode` means. I assumed it meant the file is in data mode, in which case you can do data i/o, but can't create new attributes. If so, this logic seems incorrect, because if the file is not in data mode it must be in define mode, so setting the file state to define mode is redundant.

Furthermore, all of the call sites to `put_att_pio` check for `handle%data_mode` and return before doing any work if the file is in data mode. See lines 5376, 5549, and 5719.

What am I missing?

jim-p-w: It sounds like we think the value of `data_mode`may be incorrect (that is, it isn't set even though the file is in data mode. If that is the case, line 5099 should not check the value of `data_mode`.

amstokely: Yes this seems to be the cause of this problem. I'm looking into all the points you brought above.

src/framework/mpas_io.F
amstokely: There is a bug in how data_mode is being set I'm fairly certain, as this "fix" checks if the file is in define mode and then calls PIO_redef, which should return an error.

src/framework/mpas_io.F
amstokely: @jim-p-w I confirmed that this code only works when data mode is false. As you suspected, this means that there is a bug in the code responsible for setting data mode.

src/framework/mpas_io.F
amstokely: I removed the data mode dep in the retry logic so this shouldn't be necessary.

mgduda:
@amstokely I've un-resolved several comments that were marked as resolved but didn't appear to actually have been addressed. Can you take another look at these?

PR #1237, opened 2024-10-17
Enable GPU exection of atm_divergence_damping_3d via OpenACC

abishekg7:
This PR enables the GPU execution of atm_divergence_damping_3d subroutine.

Tested with a real and idealized test cases.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -212,6 +212,7 @@ module atm_time_integration
       real (kind=RKIND), dimension(:), pointer :: invAreaCell
       integer, dimension(:), pointer :: bdyMaskCell
       integer, dimension(:), pointer :: bdyMaskEdge
+      real (kind=RKIND), dimension(:), pointer :: specZoneMaskEdge
       real (kind=RKIND), dimension(:), pointer :: invDvEdge
       real (kind=RKIND), dimension(:), pointer :: dcEdge
       real (kind=RKIND), dimension(:), pointer :: invDcEdge
@@ -294,6 +295,9 @@ module atm_time_integration
 
       call mpas_pool_get_array(mesh, 'bdyMaskEdge', bdyMaskEdge)
       !$acc enter data copyin(bdyMaskEdge)
+
+      call mpas_pool_get_array(mesh, 'specZoneMaskEdge', specZoneMaskEdge)
+      !$acc enter data copyin(specZoneMaskEdge)
 
       call mpas_pool_get_array(mesh, 'invDvEdge', invDvEdge)
       !$acc enter data copyin(invDvEdge)
@@ -412,6 +416,7 @@ module atm_time_integration
       real (kind=RKIND), dimension(:), pointer :: invAreaCell
       integer, dimension(:), pointer :: bdyMaskCell
       integer, dimension(:), pointer :: bdyMaskEdge
+      real (kind=RKIND), dimension(:), pointer :: specZoneMaskEdge
       real (kind=RKIND), dimension(:), pointer :: invDvEdge
       real (kind=RKIND), dimension(:), pointer :: dcEdge
       real (kind=RKIND), dimension(:), pointer :: invDcEdge
@@ -494,6 +499,9 @@ module atm_time_integration
 
       call mpas_pool_get_array(mesh, 'bdyMaskEdge', bdyMaskEdge)
       !$acc exit data delete(bdyMaskEdge)
+
+      call mpas_pool_get_array(mesh, 'specZoneMaskEdge', specZoneMaskEdge)
+      !$acc exit data delete(specZoneMaskEdge)
 
       call mpas_pool_get_array(mesh, 'invDvEdge', invDvEdge)
       !$acc exit data delete(invDvEdge)
@@ -2696,8 +2704,10 @@ module atm_time_integration
       real (kind=RKIND), dimension(:), pointer :: specZoneMaskEdge
 
       integer, dimension(:,:), pointer :: cellsOnEdge
-      integer, pointer :: nCellsSolve
-      integer, pointer :: nVertLevels
+      integer, pointer :: nCellsSolve_ptr
+      integer, pointer :: nVertLevels_ptr
+      integer :: nCellsSolve
+      integer :: nVertLevels
 
       real (kind=RKIND) :: divCell1, divCell2, rdts, coef_divdamp
       integer :: cell1, cell2, iEdge, k
@@ -2710,8 +2720,8 @@ module atm_time_integration
       call mpas_pool_get_array(diag, 'rtheta_pp_old', rtheta_pp_old)
       call mpas_pool_get_array(diag, 'ru_p', ru_p)
 
-      call mpas_pool_get_dimension(mesh, 'nCellsSolve', nCellsSolve)
-      call mpas_pool_get_dimension(mesh, 'nVertLevels', nVertLevels)
+      call mpas_pool_get_dimension(mesh, 'nCellsSolve', nCellsSolve_ptr)
+      call mpas_pool_get_dimension(mesh, 'nVertLevels', nVertLevels_ptr)
 
       call mpas_pool_get_config(configs, 'config_smdiv', smdiv) 
       call mpas_pool_get_config(configs, 'config_len_disp', config_len_disp)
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -2599,6 +2599,13 @@ subroutine atm_divergence_damping_3d( state, diag, mesh, configs, dts, edgeStart
       rdts = 1.0_RKIND / dts
       coef_divdamp = 2.0_RKIND * smdiv * config_len_disp * rdts
 
+      MPAS_ACC_TIMER_START('atm_divergence_damping_3d [ACC_data_xfer]')
+      !$acc enter data copyin(ru_p, rtheta_pp, rtheta_pp_old, specZoneMaskEdge, &
+      !$acc         theta_m, cellsOnEdge) 
+      MPAS_ACC_TIMER_STOP('atm_divergence_damping_3d [ACC_data_xfer]')
+
+      !$acc parallel async
```
gdicker1, line 2607: `async` could be dangerous here, I don't believe `acc exit data copyout(ru_p)` will wait for this kernel to finish before beginning to copy.

I think it would be best to remove this `async` since we don't need it (yet) and it could create issues for the kernels that come next and depend on `ru_p`

gdicker1 (changes requested):
When I tested this today for the merged branch (develop+atmosphere/port_divergence_damping_3d) versus the base branch (develop), the results matched!

However, I have some comments to address:

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Will remove the `async`. thanks!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -2599,6 +2599,13 @@ subroutine atm_divergence_damping_3d( state, diag, mesh, configs, dts, edgeStart
       rdts = 1.0_RKIND / dts
       coef_divdamp = 2.0_RKIND * smdiv * config_len_disp * rdts
 
+      MPAS_ACC_TIMER_START('atm_divergence_damping_3d [ACC_data_xfer]')
+      !$acc enter data copyin(ru_p, rtheta_pp, rtheta_pp_old, specZoneMaskEdge, &
+      !$acc         theta_m, cellsOnEdge)
```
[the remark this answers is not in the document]
abishekg7: Is that the convention we've been following so far? I assumed the routine to be ported has localized data movement clauses. I might also need to rework my other PRs if this is the case. Thanks for pointing it out.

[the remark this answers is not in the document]
abishekg7: Just to follow up.. Talked to @mgduda, and got this cleared up. Will make changes to these invariant fields. Thanks!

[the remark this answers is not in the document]
gdicker1: > Is that the convention we've been following so far?

I've been following that pattern, it was set in #1176.

I've been doing things similarly to what may happen here - do all the data movement that is needed locally for the function being ported, and then separate out those invariant fields and handle them in the `atm_dynamics_{init,finalize}` routines. Typically these invariant fields are part of the `mesh` var_struct as described in the src/core_atmosphere/Registry.xml.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: I removed this async, but wanted to note that there's another one in `atm_advance_scalars_work`. Is that fine for now?

abishekg7:
I've addressed the review comments, and also checked the output restart file for bit-for-bit reproducibility.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: > I removed this async, but wanted to note that there's another one in atm_advance_scalars_work. Is that fine for now?

I think we're good to leave that `async` clause for now, it has a matching `wait` directive/clause later on.

gdicker1 (approved):
With the fixes and my testing[^1], this PR has my approval.

[^1]: Using the regional test case for a few timesteps and comparing the restart, diag, and history output files. The baseline and "feature" branch tests matched!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -2625,6 +2641,13 @@ subroutine atm_divergence_damping_3d( state, diag, mesh, configs, dts, edgeStart
             end do
          end if ! edges for block-owned cells
       end do ! end loop over edges
+      !$acc end parallel
+
+      MPAS_ACC_TIMER_START('atm_divergence_damping_3d [ACC_data_xfer]')
+      !$acc exit data copyout(ru_p) delete(rtheta_pp, rtheta_pp_old, theta_m)
+      !$acc end data
```
mgduda, line 2648: Is this `!$acc end data` clause needed? (That's an actual, not a rhetorical, question!)

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -2599,6 +2607,13 @@ subroutine atm_divergence_damping_3d( state, diag, mesh, configs, dts, edgeStart
       rdts = 1.0_RKIND / dts
       coef_divdamp = 2.0_RKIND * smdiv * config_len_disp * rdts
 
+      MPAS_ACC_TIMER_START('atm_divergence_damping_3d [ACC_data_xfer]')
+      !$acc data present(specZoneMaskEdge, cellsOnEdge)
+      !$acc enter data copyin(ru_p, rtheta_pp, rtheta_pp_old, theta_m)
+      MPAS_ACC_TIMER_STOP('atm_divergence_damping_3d [ACC_data_xfer]')
+
+      !$acc parallel
```
mgduda, line 2615: Rather than the `!$acc data present` directive above on line 2611, it might be simpler and more robust to just add a `default(present)` clause to this `parallel` directive.

@gdicker1 Does this align with your understanding of best practices?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -2625,6 +2641,13 @@ subroutine atm_divergence_damping_3d( state, diag, mesh, configs, dts, edgeStart
             end do
          end if ! edges for block-owned cells
       end do ! end loop over edges
+      !$acc end parallel
+
+      MPAS_ACC_TIMER_START('atm_divergence_damping_3d [ACC_data_xfer]')
+      !$acc exit data copyout(ru_p) delete(rtheta_pp, rtheta_pp_old, theta_m)
```
mgduda, line 2647: For the sake of clarity, I would suggest either having separate directives for `copyout` and `delete`, or at least splitting the `copyout` and `delete` clauses onto separate lines.

mgduda (changes requested):

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: It is needed because of the `acc data present`, which is currently written as a structured data region. Structured regions in Fortran OpenACC (`parallel` and `data`) require an `acc end xxx` statement.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: I think the best practices would be neutral on this, whatever works best for our code and development. However, I agree that `default(present)` could be simpler and require less of us.

We would have to ask Pranay or other MPAS v7 OpenACC developers why they have similar function-level `acc data present ...` directives in their port. *Speculation here*: I think enclosing the kernels in a structured data region that ensures all data is present prevents the kernels from each checking their data - an optimization concern.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
mgduda: @gdicker1 Thanks for that explanation! The `!$acc data present` directive didn't register when I was scanning through the code changes.

@abishekg7 Perhaps it would be better to follow the pattern in other routines that have been ported, and to avoid the use of structured data directives for now?

mgduda: If we decide to add `default(present)` clauses to the `!$acc parallel` directives, that would obviate the need for the `!$acc data present(specZoneMaskEdge, cellsOnEdge)` directive I would think.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: done

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: I'm adding the `default(present)` for now, but noting the potential concern re. performance.

mgduda:
@abishekg7 The code changes look good to me! I'd only suggest reworking the commit history; then I think we'll be all set.

abishekg7:
Cleaned up the commit history. 1 commit seemed sufficient. Also, tested the rebased version in limited area mode and restart files are still bit identical.

mgduda (approved):

PR #1238, opened 2024-10-17
Enable GPU execution of atm_compute_moist_coefficients via OpenACC

abishekg7:
This PR enables the GPU execution of atm_compute_moist_coefficients subroutine.

Tested with a real and idealized test cases.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -1795,52 +1795,79 @@ module atm_time_integration
       integer, intent(in) :: cellSolveStart, cellSolveEnd, vertexSolveStart, vertexSolveEnd, edgeSolveStart, edgeSolveEnd
 
       integer :: iEdge, iCell, k, cell1, cell2, iq
-      integer, pointer :: nCells, nEdges, nVertLevels, nCellsSolve
+      integer, pointer :: nCells_ptr, nEdges_ptr, nVertLevels_ptr, nCellsSolve_ptr
+      integer :: nCells, nEdges, nVertLevels, nCellsSolve
       real (kind=RKIND) :: qtotal
       integer, dimension(:,:), pointer :: cellsOnEdge
-      integer, pointer :: moist_start, moist_end
+      integer, pointer :: moist_start_ptr, moist_end_ptr
+      integer :: moist_start, moist_end
       real (kind=RKIND), dimension(:,:,:), pointer :: scalars
       real (kind=RKIND), dimension(:,:), pointer :: cqw
       real (kind=RKIND), dimension(:,:), pointer :: cqu
 
-      call mpas_pool_get_dimension(dims, 'nCells', nCells)
-      call mpas_pool_get_dimension(dims, 'nEdges', nEdges)
-      call mpas_pool_get_dimension(dims, 'nVertLevels', nVertLevels)
-      call mpas_pool_get_dimension(dims, 'nCellsSolve', nCellsSolve)
-      call mpas_pool_get_dimension(state, 'moist_start', moist_start)
-      call mpas_pool_get_dimension(state, 'moist_end', moist_end)
+      call mpas_pool_get_dimension(dims, 'nCells', nCells_ptr)
+      call mpas_pool_get_dimension(dims, 'nEdges', nEdges_ptr)
+      call mpas_pool_get_dimension(dims, 'nVertLevels', nVertLevels_ptr)
+      call mpas_pool_get_dimension(dims, 'nCellsSolve', nCellsSolve_ptr)
+      call mpas_pool_get_dimension(state, 'moist_start', moist_start_ptr)
+      call mpas_pool_get_dimension(state, 'moist_end', moist_end_ptr)
 
       call mpas_pool_get_array(mesh, 'cellsOnEdge', cellsOnEdge)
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1730,39 +1730,60 @@ subroutine atm_compute_moist_coefficients( dims, state, diag, mesh, &
       call mpas_pool_get_array(diag, 'cqw', cqw)
       call mpas_pool_get_array(diag, 'cqu', cqu)
 
+      MPAS_ACC_TIMER_START('atm_compute_moist_coefficients [ACC_data_xfer]')
+      !$acc enter data copyin(qtot, cqw, cqu, scalars)
+      MPAS_ACC_TIMER_STOP('atm_compute_moist_coefficients [ACC_data_xfer]')
+
+
+      !$acc parallel
+      !$acc loop
```
gdicker1, line 1739: I think the loop directives should be more explicit (e.g. `loop gang worker` here). Just to match with other work we're doing.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1730,39 +1730,60 @@ subroutine atm_compute_moist_coefficients( dims, state, diag, mesh, &
...
+               !$acc loop seq
                do iq = moist_start, moist_end
                   qtotal = qtotal + 0.5 * ( scalars(iq, k, cell1) + scalars(iq, k, cell2) )
                end do
                cqu(k,iEdge) = 1.0 / (1.0 + qtotal)
             end do
          end if
       end do
+      !$acc end parallel
+      !$acc exit data copyout(cqw, cqu, qtot) delete(scalars)
```
gdicker1, line 1786: This `exit data` directive should be surrounded by `MPAS_ACC_TIMER_{START,STOP}` calls.

gdicker1 (changes requested):
This may need some additional work. When I compared baseline and test runs of the regional testcase, I get differences in two fields in the history file. Restart files are reported as matching.

Output from compare_netcdf.py:
```
Comparing files: f1 - f2                                                                                                                             
        f1=../2024Dec31_MPASAPR1223_OpenACC/baseline_acc_1735844488/history.2019-09-01_00.06.00.nc                                                   
        f2=test_att1735694999/history.2019-09-01_00.06.00.nc                                                                                         

            Variable  Min       Max       
```

Also a couple of suggested edits.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
mgduda: As I commented on another PR, I'd also advocate for at least splitting the `copyout` and `delete` clauses onto separate lines, or better yet, to turn them into separate directives. I think this will make it easier when it comes to shifting around and fusing data directives later, and at least for me, it just makes it easier to visually recognize that there are two things happening in this `exit data` directive.

abishekg7:
Modified the two parallel constructs to explicitly specify the level of parallelism. This also seems to lead to a performance increase for the limited area test by about 10% in the parallel clause on L1754 and about 51% on L1767. Also verified that results are still bit reproducible.

Added timers around the exit data clause and made the other suggested formatting changes. 

The differences in the diagnostic fields in the history files are expected per our previous discussion.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1710,59 +1710,91 @@ subroutine atm_compute_moist_coefficients( dims, state, diag, mesh, &
...
 
+      nCells = nCells_ptr
+      nEdges = nEdges_ptr
+      nVertLevels = nVertLevels_ptr
+      nCellsSolve = nCellsSolve_ptr
+      moist_start = moist_start_ptr
+      moist_end = moist_end_ptr
+
+      MPAS_ACC_TIMER_START('atm_compute_moist_coefficients [ACC_data_xfer]')
+      !$acc enter data copyin(qtot, cqw, cqu, scalars)
```
mgduda, line 1743: If I'm not mistaken, we only need to `create`, rather than `copyin`, the variables `qtot`, `cqw`, and `cqu`.

mgduda (changes requested):

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
mgduda: The line that you referenced only initializes the garbage cell for `qtot`. Setting aside our recent experience with the garbage cell for `theta_m`, I don't think the specific value in the garbage cell for `qtot` should matter, in which case `create` should be fine in the `!$acc data` directive.

abishekg7: Got it. I changed it the three variables to `create` now, and the results are still bit identical.

mgduda:
@abishekg7 The code changes themselves look good to me now. I'd only request that the commit history be cleaned up, and perhaps just a single commit would suffice.

abishekg7:
Rebased this PR and ran a quick check for bit reproducibility again. thanks for the review!

mgduda (approved):

mgduda:
> Rebased this PR and ran a quick check for bit reproducibility again. thanks for the review!

Nice work!

gdicker1 (approved):
This looks good to me!

PR #1240, opened 2024-10-23
correct the computation of 2-meter diagnostics with Noah-MP

ldfowler58:
This PR corrects the calculation of the 2-meter diagnostics (T2M, TH2M, and Q2) when using the Noah-MP land surface scheme. While the computation of 2-meter diagnostics is the same for Noah and Noah-MP over oceans, it is different between the two land surface schemes over land. In Noah-MP, the 2-meter diagnostics are weighted as functions of their respective diagnostics over bare soil and over vegetation. The updated diagnostics for Noah and Noah-MP are now computed in the new file mpas_atmphys_sfc_diagnostics.F.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_boundaries.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_boundaries.F
@@ -36,6 +36,11 @@ module mpas_atm_boundaries
 
     public :: nBdyZone, nSpecZone, nRelaxZone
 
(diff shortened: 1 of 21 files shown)


mgduda:
@ldfowler58 It looks like your branch may be based on the `develop` branch. Can you base your changes off of the v8.2.0 tag to avoid pulling in changes from the `develop` branch?

PR #1241, opened 2024-10-23
Enable GPU execution of atm_compute_vert_imp_coefs via OpenACC

abishekg7:
Initial OpenACC port of atm_compute_vert_imp_coefs.

  - Separating the computation of `cofrz` into a separate parallel  clause is required for correctness.
  - Variables `b_tri` and `c_tri` are declared private in lieu of extending the storage to two dimensions, like `a_tri`.
  - Back substitution/Computation of `alpha_tri` and `gamma_tri` needs to be sequential

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -229,6 +229,8 @@ module atm_time_integration
       real (kind=RKIND), dimension(:), pointer :: fEdge
       real (kind=RKIND), dimension(:), pointer :: fVertex
       real (kind=RKIND), dimension(:,:), pointer :: zz
+      real (kind=RKIND), dimension(:), pointer :: rdzw
+      real (kind=RKIND), dimension(:), pointer :: rdzu
       real (kind=RKIND), dimension(:,:,:), pointer :: zb_cell
       real (kind=RKIND), dimension(:,:,:), pointer :: zb3_cell
       real (kind=RKIND), dimension(:), pointer :: fzm
@@ -344,6 +346,12 @@ module atm_time_integration
       call mpas_pool_get_array(mesh, 'zz', zz)
       !$acc enter data copyin(zz)
 
+      call mpas_pool_get_array(mesh, 'rdzw', rdzw)
+      !$acc enter data copyin(rdzw)
+
+      call mpas_pool_get_array(mesh, 'rdzu', rdzu)
+      !$acc enter data copyin(rdzu)
+
       call mpas_pool_get_array(mesh, 'zb_cell', zb_cell)
       !$acc enter data copyin(zb_cell)
 
@@ -421,6 +429,8 @@ module atm_time_integration
       real (kind=RKIND), dimension(:), pointer :: fEdge
       real (kind=RKIND), dimension(:), pointer :: fVertex
       real (kind=RKIND), dimension(:,:), pointer :: zz
+      real (kind=RKIND), dimension(:), pointer :: rdzw
+      real (kind=RKIND), dimension(:), pointer :: rdzu
       real (kind=RKIND), dimension(:,:,:), pointer :: zb_cell
       real (kind=RKIND), dimension(:,:,:), pointer :: zb3_cell
       real (kind=RKIND), dimension(:), pointer :: fzm
@@ -535,6 +545,12 @@ module atm_time_integration
 
       call mpas_pool_get_array(mesh, 'zz', zz)
       !$acc exit data delete(zz)
+
+      call mpas_pool_get_array(mesh, 'rdzw', rdzw)
+      !$acc exit data delete(rdzw)
+
+      call mpas_pool_get_array(mesh, 'rdzu', rdzu)
+      !$acc exit data delete(rdzu)
 
       call mpas_pool_get_array(mesh, 'zb_cell', zb_cell)
       !$acc exit data delete(zb_cell)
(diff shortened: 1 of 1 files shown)


mgduda:
@abishekg7 Can you clean up the commits and add a description to this PR?

abishekg7:
> @abishekg7 Can you clean up the commits and add a description to this PR?

Thanks. Done!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1894,32 +1894,46 @@ subroutine atm_compute_vert_imp_coefs_work(nCells, moist_start, moist_end, dts,
       real (kind=RKIND) :: dtseps, c2, qtotal, rcv
       real (kind=RKIND), dimension( nVertLevels ) :: b_tri, c_tri
 
+      MPAS_ACC_TIMER_START('atm_compute_vert_imp_coefs_work [ACC_data_xfer]')
+      !$acc enter data copyin(rdzw, zz, fzm, fzp, rdzu, cqw, p, t, qtot, rb, &
+      !$acc                   rtb, rt, pb)
```
gdicker1, line 1899: Some of the fields here are invariant (across timesteps) and should be handled in the `mpas_atm_dynamics_{init,finalize}` routines instead.

Generally these invariant fields are those contained in the "invariant stream" ([L421-L524 in the atm Registry.xml](https://github.com/MPAS-Dev/MPAS-Model/blob/41e9a3fb8ca6b9250a7405209a5c60996318409f/src/core_atmosphere/Registry.xml#L421-L524)) or those accessed through the "mesh" `mpas_pool` ([L1222-L1516 in the atm Registry.xml](https://github.com/MPAS-Dev/MPAS-Model/blob/41e9a3fb8ca6b9250a7405209a5c60996318409f/src/core_atmosphere/Registry.xml#L1222-L1516))

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1894,32 +1894,46 @@ subroutine atm_compute_vert_imp_coefs_work(nCells, moist_start, moist_end, dts,
...
+      !$acc                   c_tri, alpha_tri, gamma_tri)
+      MPAS_ACC_TIMER_STOP('atm_compute_vert_imp_coefs_work [ACC_data_xfer]')
 
       !  set coefficients
       dtseps = .5*dts*(1.+epssm)
       rcv = rgas/(cp-rgas)
       c2 = cp*rcv
 
+      !$acc parallel
+      !$acc loop
```
gdicker1, line 1910: I think it would be better to be explicit with `loop` directives (e.g. `loop gang worker` here).

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1955,12 +1970,18 @@ subroutine atm_compute_vert_imp_coefs_work(nCells, moist_start, moist_end, dts,
...
             alpha_tri(k,iCell) = 1./(b_tri(k)-a_tri(k,iCell)*gamma_tri(k-1,iCell))
             gamma_tri(k,iCell) = c_tri(k)*alpha_tri(k,iCell)
          end do
 
       end do ! loop over cells
+      !$acc end parallel 
+
+      !$acc exit data copyout(cofrz, cofwr, cofwz, coftz, cofwt, a_tri, b_tri, &
+      !$acc                   c_tri, alpha_tri, gamma_tri)
+      !$acc exit data delete(rdzw, zz, fzm, fzp, rdzu, cqw, p, t, qtot, rb, rtb, rt, pb)
```
gdicker1, line 1984: Besides handling fields like `fzm` in `mpas_atm_dynamics_finalize` instead, these directives should be surrounded by `MPAS_ACC_TIMER_{START,STOP}` calls.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1894,32 +1894,46 @@ subroutine atm_compute_vert_imp_coefs_work(nCells, moist_start, moist_end, dts,
...
 
+      !$acc parallel
+      !$acc loop
 ! MGD bad to have all threads setting this variable?
       do k=1,nVertLevels
          cofrz(k) = dtseps*rdzw(k)
       end do
-
+      !$acc end parallel 
+
```
gdicker1, line 1916: This empty line contains extra spaces that should be removed. Might be good to double check for any trailing whitespace in the lines you're adding.

gdicker1 (changes requested):
This PR may need a little more work. I'm getting differences in 2 fields in the history file when comparing the regional testcase. Restart files are reported as matching.

Output from compare_netcdf.py:
```
Comparing files: f1 - f2                                                                                                                             
        f1=../2024Dec31_MPASAPR1223_OpenACC/baseline_acc_1735844488/history.2019-09-01_00.06.00.nc                                                   
        f2=test_att1735695005/history.2019-09-01_00.06.00.nc                                                                                         
                                                                                                                                                     
            Variable  Min       Max                                                                                                                  
```

Some suggested edits as well.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Done. thanks for the catch!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Done.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Done. thanks!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -506,6 +531,21 @@ subroutine mpas_atm_dynamics_finalize(domain)
 
       call mpas_pool_get_array(mesh, 'fEdge', fEdge)
       !$acc exit data delete(fEdge)
+      
+      call mpas_pool_get_array(mesh, 'rdzw', rdzw)
+      !$acc enter data copyin(rdzw)
```
abishekg7, line 536: Caught an issue here. It should be `exit data delete`, not `enter data copyin`

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Done. No performance penalties for this change, with the limited area case at least.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1979,32 +1995,45 @@ subroutine atm_compute_vert_imp_coefs_work(nCells, moist_start, moist_end, dts,
...
+      !$acc enter data create(cofrz, cofwr, cofwz, coftz, cofwt, a_tri, b_tri, &
+      !$acc                   c_tri, alpha_tri, gamma_tri)
+      MPAS_ACC_TIMER_STOP('atm_compute_vert_imp_coefs_work [ACC_data_xfer]')
 
       !  set coefficients
       dtseps = .5*dts*(1.+epssm)
       rcv = rgas/(cp-rgas)
       c2 = cp*rcv
 
+      !$acc parallel
```
mgduda, line 2009: As with other routines, should we add `default(present)` clauses to the `parallel` directives in this routine as well?

abishekg7: fixed, and thanks!

mgduda:
@abishekg7 The code changes themselves look good to me now. I'd only request that the commit history be cleaned up. If you want to keep the handling of the time-invariant fields in a separate commit, that seems fine, but just a single commit for this PR seems sufficient.

abishekg7:
Rebased and squashed to a single commit, and results are still bit identical.

gdicker1:
@abishekg7, I think everything is looking good except this sentence in the commit message

> - The parallel clauses now specify a default(present)

This seems like an incomplete thought, like there should be more after "...sent)". Maybe as short as:

> - The parallel regions use default(present) clauses.

With something about "to enforce data management" potentially being added too.

Aside: As I understand OpenACC terminology, _directives_ refer to the part of the statement right after `acc`. E.g. "parallel", "kernels", "host_data", "data", and "exit data" . If the directive applies to a region of code, the directive and the region together are a _construct_. _Clauses_ are the other statements that come after the directive (like "collapse(n)" or "async").

abishekg7:
> @abishekg7, I think everything is looking good except this sentence in the commit message
> 
> > * The parallel clauses now specify a default(present)
> 
> This seems like an incomplete thought, like there should be more after "...sent)". You could drop the "a" or replace the sentence with something like:
> 
> > * The parallel constructs use default(present) clauses.
> 
> With something about "to enforce data management" potentially being added too.
> 
> Aside: As I understand OpenACC terminology, _directives_ refer to the part of the statement right after `acc`. E.g. "parallel", "kernels", "host_data", "data", and "exit data" . If the directive applies to a region of code, the directive and the region together are a _construct_. _Clauses_ are the other parts of a statement that come after the directive (like "collapse(n)" or "async").

Thanks @gdicker1! I've been using clauses, directives and constructs interchangeably and it's good to know. Will fix this.

abishekg7:
Fixed the default(present) part of the commit message.

gdicker1 (approved):
Looks good to me now!

mgduda (approved):
Looks good to me!

PR #1242, opened 2024-10-24
correct the computation of 2-meter diagnostics with Noah-MP #1240

ldfowler58:
This PR corrects the calculation of the 2-meter diagnostics (T2M, TH2M, and Q2) when using the Noah-MP land surface scheme. While the computation of 2-meter diagnostics is the same for Noah and Noah-MP over oceans, it is different between the two land surface schemes over land. In Noah-MP, the 2-meter diagnostics are weighted as functions of their respective diagnostics over bare soil and over vegetation. The updated diagnostics for Noah and Noah-MP are now computed in the new file mpas_atmphys_sfc_diagnostics.F.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_lsm_noahmp.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_lsm_noahmp.F
@@ -84,11 +84,11 @@
 
 
 !--- local OUT pointers (with no Noah LSM equivalent as defined in WRF):
- real(kind=RKIND),dimension(:),pointer:: t2mvxy,t2mbxy,q2mvxy,q2mbxy,tradxy,neexy,gppxy,nppxy,fvegxy,runsfxy,  &
-                                         runsbxy,ecanxy,edirxy,etranxy,fsaxy,firaxy,aparxy,psnxy,savxy,sagxy,  &
-                                         rssunxy,rsshaxy,bgapxy,wgapxy,tgvxy,tgbxy,chvxy,chbxy,shgxy,shcxy,    &
-                                         shbxy,evgxy,evbxy,ghvxy,ghbxy,irgxy,ircxy,irbxy,trxy,evcxy,chleafxy,  &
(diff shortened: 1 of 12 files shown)


src/core_atmosphere/physics/mpas_atmphys_driver_lsm_noahmp.F
```diff
@@ -702,7 +706,7 @@ subroutine lsm_noahmp_toMPAS(diag_physics,diag_physics_noahmp,output_noahmp,sfc_
  real(kind=RKIND),dimension(:,:),pointer:: acc_etranixy
 
 !-----------------------------------------------------------------------------------------------------------------
-!call mpas_log_write('--- enter subroutine lsm_noahmp_toMPAS:')
+ call mpas_log_write('--- enter subroutine lsm_noahmp_toMPAS:')
```
mgduda, line 709: Could we keep this commented-out?

src/core_atmosphere/physics/mpas_atmphys_driver_lsm_noahmp.F
```diff
@@ -377,8 +377,10 @@ subroutine lsm_noahmp_fromMPAS(configs,mesh,diag,diag_physics,diag_physics_noahm
 !    see lines 242-290 in module NoahmpIOVarType.F90):
  call mpas_pool_get_array(output_noahmp,'t2mvxy'  ,t2mvxy  )
  call mpas_pool_get_array(output_noahmp,'t2mbxy'  ,t2mbxy  )
+ call mpas_pool_get_array(output_noahmp,'t2mxy'   ,t2mxy   )
```
mgduda, line 380: Is there a reason for retrieving a pointer to `t2mxy` but not using it in this routine? And similarly for `q2mxy`?

mgduda (changes requested):

ldfowler58:
Added minor changes to ./src/core_atmosphere/physics/mpas_atmphys_driver_lsm_noahmp.F, as suggested.

mgduda (approved):

PR #1244, opened 2024-11-01
Correct initialization of soil liquid water content for Noah-MP

ldfowler58:
This PR corrects the computation of the soil temperature (TSLB) in the Noah-MP land surface scheme. In this PR, we added the initialization of the soil liquid water (SH2O) in subroutine noahmp_init in module mpas_atmphys_lsm_noahmpinit.F prior to calling subroutine NoahmpInitMain.

Prior to adding the initialization of SH2O, running Noah-MP led TSLB to be set to ConstFreezePoint (273.16 K) in SoilSnowWaterPhaseChangeMod.F90 although the input TSLB was greater than ConstFreezePoint on the very first time step. This error occurred because TSLB is calculated as a function of the soil liquid (MassWatLiqTmp) and ice (MassWatIceTmp) water mass, and MassWatLiqTmp is always equal to 0 when not properly initializing SH2O.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F
@@ -10,6 +10,7 @@
  use mpas_log
  use mpas_pool_routines
 
+ use mpas_atmphys_constants,only: grav => gravity, t0 => svpt0
  use mpas_atmphys_utilities,only: physics_error_fatal
  use mpas_atmphys_vars,only     : mpas_noahmp
 
@@ -253,7 +254,11 @@
 !local variables and pointers:
  logical,pointer:: do_restart
  logical,parameter:: fndsnowh = .true.
+
  integer:: i,its,ite,ns,nsoil,nsnow,nzsnow
+
+ real(kind=RKIND),parameter:: hlice = 3.335E5
+ real(kind=RKIND):: bexp,fk,smcmax,psisat
 
 !-----------------------------------------------------------------------------------------------------------------
 !call mpas_log_write(' ')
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F
```diff
@@ -390,6 +395,40 @@ subroutine noahmp_init(configs,mesh,diag_physics,diag_physics_noahmp,output_noah
...
+
+       do ns = 1,nsoil
+          if(smois(ns,i) > smcmax) smois(ns,i) = smcmax
+       enddo
+       if(bexp.gt.0. .and. smcmax.gt.0. .and. psisat.gt.0.) then
+          do ns = 1,nsoil
+             if(tslb(ns,i) .lt. 273.149) then
+                fk = ( ((hlice/(grav*(-psisat)))*((tslb(ns,i)-t0)/tslb(ns,i)))**(-1/bexp) )*smcmax
+                fk = max(fk,0.02)
+                sh2o(ns,i) = min(fk,smois(n,i))
```
mgduda, line 419: Should `smois(n,i)` be `smois(ns,i)`?

mgduda (changes requested):
I think it would be very helpful to add some comments to the code to explain what each branch is doing. For example, which cells are affected by this condition
```fortran
       if(ivgtyp(i) == mpas_noahmp%isice_table .and. xice(i) .le. 0._RKIND) then
```
and which are affected by this condition
```fortran
       if(bexp.gt.0. .and. smcmax.gt.0. .and. psisat.gt.0.) then
```
?
It could also help to give names (and explanatory comments) for the constants used in the code (`263.15`, `273.149`, `0.02`).

ldfowler58:
I corrected the index n to ns in the computation of smois. I added a couple of comments to the initialization, as in the non-refactored version of Noah-MP (also refer to module_sf_noahmpdrv.F in WRF/phys).

mgduda (approved):

CharlesZheZhang:
Hi, dear developers, I am currently testing MPAS-NoahMP and find sh2o=0 for all the land points in init.nc after running init_atmosphere. This is related to sh2o=0 for all land points, but smois field is good. See attached figures. 
sh2o_smois_init nc

This may be related to setting sh2o to 0 for all land points in mpas_atmphys_initialize_real.F https://github.com/ldfowler58/MPAS-Model/blob/91bc0882689b3e60803e5df90636d7713e62cc2c/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F#L473

So it needs an initial value  for sh2o from sw_input, similar for smois from sm_input, also similar to real in WRF: https://github.com/wrf-model/WRF/blob/d66e442fccc04111067e29274c9f9eaccc3cef28/share/module_soil_pre.F#L1689

So, if sh2o is correctly outputed to init.nc and read in mpas_atmphys_lsm_noahmpinit.F, 
this change starting from L398  https://github.com/MPAS-Dev/MPAS-Model/blob/283c77fbd05a0d38047eeca2991355fac67301df/src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F#L398
may not be necessary - before calling NoahmpInitMain

@cenlinhe

cenlinhe:
I agree with @CharlesZheZhang , the previous proposed soil initialization bug fix is not necessary because it is included in NoahmpInitMain module. The root cause for the issue is the wrong initialization in the real step to generate the SH2O in init.nc file. This must be fixed as soon as possible. Otherwise, the model results would be wrong.

cenlinhe:
I will double check this.

ldfowler58:
Thanks for finding that "bug". I will work on it as soon as I can. We have a couple of bugs that we fixed and need to make available in the next MPAS bug fix release.

CharlesZheZhang:
Following on this PR: 
Priori to adding this commit, the mpas_noahmp%sh2o field is correctly initialized in calling NoahmpInitMain:
https://github.com/MPAS-Dev/MPAS-Model/blob/41e9a3fb8ca6b9250a7405209a5c60996318409f/src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F#L410

But they are not correctly assign to sh2o, tslb, smois in these lines:
https://github.com/MPAS-Dev/MPAS-Model/blob/41e9a3fb8ca6b9250a7405209a5c60996318409f/src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F#L422

I had some print out of the sh2o, tslb, smois fields before and after calling NoahmpInitMain and it shows the mpas_noahmp%sh2o is initialized after the call, but is not assigned to sh2o. 
image

So it seems that the additions in this commit are redundant (L398-433), but the changes from L463-465 are critical!

PR #1245, opened 2024-11-06
Removing instances of non-standard kind types in init_atmosphere

abishekg7:
Building MPAS CORE=init_atmosphere with the NAG compiler led to the discovery of a few instances of non-standard kind types being used in the init_atmosphere core, specifically in the files `mpas_init_atm_read_met.F` and `mpas_init_atm_llxy.F`.

For future reference, the build with the NAG compiler (on Izumi) failed with error messages of the kind
```
Error: src/core_init_atmosphere/mpas_init_atm_llxy.F, line 1682:
   KIND value (8) does not specify a valid representation method
```

This PR removes replaces the non-standard kind type usages with `real32` and `real64` from `iso_fortran_env` to resolve these errors.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_llxy.F
+++ b/src/core_init_atmosphere/mpas_init_atm_llxy.F
@@ -140,6 +140,7 @@ MODULE init_atm_llxy
    USE MPAS_DERIVED_TYPES, ONLY : MPAS_LOG_ERR, MPAS_LOG_CRIT
    USE MPAS_KIND_TYPES
    USE MPAS_LOG, ONLY : MPAS_LOG_WRITE
+   USE ISO_FORTRAN_ENV, ONLY: REAL64
 
    INTEGER, PARAMETER :: HH=4, VV=5
 
@@ -171,8 +172,6 @@ MODULE init_atm_llxy
    INTEGER, PUBLIC, PARAMETER  :: PROJ_ALBERS_NAD83 = 105
    INTEGER, PUBLIC, PARAMETER  :: PROJ_ROTLL = 203
 
-   ! Define some private constants
-   INTEGER, PRIVATE, PARAMETER :: HIGH = 8
  
    TYPE proj_info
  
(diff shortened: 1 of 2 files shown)


mgduda:
@abishekg7 I'd suggest we split the changes into two commits -- one commit for each file that is being modified. I think there's also an opportunity to reframe the description of the changes: rather than trying to fix compilation with NAG, we're addressing occurrences of nonstandard kind types. It's probably also helpful in each commit message to quote the compiler error message that led us to make the code changes; see, e.g., f7ceee61.

abishekg7:
@mgduda I've separated the changes to the two files into two separate PRs.

mgduda:
Apologies for not stating this better in my previous comment. I think a single PR (i.e., this one) with two commits would be fine. We can also modify the title and description of this PR to describe the general issue that we're addressing here -- the use of non-standard kind types in the `init_atmosphere` core. These issues with our code happen to have been caught by the NAG compiler, but I think it's important that the issue is framed as correcting a fundamental code issue rather than simply fixing a compilation failure with a specific compiler.

abishekg7:
Ah that makes more sense. I think your original comment was clear, in hindsight, but I just misunderstood it. Will fix that and restate the commit messages. Thanks!

mgduda:
@abishekg7 I have just two minor requests regarding the commit messages:
1. Right now, the first lines of both of the commits in this PR branch are identical, and in my opinion this isn't optimal. For example, if we were looking at history with `git log --oneline`, we'd see: 
```
2dd6f712d Removing use of non-standard kind types in init_atmosphere
b614c79fe Removing use of non-standard kind types in init_atmosphere
```
In my opinion, it's preferable to use unique one-line summaries that convey as much information as possible while keeping to 80 characters.

3. In the commit messages, it's written "This PR". I'd argue that a commit is not a PR, and the text should instead say "This commit".

abishekg7:
@mgduda  Thanks and done! The only way I could keep the first line within 80 character was to only just the source file name `mpas_init_atm_llxy.F` and not the full qualifying path `core_init_atmosphere/mpas_init_atm_llxy.F`

mgduda (approved):

PR #1250, opened 2024-12-05
added DO_PHYSICS preprocessing option in Registry.xml

ldfowler58:
added the pre-processing option DO_PHYSICS in Registry.xml to compile MPAS-Model without physics in 1) the da_state stream; and 2) to not include the Noah-MP Registry when PHYSICS = " " in ./src/core_atmosphere/Makefile.


mgduda:
It looks like these changes were already introduced by PR #1221.

PR #1254, opened 2024-12-19
Allow using the SMIOL I/O library when building with cmake.

jim-p-w:
This PR removes the restriction of only using the PIO I/O library when building with cmake.

The default behavior will build using PIO.
In order to build and use the SMIOL I/O library provide `-DMPAS_USE_SMIOL=1` on the command line when running cmake.

Test run include the mpasjedi ctests, as well as the MPAS-Workflow cylc tests:
3dvar_OIE120km_WarmStart_TEST/
3dvar_OIE120km_ColdStart_TEST/
3dvar_O30kmIE60km_ColdStart_TEST/
3denvar_O30kmIE60km_WarmStart_TEST/
4denvar_OIE120km_WarmStart_TEST/
4dhybrid_OIE120km_WarmStart_TEST/
eda_OIE120km_WarmStart_TEST/
getkf_OIE120km_WarmStart_TEST/
ForecastFromGFSAnalysesMPT_TEST/
3denvar_OIE120km_IAU_WarmStart_TEST/
  - this test fail with:
  ```
ERROR: SMIOLf_open_file failed with error -5
ERROR: NetCDF: Attempt to use feature that was not turned on when netCDF was built.
```
because the test tries to read an HDF5 data file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,12 +21,14 @@ option(MPAS_DOUBLE_PRECISION "Use double precision 64-bit Floating point." TRUE)
 option(MPAS_PROFILE "Enable GPTL profiling" OFF)
 option(MPAS_OPENMP "Enable OpenMP" OFF)
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+option(MPAS_USE_SMIOL "Build with smiol I/O library" OFF)
 
 message(STATUS "[OPTION] MPAS_CORES: ${MPAS_CORES}")
 message(STATUS "[OPTION] MPAS_DOUBLE_PRECISION: ${MPAS_DOUBLE_PRECISION}")
 message(STATUS "[OPTION] MPAS_PROFILE: ${MPAS_PROFILE}")
 message(STATUS "[OPTION] MPAS_OPENMP: ${MPAS_OPENMP}")
 message(STATUS "[OPTION] BUILD_SHARED_LIBS: ${BUILD_SHARED_LIBS}")
+message(STATUS "[OPTION] MPAS_USE_SMIOL: ${MPAS_USE_SMIOL}")
 
 # Build product output locations
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
@@ -59,7 +61,9 @@ find_package(OpenMP COMPONENTS Fortran)
 find_package(MPI REQUIRED COMPONENTS Fortran)
 find_package(NetCDF REQUIRED COMPONENTS Fortran C)
 find_package(PnetCDF REQUIRED COMPONENTS Fortran)
-find_package(PIO REQUIRED COMPONENTS Fortran C)
+if(NOT MPAS_USE_SMIOL)
+    find_package(PIO REQUIRED COMPONENTS Fortran C)
+endif()
 if(MPAS_PROFILE)
     find_package(GPTL REQUIRED)
 endif()
@@ -88,6 +92,9 @@ set(MPAS_SUBDRIVER_SRC  ${CMAKE_CURRENT_SOURCE_DIR}/src/driver/mpas_subdriver.F)
 
 ## Create targets
 add_subdirectory(src/external/ezxml) # Target: MPAS::external::ezxml
+if(MPAS_USE_SMIOL)
+    add_subdirectory(src/external/SMIOL) # Target: MPAS::external::smiol
+endif()
 if(ESMF_FOUND)
   message(STATUS "Configure MPAS for external ESMF")
   add_definitions(-DMPAS_EXTERNAL_ESMF_LIB -DMPAS_NO_ESMF_INIT)
--- a/cmake/Functions/MPAS_Functions.cmake
+++ b/cmake/Functions/MPAS_Functions.cmake
@@ -80,10 +80,16 @@ function(mpas_fortran_target target)
 
     # Global Fortran configuration
     set_target_properties(${target} PROPERTIES Fortran_FORMAT FREE)
-    set(MPAS_FORTRAN_TARGET_COMPILE_DEFINITIONS
-        _MPI=1
-        USE_PIO2=1
-    )
+    if(MPAS_USE_SMIOL)
+        set(MPAS_FORTRAN_TARGET_COMPILE_DEFINITIONS
+            MPAS_SMIOL_SUPPORT=1
+        )
+    else()
+        set(MPAS_FORTRAN_TARGET_COMPILE_DEFINITIONS
+            USE_PIO2=1
+        )
+    endif()
+    list(APPEND MPAS_FORTRAN_TARGET_COMPILE_DEFINITIONS _MPI=1)
     # Enable OpenMP support
     if(MPAS_OPENMP)
         target_link_libraries(${target} PUBLIC OpenMP::OpenMP_Fortran)
--- a/src/external/SMIOL/CMakeLists.txt
+++ b/src/external/SMIOL/CMakeLists.txt
@@ -1,0 +1,32 @@
+
+find_package(MPI REQUIRED COMPONENTS C Fortran)
+find_package(NetCDF REQUIRED COMPONENTS Fortran C)
+find_package(PnetCDF REQUIRED COMPONENTS Fortran C)
+
+# Specify the library source files
+set(SMIOL_C_SOURCES smiol.c smiol_utils.c)
+set(SMIOL_F_SOURCES smiolf.F90)
+
+# Create the C library
+add_library(smiol ${SMIOL_C_SOURCES})
+add_library(${PROJECT_NAME}::external::smiol ALIAS smiol)
+target_compile_definitions(smiol PRIVATE SMIOL_PNETCDF SINGLE_PRECISION)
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by amstokely** [comment]
Have you run any tests, that are easily reproducible, to test these changes? It would be interesting to see if the same problem in [PR 1234](https://github.com/MPAS-Dev/MPAS-Model/pull/1234) occurs when using SMIOL.

File: CMakeLists.txt [hunk 1]
```diff
@@ -59,7 +61,9 @@ find_package(OpenMP COMPONENTS Fortran)
 find_package(MPI REQUIRED COMPONENTS Fortran)
 find_package(NetCDF REQUIRED COMPONENTS Fortran C)
 find_package(PnetCDF REQUIRED COMPONENTS Fortran)
-find_package(PIO REQUIRED COMPONENTS Fortran C)
+if(NOT MPAS_USE_SMIOL)
```
jim-p-w, line 64: This doesn't seem to work. When I build with `-DMPAS_USE_SMIOL=1`  the cmake output still has:
```
-- Found PIO: /glade/work/epicufsrt/contrib/spack-stack/derecho/spack-stack-1.6.0/envs/unified-env/install/gcc/12.2.0/parallelio-2.5.10-ysh2ljm  found components: Fortran C 
-- FindPIO:
--   - PIO_PREFIX [/glade/work/epicufsrt/contrib/spack-stack/derecho/spack-stack-1.6.0/envs/unified-env/install/gcc/12.2.0/parallelio-2.5.10-ysh2ljm]
--   - PIO Components Found: Fortran;C;SHARED
```
as well as
```
-- The following REQUIRED packages have been found:
...
 * PIO
```

src/external/SMIOL/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+
+cmake_minimum_required(VERSION 3.10)
+
+find_package(MPI REQUIRED COMPONENTS C Fortran)
+find_package(NetCDF REQUIRED COMPONENTS Fortran C)
+find_package(PnetCDF REQUIRED COMPONENTS Fortran C)
+
+# Specify the library source files
+set(SMIOL_C_SOURCES smiol.c smiol_utils.c)
```
islas, line 9: `SMIOL_C_SOURCES` and `SMIOLF_SOURCES` don't quite follow the same naming convention. Doesn't matter too much, but it might be nice to align the two. My understanding is `SMIOLF_SOURCES` is meant to be effectively, `<TARGET_NAME>_SOURCES` and follow the Makefile target name.  `SMIOL_SOURCES` for the C sources follows that same convention, though it does *look* more ambiguous than the `_C_` one.

src/framework/CMakeLists.txt
```diff
@@ -38,11 +38,23 @@ set(MPAS_FRAMEWORK_SOURCES
 
 add_library(framework ${MPAS_FRAMEWORK_SOURCES})
 set_MPAS_DEBUG_flag(framework)
-set(FRAMEWORK_COMPILE_DEFINITIONS
+if (MPAS_USE_SMIOL)
+    set(FRAMEWORK_COMPILE_DEFINITIONS
+        MPAS_SMIOL_SUPPORT
+        mpas=1
```
islas, line 44: Same as before, I think breaking out common defines would be good, leaving only the specifics within the conditional branch

src/external/SMIOL/CMakeLists.txt
jim-p-w: Do you mean change `SMIOLF_SOURCES` to `SMIOL_F_SOURCES`?

islas: I think that would also work, but in keeping with the `<TARGET_NAME>_SOURCES` I propose `SMIOL_C_SOURCES` => `SMIOL_SOURCES`. If that's too ambiguous of a variable name, then yes exactly what you said instead.

cmake/Functions/MPAS_Functions.cmake
```diff
@@ -80,10 +80,17 @@ function(mpas_fortran_target target)
 
     # Global Fortran configuration
     set_target_properties(${target} PROPERTIES Fortran_FORMAT FREE)
-    set(MPAS_FORTRAN_TARGET_COMPILE_DEFINITIONS
-        _MPI=1
-        USE_PIO2=1
-    )
+    if(MPAS_USE_SMIOL)
+        set(MPAS_FORTRAN_TARGET_COMPILE_DEFINITIONS
+            _MPI=1
```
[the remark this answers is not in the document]
jim-p-w: I moved the common defines out of this if/then clauses.

src/external/SMIOL/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+
+cmake_minimum_required(VERSION 3.10)
```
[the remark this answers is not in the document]
jim-p-w: I removed the cmake_minimum_required.

src/external/SMIOL/CMakeLists.txt
jim-p-w: I prefer to leave the explicit _C_ naame, esp as c sources are rare.

src/framework/CMakeLists.txt
jim-p-w: I moved the common defines out of this if/then clauses.

src/external/SMIOL/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+add_library(smiol ${SMIOL_C_SOURCES})
+add_library(${PROJECT_NAME}::external::smiol ALIAS smiol)
+target_compile_definitions(smiol PRIVATE SMIOL_PNETCDF SINGLE_PRECISION)
+
+# Create the Fortran library
+add_library(smiolf ${SMIOL_F_SOURCES})
+enable_language(Fortran)
+mpas_fortran_target(smiolf)
+add_library(${PROJECT_NAME}::external::smiolf ALIAS smiolf)
+target_compile_definitions(smiolf PRIVATE SMIOL_PNETCDF )
```
amstokely, line 20: When compiling this branch on my local machine, the build fails because the `smiol` target cannot find `mpi.h`. I attached the build log for reference.
[log.make.txt](https://github.com/user-attachments/files/18213005/log.make.txt)
 The issue is resolved by adding `target_include_directories(smiol PRIVATE ${MPI_INCLUDE_PATH})` after `target_compile_definitions(smiol PRIVATE SMIOL_PNETCDF SINGLE_PRECISION)`.

amstokely (changes requested):

src/external/SMIOL/CMakeLists.txt
jim-p-w: Thanks for testing and finding this.

amstokely (approved):
I was able to compile this branch on my local machine, which uses openmpi-5.0.3, GCC-11.4.1, and  parallel-netcdf- 1.12.3 successfully.

src/external/SMIOL/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+# Specify the library source files
+set(SMIOL_C_SOURCES smiol.c smiol_utils.c)
+set(SMIOL_F_SOURCES smiolf.F90)
+
+# Create the C library
+add_library(smiol ${SMIOL_C_SOURCES})
+add_library(${PROJECT_NAME}::external::smiol ALIAS smiol)
+target_compile_definitions(smiol PRIVATE SMIOL_PNETCDF SINGLE_PRECISION)
+target_include_directories(smiol PRIVATE ${MPI_INCLUDE_PATH})
```
islas, line 14: I'm getting a build failure of : 
```
[  0%] Building C object src/external/SMIOL/CMakeFiles/smiol.dir/smiol.c.o
/home/aislas/mpas-model/mpas/src/external/SMIOL/smiol.c:11:10: fatal error: pnetcdf.h: No such file or directory
   11 | #include "pnetcdf.h"
      |          ^~~~~~~~~~~
compilation terminated.
gmake[2]: *** [src/external/SMIOL/CMakeFiles/smiol.dir/build.make:76: src/external/SMIOL/CMakeFiles/smiol.dir/smiol.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:263: src/external/SMIOL/CMakeFiles/smiol.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```

This can be fixed with 
`target_link_libraries( smiol PRIVATE PnetCDF::PnetCDF_Fortran )`

I would petition to just change the MPI include to also use `MPI::MPI_C` as right now `libsmiol.so` is implicitly resolving this library dependency through `libpnetcdf.so`. This would eventually break if pnetcdf is compiled as a static library.

So the final line would look like :
`target_link_libraries( smiol PRIVATE MPI::MPI_C PnetCDF::PnetCDF_Fortran )`

islas (changes requested):

src/external/SMIOL/CMakeLists.txt
jim-p-w: OK, I made the change you suggested.

islas (approved):

mgduda (approved):

Issue #1256, opened 2024-12-25
Mountain wave test case initial state wrong when init_atmosphere_model is run in parallel


MHBalsmeier:
The initial state of the mountain wave test case differs depending on whether the init_atmosphere program is run serially or in parallel. When running the program serially, everything looks fine. After five hours, the result is expected:

![image](https://github.com/user-attachments/assets/56e20d98-8f87-46d6-9f94-38fbcc1cfad8)

When running the init_atmosphere program in parallel (mpiexec -np 4 ./init_atmosphere_model), however, the result looks as follows:

![image](https://github.com/user-attachments/assets/8d576efd-e9bd-433a-8366-f89653fa9cfb)

For creating the plots, MPAS 8.2.2 was used, compiled with gfortran, with OpenMP and OpenACC switched off. Both times, the model was run in parallel.

I looked more closely at the initial state and wrote the following python program to detect differences:

```
# This script compares the input files of the serial vs. parallel initialization.

import netCDF4 as nc
import numpy as np

filename_serial = "mtn_wave_init_serial.nc"
filename_parallel = "mtn_wave_init_parallel.nc"

array_serial = {}
array_parallel = {}

ds = nc.Dataset(filename_serial, "r", format = "NETCDF4")
var_names = ds.variables.keys()
for var in var_names:
    array_serial[var] = ds[var][:]
ds.close()

ds = nc.Dataset(filename_parallel, "r", format = "NETCDF4")
for var in var_names:
    array_parallel[var] = ds[var][:]
ds.close()

for var in var_names:
    try:
        min_serial = np.round(np.nanmin(array_serial[var]), 3)
    except:
        continue
    max_serial = np.round(np.max(array_serial[var]), 3)
    min_parallel = np.round(np.min(array_parallel[var]), 3)
    max_parallel = np.round(np.max(array_parallel[var]), 3)
    if min_serial!=min_parallel or max_serial!=max_parallel:
        print("deviation detected for variable: " + var + ", serial init min, max: ", str(min_serial) + ", " +  str(max_serial) + \
        ", parallel init min, max: " + str(min_parallel) + ", " + str(max_parallel))

print("Maximum absolute zgrid diff: " + str(np.max(np.abs(array_serial["zgrid"] - array_parallel["zgrid"]))))
```

It gave the following results:

```
deviation detected for variable: zxu, serial init min, max:  -0.157, 0.157, parallel init min, max: -0.166, 0.166
[...]

PR #1257, opened 2024-12-25
Bugfix in subroutine init_atm_case_mtn_wave regarding parallel execution

MHBalsmeier:
This is a fix for issue #1256

The problem was that the center coordinate of the domain xc was not synchronized among processes which resulted in wrong orography. This is now fixed and the initial state files created in parallel and serially are identical.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -158,7 +158,7 @@ module init_atm_cases
             call mpas_pool_get_subpool(block_ptr % structs, 'diag', diag)
 
             call mpas_log_write(' calling test case setup ')
-            call init_atm_case_mtn_wave(mesh, nCells, nVertLevels, state, diag, block_ptr % configs)
+            call init_atm_case_mtn_wave(domain % dminfo, mesh, nCells, nVertLevels, state, diag, block_ptr % configs)
             call decouple_variables(mesh, nCells, nVertLevels, state, diag)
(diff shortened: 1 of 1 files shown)


mgduda:
@MHBalsmeier Thanks very much for finding this bug and providing a fix. I have just three small requests:
1. Rather than opening a PR from your `develop` branch, could you commit the fix to a new branch based on the `develop` branch and open the PR from that branch?
2. In the new PR, could you target the `hotfix-v8.2.3` branch?
3. In the commit message, could you include an explanation of the bug and how the changes in the commit fix that bug? We use the output of `git log` quite often, and so more detail in the commit messages can be valuable.

Thanks again, and if you have any questions, please don't hesitate to follow up!

MHBalsmeier:
Thank you, sure, I'll open a new one.

MHBalsmeier:
@mgduda "commit the fix to a new branch based on the develop branch and open the PR from that branch": I based my new PR on the current develop branch of this repo, and not on my develop branch ... so now it would merge all the commits in develop into the hotfix branch. It's showing 110 commits. Is that what you intended?
I might have to try a third time, sorry for the chaos.

mgduda:
@MHBalsmeier I think you'll be able to fix up PR #1312 by force-pushing changes to your existing [init_atm_cases_bugfix](https://github.com/MHBalsmeier/MPAS-Model/tree/init_atm_cases_bugfix) branch. So, there's probably not a need to make a third PR.

MHBalsmeier:
@mgduda Thank you, let's see how it looks now.

PR #1258, opened 2024-12-30
Remove duplicate allocation of indexToEdgeID in mpas_io_setup_edge_block_fields

mgduda:
This PR removes a duplicate allocation of `indexToEdgeID % array` in the `mpas_io_setup_edge_block_fields` routine that was the source of a memory leak.


gdicker1 (approved):
Approved by inspection. Nice catch

PR #1260, opened 2025-01-03
Fix data types for invariant fields transferred during init with OpenACC

gdicker1:
This PR fixes some OpenACC data transfers that occur in `mpas_atm_dynamics_init` and `mpas_atm_dynamics_finalize`. Since the data types for 2 of the variables (`edgesOnVertex_sign` and `weightsOnEdge`) were wrong, the `mpas_pool_get_array` routine would return the NULL pointer for those fields, and the OpenACC `copyin` and `delete` statements would fail to transfer the intended data.

This PR prevents some issues that could occur when the OpenACC data management approach changes to reduce transfers to/from the device during a model timestep.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -217,9 +217,9 @@ module atm_time_integration
       real (kind=RKIND), dimension(:), pointer :: invDcEdge
       integer, dimension(:,:), pointer :: edgesOnEdge
       integer, dimension(:,:), pointer :: edgesOnVertex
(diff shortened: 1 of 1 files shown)


mgduda (changes requested):
The code changes themselves look good. I'd only suggest mentioning the names of the fields whose types have been changed from `integer` to `real (kind=RKIND)` specifically in the log file and PR description. It's always nice to have grep'able names in the log to easily find which commit introduced a specific change.

gdicker1:
@mgduda does the force push and edit to the PR description address your comment?

mgduda:
> @mgduda does the force push and edit to the PR description address your comment?

Yes -- thanks very much!

mgduda (approved):

PR #1261, opened 2025-01-07
Add support for passing integer(8) values to mpas_log_write.

jim-p-w:
This PR allows callers of mpas_log_write to provide values of type integer(8).

The placeholder specification is $j for integer(8) values, as opposed to $i for integer values.

I ran a test where I interspersed integer and integer(8) values, e.g.
```
integer, dimension(:), allocatable :: intarray
integer(8), dimension(:), allocatable :: int8array
...
call mpas_log_write('SMIOLf_get_var error $i i81:$j int1:$i i82:$j int2:$i i83:$j', intArgs=[local_ierr, intarray(1), intarray(2)], int8Args=int8array, messageType=MPAS_LOG_ERR)
```

### DIFF
--- a/src/framework/mpas_log.F
+++ b/src/framework/mpas_log.F
@@ -470,15 +470,15 @@ module mpas_log
 !>                regardless of if all tasks have open log files.
 !>    flushNow: flag indicating the message should be flushed immediately.
 !>              Note: error and critical error messages are always flushed immediately.
-!>    intArgs, realArgs, logicArgs:  arrays of variable values to be inserted into the
-!>                                   message to replace the following characters: $i, $r, $l
+!>    intArgs, int8Args, realArgs, logicArgs:  arrays of variable values to be inserted into the
+!>                                   message to replace the following characters: $i, $j, $r, $l
 !>                                   See routine log_expand_string below for details.
 !>
 !
 !-----------------------------------------------------------------------
 
    recursive subroutine mpas_log_write(message, messageType, masterOnly, flushNow, &
-                 intArgs, realArgs, logicArgs, err)
+                 intArgs, int8Args, realArgs, logicArgs, err)
 
       use mpas_threading
 
@@ -492,6 +492,7 @@ module mpas_log
       logical, intent(in), optional :: masterOnly  !< Input: flag to only print message on master task
       logical, intent(in), optional :: flushNow  !< Input: flag to force a flush of the message buffer
       integer, dimension(:), intent(in), optional :: intArgs  !< Input: integer variable values to insert into message
+      integer(kind=I8KIND), dimension(:), intent(in), optional :: int8Args  !< Input: integer variable values to insert into message
       real(kind=RKIND), dimension(:), intent(in), optional :: realArgs  !< Input: real variable values to insert into message
          !< Input: exponential notation variable values to insert into message
       logical, dimension(:), intent(in), optional :: logicArgs  !< Input: logical variable values to insert into message
@@ -540,7 +541,7 @@ module mpas_log
 
 
       ! Construct message by expanding variable values as needed and inserting message type prefix
-      call log_expand_string(message, messageExpanded, intArgs=intArgs, logicArgs=logicArgs, realArgs=realArgs)
+      call log_expand_string(message, messageExpanded, intArgs=intArgs, int8Args=int8Args, logicArgs=logicArgs, realArgs=realArgs)
 
       ! Determine message prefix
       select case (messageTypeHere)
@@ -870,6 +871,7 @@ module mpas_log
    !>   The variables to be expanded are represented with a '$' symbol followed
    !>   by one of these indicators:
    !>   $i -> integer, formatted to be length of integer
+   !>   $j -> integer(kind=I8KIND), formatted to be length of integer
    !>   $l -> logical, fomatted as 'T' or 'F'
    !>   $r -> real,  formatted as 9 digits of precision for SP mode, 17 for DP mode
    !>                Floats are formatted using 'G' format which is smart about
@@ -880,13 +882,13 @@ module mpas_log
    !>   run out before the $ expansion indicators are all replaced, the remaining
    !>   expansions will be filled with a fill value ('**').  The fill value is also
    !>   used if the expansion indicator is of an unknown type, where the valid types
-   !>   are $i, $l, $r.
+   !>   are $i, $j, $l, $r.
    !>   If the user prefers more specific formatting, they have to do it external
    !>   to this routine in a local string variable.  Similarly, character variables
    !>   can be handled by the string concatenation command (//).
    !>   This routine is based off of mpas_expand_string.
    !-----------------------------------------------------------------------
-   subroutine log_expand_string(inString, outString, intArgs, logicArgs, realArgs)
+   subroutine log_expand_string(inString, outString, intArgs, int8Args, logicArgs, realArgs)
 
       implicit none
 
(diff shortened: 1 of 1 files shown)


src/framework/mpas_log.F
```diff
@@ -492,6 +492,7 @@ recursive subroutine mpas_log_write(message, messageType, masterOnly, flushNow,
       logical, intent(in), optional :: masterOnly  !< Input: flag to only print message on master task
       logical, intent(in), optional :: flushNow  !< Input: flag to force a flush of the message buffer
       integer, dimension(:), intent(in), optional :: intArgs  !< Input: integer variable values to insert into message
+      integer(8), dimension(:), intent(in), optional :: int8Args  !< Input: integer variable values to insert into message
```
mgduda, line 495: This should be declared as `integer(kind=I8KIND)` rather than as `integer(8)`.

src/framework/mpas_log.F
```diff
@@ -870,6 +871,7 @@ end subroutine log_abort
    !>   The variables to be expanded are represented with a '$' symbol followed
    !>   by one of these indicators:
    !>   $i -> integer, formatted to be length of integer
+   !>   $w -> integer(8), formatted to be length of integer
```
mgduda, line 874: As an alternative to `w`, I'd suggest considering `j`. In Fortran, variables that begin with `i, j, k, m, n` (if I recall correctly) are implicitly typed as integers (unless declared otherwise). Of course, we're not using implicit variables here, but the convention of using the aforementioned letters for integers could make `j` a bit more intuitive than `w`.

mgduda (changes requested):

mgduda:
@jim-p-w Can you add more detail to the commit message?

src/framework/mpas_log.F
jim-p-w: Done. Thanks for the suggestion.

src/framework/mpas_log.F
jim-p-w: Sounds good to me. I wasn't happy with `$w`

jim-p-w:
> @jim-p-w Can you add more detail to the commit message?

@mgduda How about:
Allow passing values of type integer(8)/integer(kind=I8KIND) to mpas_log_write.
The string `$j` is used to identify integer(kind=I8KIND) values.

I'm open to suggestions about what the commit message says.

abishekg7:
@mgduda Just for my understanding, would a core infrastructure addition like this require an addition to the unit tests also?

abishekg7:
@jim-p-w Probably a basic question, but I'm trying this code snippet out in your branch, 
```
 integer (kind=I8KIND) :: tmp_i8_var = 5123123_I8KIND

 call mpas_log_write('rk_setup i8 kind log $w', int8Args=(/tmp_i8_var/))
```
and I get this as output. Do I need to set some kind of formatting width here, or am I doing something else wrong perhaps? 
```
 rk_setup i8 kind log **
```

jim-p-w:
> @jim-p-w Probably a basic question, but I'm trying this code snippet out in your branch,
> 
> ```
>  integer (kind=I8KIND) :: tmp_i8_var = 5123123_I8KIND
> 
>  call mpas_log_write('rk_setup i8 kind log $w', int8Args=(/tmp_i8_var/))
> ```
> 
> and I get this as output. Do I need to set some kind of formatting width here, or am I doing something else wrong perhaps?
> 
> ```
>  rk_setup i8 kind log **
> ```

@abishekg7 I changed the placeholder string to `$j` instead of `$w` per Michaels suggestion.

jim-p-w:
> @mgduda Just for my understanding, would a core infrastructure addition like this require an addition to the unit tests also?

@abishekg7 @mgduda It wasn't clear to me how to create a unit test for `mpas_log_write`. When I look at the existing tests in `src/core_test` I see `mpas_log_write` is assumed to work, as it is used all over the test code.

abishekg7:
> > @mgduda Just for my understanding, would a core infrastructure addition like this require an addition to the unit tests also?
> 
> @abishekg7 @mgduda It wasn't clear to me how to create a unit test for `mpas_log_write`. When I look at the existing tests in `src/core_test` I see `mpas_log_write` is assumed to work, as it is used all over the test code.

Yeah, I also don't see any specific tests for `mpas_log_write` itself.


Otherwise, the PR looks good. I've tested with NVHPC GPU and CPU versions so far.

mgduda:
@jim-p-w Can you update the PR description to reflect the use of `$j` rather than `$w`? Also, I'd suggest adding a mention of how `I8KIND` integers are passed to the `mpas_log_write` function in the commit message.

jim-p-w:
@mgduda I can add an example usage in `test_core_run` and say "See test_core_run for example usage" in the commit comment.
Or should I not bother and make the commit message 
```
Allow passing values of type integer(8)/integer(kind=I8KIND) to mpas_log_write.
The string $j is used to identify integer(kind=I8KIND) values, e.g.
call mpas_log_write('integer: $i integer8: $j', intArgs=[local_int], int8Args=[local_int8])
```

mgduda:
> @mgduda I can add an example usage in `test_core_run` and say "See test_core_run for example usage" in the commit comment. Or should I not bother and make the commit message
> 
> ```
> Allow passing values of type integer(8)/integer(kind=I8KIND) to mpas_log_write.
> The string $j is used to identify integer(kind=I8KIND) values, e.g.
> call mpas_log_write('integer: $i integer8: $j', intArgs=[local_int], int8Args=[local_int8])
> ```
> 
> I have changed the commit message to the three lines above.

Rather than asking users to look around in the test core code, I think it would be better to just summarize in the commit message how one would print an `I8KIND` integer. I think the message you've proposed looks good, though I suppose it could be stated more explicitly that the new optional argument `int8Args` to `mpas_log_write` is used for passing `I8KIND` integers to be used for all `$j` placeholders.

src/framework/mpas_log.F
```diff
@@ -470,15 +470,15 @@ end subroutine mpas_log_open
 !>                regardless of if all tasks have open log files.
 !>    flushNow: flag indicating the message should be flushed immediately.
 !>              Note: error and critical error messages are always flushed immediately.
-!>    intArgs, realArgs, logicArgs:  arrays of variable values to be inserted into the
-!>                                   message to replace the following characters: $i, $r, $l
+!>    intArgs, int8Args,realArgs, logicArgs:  arrays of variable values to be inserted into the
```
mgduda, line 473: Could we add a space after the comma (after `int8Args,`)?

src/framework/mpas_log.F
```diff
@@ -936,6 +940,12 @@ subroutine log_expand_string(inString, outString, intArgs, logicArgs, realArgs)
          nInts = 0
       endif
 
+      if (present(int8Args)) then
+        nInt8s = size(int8Args)
+      else
+        nInt8s = 0
```
mgduda, line 946: The if-blocks before and after use three spaces of indentation. Could we do the same here for consistency?

jim-p-w: Indentation is fixed.

jim-p-w:
I will squash all commits for this pr after it's ready to be merged.

src/framework/mpas_log.F
```diff
@@ -870,6 +871,7 @@ end subroutine log_abort
    !>   The variables to be expanded are represented with a '$' symbol followed
    !>   by one of these indicators:
    !>   $i -> integer, formatted to be length of integer
+   !>   $j -> integer(kind=I8KIND), formatted to be length of integer
```
mgduda, line 874: I think `$j` should be added to the list of valid types on line 885, below.

mgduda (approved):
The changes look good to me. I'll give you a chance to modify the commit history (probably just one commit for this PR is sufficient), then I'll merge. Thanks!

jim-p-w:
I squashed 3 commits into one.

PR #1264, opened 2025-01-10
Fix memory leak from cellLimitField in mpas_block_creator_build_cell_halos

mgduda:
This PR fixes a memory leak in `mpas_block_creator_build_cell_halos` by deallocating the `cellLimitField` field before the routine returns.

The `cellLimitField` field is allocated in the `mpas_block_creator_build_cell_halos` routine and is only used in a call therein to `mpas_dmpar_get_exch_list`. If not deallocated at the end of the `mpas_block_creator_build_cell_halos` routine, the memory allocated to `cellLimitField` would be otherwise be leaked, since there is nothing outisde of the `mpas_block_creator_build_cell_halos routine` that retains a reference to it and deallocates it.

### DIFF
--- a/src/framework/mpas_block_creator.F
+++ b/src/framework/mpas_block_creator.F
@@ -715,6 +715,7 @@ module mpas_block_creator
      ! Deallocate array and field.
      deallocate(sendingHaloLayers)
      call mpas_deallocate_field(offSetField)
+     call mpas_deallocate_field(cellLimitField)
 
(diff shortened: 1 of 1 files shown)


abishekg7 (approved):
Changes look good, and model runs to the finish. However, I'm not able to pinpoint the exact differences in the valgrind outputs between the base version and this PR, as there are many other diagnostic messages as well. But this can be discussed later.

PR #1265, opened 2025-01-16
Enable executing atm_bdy_set_scalars_work on gpus

jim-p-w:
This PR enables executing the `atm_bdy_set_scalars_work` subroutine on GPUs.

   `make -j4 nvhpc  CORE=atmosphere PRECISION=single OPENACC=true`

Comparing the results stored in the `restart.*.nc` file showed no changes.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -6913,6 +6913,12 @@ module atm_time_integration
 
       !---
 
+      MPAS_ACC_TIMER_START('atm_bdy_set_scalars_work [ACC_data_xfer]')
+      !$acc enter data copyin(scalars_new, scalars_driving)
+      MPAS_ACC_TIMER_STOP('atm_bdy_set_scalars_work [ACC_data_xfer]')
+
+      !$acc parallel default(present)
+      !$acc loop gang worker
       do iCell = cellSolveStart, cellSolveEnd ! threaded over cells
 
          if ( bdyMaskCell(iCell) > nRelaxZone) then ! specified zone
@@ -6920,6 +6926,7 @@ module atm_time_integration
             !  update the specified-zone values
             !
 !DIR$ IVDEP
+           !$acc loop vector collapse(2)
             do k=1,nVertLevels
                do iScalar = 1, num_scalars
                   scalars_new(iScalar,k,iCell) = scalars_driving(iScalar,k,iCell)
@@ -6929,6 +6936,12 @@ module atm_time_integration
          end if
 
       end do  ! updates now in temp storage
+      !$acc end parallel
+
+      MPAS_ACC_TIMER_START('atm_bdy_set_scalars_work [ACC_data_xfer]')
+      !$acc exit data copyout(scalars_new)
+      !$acc exit data delete(scalars_driving)
+      MPAS_ACC_TIMER_STOP('atm_bdy_set_scalars_work [ACC_data_xfer]')
             
    end subroutine atm_bdy_set_scalars_work
 


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6913,13 +6913,17 @@ subroutine atm_bdy_set_scalars_work( scalars_driving, scalars_new, &
 
       !---
 
+      !$acc enter data copyin(scalars_new, scalars_driving)
+      !$acc parallel default(present)
```
mgduda, line 6917: As a style consideration, we've tended to leave at least a line of whitespace between the `data` directives and other directives (principally, `parallel` directives). And, it's been convention so far to add a timers around the data movement directives.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6913,13 +6913,17 @@ subroutine atm_bdy_set_scalars_work( scalars_driving, scalars_new, &
...
+      !$acc loop gang
```
mgduda, line 6918: I'd suggest adding a `worker` clause here as well unless you're finding that it alters results. In practice I have no idea if the addition of `worker` matters, but doing so would be consistent with what has been done elsewhere.

mgduda (changes requested):

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
jim-p-w: added a timer for the data copyin directive, and added blank line after the timer_stop statement.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
jim-p-w: changed `loop gang` to `loop gang worker`

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6929,6 +6936,12 @@ subroutine atm_bdy_set_scalars_work( scalars_driving, scalars_new, &
          end if
 
       end do  ! updates now in temp storage
+      !$acc end parallel
+
+      MPAS_ACC_TIMER_START('atm_init_coupled_diagnostics [ACC_data_xfer]')
+      !$acc exit data copyout(scalars_new)
+      MPAS_ACC_TIMER_STOP('atm_init_coupled_diagnostics [ACC_data_xfer]')
+      !$acc exit data delete(scalars_driving)
```
gdicker1, line 6944: Even though it shouldn't contribute, could you put this `exit data delete` inside the timer?

gdicker1 (changes requested):

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6913,13 +6913,20 @@ subroutine atm_bdy_set_scalars_work( scalars_driving, scalars_new, &
 
       !---
 
+      MPAS_ACC_TIMER_START('atm_init_coupled_diagnostics [ACC_data_xfer]')
```
abishekg7, line 6916: Could you update the string in timer calls to `atm_bdy_set_scalars_work` in all four instances.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: This has been addressed by force-push [982098f](https://github.com/MPAS-Dev/MPAS-Model/commit/982098ffad7242b7f83030ebcbe27efec19ee5c5) to [7d1d937](https://github.com/MPAS-Dev/MPAS-Model/commit/7d1d937a5b311db76885086c4dd2d1616fc1fe0f)

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: This has been addressed by force-push [982098f](https://github.com/MPAS-Dev/MPAS-Model/commit/982098ffad7242b7f83030ebcbe27efec19ee5c5) to [7d1d937](https://github.com/MPAS-Dev/MPAS-Model/commit/7d1d937a5b311db76885086c4dd2d1616fc1fe0f)

abishekg7:
Wondering if we also need a  timer start/stop around the call to `atm_bdy_set_scalars_work`. Other than that everything looks good!

gdicker1:
> Wondering if we also need a timer start/stop around the call to atm_bdy_set_scalars_work. Other than that everything looks good!

I'm wondering the same, but that would probably be another PR. @mgduda what are your thoughts about these timers? Maybe we don't need parent timers for the `_bdy_` routines since theses [ACCxfer] timers are meant to be transient?

mgduda:
The `[ACC_data_xfer]` timers are intended to be transient, and I expect we'll be removing them when we remove the data transfers for individual routines. If the `atm_bdy_set_scalars_work` routine doesn't already have a timer, then I suppose the `atm_bdy_set_scalars_work [ACC_data_xfer]` has less value. Perhaps for now, we can just include the `[ACC_data_xfer]` timer (especially since it's already been added).

@jim-p-w The code changes themselves look fine to me. Can you modify the commit message:
1. "acc" -> "OpenACC"
2. "gpus" -> "GPUs"
3. note that the commit includes a "atm_bdy_set_scalars_work [ACC_data_xfer]" timer although there is no corresponding "atm_bdy_set_scalars_work" timer that includes the computational work of the routine.

gdicker1 (approved):
This looks good to me with the commit message in f3006fa.

Aside already talked about: We could create `scalars_new` instead, but I think this copyin is justified to avoid using whatever default values the GPU would put in the unowned (halo and garbage) cells not touched by this routine.

mgduda (approved):

PR #1266, opened 2025-01-16
Enable GPU exection of atm_bdy_adjust_scalars_work via OpenACC

gdicker1:
This PR makes small code modifications and adds OpenACC directives so the atm_bdy_adjust_scalars_work routine can execute on GPU(s).

Timing information for the OpenACC data transfers in this routine is captured in the log file by a new timer: atm_bdy_adjust_scalars [ACC_data_xfer].

Invariant fields used in this routine are also copied to the device within mpas_atm_dynamics_init and are deleted in mpas_atm_dynamics_finalize.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -233,6 +233,7 @@ module atm_time_integration
       real (kind=RKIND), dimension(:,:,:), pointer :: zb3_cell
       real (kind=RKIND), dimension(:), pointer :: fzm
       real (kind=RKIND), dimension(:), pointer :: fzp
+      real (kind=RKIND), dimension(:), pointer :: meshScalingRegionalCell
 #endif
 
 
@@ -356,6 +357,9 @@ module atm_time_integration
       call mpas_pool_get_array(mesh, 'fzp', fzp)
       !$acc enter data copyin(fzp)
 
+      call mpas_pool_get_array(mesh, 'meshScalingRegionalCell', meshScalingRegionalCell)
+      !$acc enter data copyin(meshScalingRegionalCell)
+
 #endif
 
    end subroutine mpas_atm_dynamics_init
@@ -425,6 +429,7 @@ module atm_time_integration
       real (kind=RKIND), dimension(:,:,:), pointer :: zb3_cell
       real (kind=RKIND), dimension(:), pointer :: fzm
       real (kind=RKIND), dimension(:), pointer :: fzp
+      real (kind=RKIND), dimension(:), pointer :: meshScalingRegionalCell
 #endif
 
 
@@ -547,6 +552,10 @@ module atm_time_integration
 
       call mpas_pool_get_array(mesh, 'fzp', fzp)
       !$acc exit data delete(fzp)
+
+      call mpas_pool_get_array(mesh, 'meshScalingRegionalCell', meshScalingRegionalCell)
+      !$acc exit data delete(meshScalingRegionalCell)
+
 #endif
 
    end subroutine mpas_atm_dynamics_finalize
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6780,17 +6789,30 @@ subroutine atm_bdy_adjust_scalars_work( scalars_new, scalars_driving, dt, dt_rk,
...
+            !$acc loop collapse(2)
+            do k=1,nVertLevels
+               do iScalar=1,num_scalars
+                  scalars_tmp(iScalar,k,iCell) = scalars_new(iScalar,k,iCell)
+               end do
+            end do
 
             !  first, we compute the 2nd-order laplacian filter
             !
+            !$acc loop seq
```
jim-p-w, line 6815: local scalar variables are privatized by default, so it seems this loop doesn't need to execute sequentially.

jim-p-w (approved):

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6780,17 +6789,30 @@ subroutine atm_bdy_adjust_scalars_work( scalars_new, scalars_driving, dt, dt_rk,
...
+      !$acc parallel default(present)
+      !$acc loop gang worker
       do iCell = cellSolveStart, cellSolveEnd ! threaded over cells
 
          if ( (bdyMaskCell(iCell) > 1) .and. (bdyMaskCell(iCell) <= nRelaxZone) ) then ! relaxation zone
 
             laplacian_filter_coef = dt_rk*(real(bdyMaskCell(iCell)) - 1.)/real(nRelaxZone)/(10.*dt*meshScalingRegionalCell(iCell))
             rayleigh_damping_coef = laplacian_filter_coef/5.0
-            scalars_tmp(1:num_scalars,1:nVertLevels,iCell) = scalars_new(1:num_scalars,1:nVertLevels,iCell)
+
```
abishekg7, line 6805: Do we need a newline here by style conventions?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6799,6 +6821,7 @@ subroutine atm_bdy_adjust_scalars_work( scalars_new, scalars_driving, dt, dt_rk,
                cell1 = cellsOnEdge(1,iEdge)
                cell2 = cellsOnEdge(2,iEdge)
 !DIR$ IVDEP
+               !$acc loop collapse(2)
```
abishekg7, line 6824: I guess the compiler will implicitly assume a `vector` here, but wondering if we might need to also explicitly state that here for sake of convention.

abishekg7:
Other than my comments, this PR seems to be bit identical with the previous version of develop.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: Thanks for the catch. It's been addressed in the fixup commits pushed today.

mgduda: I noticed there are a few other `!$acc loop collapse(2)` directives. Should we add `vector` to those as well?

gdicker1: Thanks for the second catch. @mgduda fixup commit [9d09979](https://github.com/MPAS-Dev/MPAS-Model/pull/1266/commits/9d09979252b3d2510f3f54b94b48a05bfd2baec6) should address this.

mgduda: Thanks!

mgduda (approved):
Looks good to me!

@gdicker1 I'll merge this PR after you've had a chance to clean up the commit history.

gdicker1:
@mgduda this should be ready now!

Force-push from [9d09979](https://github.com/MPAS-Dev/MPAS-Model/commit/9d09979252b3d2510f3f54b94b48a05bfd2baec6) to [dd38bd0](https://github.com/MPAS-Dev/MPAS-Model/commit/dd38bd0a2ef4f896158784097859ded46a818f8f) to rebase fixup commits.

PR #1268, opened 2025-01-17
Enable GPU execution of atm_bdy_adjust_dynamics_speczone_tend via OpenACC

gdicker1:
This PR adds OpenACC directives so the `atm_bdy_adjust_dynamics_speczone_tend` routine can execute on GPU(s).

Timing information for the OpenACC data transfers in this routine is captured in the log file by a new timer: `atm_bdy_adjust_dynamics_speczone_tend [ACC_data_xfer]`.


mgduda:
@gdicker1 The changes themselves look good to me. I think a single commit would be sufficient for this PR.

gdicker1:
@mgduda, pushed again with only 1 commit.

abishekg7 (approved):
Looks good to me. I check for bit reproducibility against an older version of `develop` - seems fine. You may want to do a check against the latest `develop` version after rebasing.

mgduda (approved):

PR #1269, opened 2025-01-17
Initial OpenACC port of atm_bdy_adjust_dynamics_relaxzone_tend

abishekg7:
Initial OpenACC port of atm_bdy_adjust_dynamics_relaxzone_tend


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6579,14 +6619,18 @@ subroutine atm_bdy_adjust_dynamics_relaxzone_tend( config, tend, state, diag, me
...
 
       end do
+      !$acc end parallel
 
       !  Third (and last), the horizontal filter for ru
-
+      !$acc parallel default(present)
+      !$acc loop gang worker private(cell1, cell2, vertex1, vertex2, r_dc, r_dv, & 
+      !$acc    iCell, iVertex, invArea, iEdge_div, iEdge_vort, edge_sign, &
+      !$acc    laplacian_filter_coef, divergence1, divergence2, vorticity1, vorticity2)
```
gdicker1, line 6628: Here I'd only expect the arrays (`divergence1`, `divergence2`, `vorticity1`, and `vorticity2`) to be declared private. The rest should automatically be private at the worker level and passed on to their (vector) threads. Are you being explicit here, or do you need to also add the scalar-values to the private clause for correctness?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6579,14 +6619,18 @@ subroutine atm_bdy_adjust_dynamics_relaxzone_tend( config, tend, state, diag, me
...
       do iEdge = edgeStart, edgeEnd
 
          if ( (bdyMaskEdge(iEdge) > 1) .and. (bdyMaskEdge(iEdge) <= nRelaxZone) ) then ! relaxation zone
 
-            laplacian_filter_coef = dcEdge(iEdge)**2 * (real(bdyMaskEdge(iEdge)) - 1.)/   &
+            laplacian_filter_coef = dcEdge(iEdge)*dcEdge(iEdge)* (real(bdyMaskEdge(iEdge)) - 1.)/   &
```
gdicker1, line 6633: Pure curiosity about this change: could you explain your motivation a little? Does this have any effect?

gdicker1 (changes requested):
Besides a more formal commit message, here's some things I've noticed so far.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6529,37 +6550,55 @@ subroutine atm_bdy_adjust_dynamics_relaxzone_tend( config, tend, state, diag, me
...
+            !$acc loop vector
             do k=1, nVertLevels
                tend_rho(k,iCell) = tend_rho(k,iCell) - rayleigh_damping_coef * (rho_zz(k,iCell) - rho_driving_values(k,iCell))
                tend_rt(k,iCell)  = tend_rt(k,iCell)  - rayleigh_damping_coef * (rho_zz(k,iCell)*theta_m(k,iCell) - rt_driving_values(k,iCell))
             end do
          end if
       end do
+      !$acc end parallel
 
+      !$acc parallel default(present)
```
jim-p-w, line 6579: Can both of these loops be in the same parallel region and run simultaneously? They appear to be orthogonal to each other. That is, neither loop references the array(s) the other loop changes.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6529,37 +6550,55 @@ subroutine atm_bdy_adjust_dynamics_relaxzone_tend( config, tend, state, diag, me
...
+      !  Second, the horizontal filter for rtheta_m and rho_zz
+      !$acc parallel default(present)
+      !$acc loop gang worker
       do iCell = cellSolveStart, cellSolveEnd ! threaded over cells
 
          if ( (bdyMaskCell(iCell) > 1) .and. (bdyMaskCell(iCell) <= nRelaxZone) ) then ! relaxation zone
 
             laplacian_filter_coef = (real(bdyMaskCell(iCell)) - 1.)/real(nRelaxZone)/(10.*dt*meshScalingRegionalCell(iCell))
             !
+            !$acc loop seq
```
jim-p-w, line 6601: Given that local scalars are automatically privatized, it seems this loop doesn't have to execute sequentially.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6579,14 +6619,18 @@ subroutine atm_bdy_adjust_dynamics_relaxzone_tend( config, tend, state, diag, me
          end if
 
       end do
+      !$acc end parallel
 
       !  Third (and last), the horizontal filter for ru
-
+      !$acc parallel default(present)
+      !$acc loop gang worker private(cell1, cell2, vertex1, vertex2, r_dc, r_dv, &
```
jim-p-w, line 6626: Local scalar vars are automatically privatized. Additionally, there's a performance hit when explicitly declaring them private, as it causes them to be placed in global memory. [See this nvidia forum item](https://forums.developer.nvidia.com/t/openacc-private-variables/133287)

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
jim-p-w, line 6628: Why does `laplacian_filter_coef` have to be private in this loop? It's not declared private in the other parallel loops in this subroutine.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -6529,37 +6550,55 @@ subroutine atm_bdy_adjust_dynamics_relaxzone_tend( config, tend, state, diag, me
       call mpas_pool_get_array(mesh, 'nEdgesOnCell',nEdgesOnCell)
       call mpas_pool_get_array(mesh, 'verticesOnEdge', verticesOnEdge)
 
-      call mpas_pool_get_config(config, 'config_relax_zone_divdamp_coef', divdamp_coef)
+      call mpas_pool_get_config(config, 'config_relax_zone_divdamp_coef', divdamp_coef_ptr)
+
+      divdamp_coef = divdamp_coef_ptr
+      vertexDegree = vertexDegree_ptr
```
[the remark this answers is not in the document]
abishekg7: Will do.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: True. I briefly didn't remember that about scalars when I wasn't getting bit identical results. Will make these changes and try again.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: I was trying this out because I though it may be one of the obstacles to bit identical results - but probably not. Will remove this change. Although, I would also add that we could just be using the multiplication operator here instead of exponent. But that's for another time.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Thanks, yeah I'll remove the scalars from private. And did not know this about global memory - thanks!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: Agreed, I think this may be better as a multiplication than exponent but that goes against keeping to a "light touch" in this phase.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: So there's a summation going on in the innermost loop for `tend_rt` and `tend_rho`, and ignoring the order of summation, there would still be an incorrect summation if `do i=1,nEdgesOnCell(iCell)` was parallelized. 

Now we could re-write the loop to use the `acc reduction(+: tend_rt, tend_rho)` clause, but we decided to keep the code changes minimal in this round of porting.

jim-p-w: I missed that! Thanks for double checking.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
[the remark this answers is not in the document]
abishekg7: Thanks! Done.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Fusing these two parallel regions does work, but I'm wondering if we should perhaps leave this for the optimization stage.

jim-p-w: I am not knowledgable enough to weigh in on that. But if it succeeded and BFB tests passed I'd put it in now.
I'm fine with it either way, I'll leave it to you, Dylan, and Michael to decide.

jim-p-w (approved):

gdicker1:
This looks good to me now. Just waiting for any changes from Michael's review and the final commit(s)

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
mgduda: Compilers are reasonably good at spotting these opportunities for optimization. The `nvfortran` compiler turns this line:
```
pure function square(x)

    real, intent(in) :: x
    real :: square

    square = x**2

end function square
```
into this assembler:
```
	.text
	.globl	square_                         # -- Begin function square_
	.type	square_,@function
square_:                                # @square_
	vmovss	(%rdi), %xmm0                   # xmm0 = mem[0],zero,zero,zero
	vmulss	%xmm0, %xmm0, %xmm0
	retq
```

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
mgduda: I would support the fusion of these two parallel regions in this PR, especially since doing so would lead to fewer lines in the diff.

mgduda (approved):
Looks good to me.

@abishekg7 I'll merge this after you've reworked the commit history.

abishekg7:
> Looks good to me.
> 
> @abishekg7 I'll merge this after you've reworked the commit history.

Thanks! I've rebased and cleaned up the commit history. And results are still bit identical with develop.

PR #1271, opened 2025-01-19
fixed errors when compile MPAS with gfortran with -fimplicit-none

Yi-ChuanLO:
Since lots of MPAS code doesn't use `implicit none` statement, may be there'll be potential bug that relative to declaration issue.
In this PR, I declared variables explicitly and fixed errors that are detected by gfortran with flag `-fimplicit-none`


mgduda (approved):
Thank you for these changes! I think we may have introduced at least one more implicitly defined variable since this PR was opened, but I will address that with a separate PR.

PR #1272, opened 2025-01-21
Enable executing atm_zero_gradient_w_bdy_work on GPUs.

jim-p-w:
This PR enables executing the atm_zero_gradient_w_bdy_work subroutine on GPUs.

make -j4 nvhpc  CORE=atmosphere PRECISION=single OPENACC=true

Comparing the results stored in the restart.*.nc file showed no changes.

Note this commit adds "atm_zero_gradient_w_bdy_work [ACC_data_xfer]" timers to time the data transfers done in atm_zero_gradient_w_bdy_work, but there is no timer for the actual computation done in atm_zero_gradient_w_bdy_work.


gdicker1 (approved):
Looks good to me!

mgduda (approved):

PR #1277, opened 2025-01-24
Enable executing mpas_atm_get_bdy_tend on GPUs

jim-p-w:
This PR enables executing the `mpas_atm_get_bdy_tend` subroutine on GPUs.

make -j32 nvhpc  CORE=atmosphere PRECISION=single OPENACC=true

Comparing the results stored in the restart.*.nc file showed no changes.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_boundaries.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_boundaries.F
@@ -300,10 +300,10 @@ module mpas_atm_boundaries
         real (kind=RKIND), dimension(vertDim,horizDim+1) :: return_tend
 
         type (mpas_pool_type), pointer :: lbc
-        integer, pointer :: idx
+        integer, pointer :: idx_ptr
         real (kind=RKIND), dimension(:,:), pointer :: tend
         real (kind=RKIND), dimension(:,:,:), pointer :: tend_scalars
-        integer :: ierr
+        integer :: idx, i, j
 
 
         call mpas_pool_get_subpool(block % structs, 'lbc', lbc)
@@ -311,14 +311,46 @@ module mpas_atm_boundaries
         nullify(tend)
         call mpas_pool_get_array(lbc, 'lbc_'//trim(field), tend, 1)
 
+        MPAS_ACC_TIMER_START('mpas_atm_get_bdy_tend [ACC_data_xfer]')
+        !$acc enter data create(return_tend)
         if (associated(tend)) then
-            return_tend(:,:) = tend(:,:)
+            !$acc enter data copyin(tend)
         else
             call mpas_pool_get_array(lbc, 'lbc_scalars', tend_scalars, 1)
-            call mpas_pool_get_dimension(lbc, 'index_'//trim(field), idx)
+            !$acc enter data copyin(tend_scalars)
 
-            return_tend(:,:) = tend_scalars(idx,:,:)
+            ! Ensure the integer pointed to by idx_ptr is copied to the gpu device
+            call mpas_pool_get_dimension(lbc, 'index_'//trim(field), idx_ptr)
+            idx = idx_ptr
         end if
+        MPAS_ACC_TIMER_STOP('mpas_atm_get_bdy_tend [ACC_data_xfer]')
+
+        !$acc parallel default(present)
+        if (associated(tend)) then
+            !$acc loop gang vector collapse(2)
+            do j=1,horizDim+1
+                do i=1,vertDim
+                    return_tend(i,j) = tend(i,j)
+                end do
+            end do
+        else
+            !$acc loop gang vector collapse(2)
+            do j=1,horizDim+1
+                do i=1,vertDim
+                    return_tend(i,j) = tend_scalars(idx,i,j)
+                end do
+            end do
+        end if
+        !$acc end parallel
+
+        MPAS_ACC_TIMER_START('mpas_atm_get_bdy_tend [ACC_data_xfer]')
+        !$acc exit data copyout(return_tend)
+        if (associated(tend)) then
+            !$acc exit data delete(tend)
+        else
+            !$acc exit data delete(tend_scalars)
+        end if
+        MPAS_ACC_TIMER_STOP('mpas_atm_get_bdy_tend [ACC_data_xfer]')
 
     end function mpas_atm_get_bdy_tend
 


src/core_atmosphere/dynamics/mpas_atm_boundaries.F
```diff
@@ -291,25 +301,54 @@ function mpas_atm_get_bdy_tend(clock, block, vertDim, horizDim, field, delta_t)
...
-            call mpas_pool_get_dimension(lbc, 'index_'//trim(field), idx)
+            call mpas_pool_get_dimension(lbc, 'index_'//trim(field), idx_ptr)
+            !$acc enter data copyin(tend_scalars)
+        end if
+        MPAS_ACC_TIMER_STOP('mpas_atm_get_bdy_tend [ACC_data_xfer]')
 
-            return_tend(:,:) = tend_scalars(idx,:,:)
+        !$acc parallel default(present)
+        if (associated(tend)) then
+            !$acc loop vector collapse(2)
```
abishekg7, line 327: Not sure how much of a performance gain there will be, but you may also want to try `acc loop gang vector collapse(2)`. That's also more consistent with what we've been using elsewhere.

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
```diff
@@ -291,25 +301,54 @@ function mpas_atm_get_bdy_tend(clock, block, vertDim, horizDim, field, delta_t)
...
+            do j=1,horizDim+1
+                do i=1,vertDim
+                    return_tend(i,j) = tend(i,j)
+                end do
+            end do
+        else
+            idx = idx_ptr !don't use integer pointers in OpenACC code
```
abishekg7, line 334: This comment could be rephrased to be more specific, i.e. using pointers to refer to scalars (in loop bounds, or other indices) runs into problems alongside the `default(present)` clause in the parallel regions, as this clause doesn't implicitly copy non-scalar/pointer variables.

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
```diff
@@ -7,6 +7,16 @@
...
+#ifdef MPAS_OPENACC
+use mpas_timer
+#define MPAS_ACC_TIMER_START(X) call mpas_timer_start(X)
+#define MPAS_ACC_TIMER_STOP(X) call mpas_timer_stop(X)
+#else
+#define MPAS_ACC_TIMER_START(X)
+#define MPAS_ACC_TIMER_STOP(X)
+#endif
+
+
```
abishekg7, line 19: I think we have a style convention here re. only 1 new line, instead of 2.

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
```diff
@@ -291,25 +301,54 @@ function mpas_atm_get_bdy_tend(clock, block, vertDim, horizDim, field, delta_t)
         real (kind=RKIND), dimension(vertDim,horizDim+1) :: return_tend
 
         type (mpas_pool_type), pointer :: lbc
-        integer, pointer :: idx
+        integer, pointer :: idx_ptr
         real (kind=RKIND), dimension(:,:), pointer :: tend
         real (kind=RKIND), dimension(:,:,:), pointer :: tend_scalars
-        integer :: ierr
+        integer :: idx, ierr, i, j
```
abishekg7, line 307: @mgduda `ierr` seems to be unused here. Can it be removed?

abishekg7 (changes requested):
Other than the comments above, I'm getting bit identical results with the limited area case.

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
jim-p-w: what are the semantics of adding `gang` to a `loop vector` directive?

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
jim-p-w: I'm not sure I follow. Are you saying a `default(present)` statement ensures scalars referenced in the parallel region are copied from the cpu, but that scalar pointers are not? That is, the pointer will be copied, but not what the pointer points to? And the explicit assignment ensures the scalar pointed to by the pointer gets copied over?

abishekg7: So if you'd just used an `!$acc parallel` without the `default(present)`, the compiler implicitly copies both non-scalar arrays and scalars to the device. When we add the `default(present)` here, then it only copies over scalars, and array copies to device are left up to us. In the latter scenario, the scalar pointers are treated as non-scalars/arrays and aren't copied over to the device. And then we get a runtime error about the scalar pointer not present on the device. This was at least from my (limited) experience.

Now dereferencing the pointer is good because the compiler correctly copies the dereferenced scalar onto the device.

abishekg7: `That is, the pointer will be copied, but not what the pointer points to? `

I'm not a 100% sure about this part.

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
jim-p-w: I added a timer around the parallel region and ran the regional test case, before and after adding the `gang` specifier. The timing was significantly improved using `gang`
```
timer_name                                         total       calls
mpas_atm_get_bdy_tend [compute]                    3.06004       162 (without gang)
mpas_atm_get_bdy_tend [compute]                    0.01475       162 (with gang)
```

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
jim-p-w: So how does this sound:
```
! Ensure the integer pointed to by idx_ptr is copied to the gpu device
idx = idx_ptr
```

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
abishekg7: Great! and good to know!

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
mgduda: Sure, I think it would be fine to remove `ierr`.

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
jim-p-w: OK I changed the loop directive to be `loop gang vector collapse`

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
jim-p-w: `ierr` is removed.

src/core_atmosphere/dynamics/mpas_atm_boundaries.F
abishekg7: Thanks and could you also do the same for the `loop vector collapse` in the else condition.

jim-p-w: Doh! Thanks for catching that!

jim-p-w: > Thanks and could you also do the same for the `loop vector collapse` in the else condition.

@abishekg7 I've added the `gang` directive to the loop in the else condition.

abishekg7 (approved):
Looks good!

mgduda (approved):

PR #1281, opened 2025-01-28
Disable physics for MPAS dycore-only build

kuanchihwang:
When building MPAS as a dynamical core, all physics-related components are disabled. This build configuration is usually used by CAM/CAM-SIMA, and can be achieved by defining the `MPAS_CAM_DYCORE` macro in `CPPFLAGS`.

On the other hand, the `PHYSICS` Makefile variable controls whether physics are enabled in MPAS, but its logic is decoupled from the `MPAS_CAM_DYCORE` macro. Disabling physics in MPAS currently requires manual interventions.

This PR disables physics automatically when the `MPAS_CAM_DYCORE` macro is found in `CPPFLAGS`. It mainly benefits CAM-SIMA, which reuses the build infrastructure of MPAS for building it as a dynamical core.

For stand-alone MPAS and CAM, this PR is an NFC (No Functional Change).

### DIFF
--- a/src/core_atmosphere/Makefile
+++ b/src/core_atmosphere/Makefile
@@ -4,8 +4,11 @@
 # To build a dycore-only MPAS-Atmosphere model, comment-out or delete
 # the definition of PHYSICS, below
 #
-PHYSICS=-DDO_PHYSICS
-
+# If MPAS_CAM_DYCORE is found in CPPFLAGS, PHYSICS will become undefined automatically
+#
+ifeq ($(findstring MPAS_CAM_DYCORE,$(CPPFLAGS)),)
+    PHYSICS = -DDO_PHYSICS
+endif
 
 ifdef PHYSICS
     PHYSCORE = physcore


abishekg7:
Is the standalone MPAS build supposed to work if you add `MPAS_CAM_DYCORE` or `-DMPAS_CAM_DYCORE` to the `CPPFLAGS` in the root directory's Makefile?

kuanchihwang:
@abishekg7

Not really. This build configuration is usually used by CAM(-SIMA), where MPAS will be compiled *as a part of another host model*. When the `MPAS_CAM_DYCORE` macro is defined, it introduces an [external dependency](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.2.2/src/framework/mpas_constants.F#L25-L35), which comes from the host model. Therefore, MPAS cannot be compiled independently under this build configuration.

To review this PR, I would suggest the following on Derecho:

```bash
git clone https://github.com/ESCOMP/CAM-SIMA.git
cd CAM-SIMA
git checkout development
./bin/git-fleximod update
mkdir -pv ../PR1281 && cd ../PR1281

# Compile with GNU compilers
../CAM-SIMA/cime/scripts/create_newcase --compiler gnu --case PR1281-GNU --compset FKESSLER --project PROJECT_KEY --res mpasa480_mpasa480 --run-unsupported
cd PR1281-GNU
# Compile with Intel compilers
../CAM-SIMA/cime/scripts/create_newcase --compiler intel --case PR1281-Intel --compset FKESSLER --project PROJECT_KEY --res mpasa480_mpasa480 --run-unsupported
cd PR1281-Intel

./case.setup
./case.build

# Inspect the build artifacts of MPAS at:
#   "$SCRATCH/PR1281-GNU/bld/atm/obj/mpas" for GNU compilers
#   "$SCRATCH/PR1281-Intel/bld/atm/obj/mpas" for Intel compilers
```

The changes that you are reviewing have been applied as a patch: `0002-Disable-physics-for-MPAS-dycore-only-build.patch`.

abishekg7 (approved):
I've tried building standalone MPAS - seems fine. And I also followed @kuanchihwang 's instructions for CAM-SIMA. Also seems to build fine.

mgduda (approved):

PR #1282, opened 2025-01-28
Declare constants at the native precision of MPAS

kuanchihwang:
When building MPAS as a dynamical core, certain constants in the `mpas_constants` module are imported from the `physconst` module, which is a part of CAM/CAM-SIMA. However, multiple issues arise if the precision of those constants differs from MPAS.

For example, building MPAS in single precision mode with CAM-SIMA fails due to multiple occurrences of type mismatch between actual and dummy arguments.

```
mpas_geometry_utils.F:885:157:

  885 |          call mpas_log_write('$r', MPAS_LOG_ERR, realArgs=(/mpas_triangle_signed_area_sphere(a,b,c,sphereRadius) - pii/2.0_RKIND*sphereRadius*sphereRadius/))
      |                                                                                                                                                             1
Error: Type mismatch in argument 'realargs' at (1); passed REAL(8) to REAL(4)
```

Here, `pii` is declared by CAM-SIMA to be double precision, and it causes unintended floating-point promotion in the expression.

The solution is to ensure that constants in the `mpas_constants` module are declared at the native precision of MPAS. Note that CAM does not support running MPAS in single precision mode at all, so the above build error cannot be reproduced.

In addition, some best practices of modern Fortran have been implemented.

For stand-alone MPAS, the compiled executables stay bitwise identical, with or without this PR. No further tests are needed. For CAM/CAM-SIMA, it generates bitwise identical model results, with or without this PR. Additionally for CAM-SIMA, its regression tests all pass, indicating identical model results to the previous baseline.

### DIFF
--- a/src/framework/mpas_constants.F
+++ b/src/framework/mpas_constants.F
@@ -12,7 +12,7 @@
 !> \brief   MPAS Constant Module
 !> \author  Michael Duda
 !> \date    03/27/13
-!> \details 
+!> \details
 !> This module provides various constants that can be used in different parts of MPAS.
 !> They may or may not be a physical quantity.
 !
@@ -55,6 +55,7 @@ module mpas_constants
    real (kind=RKIND), parameter :: p0 = 1.0e5_RKIND                   !< Constant: 100000 Pa
    real (kind=RKIND), parameter :: prandtl = 1.0_RKIND                !< Constant: Prandtl number
 
+
    contains
 
 
@@ -65,7 +66,7 @@ module mpas_constants
 !> \brief   Computes derived constants
 !> \author  Michael Duda
 !> \date    8 May 2020
-!> \details 
+!> \details
 !>  This routine provides a place where physical constants provided by
 !>  the mpas_constants module may be computed at runtime. For example,
 !>  if some constants depend on namelist options or other runtime
--- a/src/framework/mpas_constants.F
+++ b/src/framework/mpas_constants.F
@@ -20,7 +20,7 @@
 
 module mpas_constants
 
-   use mpas_kind_types
+   use mpas_kind_types, only: RKIND
 
    implicit none
 
--- a/src/framework/mpas_constants.F
+++ b/src/framework/mpas_constants.F
@@ -22,6 +22,8 @@ module mpas_constants
 
    use mpas_kind_types
 
+   implicit none
+
 #ifdef MPAS_CAM_DYCORE
    ! Set at run-time by `mpas_constants_compute_derived`.
    real (kind=RKIND), protected :: pii     = huge(1.0_RKIND)
@@ -83,8 +85,6 @@ module mpas_constants
        use physconst, only: external_rgas => rair
        use physconst, only: external_rv => rh2o
        use physconst, only: external_cp => cpair
-
-       implicit none
 
        ! Convert external constants to the native precision of MPAS (i.e., `RKIND`).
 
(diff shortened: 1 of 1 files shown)


src/framework/mpas_constants.F
```diff
@@ -12,27 +12,33 @@
...
 
 module mpas_constants
 
-   use mpas_kind_types
+   use mpas_kind_types, only: RKIND
+
+   implicit none
+
+   public
+   private :: RKIND
```
abishekg7, line 28: Is there a significance for this line?

src/framework/mpas_constants.F
```diff
@@ -51,15 +57,14 @@ module mpas_constants
 
    contains
 
-
```
abishekg7, line 54: We may want this newline preserved I think

src/framework/mpas_constants.F
```diff
@@ -74,9 +79,25 @@ module mpas_constants
 !-----------------------------------------------------------------------
    subroutine mpas_constants_compute_derived()
 
-       implicit none
-
```
abishekg7, line 78: Is there a specific reason for removing the `implicit none` here?

abishekg7 (commented):
Overall looks good. I did have a few comments and question though.

src/framework/mpas_constants.F
abishekg7: Looking at the commit history, my question is answered

src/framework/mpas_constants.F
kuanchihwang: I think this question might also be answered by the commit message, quoted below:

> It is a best practice in Fortran to always specify the accessibility of module variables (as well as procedures).
>
> For the case here, `RKIND` is imported from `mpas_kind_types`. Due to the lack of accessibility attributes, other modules that `use mpas_constants` also have access to `RKIND`.
>
> Therefore, avoid causing unexpected namespace pollution to other modules by specifying accessibility attributes.

src/framework/mpas_constants.F
kuanchihwang: Yes, `implicit none` has been moved to the module level so it applies everywhere within.

src/framework/mpas_constants.F
abishekg7: that makes sense. Thanks!

src/framework/mpas_constants.F
kuanchihwang: I think the newline removal is warranted. Elsewhere in this module, there is only one newline around constructs and statements. The only exception is found here, there are two consecutive ones.

mgduda: In most framework and atmosphere modules, I think we tend to use two blank lines before and after a `contains` statement, and two blank lines between routines (though this module contains just one routine). So if we were to modify whitespace here, I'd suggest adding a second blank line above the `contains` statement rather than deleting the second blank line below.

kuanchihwang: I have amended the commit in https://github.com/MPAS-Dev/MPAS-Model/pull/1282/commits/a0f530b27fab8ededf97043c3f4335c7dcf268c5. It now adds a second blank line above `contains`.

abishekg7: Good to know. Thanks!

abishekg7 (approved):
Looks good!

mgduda (approved):

kuanchihwang:
I have amended the commit messages to include more specific information about what they modify. The commit contents remain the same as can be seen by [comparison](https://github.com/MPAS-Dev/MPAS-Model/compare/a0f530b27fab8ededf97043c3f4335c7dcf268c5..1353687c5b279cdd01160f4f5ba32a4810e398e1).

PR #1283, opened 2025-01-29
bug fixed in driver convection

lo-y-wni:
There is index error in variable `rvcuten` whose index is set to `(k,k)` in loops. But it should be `(k,i) ` because similar variable `rucuten` has the index `(k,i)`.
This PR will affect simulation result if cumulus scheme is enabled.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_convection.F
@@ -744,7 +744,7 @@
              rqvblten_p(i,k,j)  = rqvblten(k,i)
              rqvdynten_p(i,k,j) = rqvdynten(k,i)
              rucuten_p(i,k,j)   = rucuten(k,i)
-             rvcuten_p(i,k,j)   = rvcuten(k,k)
+             rvcuten_p(i,k,j)   = rvcuten(k,i)
           enddo
        enddo
        enddo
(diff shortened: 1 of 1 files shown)


weiwangncar:
@lo-y-wni Although it is wrong in the code, but it appears in the sections of code for Grell-Freitas scheme which actually does not have momentum tendencies or rvcuten and rucuten.

mgduda:
@weiwangncar Although the Grell-Freitas scheme in MPAS-A doesn't provide momentum tendencies, would it still be worth fixing the incorrect indexing in the code?

weiwangncar:
@mgduda Yes, it definitely should be fixed.

mgduda:
@lo-y-wni Thanks very much for finding this indexing error. Would you mind including a bit more detail in your commit message? Just adding this sentence from your PR description should be enough: "There is index error in variable rvcuten whose index is set to (k,k) in loops. But it should be (k,i) because similar variable rucuten has the index (k,i)." But, it might also be worth noting that because the Grell-Freitas does not provide momentum tendencies, there is no change to simulation results.

To modify the commit message, with your `bug_fix` branch checked out, you can run `git commit --amend`. If you have any questions at all, please don't hesitate to ask.

lo-y-wni:
Hi Michael @mgduda,
Thank you for the response. I updated the PR description as your suggestion!

mgduda:
> Hi Michael @mgduda, Thank you for the response. I updated the PR description as your suggestion!

Excellent -- thanks so much!

mgduda (approved):

PR #1284, opened 2025-01-29
Initial testing framework simplified

islas:
This PR introduces testing capabilities through a series of compartmentalized commits :
1. Setting up helper scripts to handle environment loading
2. Compilation test script able to exercise the make build
3. A git submodule to the [hpc-workflows](https://github.com/islas/hpc-workflows)  testing framework
4. A test definition config using hpc-workflows

The tests can be run using the following command :
`.ci/hpc-workflows/.ci/runner.py .ci/mpas_compilation.jsonc -t gnu`

If this is run in an environment noted to have PBS/SLURM (as seen by the line in `mpas_compilation.jsonc` : `"hsn.de.hpc" : { "submission" : "PBS" } `) an account key must be provided to utilize job submission. Local testing in within the login node can be forced but should be used with caution.

This PR is meant to supersede PR #1218 as a smaller incremental change.

### DIFF
--- a/.ci/mpas_compilation.jsonc
+++ b/.ci/mpas_compilation.jsonc
@@ -1,0 +1,48 @@
+// MPAS Compilation tests
+{
+  "submit_options" :
+  {
+    // Default values
+    "timelimit" : "00:15:00",
+    "working_directory" : "..",
+    "arguments" :
+    {
+      "base_env_numprocs"      : [ "-e", "NUM_PROCS=4" ],
+
+      ".*make.*::args_build_core"     : [ "-c", "atmosphere" ],
+      ".*debug.*::args_build_debug"     : [ "-d" ],
+      ".*make.*::args_build_options"     : [ "-b", "-j $NUM_PROCS" ],
+      ".*make.*gnu.*::args_target"     : [ "-t", "gnu" ]
+    },
+    // Derecho-specifics
+    "hsn.de.hpc"  :
+    {
+      "submission" : "PBS",
+      "queue"      : "main",
+      "hpc_arguments"  : 
+      {
+        "node_select" : { "-l " : { "select"       : 1, "ncpus" : 16 } },
+        "priority"    : { "-l " : { "job_priority" : "economy"       } }
+      },
+      "arguments"  : 
+      {
+        // We want NUM_PROCS to match ncpus
+        "base_env_numprocs"      : [ "-e", "NUM_PROCS=16" ],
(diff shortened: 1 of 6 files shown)


.ci/mpas_compilation.jsonc
```diff
@@ -0,0 +1,48 @@
...
+      {
+        "node_select" : { "-l " : { "select"       : 1, "ncpus" : 16 } },
+        "priority"    : { "-l " : { "job_priority" : "economy"       } }
+      },
+      "arguments"  : 
+      {
+        // We want NUM_PROCS to match ncpus
+        "base_env_numprocs"      : [ "-e", "NUM_PROCS=16" ],
+        "very_last_modules"       : [  "cray-mpich", "parallel-netcdf" ],
+        ".*gnu.*::test_modules"   : [ "gcc/12.2.0" ]
```
abishekg7, line 32: One part where I've had some troubles with is the ordering of the arguments that are passed to the script we are running (`build.sh` here). If I'm not careful enough, there are sometimes options I'm trying to pass to the script (say "-z option1") that get misconstrued as modules to be loaded, etc. 

Probably related to the ordering of options which are defined in these regex statements (".*make.*::args_build_options"). I usually avoid this issue by specifying the arguments locally after every command ("command" : ".ci/tests/build.sh"). Not sure if I what I wrote is very clear, but I'll try to clarify my comment with a better example.

islas: I think I understand. This is probably due to the way I have it setup where on the first unknown arg it encounters `getopts` exits and forwards the rest to the environment script. 

The order passed in will be determined by the alphabetical order of the non-regex portions, deconflicted by order of appearance in the config file. I'm not sure if you've already found this documentation (https://github.com/islas/hpc-workflows/blob/main/tutorials/AdvancedTestConfig_regex_argpacks.md#simple-regex-argpack-config) at the end of that section there is a note about this.

islas: It is, admittedly, a very fragile setup relying on argument order between what the test script needs and what the environment script needs..

Perhaps a better approach would be something like the `-e env` option where things are accumulated and then handled afterwards. In that way multiple things can be passed to then env script without relying on order

abishekg7: Thanks @islas! I don't think I've looked at this specific section of your documentation. I will go through that. 

I think it's a relatively minor issue, and as long as there's some documentation, I'd be fine with things as they are.

PR #1285, opened 2025-01-30
Prefix all namelist group and option names for MPAS dycore

kuanchihwang:
When MPAS is used as a dynamical core in a host model (e.g., CAM/CAM-SIMA), it needs to share the namelist file with other model components. As a result, MPAS namelist groups and options may not be easily recognizable at first sight. The solution, which is implemented by this PR, is to add a unique identifier to all MPAS namelist group and option names.

The following transformations are performed for each MPAS namelist group and option name:

1. Leading `config_` is removed recursively from the name. Case-insensitive.
2. Leading `mpas_` is removed recursively from the name. Case-insensitive.
3. Prepend `mpas_` to the name.

By doing so, it is now easier to distinguish MPAS namelist groups and options from host model ones. The possibility of name collisions with host model ones is also resolved once and for all. Note that only namelist I/O is affected. Internally, MPAS still refers to its namelist options by their original names due to compatibility reasons.

Compared to the implementation in CAM, this PR applies the "name mangling" logic in an algorithmic and predictable way. It works automatically even if there are additions, modifications, or deletions to MPAS namelist groups and options.

The "name mangling" logic is entirely guarded behind the `MPAS_CAM_DYCORE` macro. For stand-alone MPAS, its namelist groups and options keep their original names as in the registry. For CAM, it is not affected by this PR because it does not use the code generation functionality of MPAS for namelist reading at all. For CAM-SIMA, the Fortran include files generated by the `parse` utility stay the same. Its regression tests all pass, indicating identical model results to the previous baseline.

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -15,6 +15,10 @@
 #include "fortprintf.h"
 #include "utility.h"
 
+#ifdef MPAS_CAM_DYCORE
+#include <ctype.h>
+#endif
+
 void process_core_macro(const char *macro, const char *val, va_list ap);
 void process_domain_macro(const char *macro, const char *val, va_list ap);
 
@@ -696,8 +700,12 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 	ezxml_t nmlrecs_xml, nmlopt_xml;
 
 	const char *const_core;
-	const char *nmlrecname, *nmlrecindef, *nmlrecinsub;
-	const char *nmloptname, *nmlopttype, *nmloptval, *nmloptunits, *nmloptdesc, *nmloptposvals, *nmloptindef;
+	const char *original_nmlrecname, *nmlrecindef, *nmlrecinsub;
+	const char *original_nmloptname, *nmlopttype, *nmloptval, *nmloptunits, *nmloptdesc, *nmloptposvals, *nmloptindef;
+
+	// Fortran variable names have a length limit of 63 characters. + 1 for the terminating null character.
+	char nmlrecname[64];
+	char nmloptname[64];
 
 	char pool_name[1024];
 	char core_string[1024];
@@ -743,7 +751,9 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 
 	// Parse Namelist Records
 	for (nmlrecs_xml = ezxml_child(registry, "nml_record"); nmlrecs_xml; nmlrecs_xml = nmlrecs_xml->next){
-		nmlrecname = ezxml_attr(nmlrecs_xml, "name");
+		original_nmlrecname = ezxml_attr(nmlrecs_xml, "name");
+		mangle_name(nmlrecname, sizeof(nmlrecname), original_nmlrecname);
+
 		nmlrecindef = ezxml_attr(nmlrecs_xml, "in_defaults");
 		nmlrecinsub = ezxml_attr(nmlrecs_xml, "in_subpool");
 
@@ -777,7 +787,9 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 
 		// Define variable definitions prior to reading the namelist in.
 		for (nmlopt_xml = ezxml_child(nmlrecs_xml, "nml_option"); nmlopt_xml; nmlopt_xml = nmlopt_xml->next){
-			nmloptname = ezxml_attr(nmlopt_xml, "name");
+			original_nmloptname = ezxml_attr(nmlopt_xml, "name");
+			mangle_name(nmloptname, sizeof(nmloptname), original_nmloptname);
+
 			nmlopttype = ezxml_attr(nmlopt_xml, "type");
 			nmloptval = ezxml_attr(nmlopt_xml, "default_value");
 			nmloptunits = ezxml_attr(nmlopt_xml, "units");
@@ -809,7 +821,9 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 		// Define the namelist block, to read the namelist record in.
 		fortprintf(fd, "      namelist /%s/ &\n", nmlrecname);
 		for (nmlopt_xml = ezxml_child(nmlrecs_xml, "nml_option"); nmlopt_xml; nmlopt_xml = nmlopt_xml->next){
-			nmloptname = ezxml_attr(nmlopt_xml, "name");
+			original_nmloptname = ezxml_attr(nmlopt_xml, "name");
+			mangle_name(nmloptname, sizeof(nmloptname), original_nmloptname);
+
 			if(nmlopt_xml->next){
 				fortprintf(fd, "         %s, &\n", nmloptname);
 			} else {
@@ -840,7 +854,9 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 		// Define broadcast calls for namelist values.
 		fortprintf(fd, "      if (ierr <= 0) then\n");
 		for (nmlopt_xml = ezxml_child(nmlrecs_xml, "nml_option"); nmlopt_xml; nmlopt_xml = nmlopt_xml->next){
-			nmloptname = ezxml_attr(nmlopt_xml, "name");
+			original_nmloptname = ezxml_attr(nmlopt_xml, "name");
+			mangle_name(nmloptname, sizeof(nmloptname), original_nmloptname);
+
 			nmlopttype = ezxml_attr(nmlopt_xml, "type");
 
 			if(strncmp(nmlopttype, "real", 1024) == 0){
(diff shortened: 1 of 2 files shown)


kuanchihwang:
> Just started looking at this. One question I have is why the leading `mpas_` and `config_ ` strings need to be removed recursively and not just once.

Performing the removal operation only once would also work. However, the recursion should make it more robust in the unlikely case that a namelist option with duplicate prefixes is introduced to MPAS registry.

abishekg7:
> > Just started looking at this. One question I have is why the leading `mpas_` and `config_ ` strings need to be removed recursively and not just once.
> 
> Performing the removal operation only once would also work. However, the recursion should make it more robust in the unlikely case that a namelist option with duplicate prefixes is introduced to MPAS registry.

Thanks @kuanchihwang. The code itself looks fine, but I was wondering how to go about testing this implementation. You provided some CAM-SIMA build instructions earlier - can I use those?

kuanchihwang:
@abishekg7

Sure! I will also provide the instructions here for clarity and convenience:

1. Output the changes in this PR as a patch by `git format-patch` (`git diff` also works).
2. Follow the steps below on Derecho:

```bash
git clone https://github.com/ESCOMP/CAM-SIMA.git
cd CAM-SIMA
git checkout development
./bin/git-fleximod update
# Replace the patch at "src/dynamics/mpas/assets/0001-Prefix-all-MPAS-namelist-group-and-option-names.patch" with the one that you generated in step 1.
mkdir -pv ../PR1285 && cd ../PR1285

# Compile with GNU compilers
../CAM-SIMA/cime/scripts/create_newcase --compiler gnu --case PR1285-GNU --compset FKESSLER --project PROJECT_KEY --res mpasa480_mpasa480 --run-unsupported
cd PR1285-GNU
# Compile with Intel compilers
../CAM-SIMA/cime/scripts/create_newcase --compiler intel --case PR1285-Intel --compset FKESSLER --project PROJECT_KEY --res mpasa480_mpasa480 --run-unsupported
cd PR1285-Intel

./case.setup
./case.build

# Inspect the build artifacts of MPAS at:
#   "$SCRATCH/PR1285-GNU/bld/atm/obj/mpas" for GNU compilers
#   "$SCRATCH/PR1285-Intel/bld/atm/obj/mpas" for Intel compilers
#
# Specifically, inspect the "namelist_call.inc" and "namelist_defines.inc" include files at:
#   "$SCRATCH/PR1285-GNU/bld/atm/obj/mpas/core_atmosphere/inc" for GNU compilers
#   "$SCRATCH/PR1285-Intel/bld/atm/obj/mpas/core_atmosphere/inc" for Intel compilers
#
# You will find that MPAS group and option names have been prefixed with "mpas_".
```

Similarly, this PR is also an attempt to upstream that particular patch so we do not have to maintain it downstream at CAM-SIMA. Thanks so much for your time!

abishekg7:
Thanks for the instructions @kuanchihwang . I follow them and inspected the `namelist_defines.inc` for both the builds - Looks good to me. 

The only comment I have is re. the commit message. Could you reference (CAM/CAM-SIMA) in the title of the commit message and also the first line inside the message. i.e it ends with "When MPAS is used as a dynamical core". Could you add  "in a host model (e.g., CAM/CAM-SIMA)" right after this, just like in the PR description.

kuanchihwang:
@abishekg7

Thanks for the comments! I have updated the commit title and message to mention CAM/CAM-SIMA.

abishekg7 (approved):
Looks good!

src/tools/registry/gen_inc.c
```diff
@@ -699,6 +703,14 @@ int parse_namelist_records_from_registry(ezxml_t registry)/*{{{*/
 	const char *nmlrecname, *nmlrecindef, *nmlrecinsub;
 	const char *nmloptname, *nmlopttype, *nmloptval, *nmloptunits, *nmloptdesc, *nmloptposvals, *nmloptindef;
 
+#ifdef MPAS_CAM_DYCORE
```
mgduda, line 706: I'm interested in discussing the idea further, but rather than adding `#ifdef` directives in many places throughout the code, what if we were to frame the `transform_name` function in a more general way, as something that transforms namelist variable and record names as appropriate for the host model. For CAM-SIMA, this means removing any `config_` and `mpas_` prefixes, then prepending `mpas_`. For stand-alone MPAS-A, this means returning the original namelist variable or record name. Then, we would always invoke `transform_name`, and the implementation of the `transform_name` function would be the only place where we have `#ifdef MPAS_CAM_DYCORE`?

src/tools/registry/gen_inc.c
```diff
@@ -2532,3 +2590,32 @@ int parse_structs_from_registry(ezxml_t registry)/*{{{*/
 
 	return 0;
 }/*}}}*/
+
+
+#ifdef MPAS_CAM_DYCORE
+// Perform transformations for namelist group and option names.
```
mgduda, line 2596: We have obviously not done a good job of documenting functions in the past, but I think we're getting better (see, e.g., 3bd1cdd4, which modifies the `src/registry/utility.c` code). Especially if we rethink the usage of the `transform_name` function, I think it might be nice to include more extensive documentation here.

src/tools/registry/gen_inc.c
```diff
@@ -2532,3 +2590,32 @@ int parse_structs_from_registry(ezxml_t registry)/*{{{*/
...
+void transform_name(char *new_name, const size_t new_name_size, const char *old_name) {
```
mgduda, line 2597: Although many of the original functions in this file place the opening `{` on the same line, newer code adheres to this guidance from the Linux kernel coding style:

> However, there is one special case, namely functions: they have the opening brace at the beginning of the next line, thus:
> ```
> int function(int x)
> {
>         body of function
> }
> ```
> Heretic people all over the world have claimed that this inconsistency is ... well ... inconsistent, but all right-thinking people know that (a) K&R are right and (b) K&R are right. Besides, functions are special anyway (you can’t nest them in C).

src/tools/registry/gen_inc.c
```diff
@@ -2532,3 +2590,32 @@ int parse_structs_from_registry(ezxml_t registry)/*{{{*/
...
+
+    // Remove all leading "config_" by moving pointer forward.
+    while (strncasecmp(old_name, old_prefix, strlen(old_prefix)) == 0) old_name += strlen(old_prefix);
+
+    // Remove all leading "mpas_" by moving pointer forward.
+    while (strncasecmp(old_name, new_prefix, strlen(new_prefix)) == 0) old_name += strlen(new_prefix);
+
+    *new_name = '\0';
+    size = snprintf(NULL, 0, "%s%s", new_prefix, old_name) + 1;
+    snprintf(new_name, size > new_name_size ? new_name_size : size, "%s%s", new_prefix, old_name);
```
mgduda, line 2615: Since we are given that the size of `new_name` is `new_name_size`, is there any need for the ternary operator here (as well as the preceding line to compute `size`)? Would the following work just as well?
```
    snprintf(new_name, new_name_size, "%s%s", new_prefix, old_name);
```

mgduda (changes requested):

src/tools/registry/gen_inc.c
kuanchihwang: I have removed the redundant computation for `size` in https://github.com/MPAS-Dev/MPAS-Model/pull/1285/commits/2932c7f13ed55ff4b1ac2118c4ae3efa3edf5da4.

src/tools/registry/gen_inc.c
kuanchihwang: I have put the opening brace on its own line in https://github.com/MPAS-Dev/MPAS-Model/pull/1285/commits/e04b8872b9def730b2a62467fcefec09c39627b6.

src/tools/registry/gen_inc.c
kuanchihwang: As above, collectively addressed in https://github.com/MPAS-Dev/MPAS-Model/pull/1285/commits/7e12d19d06a483144553a2798d02df83614f0a2c.

src/tools/registry/gen_inc.c
kuanchihwang: I have renamed and reworked the `mangle_name` function in https://github.com/MPAS-Dev/MPAS-Model/pull/1285/commits/7e12d19d06a483144553a2798d02df83614f0a2c. Now, it is always called for each MPAS namelist group and option name. If `MPAS_CAM_DYCORE` is defined, the name mangling will be effective. Otherwise, if `MPAS_CAM_DYCORE` is not defined, MPAS namelist groups and options will keep their original names as in the registry.

mgduda: Thanks for the updates! I think having fewer `#ifdef`s in the code will help readability.

src/tools/registry/gen_inc.c
mgduda: Thanks! The documentation for `mangle_name` looks great.

mgduda (approved):
Thanks for the changes -- they look good to me!

Before I merge this PR, I think it would be fine to squash the fixup commits into the first commit of the branch.

kuanchihwang:
Thanks! I have squashed the fixup commits into the first one. Additionally, the commit message has been updated to include the clarification that for stand-alone MPAS, its namelist groups and options keep their original names as in the registry. The commit contents remain the same as can be seen by [comparison](https://github.com/MPAS-Dev/MPAS-Model/compare/7e12d19d06a483144553a2798d02df83614f0a2c..71f3d2220af9d0e3c28c98c55945643f2e41cc29).

PR #1290, opened 2025-02-20
Provide consistency in setting of prevRingTime for alarms when changing ref time

mgduda:
This PR aims to provide consistency in the ringing behavior of recurring alarms after their reference time has been adjusted with a call to `mpas_adjust_alarm_to_reference_time`. Now, adjusting the reference time for an alarm will always leave that alarm in a state such that it is considered by the `mpas_is_alarm_ringing` routine to be ringing at the current time.

For simplicity of discussion, consider only a clock that is running forward in time. The objective of the `mpas_adjust_alarm_to_reference_time` routine is to set the `prevRingTime` (previous ring time) of a recurring alarm to (1) a time that differs from the specified reference time by an integer multiple of the alarm interval; and (2) the latest such time that is not after the current time on the clock.

Prior to this PR, the logic in the `mpas_adjust_alarm_to_reference_time` routine resulted in one of two outcomes:

(1) If the difference between the reference time and the current time is divisible by the alarm interval, the `prevRingTime` becomes the current time, as illustrated below.
```
    ring            ring            ring            ring
     v               v               v               v
  -----------------------------------------------------------
                                     ^
                                    now
                                prevRingTime
```
(2) Otherwise, the `prevRingTime` becomes the latest time before the current time that lies on an integer multiple of the alarm interval away from the reference time, as illustrated below.
```
    ring            ring            ring            ring
     v               v               v               v
  -----------------------------------------------------------
                     ^           ^
                prevRingTime    now
```
To determine whether a recurring alarm is ringing, the alarm's interval is added to the alarm's `prevRingTime`. If the result is on or before the current time, the alarm is ringing; otherwise, if the result is later than the current time, the alarm is not ringing. As a consequence, outcome (1) from the `mpas_adjust_alarm_to_reference_time` leads to an alarm that is ringing after the call to the routine, while outcome(2) leads to an alarm that is not yet ringing.

In order to rectify the inconsistency in whether an alarm is ringing depending on where its reference time is set relative to the current time, the `prevRingTime` for an alarm is always set to be at least one full alarm interval before the current time. Whether the difference between the current time and the reference time is evenly divisible by the alarm's ring interval or not, a query of the alarm's status will always show that it is ringing.

This PR makes changes to the logic for both a forward and a backward running clock in the `mpas_adjust_alarm_to_reference_time`, although it appears that no code actually makes use of a backward running clock at present.

This PR also adds unit tests for the `mpas_adjust_alarm_to_reference_time` routine in the `test` core.

### DIFF
--- a/src/core_test/mpas_test_core_timekeeping_tests.F
+++ b/src/core_test/mpas_test_core_timekeeping_tests.F
@@ -5,6 +5,9 @@
 ! Additional copyright and license information can be found in the LICENSE file
 ! distributed with this code, or at http://mpas-dev.github.com/license.html
 !
+
+#define MPAS_ADJUST_ALARM_VERBOSE( M ) ! M
+
 module test_core_timekeeping_tests
 
    use mpas_derived_types
@@ -19,7 +22,8 @@ module test_core_timekeeping_tests
    implicit none
    private
 
-   public :: test_core_test_intervals
+   public :: test_core_test_intervals, &
+             mpas_adjust_alarm_tests
 
    contains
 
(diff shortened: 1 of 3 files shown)


mgduda:
@abishekg7 @gdicker1 For more background, the fix in this PR to the `mpas_adjust_alarm_to_reference_time` routine are needed so that we can properly restart limited-area simulations between LBC update times (for example, restart a simulation at 01:30 UTC when LBCs are updated every three hours at 00:00, 03:00, 06:00, etc.)

mgduda:
When reviewing the changes to the `mpas_adjust_alarm_to_reference_time` routine, the diff may be a bit confusing, so it may be easier to just look at the new state of the routine (and then to take out a pencil and paper and make some drawings).

gdicker1:
After noodling this around a bit, the changes make sense.

I think the first commit message could use some work. It would at least be nice if the newly consistent behavior was added to the first paragraph, even if the terms will be defined or expanded on more later. That said, I'm having trouble phrasing it. Maybe something like:

> ... for a recurring alarm. After mpas_adjust_alarm_to_reference_time, the prevRingTime for the alarm is now consistently set so that it is one ringing interval away from the current time and is now set so that the alarm is ringing.

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -178,4 +182,283 @@ subroutine test_core_interval_test(ref_str, int1_str, int2_str, expected_divs, e
...
+      write(test_mesg, '(a)') '  forward clock, ref_time < now, now is NOT on new alarm time grid: '
+      MPAS_ADJUST_ALARM_VERBOSE( call mpas_log_write('=================================') )
+      MPAS_ADJUST_ALARM_VERBOSE( call mpas_log_write('Setting ref time to 1999-06-15_08') )
+      MPAS_ADJUST_ALARM_VERBOSE( call mpas_log_write('---------------------------------') )
+      call mpas_set_time(test_refTime, YYYY=1999, MM=6, DD=15, H=8, M=0, S=0, S_n=0, S_d=0, ierr=ierr_local)
+      call mpas_adjust_alarm_to_reference_time(test_clock, 'foobar', test_refTime, ierr_local)
+      MPAS_ADJUST_ALARM_VERBOSE( call mpas_print_alarm(test_clock, 'foobar', ierr_local) )
+      if (mpas_is_alarm_ringing(test_clock, 'foobar', ierr=ierr_local)) then
+          MPAS_ADJUST_ALARM_VERBOSE( call mpas_log_write('-> Alarm is RINGING') )
+          test_mesg = trim(test_mesg)//' SUCCESS'
```
abishekg7, line 340: I think the non-verbose output message could also include some context about which test exactly was successful, i.e. Line 339 could be combined with 340 to be the regular (non-verbose) output.

abishekg7:
The changes to `mpas_adjust_alarm_to_reference_time` and the unit test look good, just had a minor comment.

mgduda:
> After noodling this around a bit, the changes make sense.
> 
> I think the message for the commit modifying mpas_adjust_alarm_to_reference_time could use some work. I think it would at least be nice if the newly consistent behavior was added to the first paragraph, even if the terms will be defined or expanded on more later. Though, I'm having trouble phrasing it (maybe not understanding as well as I'd like). Could be something like:
> 
> > ... for a recurring alarm. After mpas_adjust_alarm_to_reference_time, the prevRingTime for the alarm is now consistently set so that it is one ringing interval away from the current time and prevRingTime is now set so that the alarm is ringing.

How about the following for the first part of the commit message?
```
This commit aims to provide consistency in the ringing behavior of recurring 
alarms after their reference time has been adjusted with a call to
`mpas_adjust_alarm_to_reference_time`. Now, adjusting the reference time for an
alarm will always leave that alarm in a state such that it is considered by the
`mpas_is_alarm_ringing` routine to be ringing at the current time.
```

src/core_test/mpas_test_core_timekeeping_tests.F
mgduda: Here's what the log output looks like for these tests at present:
```
 Testing mpas_adjust_alarm_to_reference_time:
   forward clock, ref_time < now, now is on new alarm time grid: SUCCESS
   forward clock, ref_time > now, now is on new alarm time grid: SUCCESS
   forward clock, ref_time = now, now is on new alarm time grid: SUCCESS
   forward clock, ref_time < now, now is NOT on new alarm time grid: SUCCESS
   forward clock, ref_time > now, now is NOT on new alarm time grid: SUCCESS
   backward clock, ref_time < now, now is on new alarm time grid: SUCCESS
   backward clock, ref_time > now, now is on new alarm time grid: SUCCESS
   backward clock, ref_time = now, now is on new alarm time grid: SUCCESS
   backward clock, ref_time < now, now is NOT on new alarm time grid: SUCCESS
   backward clock, ref_time > now, now is NOT on new alarm time grid: SUCCESS
 * mpas_adjust_alarm_tests tests - all tests passed: SUCCESS
```
And here's what the log output would look like if we added the `-> Alarm is RINGING` messages:
```
 Testing mpas_adjust_alarm_to_reference_time:
 -> Alarm is RINGING
   forward clock, ref_time < now, now is on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   forward clock, ref_time > now, now is on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   forward clock, ref_time = now, now is on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   forward clock, ref_time < now, now is NOT on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   forward clock, ref_time > now, now is NOT on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   backward clock, ref_time < now, now is on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   backward clock, ref_time > now, now is on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   backward clock, ref_time = now, now is on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   backward clock, ref_time < now, now is NOT on new alarm time grid: SUCCESS
 -> Alarm is RINGING
   backward clock, ref_time > now, now is NOT on new alarm time grid: SUCCESS
 * mpas_adjust_alarm_tests tests - all tests passed: SUCCESS
```

abishekg7: Right. The second version may not be necessarily clearer, so we can just keep it as it currently is.

gdicker1:
@mgduda, that looks great!

gdicker1 (approved):
Trusting that the commit message will be updated, this PR has my approval.

abishekg7 (approved):

mgduda:
> Trusting that the commit message will be updated, this PR has my approval.

I've updated the commit message as well as the PR description.

PR #1291, opened 2025-02-26
Fix error reporting when config_nfglevels is insufficiently large

abishekg7:
In init_atmosphere, during the interpolation of first-guess meteorological fields from the intermediate files (`config_init_case` = 7 or 9), the value of `config_nfglevels` must be set to a minimum of the number of vertical levels in the intermediate file. Currently, the routines `init_atm_case_gfs` and `init_atm_case_lbc` have a logic to throw an error and exit if `config_nfglevels` is not sufficiently large. In practice, this error message is not reported consistently with intermediate files containing many levels and/or many MPI ranks being used.

This PR fixes the inconsistent error reporting issue by adding the required `deallocate` statements before the repeated calls to read_next_met_field in the case of `config_nfglevels` being insufficiently large. Without the `deallocate` one or more MPI ranks crash during the read of the `field % slab` from the intermediate file. This typically results in a crash of init_atmosphere without any useful error messages in the log files.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -3547,6 +3547,7 @@ module init_atm_cases
                ! many unique levels are found using the code above
                !
                if (too_many_fg_levs) then
+                  deallocate(field % slab)
                   call read_next_met_field(field, istatus)
                   cycle
                end if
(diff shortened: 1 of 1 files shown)


abishekg7:
Tested `config_init_case=7` with 24km and 480km grids using an `ERA5:2010-10-23_00` intermediate file with 138 levels, both cases running 128 MPI ranks on 1 node. The tests with the `develop` branch failed with no error messages, whereas the test with this PR crashed after correctly reporting the error message about `config_nfglevels` being smaller than 138.

mgduda (approved):

PR #1294, opened 2025-03-21
Enable GPU exection of summarize_timestep via OpenACC

gdicker1:
This PR modifies the code and adds OpenACC directives to allow the loops in the `summarize_timestep` routine can execute on GPUs.

Timing information for the temporary OpenACC data transfers in this routine is captured in the log file by a new timer: 'summarize_timestep [ACC_data_xfer]'.

This PR includes a re-write of the loops associated with `config_print_detailed_minmax_vel` to ensure consistent locations are reported for the minimum and maximum values for the wind speeds.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -7265,6 +7265,7 @@ module atm_time_integration
        real (kind=RKIND), dimension(5) :: localVals, globalVals
 
        real (kind=RKIND), dimension(:,:), allocatable :: spd
+       !$acc declare create(spd)
        real (kind=RKIND), dimension(:,:), pointer :: w
        real (kind=RKIND), dimension(:,:), pointer :: u, v, uReconstructZonal, uReconstructMeridional, uReconstructX, uReconstructY, uReconstructZ
        real (kind=RKIND), dimension(:,:,:), pointer :: scalars, scalars_1, scalars_2
@@ -7290,6 +7291,17 @@ module atm_time_integration
       nVertLevels = nVertLevels_ptr
       num_scalars = num_scalars_ptr
 
+      MPAS_ACC_TIMER_START('summarize_timestep [ACC_data_xfer]')
+      if (config_print_detailed_minmax_vel) then
+         !$acc enter data copyin(w,u,v)
+      else if (config_print_global_minmax_vel) then
+         !$acc enter data copyin(w,u)
+      end if
+      if (config_print_global_minmax_sca) then
+         !$acc enter data copyin(scalars)
+      end if
+      MPAS_ACC_TIMER_STOP('summarize_timestep [ACC_data_xfer]')
+
       if (config_print_detailed_minmax_vel) then
          call mpas_log_write('')
 
@@ -7309,14 +7321,19 @@ module atm_time_integration
          kMax = -1
          latMax = 0.0
          lonMax = 0.0
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:scalar_min)
          do iCell = 1, nCellsSolve
          do k = 1, nVertLevels
             scalar_min = min(scalar_min, w(k,iCell))
          end do
          end do
+         !$acc end parallel
 
          ! This second loop using offset_1d ensures the same (kMax,indexMax) are reported with the scalar_min
          ! Especially when using OpenACC
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:offset_1d)
          do iCell = 1, nCellsSolve
          do k = 1, nVertLevels
             if (w(k,iCell) == scalar_min) then
@@ -7325,6 +7342,7 @@ module atm_time_integration
             end if
          end do
          end do
+         !$acc end parallel
          kMax = mod(offset_1d, size(w,1)) + 1
          indexMax = (offset_1d / size(w,1)) + 1 ! Integer divide
          latMax = latCell(indexMax)
@@ -7355,12 +7373,17 @@ module atm_time_integration
          kMax = -1
          latMax = 0.0
          lonMax = 0.0
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(max:scalar_max)
          do iCell = 1, nCellsSolve
          do k = 1, nVertLevels
             scalar_max = max(scalar_max, w(k,iCell))
          end do
          end do
+         !$acc end parallel
 
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:offset_1d)
          do iCell = 1, nCellsSolve
          do k = 1, nVertLevels
             if (w(k,iCell) == scalar_max) then
@@ -7368,6 +7391,7 @@ module atm_time_integration
             end if
          end do
          end do
+         !$acc end parallel
          kMax = mod(offset_1d, size(w,1)) + 1
          indexMax = (offset_1d / size(w,1)) + 1
          latMax = latCell(indexMax)
@@ -7398,12 +7422,17 @@ module atm_time_integration
          kMax = -1
          latMax = 0.0
          lonMax = 0.0
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:scalar_min)
          do iEdge = 1, nEdgesSolve
          do k = 1, nVertLevels
             scalar_min = min(scalar_min, u(k,iEdge))
          end do
          end do
+         !$acc end parallel
 
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:offset_1d)
          do iEdge = 1, nEdgesSolve
          do k = 1, nVertLevels
             if (u(k,iEdge) == scalar_min) then
@@ -7411,6 +7440,7 @@ module atm_time_integration
             end if
          end do
          end do
+         !$acc end parallel
          kMax = mod(offset_1d, size(u,1)) + 1
          indexMax = (offset_1d / size(u,1)) + 1
          latMax = latEdge(indexMax)
@@ -7441,12 +7471,17 @@ module atm_time_integration
          kMax = -1
          latMax = 0.0
          lonMax = 0.0
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(max:scalar_max)
          do iEdge = 1, nEdgesSolve
          do k = 1, nVertLevels
             scalar_max = max(scalar_max, u(k,iEdge))
          end do
          end do
+         !$acc end parallel
 
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:offset_1d)
          do iEdge = 1, nEdgesSolve
          do k = 1, nVertLevels
             if (u(k,iEdge) == scalar_max) then
@@ -7454,6 +7489,7 @@ module atm_time_integration
             end if
          end do
          end do
+         !$acc end parallel
          kMax = mod(offset_1d, size(u,1)) + 1
          indexMax = (offset_1d / size(u,1)) + 1
          latMax = latEdge(indexMax)
@@ -7484,13 +7520,18 @@ module atm_time_integration
          kMax = -1
          latMax = 0.0
          lonMax = 0.0
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(max:scalar_max)
          do iEdge = 1, nEdgesSolve
          do k = 1, nVertLevels
             spd(k,iEdge) = sqrt(u(k,iEdge)*u(k,iEdge) + v(k,iEdge)*v(k,iEdge))
             scalar_max = max(scalar_max, spd(k,iEdge))
          end do
          end do
+         !$acc end parallel
 
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:offset_1d)
          do iEdge = 1, nEdgesSolve
          do k = 1, nVertLevels
             if (spd(k,iEdge) == scalar_max) then
@@ -7498,6 +7539,7 @@ module atm_time_integration
             end if
          end do
          end do
+         !$acc end parallel
          kMax = mod(offset_1d, size(spd,1)) + 1
          indexMax = (offset_1d / size(spd,1)) + 1
          latMax = latEdge(indexMax)
@@ -7527,6 +7569,8 @@ module atm_time_integration
          !
          found_NaN = .false.
 
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(.or.:found_NaN)
          do iCell = 1, nCellsSolve
          do k = 1, nVertLevels
             if (ieee_is_nan(w(k,iCell))) then
@@ -7534,12 +7578,15 @@ module atm_time_integration
             end if
          end do
          end do
+         !$acc end parallel
          if (found_NaN) then
             call mpas_log_write('NaN detected in ''w'' field.', messageType=MPAS_LOG_CRIT)
          end if
 
          found_NaN = .false.
 
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(.or.:found_NaN)
          do iEdge = 1, nEdgesSolve
          do k = 1, nVertLevels
             if (ieee_is_nan(u(k,iEdge))) then
@@ -7547,6 +7594,7 @@ module atm_time_integration
             end if
          end do
          end do
+         !$acc end parallel
          if (found_NaN) then
             call mpas_log_write('NaN detected in ''u'' field.', messageType=MPAS_LOG_CRIT)
          end if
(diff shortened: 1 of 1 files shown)


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7262,11 +7263,19 @@ subroutine summarize_timestep(domain)
        real (kind=RKIND) :: lonMax, lonMax_global
        real (kind=RKIND), dimension(5) :: localVals, globalVals
 
-       real (kind=RKIND) :: spd
+       real (kind=RKIND), dimension(:,:), allocatable :: spd
+       !$acc declare create(spd)
```
abishekg7, line 7267: Everything looks good, but I had a question about why the declare create is required here. Wasn't fully clear about that from our previous meetings.

gdicker1: I did this just because it seems to be what is preferred by the OpenACC standard for Fortran allocatable arrays.

`acc declare create(spd)` allows the compiled code to "just do the right thing" whenever it encounters the `allocate(spd(...))` or `deallocate(spd)` statements later on - in this case, allocate this array on both device and host.

gdicker1: Also, since `spd` won't be fused into any data region (and thus isn't a type of temporary data movement) it doesn't need to be tracked in any `MPAS_ACC_TIMER_...` region.

abishekg7: Got it. Thanks for the explanation!

abishekg7:
The code changes look reasonable and I've only compared the log file to the reference run, but that seems to match. 

Perhaps the commit history could be squashed into fewer commits, but also fine as it is.

abishekg7 (approved):

gdicker1:
Force-push from [256463e](https://github.com/MPAS-Dev/MPAS-Model/commit/256463e1b1ace44204069a879ec1af0282dfd512) to [a21db2c](https://github.com/MPAS-Dev/MPAS-Model/commit/a21db2c350dd41ac9cb0dbac6ca9306f3dc37896) has no functional changes. I just corrected to having separate commits for adding OpenACC for `config_print_global_minmax_vel` section and for `config_print_detailed_minmax_vel` section..

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7242,7 +7242,8 @@ subroutine summarize_timestep(domain)
        logical, pointer :: config_print_global_minmax_sca
 
        integer :: iCell, k, iEdge, iScalar
-       integer, pointer :: num_scalars, nCellsSolve, nEdgesSolve, nVertLevels
+       integer, pointer :: num_scalars_ptr, nCellsSolve_ptr, nEdgesSolve_ptr, nVertLevels_ptr
+       integer ::          num_scalars, nCellsSolve, nEdgesSolve, nVertLevels
```
mgduda, line 7246: I'd suggest we leave out the extra whitespace padding after the `::`.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7262,11 +7263,19 @@ subroutine summarize_timestep(domain)
...
        real (kind=RKIND), dimension(:,:), pointer :: w
        real (kind=RKIND), dimension(:,:), pointer :: u, v, uReconstructZonal, uReconstructMeridional, uReconstructX, uReconstructY, uReconstructZ
        real (kind=RKIND), dimension(:,:,:), pointer :: scalars, scalars_1, scalars_2
 
+       integer :: lin_indexMax ! Position of min/max in some array as a single integer.
+                               ! If multiple locations match the extrema, always pick the lowest lin_indexMax.
+                               ! array(k,idx) -> lin_indexMax = k+(idx-1)*size(array,1)
+                               ! array(5,4) with size(array,1) = 10 -> lin_indexMax = 35
+
+       logical found_NaN
```
mgduda, line 7277: I think we need a `::` between `logical` and `found_NaN` for consistency of style.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7262,11 +7263,19 @@ subroutine summarize_timestep(domain)
...
 
-       real (kind=RKIND) :: spd
+       real (kind=RKIND), dimension(:,:), allocatable :: spd
+       !$acc declare create(spd)
        real (kind=RKIND), dimension(:,:), pointer :: w
        real (kind=RKIND), dimension(:,:), pointer :: u, v, uReconstructZonal, uReconstructMeridional, uReconstructX, uReconstructY, uReconstructZ
        real (kind=RKIND), dimension(:,:,:), pointer :: scalars, scalars_1, scalars_2
 
+       integer :: lin_indexMax ! Position of min/max in some array as a single integer.
+                               ! If multiple locations match the extrema, always pick the lowest lin_indexMax.
```
mgduda, line 7273: Some of this explanation seems like it would be better placed near the point where this variable is used, or it may be entirely superfluous. The use of `Max` in the variable name might also be confusing: does this refer to the value at the index, or to the index itself?

What if we called this variable something like `offset_1d`, and declare it as:
```fortran
integer :: offset_1d    ! Offset into a multi-dimensional array, if that array were viewed as a contiguous 1-d array
```

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7286,26 +7295,48 @@ subroutine summarize_timestep(domain)
...
-               scalar_min = w(k,iCell)
-               indexMax = iCell
-               kMax = k
-               latMax = latCell(iCell)
-               lonMax = lonCell(iCell)
+            scalar_min = min(scalar_min, w(k,iCell))
+         end do
+         end do
+         !$acc end parallel
+         !$acc parallel default(present)
```
mgduda, line 7325: I think a blank line between parallel regions here and elsewhere could help readability. Parallelism ends and begins again, as do the contiguous lines of code.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7474,55 +7553,93 @@ subroutine summarize_timestep(domain)
...
+               found_NaN = .true.
             end if
          end do
          end do
+         !$acc end parallel
+         if (found_NaN) then
+            call mpas_log_write('NaN detected in ''w'' field.', messageType=MPAS_LOG_CRIT)
+         end if
 
+
```
mgduda, line 7572: If we reach this line with the code as written, `found_NaN` is guaranteed to be `.false.`. However, if we were ever to change the log messages from critical error messages (that halt execution) to regular error messages, we may then get false messages about NaNs in the `u` field below. To improve robustness of the code, should we reset `found_NaN` to `.false.` here?

mgduda (changes requested):

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: > The use of Max in the variable name might also be confusing ...

I don't agree with this given the context of `kMax`, `latMax`, etc. At least it's not _more_ confusing than those. Maybe just positioning the declarations closer together would help. However, the idea of a "linear index" isn't well communicated by my comments. Let it be `offset_1d`!

I'll re-work this so the example with `array(5,4)` and other comments is closer to the first usage of `offset_1d`.

gdicker1:
@mgduda this should be ready for re-review. I incorporated your changes and did it as the force-push from [a21db2c](https://github.com/MPAS-Dev/MPAS-Model/commit/a21db2c350dd41ac9cb0dbac6ca9306f3dc37896) to [48d0014](https://github.com/MPAS-Dev/MPAS-Model/commit/48d0014001455eba3d41cb799da5dcdb5a539252) to adjust the commit message mentioning "linear index" a little too.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
mgduda: Can you make the same change elsewhere in this routine where we have 
```
         !$acc end parallel
```
immediately followed by
```
         !$acc parallel default(present)
```
with no intervening whitespace?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7286,26 +7292,50 @@ subroutine summarize_timestep(domain)
...
-               latMax = latCell(iCell)
-               lonMax = lonCell(iCell)
+            scalar_min = min(scalar_min, w(k,iCell))
+         end do
+         end do
+         !$acc end parallel
+
+         ! This second loop using offset_1d ensures the same (kMax,indexMax) are reported with the scalar_min
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:lin_indexMax)
```
mgduda, line 7325: `lin_indexMax` is no longer a variable in this routine.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7327,21 +7357,36 @@ subroutine summarize_timestep(domain)
...
+            if (w(k,iCell) == scalar_max) then
+               offset_1d = min(offset_1d, k + size(w,1)*(iCell-1))
             end if
          end do
          end do
+         !$acc end parallel
+         kMax = modulo(offset_1d, size(w,1))
+         indexMax = ((offset_1d - kMax) / size(w,1)) + 1
+         kMax = modulo(lin_indexMax, size(w,1))
+         indexMax = ((lin_indexMax - kMax) / size(w,1)) + 1
```
mgduda, line 7387: It looks like these two lines (using `lin_indexMax`) should be replaced by the two lines above them (using `offset_1d`).

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7399,21 +7457,34 @@ subroutine summarize_timestep(domain)
...
+         !$acc loop collapse(2) gang vector reduction(min:lin_indexMax)
+         do iEdge = 1, nEdgesSolve
+         do k = 1, nVertLevels
+            if (u(k,iEdge) == scalar_max) then
+               offset_1d = min(offset_1d, k + size(u,1)*(iEdge-1))
             end if
          end do
          end do
+         !$acc end parallel
+         kMax = modulo(offset_1d, size(u,1))
```
mgduda, line 7484: I don't think the logic is completely correct if the min/max value is at `k = nVertLevels`. Instead, I think the logic should be:
```fortran
    do iEdge = 1, nEdgesSolve
    do k = 1, nVertLevels
       if (u(k,iEdge) == scalar_max) then
          offset_1d = min(offset_1d, (k-1) + size(u,1)*(iEdge-1))
       end if
    end do
    end do

    kMax = mod(offset_1d, size(u,1)) + 1
    indexMax = offset_1d / size(u,1) + 1
```

gdicker1: I see what you mean. Update coming in soon.

E.g. if max is at `(k,iEdge) = (4,2)` and `(nVertLevels,nEdges) = (4,3)`, the current method would wind up with `kMax = 0` and `indexMax = 9`...

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: (I decided to just delete the example)

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: Apologies for the miss, addressed by force push [48d0014](https://github.com/MPAS-Dev/MPAS-Model/commit/48d0014001455eba3d41cb799da5dcdb5a539252) to [4c1dd04](https://github.com/MPAS-Dev/MPAS-Model/commit/4c1dd042f528f17e7fbb554ace5bc3b6435a6480)

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: Should be fixed as part of force-push [27dca1f](https://github.com/MPAS-Dev/MPAS-Model/commit/27dca1f1753d31d791471081aba358ae60065b2e) to [393492e](https://github.com/MPAS-Dev/MPAS-Model/commit/393492e813271638a1c9d48a76e0b010f26ea714). Let me know if not.

gdicker1:
@mgduda this should be ready for re-review!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7435,22 +7504,35 @@ subroutine summarize_timestep(domain)
...
+         end do
+         end do
+         !$acc end parallel
+
+         !$acc parallel default(present)
+         !$acc loop collapse(2) gang vector reduction(min:offset_1d)
+         do iEdge = 1, nEdgesSolve
+         do k = 1, nVertLevels
+            if (spd(k,iEdge) == scalar_max) then
+               offset_1d = min(offset_1d, (k-1) + size(u,1)*(iEdge-1))
```
mgduda, line 7527: While `u` and `spd` do have the same dimensions, I'd still advocate for the use of `size(spd,1)` here (and in the code below to compute `kMax` and `indexMax`.

gdicker1: Good point. Addressed by force-push from [393492e](https://github.com/MPAS-Dev/MPAS-Model/commit/393492e813271638a1c9d48a76e0b010f26ea714) to [33b0db0](https://github.com/MPAS-Dev/MPAS-Model/commit/33b0db038b47664f4c9db12345bbb95a2425a831)

gdicker1:
Force-push from [33b0db0](https://github.com/MPAS-Dev/MPAS-Model/commit/33b0db038b47664f4c9db12345bbb95a2425a831) to [b2cfcce](https://github.com/MPAS-Dev/MPAS-Model/commit/b2cfccee02d32000f45a09722cd2001951d90ca5) because of a whitespace change that was bugging me.

mgduda:
I think the changes look good, and I'd only offer one additional point to consider. What if we were to move all of the data movements to the beginning and end of the `summarize_timestep` routine? For example, the beginning of the routine could use something like:
```fortran
     call mpas_pool_get_array(state, 'w', w, 2)
     call mpas_pool_get_array(state, 'u', u, 2)
     call mpas_pool_get_array(diag, 'v', v)
     call mpas_pool_get_array(state, 'scalars', scalars, 2)

     MPAS_ACC_TIMER_START('summarize_timestep [ACC_data_xfer]')
     if (config_print_detailed_minmax_vel) then
        !$acc enter data copyin(w,u,v)
     else if (config_print_global_minmax_vel) then
        !$acc enter data copyin(w,u)
     end if
     if (config_print_global_minmax_sca) then
        !$acc enter data copyin(scalars)
     end if
     MPAS_ACC_TIMER_STOP('summarize_timestep [ACC_data_xfer]')
```
Note that we'd need to retrieve the pointers to `u`, `v`, `w`, and `scalars` earlier than we currently do in the code so that the `copyin` clauses would have valid pointers to work with.

gdicker1:
> I think the changes look good, and I'd only offer one additional point to ...

Sure, that makes sense. I think I avoided it before to not repeat the `if (config_print_detailed_minmax_vel) ...` structures. I'll make those edits tomorrow.

gdicker1:
@mgduda should be ready for a re-review!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7258,19 +7259,45 @@ subroutine summarize_timestep(domain)
...
        real (kind=RKIND), dimension(:,:), pointer :: u, v, uReconstructZonal, uReconstructMeridional, uReconstructX, uReconstructY, uReconstructZ
        real (kind=RKIND), dimension(:,:,:), pointer :: scalars, scalars_1, scalars_2
 
+       logical :: found_NaN
+
        call mpas_pool_get_config(block % configs, 'config_print_global_minmax_vel', config_print_global_minmax_vel)
        call mpas_pool_get_config(block % configs, 'config_print_detailed_minmax_vel', config_print_detailed_minmax_vel)
        call mpas_pool_get_config(block % configs, 'config_print_global_minmax_sca', config_print_global_minmax_sca)
 
+      ! Fetch pointers early for temporary OpenACC data transfers
```
mgduda, line 7279: I think this comment is unnecessary, and at a minimum, I'd suggest removing the words "early" and "temporary".

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -7258,19 +7259,45 @@ subroutine summarize_timestep(domain)
...
+      call mpas_pool_get_subpool(block % structs, 'state', state)
+      call mpas_pool_get_subpool(block % structs, 'diag', diag)
+      call mpas_pool_get_array(state, 'w', w, 2)
+      call mpas_pool_get_array(state, 'u', u, 2)
+      call mpas_pool_get_array(diag, 'v', v)
+      call mpas_pool_get_array(state, 'scalars', scalars, 2)
+      if (config_print_detailed_minmax_vel) then
+         MPAS_ACC_TIMER_START('summarize_timestep [ACC_data_xfer]')
```
mgduda, line 7287: It would lead to less code (and, in my opinion, more readable code) if we moved the `MPAS_ACC_TIMER_{START,STOP}` macros outside of the conditional blocks, e.g.,
```fortran
     MPAS_ACC_TIMER_START('summarize_timestep [ACC_data_xfer]')
     if (config_print_detailed_minmax_vel) then
        !$acc enter data copyin(w,u,v)
     else if (config_print_global_minmax_vel) then
        !$acc enter data copyin(w,u)
     end if
     if (config_print_global_minmax_sca) then
        !$acc enter data copyin(scalars)
     end if
     MPAS_ACC_TIMER_STOP('summarize_timestep [ACC_data_xfer]')
```

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
gdicker1: I assumed these `..._get_array` calls were as temporary as the `acc (enter|exit) data` directives they enable, so I just duplicated them to keep the code modifications to a minimum.

I'll get the comment and the later `...get_array` calls removed.

mgduda: I think you make a fair point that we're only moving these `mpas_pool_get_array` calls so that we can place the `!$acc enter data` directives at the beginning of the subroutine, and our goal is to eventually eliminate those data movement directives. On the one hand, we could move the `mpas_pool_get_array` calls back to where they originally were (either now, or after eliminating data movement directives); or, we could just leave them here at the beginning of the routine, which might be ultimately be cleaner.

gdicker1: I'm all here for cleaner code! Given our talk earlier this week about there being nearly 0 impact from `mpas_pool_get_array` calls, I don't see a problem with occasionally fetching more than we need.

I don't think there would be any impacts on the GPU execution to fetch the (host-side) pointers if we don't use them. There should only be data movement once something causes an attach action - a compute region (esp. one with `default(present)`, a `copyin`, a `present` clause on a data region, etc.

mgduda (changes requested):
It looks like we no longer get printout for scalars when `config_print_global_minmax_sca` is `true`, possibly as a result of `num_scalars` never getting set.

gdicker1:
@mgduda, thanks for the catch. Force-push from [5c5fc5a](https://github.com/MPAS-Dev/MPAS-Model/commit/5c5fc5a592a9190dbd9486df975411f2a972e01e) to [fe1638a](https://github.com/MPAS-Dev/MPAS-Model/commit/fe1638ae99d4f6b4f50c030414d39315b6e6754b) addresses this. The print outs for scalars are now present and match my reference run.

mgduda (approved):

PR #1298, opened 2025-04-12
Capability for understanding ICON soil variables

MHBalsmeier:
The ICON models have a soil grid that is different from the Noah LSM grid, which is why this pull request contains the capability for understanding 16 more soil variables in the output files of ungrib.

The new variables are listed in the [Vtable](https://github.com/wrf-model/WPS/blob/master/ungrib/Variable_Tables/Vtable.ICONm) of ICON.

ICON data can be found [here](https://opendata.dwd.de/weather/nwp).

If this pull request is successful, I can do the same for the soil output of RAP/HRRR.


mgduda (approved):
Thanks for the changes -- these look good!

PR #1302, opened 2025-04-29
Bugfix for horizontal 2nd-order filter for the cam upper absorbing layer

skamaroc:
This PR fixes a bug in the horizontal 2nd-order filter for the CAM upper absorbing layer, where the wrong level in the kdiff field was being used when enforcing a lower-bound on kdiff. This absorbing layer is active only when `config_mpas_cam_coef > 0.0`.


mgduda (approved):

adamrher:
@mgduda mind telling me which version of MPAS this bug fix came in to? I believe CAM is currently using 8.2.1.

mgduda:
@adamrher I believe this fix was included in MPAS v8.2.3.

PR #1305, opened 2025-05-06
Eliminate host-device copies of the tendencies provided by mpas_atm_get_bdy_tend

mgduda:
This PR converts the `mpas_atm_get_bdy_tend` function into a subroutine, thereby enabling the tendencies provided by this routine to be entirely resident on the GPU when MPAS-A is built with OpenACC. The `ru_driving_tend`, `rt_driving_tend`, and `rho_driving_tend` allocatable arrrays that are set by the `mpas_atm_get_bdy_tend` subroutine are automatically allocated on the GPU through `!$acc declare create` directives in the `atm_time_integration` module, and the `create` and `copyout` of these variables can be eliminated from the `mpas_atm_get_bdy_tend` routine, as can the `copyin` and `delete` of these variables from the `atm_bdy_adjust_dynamics_speczone_tend` routine.


abishekg7 (approved):
Changes look good. And verified that it produces bit identical results for the CONUS limited area case.

PR #1306, opened 2025-05-06
Eliminate host-device copying of LBC driving data from mpas_atm_get_bdy_state

mgduda:
This PR eliminates the host-device copying of LBC driving data provided by the `mpas_atm_get_bdy_state_{2d,3d}` routines.

In order to eliminate host-device copying of the LBC driving data provided by the `mpas_atm_get_bdy_state_{2d,3d}` routines, the allocatable arrays `scalars_driving`, `ru_driving_values`, `rt_driving_values`, and `rho_driving_values` in the `atm_time_integration` module are now automatically allocated on the device through `!$acc declare create` directives.
    
Accordingly, the `mpas_atm_get_bdy_state_{2d,3d}` routines no longer need `create` / `copyout` data directives for the `return_state` array produced by those routines, since the actual arguments to the `return_state` dummy argument are always one of the four arrays (`scalars_driving`, `ru_driving_values`, `rt_driving_values`, and `rho_driving_values`) that are allocated on the device before calls to these routines.
    
Also, all other code in the `atm_time_integration` module that employs the LBC driving data arrays no longer needs to use `copyin` / `delete` data directives for those arrays.

As part of the work to eliminate all host-device copying of the LBC driving data arrays, two loop in the `atm_srk3` routine for setting `u` and `ru` in the specified zone have been ported to GPUs with OpenACC, and those ported loops use the on-device copies of `ru_driving_values`.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_boundaries.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_boundaries.F
@@ -359,11 +359,11 @@ module mpas_atm_boundaries
     !
     !  routine mpas_atm_get_bdy_state_2d
     !
-    !> \brief   Returns LBC state at a specified delta-t in the future
+    !> \brief   Provides LBC state at a specified delta-t in the future
     !> \author  Michael Duda
     !> \date    28 September 2016
     !> \details 
-    !>  This function returns an array providing the state for the requested
+    !>  This subroutine returns an array providing the state for the requested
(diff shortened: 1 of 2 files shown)


jim-p-w (approved):
Looks good. And I verified the results from the regional test case  don't change.

abishekg7 (approved):
Changes look good. Verified that results are bit identical with the limited area CONUS case.

PR #1307, opened 2025-05-08
[Draft] Consolidating OpenACC device-host memory transfers

abishekg7:
Current work in progress of the effort to consolidate OpenACC device <-> host memory transfers to just before and after the call to `atm_srk3` in the subroutine `atm_timestep` in `mpas_atm_time_integration.F`


abishekg7:
Replaced by https://github.com/MPAS-Dev/MPAS-Model/pull/1315

PR #1309, opened 2025-05-08
Add MUSICA library as a build option

mattldawson:
This PR adds a build option to include the [MUSICA-Fortran](https://github.com/NCAR/musica) library in the atmosphere model build. This is intended as a first step towards being able to use MUSICA chemistry components ([MICM](https://github.com/NCAR/micm), [TUV-x](https://github.com/NCAR/tuv-x)) in MPAS-A.

To include the MUSICA-Fortran library in the build, MUSICA-Fortran and its dependencies must be installed locally. Include MUSICA in the MPAS-A build like this:
```
make gnu CORE=atmosphere MUSICA=true
```

Omitting the `MUSICA` flag does not attempt to link against the MUSICA libraries.

I'm happy to make any changes. I did my best to follow existing patterns in the codebase, but I'm brand-new to MPAS-A.

@mgduda @rkumar-ncar @marybarthbrock @pfisterggp @k20shores

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -690,8 +690,8 @@ ifeq "$(shell echo $(MUSICA) | tr '[:upper:]' '[:lower:]')" "true"
 ifeq ($(shell pkg-config --exists musica-fortran && echo yes || echo no), no)
 $(error "musica-fortran package is not installed. Please install it to proceed.")
 endif
-	FCINCLUDES += $(shell pkg-config --cflags musica-fortran)
-	LIBS += $(shell pkg-config --libs musica-fortran)
+	MUSICA_FCINCLUDES += $(shell pkg-config --cflags musica-fortran)
+	MUSICA_LIBS += $(shell pkg-config --libs musica-fortran)
 	MUSICA_FFLAGS = -DMPAS_USE_MUSICA
 endif
 
@@ -882,6 +882,8 @@ $(if $(PRECISION),$(info NOTE: PRECISION=single is unnecessary, single is the de
 	PRECISION_MESSAGE="MPAS was built with default single-precision reals."
 endif #PRECISION IF
 
+FCINCLUDES += $(MUSICA_FCINCLUDES)
+LIBS += $(MUSICA_LIBS)
 FFLAGS += $(MUSICA_FFLAGS)
 
 ifeq "$(USE_PAPI)" "true"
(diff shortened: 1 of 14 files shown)


K20shores (approved):

boulderdaze (approved):
I reviewed it and it looks good to me, but I also just started learning about the MPAS way to creating makefiles by reviewing this PR, so I'd love to see comments or feedbacks from others who are more familiar with MPAS.

Makefile
```diff
@@ -685,6 +685,16 @@ CPPINCLUDES =
...
+# Optional MUSICA support for chemistry
+ifeq "$(shell echo $(MUSICA) | tr '[:upper:]' '[:lower:]')" "true"
+ifeq ($(shell pkg-config --exists musica-fortran && echo yes || echo no), no)
+$(error "musica-fortran package is not installed. Please install it to proceed.")
+endif
+	MUSICA_FCINCLUDES += $(shell pkg-config --cflags musica-fortran)
+	MUSICA_LIBS += $(shell pkg-config --libs musica-fortran)
+	MUSICA_FFLAGS = -DMPAS_USE_MUSICA
+endif
+
```
mgduda, line 697: It could help to keep the MUSICA changes more consolidated if we moved this logic around line 885 (new line number) where we have the logic to append to `FCINCLUDES`, `LIBS`, and `FFLAGS`.

Makefile
```diff
@@ -1374,6 +1388,31 @@ mpi_f08_test:
...
+		musica_fortran_status=$$?; $\
+		rm -f test_musica_fortran.f90 test_musica_fortran.x; $\
+		if [ $$musica_fortran_status -eq 0 ]; then $\
+			printf "1"; $\
+		else $\
+			printf "0"; $\
+		fi $\
+	))
+	$(if $(findstring 0,$(MUSICA_FORTRAN_TEST)), $(error Could not build a simple test program with MUSICA-Fortran))
+	$(if $(findstring 1,$(MUSICA_FORTRAN_TEST)), $(eval CHEMISTRY_MESSAGE = "MUSICA-Fortran library available."), )
```
mgduda, line 1415: So that we don't set `CHEMISTRY_MESSAGE` (perhaps renamed `MUSICA_MESSAGE`) in two different places, here it might suffice to report the MUSICA library version that was found?

Makefile
```diff
@@ -1374,6 +1388,31 @@ mpi_f08_test:
 	$(if $(findstring 1,$(MPAS_MPI_F08)), $(eval MPI_F08_MESSAGE = "Using the mpi_f08 module."), )
 	$(if $(findstring 1,$(MPAS_MPI_F08)), $(info mpi_f08 module detected.))
 
+musica_fortran_test:
+	@#
+	@# Create a Fortran test program that will link against the MUSICA library
+	@#
+	$(info Building a simple program with the MUSICA-Fortran library...)
```
mgduda, line 1395: Just for consideration, we could say something like `Checking for a working MUSICA-Fortran library...`, as that's really the point of this build target. How that's done -- whether through the use of a simple program or otherwise -- probably matters less to users.

mgduda (changes requested):

Makefile
```diff
@@ -695,7 +705,7 @@ ifneq ($(wildcard $(PIO)/lib), )
 else
 	PIO_LIB = $(PIO)
 endif
-LIBS = -L$(PIO_LIB)
+LIBS += -L$(PIO_LIB)
```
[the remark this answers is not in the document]
mattldawson: It ended up not being necessary for the MUSICA linking (it was needed in a previous iteration).

I left it because `LIBS =` is included several lines above this, and it seemed like this should be appending to `LIBS` instead of redefining it. But, I can remove it if you prefer.

Makefile
```diff
@@ -1424,6 +1463,12 @@ IO_MESSAGE = "Using the SMIOL library."
 override CPPFLAGS += "-DMPAS_SMIOL_SUPPORT"
 endif
 
+ifneq "$(MUSICA_FFLAGS)" ""
+MAIN_DEPS += musica_fortran_test
+else
+CHEMISTRY_MESSAGE = "No external chemistry libraries available."
```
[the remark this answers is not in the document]
mattldawson: updated as suggested

Makefile
mattldawson: so moved

Makefile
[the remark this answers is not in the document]
mgduda: I think it might be better to make a separate PR to clean up some of the Makefile logic, and so for this PR, we can aim to minimize the changes and to keep them focused only on what is required for linking with MUSICA.

Makefile
```diff
@@ -872,6 +872,20 @@ $(if $(PRECISION),$(info NOTE: PRECISION=single is unnecessary, single is the de
...
+$(error "musica-fortran package is not installed. Please install it to proceed.")
+endif
+	MUSICA_FCINCLUDES += $(shell pkg-config --cflags musica-fortran)
+	MUSICA_LIBS += $(shell pkg-config --libs musica-fortran)
+	MUSICA_FFLAGS = -DMPAS_USE_MUSICA
+endif
+
+FCINCLUDES += $(MUSICA_FCINCLUDES)
+LIBS += $(MUSICA_LIBS)
+FFLAGS += $(MUSICA_FFLAGS)
```
mgduda, line 887: Could these three lines be moved inside the `ifeq` block above, since the MUSICA variables are only set if `MUSICA=true` was specified?

Makefile
```diff
@@ -1456,6 +1503,7 @@ mpas_main: $(MAIN_DEPS)
 	@echo $(PARALLEL_MESSAGE)
 	@echo $(MPI_F08_MESSAGE)
 	@echo $(PAPI_MESSAGE)
+	@echo $(MUSICA_MESSAGE)
```
mgduda, line 1506: We could move this message after the `OPENACC_MESSAGE` on line 1510 rather than placing it between the PAPI and TAU messages (PAPI and TAU are for performance measurement and profiling, and so serve related roles).

Makefile
mattldawson: moved!

mgduda (approved):

PR #1311, opened 2025-05-12
Remote redundant part for noahmp soil init

CharlesZheZhang:
Remove redundant soil liquid water init from NoahMP, based on this previous PR:
https://github.com/MPAS-Dev/MPAS-Model/pull/1244

With this previous PR: https://github.com/MPAS-Dev/MPAS-Model/pull/1244,
the mpas_noahmp%sh2o field is correctly initialized in calling NoahmpInitMain and correctly assigned to sh2o. 

This PR will remove the redundant part (L398-433).

@cenlinhe @barlage

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_lsm_noahmpinit.F
@@ -395,42 +395,6 @@
  call mpas_pool_get_array(output_noahmp,'t2mvxy',t2mvxy  )
  call mpas_pool_get_array(output_noahmp,'qtdrain',qtdrain)
 
-!--- initialization of the soil liquid water content:
- do i = its,ite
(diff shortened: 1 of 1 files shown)


mgduda:
@CharlesZheZhang Based on a global 120-km test, the changes in this PR appear to have no impact on results -- which I think we should expect if the removed code is truly redundant -- and I can see where the redundant code in `noahmp_init` is also found in `NoahmpInitMain`.

Could you amend the message in commit adfcc4a39a with a more detailed explanation of the changes in the commit? Ideally, the commit message would stand on its own and not reference another PR or commit. Regarding commit message format, we typically use the following: A one-line summary (80 chars max) followed by a blank line, then one or more paragraphs (wrapped at 80 chars) explaining the changes in the commit. We preserve commit history when merging PRs in MPAS-Model, and it's possible that others might cherry-pick commits into their own feature branches, and so it's helpful to include details in each commit (especially for anyone reading through commit history in future trying to track down a bug).

PR #1312, opened 2025-05-12
Bugfix in subroutine init_atm_case_mtn_wave regarding parallel execution - v2

MHBalsmeier:
This is a fix for issue https://github.com/MPAS-Dev/MPAS-Model/issues/1256

The problem was that the center coordinate of the domain xc was not synchronized among processes which resulted in wrong orography. This is now fixed and the initial state files created in parallel and serially are identical.


mgduda (approved):
Nice work finding this bug, and thanks for the fix!

PR #1314, opened 2025-05-13
Enable GPU execution of loops in atm_srk3 involving module level variables

abishekg7:
This PR ports some of the loops in the `mpas_atm_time_integration` and `mpas_atm_core` modules, which initialize module level variables belonging to the `mpas_atm_time_integration` module, to OpenACC in preparation for the consolidating all data transfers between host and device to before and after each dynamics call.

In order to do this, we also declare the allocatable module level variables in this scope using the OpenACC `declare create` statement, which instructs the nvhpc compiler to automatically create and delete the variable whenever it encounters an allocate or deallocate statement, respectively. This commit also removes these variables from manual data movement statements as required.

This PR also introduces some integers for loop bounds, so as to dereference scalar integer pointers which the OpenACC parallel regions do not correctly copy to device memory in the presence of a default(present) clause.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -65,9 +65,13 @@ module atm_time_integration
 !   real (kind=RKIND), allocatable, dimension(:,:) :: delsq_circulation    ! no longer used -> removed 
    real (kind=RKIND), allocatable, dimension(:,:) :: delsq_vorticity
    real (kind=RKIND), allocatable, dimension(:,:) :: dpdz
+   !$acc declare create(qtot)
+   !$acc declare create(delsq_theta, delsq_w, delsq_divergence)
+   !$acc declare create(delsq_u, delsq_vorticity, dpdz)
 
    ! Used in atm_advance_scalars
    real (kind=RKIND), dimension(:,:,:), allocatable :: horiz_flux_array
+   !$acc declare create(horiz_flux_array)
 
    ! Used in atm_advance_scalars_mono
    real (kind=RKIND), dimension(:,:), allocatable :: scalar_old_arr, scalar_new_arr
@@ -77,6 +81,10 @@ module atm_time_integration
    real (kind=RKIND), dimension(:,:), allocatable :: flux_tmp_arr
    real (kind=RKIND), dimension(:,:), allocatable :: wdtn_arr
    real (kind=RKIND), dimension(:,:), allocatable :: rho_zz_int
+   !$acc declare create(scalar_old_arr, scalar_new_arr)
+   !$acc declare create(s_max_arr, s_min_arr)
+   !$acc declare create(flux_array, flux_upwind_tmp_arr)
+   !$acc declare create(flux_tmp_arr, wdtn_arr)
 
    real (kind=RKIND), dimension(:,:), allocatable :: ru_driving_tend ! regional_MPAS addition 
    real (kind=RKIND), dimension(:,:), allocatable :: rt_driving_tend ! regional_MPAS addition 
@@ -100,6 +108,7 @@ module atm_time_integration
    ! Used in compute_solve_diagnostics
    real (kind=RKIND), allocatable, dimension(:,:) :: ke_vertex
    real (kind=RKIND), allocatable, dimension(:,:) :: ke_edge
+   !$acc declare create(ke_vertex, ke_edge)
 
    type (MPAS_Clock_type), pointer, private :: clock
    type (block_type), pointer, private :: block
@@ -842,9 +851,10 @@ module atm_time_integration
       character (len=StrKIND), pointer :: config_microp_scheme
       character (len=StrKIND), pointer :: config_convection_scheme
 
-      integer, pointer :: num_scalars, index_qv, nCells, nCellsSolve, nVertices, nVerticesSolve
-      integer, pointer :: nEdges_ptr, nEdgesSolve_ptr, nVertLevels_ptr
-      integer :: nEdges, nEdgesSolve, nVertLevels
+      integer, pointer :: index_qv, nCellsSolve, nVertices_ptr, nVerticesSolve
+      integer, pointer :: nCells_ptr, nEdges_ptr, nEdgesSolve_ptr
+      integer, pointer :: nVertLevels_ptr, num_scalars_ptr
+      integer :: nCells, nEdges, nVertices, nEdgesSolve, nVertLevels, num_scalars
 
       character(len=StrKIND), pointer :: config_IAU_option
 
@@ -907,10 +917,10 @@ module atm_time_integration
       ! Retrieve dimensions
       ! Note: nCellsSolve and nVerticesSolve are not currently used in this function
       !
-      call mpas_pool_get_dimension(mesh, 'nCells', nCells)
+      call mpas_pool_get_dimension(mesh, 'nCells', nCells_ptr)
       call mpas_pool_get_dimension(mesh, 'nCellsSolve', nCellsSolve)
       call mpas_pool_get_dimension(mesh, 'nEdges', nEdges_ptr)
-      call mpas_pool_get_dimension(mesh, 'nVertices', nVertices)
+      call mpas_pool_get_dimension(mesh, 'nVertices', nVertices_ptr)
       call mpas_pool_get_dimension(mesh, 'nVertLevels', nVertLevels_ptr)
 
       !call mpas_pool_get_dimension(mesh, 'nCellsSolve', nCellsSolve)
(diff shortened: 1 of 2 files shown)


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -945,13 +957,18 @@ subroutine atm_srk3(domain, dt, itimestep, exchange_halo_group)
 #ifdef DO_PHYSICS
       call mpas_pool_get_dimension(state, 'index_qv', index_qv)
 #endif
-      call mpas_pool_get_dimension(state, 'num_scalars', num_scalars)
-
+      call mpas_pool_get_dimension(state, 'num_scalars', num_scalars_ptr)
+      num_scalars = num_scalars_ptr
```
mgduda, line 961: Could we add a blank line after this. The comment below is unrelated to `num_scalars`, so a blank line to separate these unrelated code blocks seems appropriate.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1130,19 +1147,49 @@ subroutine atm_srk3(domain, dt, itimestep, exchange_halo_group)
             call mpas_timer_start('atm_compute_dyn_tend')
    
             allocate(delsq_theta(nVertLevels,nCells+1))
-            delsq_theta(:,nCells+1) = 0.0_RKIND
+            !$acc parallel default(present)
+            !$acc loop vector
+            do k = 1, nVertLevels
+            delsq_theta(k,nCells+1) = 0.0_RKIND
```
mgduda, line 1153: Here and in several of the loops below, could we indent the loop body?

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1130,19 +1147,49 @@ subroutine atm_srk3(domain, dt, itimestep, exchange_halo_group)
...
+            end do
+            !$acc end parallel
             allocate(dpdz(nVertLevels,nCells+1))
-            dpdz(:,nCells+1) = 0.0_RKIND
+            !$acc parallel default(present)
+            !$acc loop vector
+            do k = 1, nVertLevels
+            dpdz(k,nCells+1) = 0.0_RKIND
+            end do
+            !$acc end parallel
```
mgduda, line 1192: Rather than having six `parallel` regions for initializing variables, we could reorganize the code to allocate the variables, then initialize the garbage element for all of them in a single `parallel` region.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1786,18 +1848,59 @@ subroutine advance_scalars(field_name, domain, rk_step, rk_timestep, config_mono
...
+      !$acc end parallel
+
       allocate(wdtn_arr(nVertLevels+1,nCells+1))
-      wdtn_arr(:,nCells+1) = 0.0_RKIND
+      !$acc parallel default(present)
+      !$acc loop vector
+      do k = 1, nVertLevels
+         wdtn_arr(k,nCells+1) = 0.0_RKIND
+      end do
+      !$acc end parallel
```
mgduda, line 1902: A single `parallel` region to initialize the garbage elements rather than six `parallel` regions might be better (and probably would be more optimal).

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1806,12 +1909,29 @@ subroutine advance_scalars(field_name, domain, rk_step, rk_timestep, config_mono
...
+         end do
+         !$acc end parallel
          allocate(flux_tmp_arr(nVertLevels,nEdges+1))
-         flux_tmp_arr(:,nEdges+1) = 0.0_RKIND
+         !$acc parallel default(present)
+         !$acc loop vector
+         do k = 1, nVertLevels
+            flux_tmp_arr(k,nEdges+1) = 0.0_RKIND
+         end do
+         !$acc end parallel
```
mgduda, line 1934: One `parallel` region rather than two might be better here as well.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: thanks for the catch!

abishekg7:
Push some changes addressing the review comments so far, and also added acc directives for the two loops left behind in `atm_mpas_init_block `.

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -1806,12 +1850,24 @@ subroutine advance_scalars(field_name, domain, rk_step, rk_timestep, config_mono
       end if
       if (rk_step < 3 .or. (.not. config_monotonic .and. .not. config_positive_definite)) then
          allocate(horiz_flux_array(num_scalars,nVertLevels,nEdges+1))
-         horiz_flux_array(:,:,nEdges+1) = 0.0_RKIND
+         !$acc parallel default(present)
+         !$acc loop gang vector collapse(2)
+         do iScalar = 1,num_scalars
+            do k = 1, nVertLevels
```
mgduda, line 1856: Although in practice compilers will probably spot this opportunity for optimization, it could be better for CPU execution, and more in line with how we've written nested loops, to transpose the `iScalar` and `k` loops so that the innermost loop corresponds to the left-most array dimension.

Also, though a minor detail, I'd suggest consistently using a space after the comma in the loop bounds (e.g., `1, num_scalars` rather than `1,num_scalars`).

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: thanks for the catch!

src/core_atmosphere/dynamics/mpas_atm_time_integration.F
abishekg7: Good point. Fixed it in the new commit.

src/core_atmosphere/mpas_atm_core.F
```diff
@@ -405,7 +405,9 @@ subroutine atm_mpas_init_block(dminfo, stream_manager, block, mesh, dt)
       real (kind=RKIND), dimension(:), pointer :: dvEdge, invDvEdge
       real (kind=RKIND), dimension(:), pointer :: dcEdge, invDcEdge
       real (kind=RKIND), dimension(:), pointer :: areaTriangle, invAreaTriangle
-      integer, pointer :: nCells, nEdges, nVertices, nVertLevels, nEdgesSolve
+      integer, pointer :: nCells_ptr, nEdges_ptr, nVertices_ptr, nVertLevels_ptr, nEdgesSolve
+      integer :: nCells, nEdges, nVertices, nVertLevels
+      integer :: k
```
gdicker1, line 410: Seems cleaner to put `k` with the other loop counters above; do `integer :: iCell,iEdge,iVertex,k`.

gdicker1 (commented):
Nothing major from me. This is looking good!

(Just holding off on my review state while the PR is being iterated on)

src/core_atmosphere/mpas_atm_core.F
abishekg7: done

mgduda (approved):
Looks good to me. Thanks!

abishekg7:
Thanks for the review! Rebased and squashed to a single commit.

gdicker1:
> This commit also introduces some integers for loop bounds, so as to dereference
scalar integer pointers which the OpenACC parallel regions do not correctly
copy to device memory in the presence of a default(present) clause.

Should this be reworded in the commit message? My current understanding is that the issue doesn't involve `default(present)` clauses but rather the _implicit_ copy of these scalar pointers. Or do you understand differently?

gdicker1:
Agreed on rewording, I remember these pointers also causing CUDA_ERROR_ILLEGAL_ADDRESS without `default(present)` (see [this comment in PR#1289](https://github.com/MPAS-Dev/MPAS-Model/pull/1289#discussion_r2061132776)). I think it would be better to focus on the implicit copy or have a little less detail.

Simple option, say less:

> This commit also introduces some integers for loop bounds, so as to dereference scalar integer pointers which the OpenACC parallel regions do not correctly copy to device memory.

Something like the following would be really specific:

> This commit also introduces some integers for loop bounds, so as to dereference the pointers fetched by mpas_pool_get_dimension. Issues occur at run-time with OpenACC when scalar pointers are moved to the device by the implicit copy mechanism and used in a compute region.

Another option:

> This commit also dereferences scalar integer pointers used as loop bounds in OpenACC parallel regions. This prevents run-time issues when non-array pointers in compute regions are moved to the device by an implicit copy.

abishekg7:
> > This commit also introduces some integers for loop bounds, so as to dereference scalar integer pointers which the OpenACC parallel regions do not correctly copy to device memory.

This is useful - thanks! I don't think I have encountered the `CUDA_ERROR_ILLEGAL_ADDRESS` error in this context yet, but good to know. I've updated the commit message now (and kept it simple)

gdicker1 (approved):
Looks good now

PR #1316, opened 2025-05-14
Use the physics from https://github.com/NOAA-GSL/UGWP.git with cmake build

jim-p-w:
There is a new dependency from the physics model to https://github.com/NOAA-GSL/UGWP.git
This PR adds the necessary elements to src/core_atmosphere/CMakeLists.txt to 

This was tested by building with the following commands:
```
cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DMPAS_DOUBLE_PRECISION=OFF ../MPAS-Model-develop |& tee cmake.log
qcmd -A nmmm0043 -q develop -- make -j4 |&tee make.log
```

### DIFF
--- a/src/core_atmosphere/CMakeLists.txt
+++ b/src/core_atmosphere/CMakeLists.txt
@@ -51,6 +51,7 @@ set(ATMOSPHERE_CORE_PHYSICS_WRF_SOURCES
     cu_ntiedtke_pre.F
     module_bep_bem_helper.F
     module_bl_gwdo.F
+    module_bl_ugwp_gwdo.F
     module_bl_ysu.F
     module_cam_error_function.F
     module_cam_shr_kind_mod.F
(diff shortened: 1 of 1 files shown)


mgduda:
@jim-p-w Can you take another look at the commit message? "Use the physics from https://github.com/NOAA-GSL/UGWP.git when building with cmake." doesn't seem to me like it fully explains why we're making the changes in the commit. 

It's probably worth mentioning the point (hash) on the `develop` branch at which new files were added, and that because the CMakeLists.txt files weren't updated to account for the new code, the builds with CMake were broken. It can sometimes also be worth including the build error that the changes in the commit aim to fix.

jim-p-w:
I have updated the commit message.

amstokely (approved):

mgduda:
> I have updated the commit message.

Thanks -- the new message looks good!

mgduda (approved):

jim-p-w:
> Ah... does this break the mpas-bundle build ? Do I also need #1317 if I want to build the latest mpas-bundle, @jim-p-w ?

@byoung-joo Yes the mpas-bundle build is broken. Yes you need the changes from both this PR and PR 1317. With both sets of changes the mpas_bundle build works.

PR #1317, opened 2025-05-15
This change updates the core_init_atmosphere cmake file to add new files

jim-p-w:
This change updates the core_init_atmosphere cmake file to add new files added in commit d2c068015  (PR #1276).
That PR updated the makefiles, but not the cmake files. That resulted in compile errors when building with cmake:
src/core_init_atmosphere/mpas_init_atm_cases.F:48:11:

   48 |       use mpas_init_atm_gwd_gsl, only : calc_gsl_oro_data
      |           1
Fatal Error: Cannot open module file 'mpas_init_atm_gwd_gsl.mod' for reading at
(1): No such file or directory

Tested by building mpas_bundle.

### DIFF
--- a/src/core_init_atmosphere/CMakeLists.txt
+++ b/src/core_init_atmosphere/CMakeLists.txt
@@ -23,12 +23,15 @@ set(init_atm_core_srcs
         mpas_atmphys_initialize_real.F
         mpas_atmphys_utilities.F
         mpas_geotile_manager.F
+        mpas_gsl_oro_data_sm_scale.F
+        mpas_gsl_oro_data_lg_scale.F
         mpas_init_atm_bitarray.F
         mpas_init_atm_cases.F
         mpas_init_atm_core.F
         mpas_init_atm_core_interface.F
-	mpas_init_atm_thompson_aerosols.F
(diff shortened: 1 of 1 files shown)


mgduda:
@jim-p-w Is the branch name for this PR the same as in #1316? If so, it might be preferable to use a new branch name for this PR, perhaps `init_atmosphere/cmake_noaa_ugwp`. The branch name ends up in the first line of a merge commit message, so we'd otherwise end up with these two lines in the commit log for the `develop` branch:
```
Merge branch 'atmosphere/cmake_noaa_ugwp' into develop (PR #1316)
```
and
```
Merge branch 'atmosphere/cmake_noaa_ugwp' into develop (PR #1317)
```
which might be confusing.

amstokely (approved):

jim-p-w:
@mgduda I don't see how to change the source branch for this PR. It appears only the target branch can be changed. I think I will need to create a new PR and close this one?

jim-p-w:
This PR is superseded by https://github.com/MPAS-Dev/MPAS-Model/pull/1320.

PR #1318, opened 2025-05-15
Workaround for reading block decomposition files with OpenMPI v5.x

gdicker1:
This PR adds a workaround for problems encountered on systems using OpenMPI v5.x (observed with 5.0.7).  The inlist argument to mpas_dmpar_scatter_ints seems to be affected by the MPI_ScatterV call, which then causes run-time fails in mpas_block_decomp_cells_for_proc when re-reading the block decomposition file. Making global_list an allocatable, associating a pointer with it, and using that pointer as the inlist argument to mpas_dmpar_scatter_ints seems to resolve the issue.

NOTE: So this PR can be applied broadly it is based on a very, very old commit. I think around v4.0, at least as old as the v6.0 tag.

### DIFF
--- a/src/framework/mpas_block_decomp.F
+++ b/src/framework/mpas_block_decomp.F
@@ -63,7 +63,8 @@ module mpas_block_decomp
       character (len=*), intent(in) :: blockFilePrefix, & !< Input: File prefix for block decomposition
                                        procFilePrefix !< Input: File prefix for processor decomposition
 
-      integer, dimension(:), pointer :: global_list
+      integer, dimension(:), allocatable, target :: global_list
+      integer, dimension(:), pointer :: global_list_ptr
       integer, dimension(:), pointer :: global_start
 
       integer, dimension(:), allocatable :: local_block_list
@@ -93,7 +94,8 @@ module mpas_block_decomp
         allocate(local_nvertices(dminfo % nprocs))
         allocate(global_start(dminfo % nprocs))
         allocate(global_list(partial_global_graph_info % nVerticesTotal))
-  
+        global_list_ptr => global_list
+
         if (dminfo % my_proc_id == IO_NODE) then
   
            iunit = 50 + dminfo % my_proc_id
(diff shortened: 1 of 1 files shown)


gdicker1:
The problems in this PR were noted first when a collaborator on the EarthWorks project who was working on TACC's Vista system ran into issues getting MPAS-A to work as a step towards running EarthWorks (CESM). Around August of 2024.

Another EarthWorks user posted about this in https://github.com/EarthWorksOrg/EarthWorks/issues/109 and proposed this fix. When they were running on the Narval system in Canada, runs would die with a "FIO-F-231/list-directed read/unit=1/error on data conversion." message which pointed back to mpas_block_decomp.F near [line 181](https://github.com/MPAS-Dev/MPAS-Model/blob/3fe7ae1678d48a44569c7d461d2a1788deaa50f6/src/framework/mpas_block_decomp.F#L181).

On Derecho, I ran the develop branch (from the merge of PR#1298) with nvhpc/25.1 and openmpi/5.0.7 modules. This reported a "double free or corruption (!prev)" which crashed the model. Examining this with gdb seems to point back to mpas_block_decomp.F [line 261](https://github.com/MPAS-Dev/MPAS-Model/blob/3fe7ae1678d48a44569c7d461d2a1788deaa50f6/src/framework/mpas_block_decomp.F#L261). Once the fix was applied, I could not recreate this error.

mgduda:
I think I'm able to reproduce the issue that this PR aims to resolve. An alternative fix to the one proposed here may involve simply changing the `inlist` argument to the `mpas_dmpar_scatter_ints` routine to a regular array (removing the `pointer` attribute); i.e, changing the [inlist dummy argument](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.3.1/src/framework/mpas_dmpar.F#L1496) so that it is declared as
```fortran
      integer, dimension(:), intent(in) :: inlist !< Input: List of integers to send
```

mgduda:
PR #1361 implements an alternative fix that has the positive side effect of allowing the `mpas_dmpar_scatter_ints` routine to work with non-pointer send buffer arrays.

PR #1319, opened 2025-05-15
Capability for understanding data in Lambert conformal projection implemented

MHBalsmeier:
As a next step towards being able to nest MPAS into RAP and HRRR this pull request includes changes to mpas_init_atm_cases.F. In three subroutines (blend_bdy_terrain, init_atm_case_gfs and init_atm_case_lbc), the case field % iproj == PROJ_LC is now included.

Disclaimer: this is not fully tested yet, since the HRRR/RAP soil grid can not yet be understood by MPAS, which I plan on fixing in a future pull request. However, the changes made in this commit won't break anything.


MHBalsmeier:
@mgduda Could you take a look at this? I'd like to get the RUC soil grid variables done before the next release also but I'd like to have this one merged before implementing that because the HRRR/RAP nesting only works with the Lambert projection implemented. Thank you.

PR #1321, opened 2025-05-22
Enable package logic code to be automatically generated through new XML attribute, active_when

mgduda:
This PR enables the package logic code that determines when a package is active
to be automatically generated by the registry through the use of a new XML attribute
for `<package>` elements, `active_when`.

Prior to the changes in this PR, defining a new package in the Registry with
a `<package>` element required code to be written (typically in a core's
`setup_packages` routine) to determine whether that package was active. For simple
conditions -- e.g., a package being active if and only if a logical namelist
option is true -- the need to write boilerplate-type Fortran code to retrieve
package and namelist pointers, check that those pointers are valid, and to
finally set the package variable, was unnecessarily cumbersome.

This PR introduces a new XML attribute, `active_when`, that may be added to
`<package>` elements in the Registry, along with new functionality in the registry
processor to automatically generate package-logic code (the equivalent to what
would otherwise need to be written by hand in a core's `setup_packages` routine)
at compile time.

If provided, the `active_when` attribute contains a Fortran expression that
evaluates to a logical, and that may use any literal constants and namelist
options (which are assumed to begin with the string 'config'). The expression
from the `active_when` attribute is essentially used verbatim in Fortran code to
set the package variable.

For example, consider a package named `foo` that is active if and only if the
namelist option `config_use_physics` is true. This package might have the
following definition in the Registry:
```
  <package name="foo" active_when="config_do_physics"/>
```
Logically, the generated code is equivalent to:
```
  fooActive = ( config_do_physics )
```
As another example, consider the following package definition in the Registry:
```
  <package name="bar" active_when="config_nsteps > 0 .and. .not. config_dryrun"/>
```
The generated Fortran logic would be:
```
  barActive = ( config_nsteps > 0 .and. .not. config_dryrun )
```
Note that in both examples, the package's `active_when` logic is a valid Fortran
logical expression that makes uses of namelist options and literal constants.

In order to make use of this functionality introduced in this PR, a core
must make two changes in its `mpas_<CORE>_core_interface` module:

1) The module should include through a pre-processor directive the
   registry-generated package-logic code, which resides in a file named
   `setup_packages.inc` in the core's `inc/` directory; e.g.,
```
   #include "setup_packages.inc"
```
2) The setup_packages routine in the module should make a call to the generated
   `<CORE>_setup_packages_when` routine, which is defined in the
   `setup_packages.inc` file; e.g.,
```
   call atm_setup_packages_when(configs, packages)
```
The registry-generated package-logic code for packages that employ an
`active_when` attribute also contains calls to `mpas_log_write` to record
information about the activity of packages to a core's log file. For example,
the following messages are now written to the log file if two packages, `foo` and
`bar`, were active and inactive, respectively:
```
 Configuring registry-specified packages...
   foo : T
   bar : F
 ----- done configuring registry-specified packages -----
```

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -27,6 +27,8 @@ const char * nmlopt_type(ezxml_t registry, const char *nmlopt);
 int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
                           const char *packagename, const char *packagewhen,
                           ezxml_t registry);
+void gen_pkg_debug_info(FILE *fd, regex_t *preg, ezxml_t registry,
+                        const char *packagename, const char *packagewhen);
 
 #define NUM_MODIFIED_ATTRS 2
 #define NUM_IGNORED_ATTRS 9
@@ -2647,6 +2649,13 @@ int generate_package_logic(ezxml_t registry)/*{{{*/
 
 	fd = fopen("setup_packages.inc", "w+");
 
+	fortprintf(fd, "#ifdef MPAS_DEBUG\n");
+	fortprintf(fd, "#define COMMA ,\n");
+	fortprintf(fd, "#define PACKAGE_LOGIC_PRINT(M) call mpas_log_write(M)\n");
+	fortprintf(fd, "#else\n");
+	fortprintf(fd, "#define PACKAGE_LOGIC_PRINT(M) ! M\n");
+	fortprintf(fd, "#endif\n\n");
+
 	fortprintf(fd, "   !\n");
 	fortprintf(fd, "   ! WARNING: This function is automatically generated at compile time.\n");
         fortprintf(fd, "   !          Any modifications to this code will be lost when MPAS is recompiled.\n");
@@ -2812,6 +2821,9 @@ int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
 	fortprintf(fd, "      nullify(%sActive)\n", packagename);
 	fortprintf(fd, "      call mpas_pool_get_package(packagePool, '%sActive', %sActive)\n", packagename, packagename);
 	fortprintf(fd, "\n");
+
+	gen_pkg_debug_info(fd, preg, registry, packagename, packagewhen);
+
 	fortprintf(fd, "      %sActive = ( %s )\n", packagename, packagewhen);
 	fortprintf(fd, "      call mpas_log_write('  %s : $l', logicArgs=[%sActive])\n", packagename, packagename);
 	fortprintf(fd, "\n");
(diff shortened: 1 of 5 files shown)


src/tools/registry/gen_inc.c
```diff
@@ -2603,3 +2608,307 @@ void mangle_name(char *new_name, const size_t new_name_size, const char *old_nam
...
+int package_logic_routine(FILE *fd, regex_t *preg, const char *corename, const char *packagename, const char *packagewhen, ezxml_t registry)/*{{{*/
+{
+	ezxml_t packages_xml, package_xml;
+	char *match;
+	regoff_t next;
+
+
+	fortprintf(fd, "\n");
+	fortprintf(fd, "   !\n");
+	fortprintf(fd, "   ! WARNING: This is subroutine is automatically generated at compile time.\n");
```
mattldawson, line 2744:
```suggestion
	fortprintf(fd, "   ! WARNING: This subroutine is automatically generated at compile time.\n");
```

mattldawson (approved):
Thanks for including me in the review @mgduda! Looks great to me, although I will admit I don't have a lot of experience writing code generators, so my suggestions are pretty trivial.

One corner case I was wondering about, is how a user would get this type of expression to set the active flag:
```
fooActive = ( trim( string_might_be_foo ) == "FOO" )
```
which contains double quotes? Are there escape characters, or would you use single quotes to enclose the conditional logic in the XML file? (full disclosure, I'm also not that familiar with XML)

Anyway, looks like a useful feature to me!

mgduda:
> Thanks for including me in the review @mgduda! Looks great to me, although I will admit I don't have a lot of experience writing code generators, so my suggestions are pretty trivial.
> 
> One corner case I was wondering about, is how a user would get this type of expression to set the active flag:
> 
> ```
> fooActive = ( trim( string_might_be_foo ) == "FOO" )
> ```
> 
> which contains double quotes? Are there escape characters, or would you use single quotes to enclose the conditional logic in the XML file? (full disclosure, I'm also not that familiar with XML)
> 
> Anyway, looks like a useful feature to me!

I should add this to the documentation or commit message somewhere, but for string literals, you'll need to use single quotes; e.g.,
```
<package name="blaz" active_when="trim(config_physics_suite) == 'mesoscale'"/>
```

src/tools/registry/gen_inc.c
mgduda: Good catch -- I've just fixed this!

src/tools/registry/gen_inc.c
```diff
@@ -2603,3 +2608,307 @@ void mangle_name(char *new_name, const size_t new_name_size, const char *old_nam
...
+
+	fortprintf(fd, "#ifdef MPAS_DEBUG\n");
+	fortprintf(fd, "#define COMMA ,\n");
+	fortprintf(fd, "#define PACKAGE_LOGIC_PRINT(M) call mpas_log_write(M)\n");
+	fortprintf(fd, "#else\n");
+	fortprintf(fd, "#define PACKAGE_LOGIC_PRINT(M) ! M\n");
+	fortprintf(fd, "#endif\n\n");
+
+	fortprintf(fd, "   !\n");
+	fortprintf(fd, "   ! WARNING: This is function is automatically generated at compile time.\n");
```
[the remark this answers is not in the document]
mgduda: Good catch -- I've just fixed this!

mgduda:
> I think the code changes look good overall, and I've done some testing - seems fine. I can approve when this PR is finalized.

Ok, this PR should be ready for a final review now.

PR #1252, opened 2024-12-12
Addition of the new static information: BNU soil types and 15" MODIS landuse

tanyasmirnova:
The BNU soil dataset has the same soil categories as in STATSGO with differences in spatial variations.
The MODIFIED_IGBP_MODIS_NOAH_15s has the same categories as in MODIFIED_IGBP_MODIS_NOAH.
The additional static information is introduced via the namelist.init_atmosphere
    config_landuse_data = 'MODIFIED_IGBP_MODIS_NOAH_15s'
    config_soilcat_data = 'BNU'

PR #1322, opened 2025-05-22
Addition of new static datasets: 30" BNU soil category and 15" MODIS land use

mgduda:
This PR adds support for two new static datasets in the `init_atmosphere` core: 30" BNU soil category and 15" MODIS land use. These datasets may be selected through namelist options when generating "static" files with the `init_atmosphere` core with `config_static_interp = true` in the `namelist.init_atmosphere` file.

The 30" BNU soil category dataset can be selected by setting the new namelist option `config_soilcat_data` to `'BNU'` in the `&data_sources` namelist group.

The 15" MODIS land use dataset may be selected by setting the existing namelist option `config_landuse_data` to `'MODIFIED_IGBP_MODIS_NOAH_15s'` in the `&data_sources` namelist group.

Additionally, this PR adds a new namelist option, `config_lu_supersample_factor`, to control the super-sampling of land use data, which may now be on either a 30" or a 15" grid, depending on the choice of dataset. The existing namelist option
`config_30s_supersample_factor` now controls the super-sampling for 30" terrain, soil category, and MODIS FPAR monthly vegetation fraction data only.

### DIFF
--- a/src/core_init_atmosphere/Registry.xml
+++ b/src/core_init_atmosphere/Registry.xml
@@ -196,9 +196,14 @@
                      description="The supersampling factor to be used for MODIS maximum snow albedo and monthly albedo datasets (case 7 only)"
                      possible_values="Positive integer values"/>
 
+	        <nml_option name="config_lu_supersample_factor" type="integer"       default_value="1"
+                     units="-"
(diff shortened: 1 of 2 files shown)


mgduda:
This PR replaces #1252.

PR #1325, opened 2025-05-28
Changing default value for three namelists

weiwangncar:
This PR changes the default value for the following namelists: config_sfc_albedo (new value = false), config_microp_re (new value = true) and config_ysu_pblmix (new value = true).

With config_sfc_albedo set to false, the model will use table values to define albedo based on landuse values. This should be a better representation for models that have resolution better than the climatological albedo which is at about 5.5 km resolution.

With config_microp_re set to true, this will allow model to compute effective radii from microphysics variables and use them in radiation. This should better represent interactions between microphysics and radiation.

With config_ysu_pblmix set to true, it will account for top-down radiation-driven mixing at the PBL top.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -2079,7 +2079,7 @@
                      description="logical for configuration of fractional sea-ice"
                      possible_values=".true. for sea-ice between 0 or 1; .false. for sea-ice equal to 0 or 1 (flag)."/>
 
-                <nml_option name="config_sfc_albedo" type="logical" default_value="true" in_defaults="false"
+                <nml_option name="config_sfc_albedo" type="logical" default_value="false" in_defaults="false"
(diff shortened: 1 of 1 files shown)


weiwangncar:
I agree that MPAS does not need to follow everything that WRF does. But if an option represents good science for a particular physics, it should be employed whether it is in WRF or MPAS. Even though the options are documented, most users would not go through them and they would use what are in default. This comes from our experience working with WRF users. Yes, WRF has changed default values for a number of options, the ones regarding effective radii and top-down mixing in YSU PBL are two of them. When we change them, we do so and announce them in major releases.

mgduda:
We could at least include `config_sfc_albedo`, `config_microp_re`, and `config_ysu_pblmix` in the default `namelist.atmosphere` file so that users will be aware that these options exist.

PR #1326, opened 2025-05-28
Add new Python script for setting up an MPAS-Atmosphere run directory

mgduda:
This PR adds a new Python script for setting up an MPAS-Atmosphere run directory.

The `setup_run_dir.py` script, located in the `testing_and_setup/atmosphere` directory, may be used to set up a run directory for MPAS-A by copying and linking executables, namelist files, streams files, and physics look-up tables from the `MPAS-Model` directory to a specified directory.

The usage of the `setup_run_dir.py` script is as follows:
```
usage: setup_run_dir.py [-h] [-o] [-i | -a] run_dir

positional arguments:
  run_dir          the run directory to set up

options:
  -h, --help       show this help message and exit
  -o, --overwrite  overwrite any existing files and symbolic links
  -i, --init-only  only set up files for the init_atmosphere core
  -a, --atm-only   only set up files for the atmosphere core
```


testing_and_setup/atmosphere/setup_run_dir.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+if __name__ == '__main__':
+    import os
+    import sys
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('run_dir', help='the run directory to set up')
+    parser.add_argument('-o', '--overwrite', action='store_true', help='overwrite any existing files and symbolic links')
```
abishekg7, line 92: PEP8 also recommends a max line length of 79 chars. Suggested, but fine to leave it as is. https://peps.python.org/pep-0008/#maximum-line-length

testing_and_setup/atmosphere/setup_run_dir.py
```diff
@@ -0,0 +1,226 @@
...
+    if not os.access(run_dir, os.W_OK):
+        print_color(RED, 'Error: No write permission for directory ' + run_dir)
+        sys.exit(1)
+
+    if not args.atm_only:
+        print('')
+        print_color(GREEN, 'Setting up init_atmosphere core files...')
+        print_color(GREEN, '----------------------------------------')
+        for f in init_files_to_copy:
+            copy_file(mpas_root + '/' + f, run_dir + '/' + os.path.basename(f), overwrite=args.overwrite)
```
abishekg7, line 209: Could use `os.path.join` here
```
copy_file(os.path.join(mpas_root, f, run_dir, os.path.basename(f)), overwrite=args.overwrite)
```

testing_and_setup/atmosphere/setup_run_dir.py
mgduda: I've wrapped some lines, though I'm inclined to leave others (mostly lines that involve long string literals) as is.

PR #1331, opened 2025-06-01
Copy streams.init_atmosphere in the setup_run_dir.py script for MPAS-A

mgduda:
This PR updates the `setup_run_dir.py` script so that it also copies the `streams.init_atmosphere` file.

The `setup_run_dir.py` script for setting up an MPAS-A run directory previously omitted a copy of the `streams.init_atmosphere` file. This PR adds `streams.init_atmosphere` to the list of files to copy to the run directory for the MPAS-A `init_atmosphere` core.


abishekg7 (approved):
Looks good.

PR #1333, opened 2025-06-02
Note in description of config_lu_supersample_factor applicability to USGS

mgduda:
This PR updates the description of the `config_lu_supersample_factor` namelist option to make it clearer that the supersampling will apply to any of the three land use datasets selected by the `config_landuse_data` namelist option, including 30" USGS land use.


gdicker1 (approved):
Approved on visual inspect.

PR #1334, opened 2025-06-02
Add 13 to the possible values for config_init_case in init_atmosphere Registry

mgduda:
This PR adds initialization case 13 (CAM-MPAS 3-d grid with specified topography and zeta levels) to the `possible_values` attribute for the `config_init_case` namelist option in the `init_atmosphere` core `Registry.xml` file.


gdicker1 (approved):
Approved by inspection.

PR #1338, opened 2025-06-24
Add a chemistry package structure

K20shores:
This is continued from where Matt left off since it already seemed mostly correct to me. I'll update the PR description when we know if this is an expected structure

will refactor this to do this:

- single folder called chemistry
  - single file called chemistry.F with an `init`, `run` and `finalize`
  - This file will call musica things, which maybe is placed in chemistry/musica
  - All chemistry things should be here and called from core_atmosphere or maybe init_atmosphere
  - It would be nice to have both direct updates of the concentrations as well as tendencies

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -685,16 +685,6 @@ CPPINCLUDES =
 FCINCLUDES =
 LIBS =
 
-# Optional MUSICA support for chemistry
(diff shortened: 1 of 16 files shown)


mgduda:
@K20shores I think this PR has been effectively replaced by PR #1360 . If you agree, could you close this PR?

PR #1353, opened 2025-08-04
Fix duplicate stream detection in mpas_stream_list_insert

amstokely:
This PR fixes a bug in the `mpas_stream_list` module where adjacent duplicate streams were allowed to be inserted into the list. The issue was in the `MPAS_stream_list_insert` logic, which has been updated to properly reject adjacent duplicates. The fix has been confirmed by the `mpas_test_insert_duplicate_at_begin` and `mpas_test_insert_duplicate_at_end` tests. Additionally, the unit test suite covers all core subroutines in the `mpas_stream_list` module, ensuring there are no other similar bugs.

### DIFF
--- a/src/core_test/mpas_test_core_stream_list.F
+++ b/src/core_test/mpas_test_core_stream_list.F
@@ -1,0 +1,653 @@
+! Copyright (c) 2025 The University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
+! Additional copyright and license information can be found in the LICENSE file
+! distributed with this code, or at https://mpas-dev.github.io/license.html .
+!
+module mpas_test_core_stream_list
+
+   use mpas_derived_types
+   use mpas_log
+   use mpas_stream_list
+
+   implicit none
+   private
+
+   public :: mpas_test_stream_list
+
+contains
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_create_list
+   !
+   !> \brief   Test creating an empty stream list and verify initialization.
+   !>
+   !> \details This subroutine tests the creation of a new MPAS stream list, ensuring
+   !>          that the list is properly initialized with zero items and no head stream.
+   !>          The list is then destroyed to clean up any allocated memory.
+   !>
+   !> \param err  The error code that indicates the result of the test.
+   !
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_create_list(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list
+
+      err = 0
+
+      call MPAS_stream_list_create(list)
+
+      if (.not. associated(list)) then
+         err = err + 1
+      end if
+
+      if (list%nItems /= 0) then
+         err = err + 1
+      end if
+
+      if (associated(list%head)) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_destroy(list)
+   end subroutine mpas_test_create_list
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_insert_single
+   !
+   !> \brief   Test inserting a single stream into the list and verify correct insertion.
+   !>
+   !> \details This subroutine tests the insertion of a single stream into an MPAS
+   !>          stream list. It verifies that the stream is correctly added to the
+   !>          list, and checks that the list contains the expected stream with the
+   !>          correct number of items.
+   !>
+   !> \param err  The error code that indicates the result of the test.
+   !
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_insert_single(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list, stream
+      integer :: ierr
+
+      err = 0
+
+      call MPAS_stream_list_create(list)
+      allocate(stream)
+      stream%name = 'stream1'
+
+      call MPAS_stream_list_insert(list, stream, ierr)
+
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      if (.not. associated(list%head)) then
+         err = err + 1
+      end if
+
+      if (list%nItems /= 1) then
+         err = err + 1
+      end if
+
+      if (trim(list%head%name) /= 'stream1') then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_destroy(list)
+   end subroutine mpas_test_insert_single
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_query_exact_match
+   !
+   !> \brief   Test querying for an exact stream match and ensure correct stream is found.
+   !>
+   !> \details This subroutine tests querying a stream list for an exact match of a
+   !>          stream's name. It ensures that the correct stream is found and that
+   !>          the query operation behaves as expected.
+   !>
+   !> \param err  The error code that indicates the result of the test.
+   !
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_query_exact_match(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list, stream, found
+      logical :: matched
+
+      err = 0
+
+      call MPAS_stream_list_create(list)
+      allocate(stream)
+      stream%name = 'stream1'
+      call MPAS_stream_list_insert(list, stream)
+      found => null()
+
+      matched = MPAS_stream_list_query(list, 'stream1', found)
+
+      if (.not. matched) then
+         err = err + 1
+      end if
+
+      if (.not. associated(found)) then
+         err = err + 1
+      else if (trim(found%name) /= 'stream1') then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_destroy(list)
+   end subroutine mpas_test_query_exact_match
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_remove_existing_streams
+   !
+   !> \brief   Test removing streams from the beginning, middle, and end of a list.
+   !>
+   !> \details This subroutine verifies that removing streams from different positions
+   !>          in an MPAS stream list works as expected. It inserts three streams into
+   !>          the list, then removes one from the middle, one from the end, and one
+   !>          from the beginning, checking that the correct stream is removed in each
+   !>          case and that the operation returns a success code.
+   !>
+   !> \param err  The error code that indicates the result of the test.
+   !>
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_remove_existing_streams(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list, s1, s2, s3, &
+            removed1, removed2, removed3
+      integer :: ierr
+
+      err = 0
+
+      allocate(s1)
+      s1%name = 'stream1'
+      allocate(s2)
+      s2%name = 'stream2'
+      allocate(s3)
+      s3%name = 'stream3'
+
+      call MPAS_stream_list_create(list, ierr=ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s1, ierr=ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s2, ierr=ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s3, ierr=ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      ! Remove from the middle
+      call MPAS_stream_list_remove(list, 'stream2', removed2, ierr=ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+      if (.not. associated(removed2)) then
+         err = err + 1
+      end if
+      if (trim(removed2%name) /= 'stream2') then
+         err = err + 1
+      end if
+
+      ! Remove from the end
+      call MPAS_stream_list_remove(list, 'stream3', removed3, ierr=ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+      if (.not. associated(removed3)) then
+         err = err + 1
+      end if
+      if (trim(removed3%name) /= 'stream3') then
+         err = err + 1
+      end if
+
+      ! Remove from the beginning
+      call MPAS_stream_list_remove(list, 'stream1', removed1, ierr=ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+      if (.not. associated(removed1)) then
+         err = err + 1
+      end if
+      if (trim(removed1%name) /= 'stream1') then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_destroy(list)
+      deallocate(removed1)
+      deallocate(removed2)
+      deallocate(removed3)
+   end subroutine mpas_test_remove_existing_streams
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_insert_non_adjacent_duplicate
+   !
+   !> \brief   Test inserting a non-adjacent duplicate of the first item added to the list.
+   !>
+   !> \details This subroutine verifies that inserting a duplicate of the first stream
+   !>          added to an MPAS stream list results in the correct duplicate error
+   !>          code when the duplicate is not inserted immediately after the original.
+   !>          It does so by inserting two unique streams into the list, then attempting
+   !>          to insert the first stream again. This confirms that duplicate detection
+   !>          works for non-adjacent duplicates in insertion order.
+   !>
+   !> \param err  The error code that indicates the result of the test.
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_insert_non_adjacent_duplicate(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list
+      type(MPAS_stream_list_type), pointer :: s1, s2
+      integer :: ierr
+
+      err = 0
+      call MPAS_stream_list_create(list)
+
+      allocate(s1)
+      s1%name = 'stream1'
+      allocate(s2)
+      s2%name = 'stream2'
+
+      call MPAS_stream_list_insert(list, s1, ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s2, ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s1, ierr)
+      if (ierr /= MPAS_STREAM_LIST_DUPLICATE) then
+         err = err + 1
+      end if
+      if (MPAS_stream_list_length(list) /= 2) then
+         err = err + 1
+      end if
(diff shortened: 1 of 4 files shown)


src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
+! Copyright (c) 2023, University Corporation for Atmospheric Research (UCAR)
```
mgduda: At a minimum, I think we should set the year here to 2025. But, I think we should also match the formatting used in other files, e.g., [mpas_stream_inquiry.F](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.3.1/src/framework/mpas_stream_inquiry.F#L1-L6).

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+      err = 0
+
+      call MPAS_stream_list_create(list)
+      allocate(stream)
+      stream%name = 'stream1'
+      nullify(stream%next)
+
+      call MPAS_stream_list_insert(list, stream, ierr)
+
+      if (ierr /= 0) then
```
mgduda: The `mpas_stream_list` uses `MPAS_STREAM_LIST_NOERR` to indicate no error, so it's probably best to compare the return error code against `MPAS_STREAM_LIST_NOERR` rather than `0`.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list, stream
+      integer :: ierr
+
+      err = 0
+
+      call MPAS_stream_list_create(list)
+      allocate(stream)
+      stream%name = 'stream1'
+      nullify(stream%next)
```
mgduda: It may reduce the amount of code in this module if we omit the nullification of `stream % next` here and elsewhere; the `next` member of the `MPAS_stream_list_type` receives a default initialization of `null()`: https://github.com/MPAS-Dev/MPAS-Model/blob/v8.3.1/src/framework/mpas_stream_list_types.inc#L41 .

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+   subroutine mpas_test_insert_duplicate_at_begin_and_end(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list
+      type(MPAS_stream_list_type), pointer :: s1, s2, s1_dup
+      integer :: ierr
+
+      err = 0
+      call MPAS_stream_list_create(list)
+
+      allocate(s1);       s1%name = 'stream1';      nullify(s1%next)
```
mgduda: As a matter of style, I'd generally suggest not combining statements on the same line with a semicolon.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+   !> \brief   Test attempting to remove a stream not found in the list, expect error.
+   !>
+   !> \details This subroutine tests the behavior of attempting to remove a stream
+   !>          that does not exist in the list. It ensures that the correct error
+   !>          is returned when the stream is not found.
+   !>
+   !> \param err  The error code that indicates the result of the test.
+   !
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_remove_not_found(err)
```
mgduda, line 327: This test searches an empty stream list, and as such it tests one special case. Would it be worth adding another test for removal of a non-existent item from a non-empty list?

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+
+      err = 0
+
+      call MPAS_stream_list_create(list)
+      allocate(stream)
+      stream%name = 'stream1'
+      nullify(stream%next)
+      call MPAS_stream_list_insert(list, stream)
+
+      call MPAS_stream_list_remove(list, 'stream1', removed)
```
mgduda: Don't forget to deallocate the removed node.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+      if (ierr /= 0) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s2, ierr)
+      if (ierr /= 0) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s1_dup, ierr)
```
mgduda: Since `s1_dup` should not be added to the list, I think you'll need to manually deallocate it.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+      allocate(s1_dup)
+      s1_dup%name = 'stream1'
+      nullify(s1_dup%next)
+
+      call MPAS_stream_list_insert(list, s1, ierr)
+      if (ierr /= 0) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s1_dup, ierr)
```
mgduda: If `s1_dup` is not added to the list here (as it shouldn't be), you'll need to manually deallocate it.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,530 @@
...
+      if (ierr /= 0) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s2, ierr)
+      if (ierr /= 0) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s2_dup, ierr)
```
mgduda: Don't forget to deallocate `s2_dup`.

mgduda (changes requested):

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,596 @@
...
+      if (ierr /= MPAS_STREAM_LIST_DUPLICATE) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_destroy(list)
+      deallocate(s1_dup)
+   end subroutine mpas_test_insert_duplicate_at_begin_and_end
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_remove_not_found
```
mgduda: I think the name here needs to be updated to `mpas_test_remove_from_empty_list`.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,596 @@
...
+      integer :: ierr
+
+      err = 0
+
+      allocate(s1)
+      s1%name = 'stream1'
+      allocate(s2)
+      s2%name = 'stream2'
+      allocate(s3)
+      s3%name = 'stream3'
```
mgduda: Aside from being deallocated at the end, is `s3` ever used? If not, perhaps we can just eliminate it altogether from this routine?

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,596 @@
...
+
+      call MPAS_stream_list_insert(list, s1)
+      call MPAS_stream_list_insert(list, s2)
```
mgduda, line 393: Perhaps there's an opportunity here to verify the length of the list after each insertion? The checking code could be made more compact by eliminating the use of a local variable `len`; e.g.,
```fortran
   if (MPAS_stream_list_length(list) /= N) then
      err = err + 1
   end if
```

mgduda (changes requested):

src/framework/mpas_stream_list.F
```diff
@@ -118,15 +118,17 @@ subroutine MPAS_stream_list_insert(list, stream, ierr) !{{{
             list % head => stream
         else
             node => list % head
-            do while (associated(node % next))
+            do while (associated(node))
                 if (node % name == stream % name) then
                     if (present(ierr)) ierr = MPAS_STREAM_LIST_DUPLICATE
-                    LIST_ERROR_WRITE('Found duplicate item '//trim(stream % name)//' in list.')
```
mgduda, line 124: I don't think we need to remove this line, as the `LIST_ERROR_WRITE` macro is defined as a commented-out call to `mpas_log_write`.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,598 @@
...
+   !
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_remove_from_empty_list(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list, s1, removed
+      integer :: ierr
+
+      err = 0
+      allocate(s1)
+      s1%name = 'stream1'
```
mgduda: It looks like we don't actually need `s1` in this routine.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,598 @@
...
+      end if
+
+      call MPAS_stream_list_destroy(list)
+      deallocate(removed)
+   end subroutine mpas_test_remove_head
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_insert_duplicate_at_begin_and_end
+   !
+   !> \brief   Test inserting a duplicate stream at both the beginning and the end of the list.
```
mgduda: It's not clear to me that what this routine actually does is to insert a duplicate at both the beginning and the end of a list. Could you either clarify what this routine actually does, or modify the code so that it matches this description?

mgduda (changes requested):
Aside from a couple of comments that I've left, it might be worth adding at least one test to verify that we can successfully remove items from the middle or the end of a list. Currently, it seems that we only test:
1. Removing the only item (head) of a list
2. Attempting to remove an item from an empty list (which should yield an error)
3. Attempting to remove a non-existent item from a list (which should yield an error)

src/core_test/mpas_test_core_stream_list.F
amstokely: This test verifies that mpas_stream_list_insert returns MPAS_STREAM_LIST_DUPLICATE when the first and last streams in the list are duplicates (e.g., [s1, s2, s1_dup]). It covers the non-adjacent case, with one duplicate at the start of the list and the other at the end. Instead of inserting s1_dup (which is treated as a duplicate because it has the same name as s1), reinserting s1 directly would make the test’s intent clearer?

mgduda: Ok, now I think I understand the meaning of "inserting a duplicate stream at both the beginning and the end". In principle, we shouldn't know anything about how the list is actually storing the items that we add (e.g., do we always insert at the head of the list, or that the tail of the list? that details is internal to the list implementation). How about explaining this test as "inserting a duplicate of the first item added to the list"?

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,635 @@
...
+   !>          updated after multiple streams are inserted. It verifies that the length
+   !>          matches the expected value.
+   !>
+   !> \param err  The error code that indicates the result of the test.
+   !
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_list_length(err)
+      integer, intent(out) :: err
+      type(MPAS_stream_list_type), pointer :: list, s1, s2, s3
+      integer :: len
```
mgduda: It doesn't look like we need the `len` variable anymore.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,635 @@
...
+      if (trim(removed1%name) /= 'stream1') then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_destroy(list)
+      deallocate(removed1)
+      deallocate(removed2)
+      deallocate(removed3)
+   end subroutine mpas_test_remove_existing_streams
+   !**************************************************************************************
```
mgduda, line 231: Let's add a blank line before this documentation block.

src/framework/mpas_stream_list.F
```diff
@@ -112,21 +112,23 @@ subroutine MPAS_stream_list_insert(list, stream, ierr) !{{{
...
                     return
                 end if
+                if (.not. associated(node % next)) then
+                    node % next => stream
+                    exit
+                end if
                 node => node % next
             end do
-            node % next => stream
+            nullify(stream % next)
```
mgduda: Should we also nullify `stream % next` after setting `list % head => stream` on line 116?

src/core_test/mpas_test_core_stream_list.F
mgduda: Just a reminder to add a blank line above...

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,638 @@
...
+      call MPAS_stream_list_insert(list, s2, ierr)
+      if (ierr /= MPAS_STREAM_LIST_NOERR) then
+         err = err + 1
+      end if
+
+      call MPAS_stream_list_insert(list, s1, ierr)
+      if (ierr /= MPAS_STREAM_LIST_DUPLICATE) then
+         err = err + 1
+      end if
+      if (.not. associated(s1%next)) then
```
mgduda, line 277: Without knowing the details of how the `MPAS_stream_list_insert` routine is implemented (e.g., do we add new items at the head of the list, or at the tail of the list?), it might not be obvious to the reader why we are testing that `s1%next` is associated here. I think a comment would be helpful here.

src/framework/mpas_stream_list.F
```diff
@@ -112,21 +112,23 @@ subroutine MPAS_stream_list_insert(list, stream, ierr) !{{{
...
             node => list % head
-            do while (associated(node % next))
+            do while (associated(node))
                 if (node % name == stream % name) then
                     if (present(ierr)) ierr = MPAS_STREAM_LIST_DUPLICATE
                     LIST_ERROR_WRITE('Found duplicate item '//trim(stream % name)//' in list.')
                     return
                 end if
+                if (.not. associated(node % next)) then
+                    node % next => stream
```
mgduda, line 127: As part of 999c449c, perhaps it would have led to slightly more understandable code (in my opinion) if we had left the `node % next => stream` statement in its original location below line 130 rather than moving it here, especially considering that we `nullify(stream % next)` only after adding the stream to the list. What do you think?

amstokely: I think it would be clearer to move the second nullify(stream % next) call directly below the current node % next => stream line. That way, all of the insertion logic is in the same conditional block, making the flow easier to follow.

src/framework/mpas_stream_list.F
amstokely: Yes we should...good catch!

src/framework/mpas_stream_list.F
mgduda: That's fine, too. Whatever we do, it would be best to have the logic all together in the same place.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,639 @@
...
+      logical :: matched
+      integer :: ierr
+
+      err = 0
+      call MPAS_stream_list_create(list)
+      allocate(stream)
+      stream%name = 'stream1'
+      call MPAS_stream_list_insert(list, stream)
+
+      matched = MPAS_stream_list_query(list, 'stream', found, ierr)
```
mgduda, line 431: Should we check that `ierr` is returned as `MPAS_STREAM_LIST_NOERR`?

mgduda: Also, I think we need to ensure that the `found` variable has been nullified before calling `MPAS_stream_list_query`; otherwise, valgrind reports the use of an uninitialized value:
```
==944499== Conditional jump or move depends on uninitialised value(s)
==944499==    at 0x4F1540: __mpas_stream_list_MOD_mpas_stream_list_query (mpas_stream_list.F:266)
==944499==    by 0x444F3C: __mpas_test_core_stream_list_MOD_mpas_test_query_partial_match (mpas_test_core_stream_list.F:426)
==944499==    by 0x4445BD: __mpas_test_core_stream_list_MOD_mpas_test_stream_list (mpas_test_core_stream_list.F:531)
==944499==    by 0x42A99E: __test_core_MOD_test_core_run (mpas_test_core.F:257)
==944499==    by 0x405CFE: __mpas_subdriver_MOD_mpas_run (mpas_subdriver.F:416)
==944499==    by 0x4048E2: MAIN__ (mpas.F:20)
==944499==    by 0x404943: main (mpas.F:10)
```
From the documentation for `MPAS_stream_list_query`: *"In the first call to this function, the stream argument should be a null pointer."*

mgduda: It looks like you've added the check for `ierr` in the wrong place.

src/core_test/mpas_test_core_stream_list.F
```diff
@@ -0,0 +1,694 @@
...
+      end if
+
+      call MPAS_stream_list_destroy(list)
+   end subroutine mpas_test_insert_duplicate_at_end
+
+   !**************************************************************************************
+   !  Subroutine mpas_test_insert_duplicate_pointer_with_different_name
+   !
+   !--------------------------------------------------------------------------------------
+   subroutine mpas_test_insert_duplicate_pointer_with_different_name(err)
```
mgduda: I'm not convinced of the value of this test, and so I'd suggest we don't add it.

mgduda: In any case, it seems that this routine isn't being called:
```
mpas_test_core_stream_list.F:535:68:

  535 |    subroutine mpas_test_insert_duplicate_pointer_with_different_name(err)
      |                                                                    ^
Warning: ‘mpas_test_insert_duplicate_pointer_with_different_name’ defined but not used [-Wunused-function]
```

amstokely: I meant to remove this test.

src/framework/mpas_stream_list.F
```diff
@@ -2,8 +2,8 @@ module mpas_stream_list
 
 #define COMMA ,
 #define LIST_DEBUG_WRITE(M) ! call mpas_log_write(M)
-#define LIST_WARN_WRITE(M) call mpas_log_write( M , messageType=MPAS_LOG_WARN)
-#define LIST_ERROR_WRITE(M) call mpas_log_write( M , messageType=MPAS_LOG_ERR)
+#define LIST_WARN_WRITE(M) ! call mpas_log_write( M , messageType=MPAS_LOG_WARN)
+#define LIST_ERROR_WRITE(M) ! call mpas_log_write( M , messageType=MPAS_LOG_ERR)
```
mgduda: These changes are already in the `develop` branch (and I suppose they were inadvertently re-introduced here), so perhaps this PR branch can be rebased on the current HEAD of the `develop` branch and these changes omitted.

mgduda:
@amstokely I've left just one more comment regarding changes to the `LIST_WARN_WRITE` and `LIST_ERROR_WRITE` macros. Otherwise, the code looks fine to me, and I think we can rework the commit history to contain just a few commits:
1. Add unit tests for the `mpas_stream_list` module
2. Fix the bug related to insertion of duplicate entries.
3. Fix the bug related the premature nullification of `stream % next` pointer.

mgduda (changes requested):
Can you try again to rework the commit history? Currently, 1b24f2f7 not only fixes the duplicate insertion bug, but it also makes many changes to the unit tests introduced in f8eb09cb. Also, can you add some details to the commit message that is currently in 7a87f048 and ensure that any changes to the unit test module are moved into the first commit?

amstokely:
> Can you try again to rework the commit history? Currently, [1b24f2f](https://github.com/MPAS-Dev/MPAS-Model/commit/1b24f2f7da66e0e5924a230c0f74287081e5dae1) not only fixes the duplicate insertion bug, but it also makes many changes to the unit tests introduced in [f8eb09c](https://github.com/MPAS-Dev/MPAS-Model/commit/f8eb09cb42d40709f8a7e7ccc2b4b1c3a414e5fc). Also, can you add some details to the commit message that is currently in [7a87f04](https://github.com/MPAS-Dev/MPAS-Model/commit/7a87f04848eec0afb6c13a7cfb1c041e1c2d6016) and ensure that any changes to the unit test module are moved into the first commit?

I'm working on it right now!

mgduda (approved):

PR #1354, opened 2025-08-07
Add error handling to MPAS_stream_mgr_remove_alarm and remove redundant logging

amstokely:
This PR introduces error checking to the `MPAS_stream_mgr_remove_alarm` subroutine in `mpas_stream_manager.F`, ensuring that errors returned by the `mpas_stream_list_remove` subroutine are correctly handled. Previously, specific logic paths (e.g., passing an incorrect direction to `MPAS_stream_mgr_remove_alarm`) would log an error without affecting the state of the code. This issue has been addressed by adding the necessary error handling, which now ensures the correct error codes are set and logged when errors occur.

Additionally, redundant error logging in the `mpas_stream_list_module` (specifically in the `mpas_stream_list_remove` subroutine) has been removed, as the newly added error checking now handles and logs these errors correctly.

### Testing:
The changes can be verified by running the test core. The stream tests, which tested this error checking, in commit [7bc8ea0f518d59a9f30557446b70bb656d452e5f](https://github.com/MPAS-Dev/MPAS-Model/tree/7bc8ea0f518d59a9f30557446b70bb656d452e5f) (the first commit in this PR), are now passing, confirming that the issue has been resolved.

### DIFF
--- a/src/core_test/mpas_test_core_streams.F
+++ b/src/core_test/mpas_test_core_streams.F
@@ -19,7 +19,7 @@ module test_core_streams
 
    !***********************************************************************
    !
-   !  routine test_core_streams_test
+   !  routine test_read_write_real_streams
    !
    !> \brief   tests reading/writing single- and double-precision streams
    !> \author  Michael Duda
@@ -35,7 +35,7 @@ module test_core_streams
    !>  that are created by this routine.
    !
    !-----------------------------------------------------------------------
-   subroutine test_core_streams_test(domain, threadErrs, ierr)
+   subroutine test_read_write_real_streams(domain, threadErrs, ierr)
 
       use mpas_stream_manager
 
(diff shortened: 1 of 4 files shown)


src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,197 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+
+   !***********************************************************************
+   !  Subroutine test_remove_input_alarm
+   !
+   !> \brief   Wrapper for testing the removal of an input alarm from a stream.
+   !>
+   !> \details This subroutine calls the `test_remove_alarm` subroutine,
+   !>          passing the `MPAS_STREAM_INPUT` as the stream type to test
+   !>          the functionality of removing an input alarm from a stream.
+   !
```
mgduda, line 515: It looks like the commit in the lines above should go with the `test_remove_input_alarm` subroutine, rather than with the `test_core_streams_test` subroutine in the line immediately below.

src/core_test/mpas_test_core_streams.F
```diff
@@ -35,7 +35,7 @@ module test_core_streams
    !>  that are created by this routine.
    !
    !-----------------------------------------------------------------------
-   subroutine test_core_streams_test(domain, threadErrs, ierr)
+   subroutine test_read_write_real_streams(domain, threadErrs, ierr)
```
mgduda, line 38: It would be good to update the routine name on line 22 in the comment block above.

src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,197 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+      call mpas_log_write('Testing removing an output alarm from a stream.')
+      call test_remove_output_alarm(domain, threadErrs, ierr2)
+      if (ierr2 == 0) then
+         call mpas_log_write('Removing output alarm test: SUCCESS')
+      else
+         call mpas_log_write('Removing output alarm test: FAILURE', MPAS_LOG_ERR)
+      end if
+
+      call mpas_log_write('Testing removing an input alarm from a stream.')
+      call test_remove_input_alarm(domain, threadErrs, ierr)
```
mgduda: Did you intend for the third argument to be `ierr3` in this call?

src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,197 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+      if (ierr3 == 0) then
+         call mpas_log_write('Removing input alarm test: SUCCESS')
+      else
+         call mpas_log_write('Removing input alarm test: FAILURE', MPAS_LOG_ERR)
+      end if
+
+      ierr = abs(ierr1) + abs(ierr2) + abs(ierr3)
```
mgduda: Rather than having a new `ierrX` variable for every test, I think we could just use a local `ierr` variable and immediately add its absolute value to the return `ierr` after each test call.

mgduda (changes requested):

mgduda:
@amstokely Rather than combining the changes to `mpas_stream_manager.F` and `mpas_stream_list.F` in the same commit, it may be better to split those changes into two commits: one to ensure complete error handling in the stream manager for all calls to the `mpas_stream_list` module that might return an error code, and a second one to then remove the redundant error logging in the `mpas_stream_list` module.

src/core_test/mpas_test_core_streams.F
amstokely: I did not...good catch!

src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,205 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+   !>          of alarms in different scenarios. It ensures that alarms
+   !>          cannot be removed from the opposite list and can be removed
+   !>          from the correct list. The stream type (input or output)
+   !>          is passed as an argument to test both scenarios.
+   !>
+   !> \param domain      The domain object that contains the stream manager
+   !>                    and clock.
+   !> \param ierr        The error code that indicates the result of the test.
+   !> \param streamType  The type of the stream being tested (input or output).
+   !
```
mgduda, line 413: As on line 37, can we add a line of hyphens to terminate these documentation blocks? It would be nice to maintain consistent style.

src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,205 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+
+      use mpas_log
+
+      implicit none
+
+      type (domain_type), intent(inout) :: domain
+      integer, dimension(:), intent(out) :: threadErrs
+      integer, intent(out) :: ierr
+
+      integer, dimension(3) :: test_status
```
mgduda: What's the rationale for making `test_status` an array? Would it be better to just follow what you've done in `test_remove_alarm`, and to have a local scalar to capture the return error code from each test routine, and to add that to the overall `ierr` return code after each test?

src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,205 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+   !>          cannot be removed from the opposite list and can be removed
+   !>          from the correct list. The stream type (input or output)
+   !>          is passed as an argument to test both scenarios.
+   !>
+   !> \param domain      The domain object that contains the stream manager
+   !>                    and clock.
+   !> \param ierr        The error code that indicates the result of the test.
+   !> \param streamType  The type of the stream being tested (input or output).
+   !
+   subroutine test_remove_alarm(domain, ierr, streamType)
```
mgduda: Just as a style suggestion, we almost invariably let the return error code (`ierr`) be the last argument to a subroutine.

src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,205 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+
+      integer, dimension(3) :: test_status
+
+      ierr = 0
+      test_status = 0
+
+      call mpas_log_write('Testing reading/writing real-valued streams')
+
+      call test_read_write_real_streams(domain, threadErrs, test_status(1))
+
```
mgduda: Here and on line 567, removing the blank lines would give more consistent whitespace with the blocks of code for alarm removal tests below.

mgduda (changes requested):

src/core_test/mpas_test_core_streams.F
amstokely: I initially thought this approach would allow the error code to be set to zero in one place. However, since the `ierr` argument is `intent(out)`, it must be initialized in the test anyway, so using an array doesn't offer any advantage. I'll update this in my next commit.

src/core_test/mpas_test_core_streams.F
mgduda: It looks like this documentation block still needs a line of hyphens.

src/core_test/mpas_test_core_streams.F
```diff
@@ -391,6 +391,208 @@ subroutine test_core_streams_test(domain, threadErrs, ierr)
...
+   !
+   !-----------------------------------------------------------------------
+   subroutine test_remove_output_alarm(domain, ierr)
+      use mpas_stream_manager
+      implicit none
+      type(domain_type), intent(inout) :: domain
+      integer, intent(out) :: ierr
+
+      call test_remove_alarm(domain, MPAS_STREAM_OUTPUT, ierr)
+   end subroutine test_remove_output_alarm
```
mgduda: Since this routine is just calling `test_remove_alarm` with a specific `streamType` argument and nothing more, what do you think about just eliminating this routine as well as `test_remove_input_alarm` and calling `test_remove_alarm` directly from `test_core_streams_test`?

amstokely: Works for me!

mgduda:
In the commit message for 44d15ed2, you state:
```
Tests fail when adding input alarm to an output list or output alarm to an
input list due to absence of error handling.
```
But it looks to me like the failure happens when we try to remove an alarm that doesn't exist. Is that right? If so, can you update the commit message?

If you'll be amending commit 44d15ed2, it might be nice to declare the dummy arguments to the `test_remove_alarm` routine in the order that they appear in the argument list.

mgduda:
Also, if you'll be reworking the commit history to address the above comment, can you also give another try at keeping the one-line summary in each commit to one line (and ~80 characters or less)? For example, I think the summary lines for 0d4a7fe6
```
Added error checking to MPAS_stream_mgr_remove_alarm to handle errors
returned by mpas_stream_list_remove.
```
could be shortened to
```
Handle errors returned by MPAS_stream_list_remove in MPAS_stream_mgr_remove_alarm
```
.

mgduda (approved):

PR #1315, opened 2025-05-13
Consolidating OpenACC device-host memory transfers

abishekg7:
This PR consolidates much of the OpenACC host and device data transfers during the course of the dynamical execution to two subroutines `mpas_atm_pre_dynamics_h2d` and `mpas_atm_post_dynamics_d2h` that are called before and after the call to `atm_srk3` subroutine. Due to `atm_compute_solve_diagnostics` also being called once before the start of model run, we also have a pair of subroutines `mpas_atm_pre_computesolvediag_h2d` and `mpas_atm_post_computesolvediag_d2h` to handle data movements around the first call to `atm_compute_solve_diagnostics`. Any fields copied onto the device in these subroutines are removed from explicit data movement statements in the dynamical core. 

The mesh/time-invariant fields are still copied onto the device in `mpas_atm_dynamics_init` and removed from the device in `mpas_atm_dynamics_finalize`, with the exception of select fields moved in `mpas_atm_pre_computesolvediag_h2d` and `mpas_atm_post_computesolvediag_d2h`. This is a special case due to `atm_compute_solve_diagnostics` being called for the first time before the call to `mpas_atm_dynamics_init`

This PR also includes explicit host-device data transfers in the `mpas_atm_iau`, `mpas_atmphys_interface` and `mpas_atmphys_todynamics` modules to ensure that the physics and IAU regions, which run on CPU, use the latest values from the dynamical core running on GPUs, and vice versa. In addition, this PR also includes explicit data transfers around halo exchanges in the `atm_srk3` subroutine.

These subroutines for data routines, and the `acc update` statements are an interim solution until we have a book-keeping method in place. 

This PR also introduces a couple of new timers to keep track of the cost of data transfers.

PR #1355, opened 2025-08-07
OpenACC port of halo exchanges 

abishekg7:
This PR enables execution of halo exchanges on GPUs via OpenACC directives, when the MPAS atmosphere core has been built with an appropriate GPU-aware MPI library. This PR builds on the OpenACC memory consolidation introduced in https://github.com/MPAS-Dev/MPAS-Model/pull/1315. 

Module `mpas_halo` is modified in the following ways to enable GPU-aware halo exchanges:
- In the call to `mpas_halo_exch_group_complete`, OpenACC directives copy to device all the relevant fields and metadata that are required for the packing and unpacking loops later.
- OpenACC directives are introduced around the packing and unpacking loops to perform the field to/from send/recv buffer operations on the device. The `attach` clauses introduced to the parallel constructs ensures that the device pointers are attached to the device targets at the start of the parallel region and detached at the end of the region.
- The actual `MPI_Isend` and `MPI_Irecv` operations use GPU-aware MPI, by wrapping these calls within
```
!$acc host_data use_device(pointer_to_buffer)

!$acc end host_data
```

In addition, this PR introduces a new namelist option under the `development` group, `config_gpu_aware_mpi`. When set to `true`, it switches on GPU-direct halo exchanges. The default setting is `false`, in which halo exchanges occur on the host, which necessitates host<->device data transfers around the halo exchanges as necessary.

Note: This feature is not available to use when config_halo_exch_method is set to 'mpas_dmpar'

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -443,6 +443,11 @@
                      units="-"
                      description="Method to use for exchanging halos"
                      possible_values="`mpas_dmpar', `mpas_halo'"/>
+
+               <nml_option name="config_gpu_aware_mpi" type="logical" default_value="false"
+                     units="-"
+                     description="Whether to use GPU-aware MPI for halo exchanges. GPU-aware MPI is only implemented for config_halo_exch_method='mpas_halo'"
+                     possible_values=".true. or .false."/>
         </nml_record>
 
 #ifdef MPAS_USE_MUSICA
--- a/src/framework/mpas_dmpar.F
+++ b/src/framework/mpas_dmpar.F
@@ -7450,10 +7450,11 @@ subroutine mpas_dmpar_get_exch_list(haloLayer, ownedListField, neededListField, 
 !>  exchange is complete.
 !
 !-----------------------------------------------------------------------
-   subroutine mpas_dmpar_exch_group_full_halo_exch(domain, groupName, iErr)!{{{
+   subroutine mpas_dmpar_exch_group_full_halo_exch(domain, groupName, withGPUAwareMPI, iErr)!{{{
 
       type (domain_type), intent(inout) :: domain
       character (len=*), intent(in) :: groupName
+      logical, optional, intent(in) :: withGPUAwareMPI
       integer, optional, intent(out) :: iErr
 
       type (mpas_exchange_group), pointer :: exchGroupPtr
@@ -7461,6 +7462,12 @@ subroutine mpas_dmpar_get_exch_list(haloLayer, ownedListField, neededListField, 
 
       if ( present(iErr) ) then
          iErr = MPAS_DMPAR_NOERR
+      end if
+
+      if (present(withGPUAwareMPI)) then
+        if (withGPUAwareMPI) then
+           call mpas_log_write(' GPU-aware MPI not implemented in this module', MPAS_LOG_CRIT)
+        end if
       end if
 
       nLen = len_trim(groupName)
--- a/src/framework/mpas_halo.F
+++ b/src/framework/mpas_halo.F
@@ -280,6 +280,29 @@ module mpas_halo
 
         call refactor_lists(domain, groupName, iErr)
 
+        ! Logic to return early if there are no neighbors to send to
+        if ( newGroup% nGroupSendNeighbors <=0 ) then
+            return
+        end if
+
+
+        ! Always copy in the main data member first
+        !$acc enter data copyin(newGroup)
+        ! Then the data in the members of the type
+        !$acc enter data copyin(newGroup % recvBuf(:), newGroup % sendBuf(:))
+        !$acc enter data copyin(newGroup % fields(:))
+        do i = 1, newGroup % nFields
+            !$acc enter data copyin(newGroup % fields(i))
+            !$acc enter data copyin(newGroup % fields(i) % nSendLists(:,:))
+            !$acc enter data copyin(newGroup % fields(i) % packOffsets(:))
+            !$acc enter data copyin(newGroup % fields(i) % sendListSrc(:,:,:))
+            !$acc enter data copyin(newGroup % fields(i) % sendListDst(:,:,:))
+            !$acc enter data copyin(newGroup % fields(i) % nRecvLists(:,:))
+            !$acc enter data copyin(newGroup % fields(i) % unpackOffsets(:))
+            !$acc enter data copyin(newGroup % fields(i) % recvListSrc(:,:,:))
+            !$acc enter data copyin(newGroup % fields(i) % recvListDst(:,:,:))
+        end do
+
     end subroutine mpas_halo_exch_group_complete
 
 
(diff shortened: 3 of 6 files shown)


src/framework/mpas_halo.F
```diff
@@ -508,6 +547,7 @@ subroutine mpas_halo_exch_group_full_halo_exch(domain, groupName, iErr)
         use mpas_derived_types, only : domain_type, mpas_halo_group, MPAS_HALO_REAL, MPAS_LOG_CRIT
         use mpas_pool_routines, only : mpas_pool_get_array
         use mpas_log, only : mpas_log_write
+        use mpas_timer, only : mpas_timer_start, mpas_timer_stop
```
mgduda: Don't forget to update the `Makefile` to add a dependency on `mpas_timer.o` for the `mpas_halo.o` target.

src/core_atmosphere/Registry.xml
```diff
@@ -443,6 +443,10 @@
                      units="-"
                      description="Method to use for exchanging halos"
                      possible_values="`mpas_dmpar', `mpas_halo'"/>
+               <nml_option name="config_gpu_aware_mpi" type="logical" default_value="false"
```
mgduda, line 447: For consistency and readability, let's consider adding a blank line above this. (I do see that the `IAU` namelist record doesn't use a blank line between options, but we can address that in a separate clean-up PR at a later time.)

src/core_atmosphere/Registry.xml
```diff
@@ -443,6 +443,10 @@
...
+                     units="-"
+                     description="Whether to use GPU-aware MPI for halo exchanges"
```
mgduda: In the description, we could also mention that this only works when `config_halo_exch_method` is set to `mpas_halo`.

src/framework/mpas_dmpar.F
```diff
@@ -7450,19 +7450,28 @@ end subroutine mpas_dmpar_exch_group_end_halo_exch!}}}
...
 
       type (mpas_exchange_group), pointer :: exchGroupPtr
       integer :: nLen
+      logical :: useGPUAwareMPI
 
       if ( present(iErr) ) then
          iErr = MPAS_DMPAR_NOERR
       end if
 
+      useGPUAwareMPI = .false.
```
mgduda: Is this local variable actually needed?

src/core_atmosphere/mpas_atm_halos.F
```diff
@@ -48,18 +48,23 @@ subroutine atm_build_halo_groups(domain, ierr)
...
       call mpas_pool_get_config(domain % blocklist % configs, 'config_halo_exch_method', config_halo_exch_method)
+      call mpas_pool_get_config(domain % blocklist % configs, 'config_gpu_aware_mpi', config_gpu_aware_mpi)
 
       if (trim(config_halo_exch_method) == 'mpas_dmpar') then
          call mpas_log_write('')
          call mpas_log_write('*** Using ''mpas_dmpar'' routines for exchanging halos')
          call mpas_log_write('')
 
+         if (config_gpu_aware_mpi) then
+            call mpas_log_write('GPU-aware MPI is not presently supported with config_halo_exch_method = mpas_dmpar',MPAS_LOG_CRIT)
```
mgduda: This seems like it's redundant given the check that exists in the `mpas_dmpar_exch_group_full_halo_exch` routine.

src/framework/mpas_halo.F
```diff
@@ -555,11 +596,15 @@ subroutine mpas_halo_exch_group_full_halo_exch(domain, groupName, iErr)
         integer, dimension(:,:,:), CONTIGUOUS pointer :: recvListSrc, recvListDst
         integer, dimension(:), CONTIGUOUS pointer :: unpackOffsets
 
-
```
mgduda, line 558: I'd suggest leaving this blank line in place, as this change isn't necessary to the objective of this PR and this line wouldn't otherwise be modified.

src/framework/mpas_halo.F
```diff
@@ -577,6 +622,12 @@ subroutine mpas_halo_exch_group_full_halo_exch(domain, groupName, iErr)
                                 messageType=MPAS_LOG_CRIT)
         end if
 
+        ! Logic to return early if there are no neighbors to send to
+        if ( group% nGroupSendNeighbors <= 0 ) then
+            return
+        end if
+
+        call mpas_timer_start('full_halo_exch')
```
mgduda: In the past, my experience has been that timing for halo exchanges may not be especially valuable. Load imbalance can show up as time apparently spent in halo communications, when it's really just that one MPI task had to finish some extra work before participating in the halo exchange. What do you think about removing these timers?

src/framework/mpas_halo.F
```diff
@@ -589,6 +640,7 @@ subroutine mpas_halo_exch_group_full_halo_exch(domain, groupName, iErr)
 #endif
         rank = group % fields(1) % compactHaloInfo(8)
 
+        !$acc data present(group % recvBuf(:), group % sendBuf(:)) if(useGPUAwareMPI)
```
mgduda: If the `parallel` directives in this routine all use `default(present)`, is this directive necessary?

mgduda (changes requested):

src/framework/mpas_halo.F
abishekg7: That's true. I'm fine with removing it for now and revisiting it as needed.

src/core_atmosphere/mpas_atm_halos.F
abishekg7: I will retain this check and remove the redundant one in `mpas_dmpar_exch_group_full_halo_exch` so that it will terminate earlier.

mgduda: Actually, I think the check in `mpas_dmpar_exch_group_full_halo_exch` is preferable, since it would catch uses of the routine with `withGPUAwareMPI = .true.` in cores other than the `atmosphere` core.

src/framework/mpas_halo.F
abishekg7: Doesn't appear to be required, and will remove it.

src/core_atmosphere/mpas_atm_halos.F
```diff
@@ -48,12 +48,14 @@ subroutine atm_build_halo_groups(domain, ierr)
...
       ! Local variables
       character(len=StrKIND), pointer :: config_halo_exch_method
+      logical, pointer :: config_gpu_aware_mpi
 
 
       !
       ! Determine from the namelist option config_halo_exch_method which halo exchange method to employ
       !
       call mpas_pool_get_config(domain % blocklist % configs, 'config_halo_exch_method', config_halo_exch_method)
+      call mpas_pool_get_config(domain % blocklist % configs, 'config_gpu_aware_mpi', config_gpu_aware_mpi)
```
mgduda: It doesn't look like `config_gpu_aware_mpi` is subsequently used in this routine. Should we just remove it altogether?

mgduda (changes requested):

src/core_atmosphere/mpas_atm_halos.F
abishekg7: Thanks! Fixed.

mgduda (approved):

PR #1356, opened 2025-08-08
Fix Linking of PnetCDF in CMake

amstokely:
**Summary:**
This PR resolves an issue with linking the PnetCDF C library in the CMake build system. It ensures that the `PnetCDF::PnetCDF_C` target is correctly created and linked by updating the `FindPnetCDF.cmake` file.

**Changes:**
The primary fix addresses insufficient linking when using `INTERFACE_LINK_DIRECTORIES`. Previously, `PnetCDF::PnetCDF_C` was only setting `INTERFACE_LINK_DIRECTORIES`, which specified the directory for the linker but did not actually link the necessary `pnetcdf` library. This prevented CMake targets from linking with parallel-netcdf when using `target_link_libraries(<target> PnetCDF::PnetCDF_C)`. The update now defines `INTERFACE_LINK_LIBRARIES` as `pnetcdf`, ensuring proper linkage.

**Tests:**
The validity of this change can be verified by compiling the develop branch of MPAS with CMake, enabling SMIOL support, and building it as a shared library. When running `ldd` on `libsmiol.so`, `pnetcdf` should not be listed on the develop branch. However, with this branch, `ldd` will confirm that `libsmiol.so` is correctly linked against parallel-netcdf.

### DIFF
--- a/cmake/Modules/FindPnetCDF.cmake
+++ b/cmake/Modules/FindPnetCDF.cmake
@@ -133,19 +133,20 @@ set(_new_components)
 if(PnetCDF_Fortran_FOUND AND NOT TARGET PnetCDF::PnetCDF_Fortran)
     add_library(PnetCDF::PnetCDF_Fortran INTERFACE IMPORTED)
     set_target_properties(PnetCDF::PnetCDF_Fortran PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${PnetCDF_INCLUDE_DIR}
-                                                              INTERFACE_LINK_DIRECTORIES ${PnetCDF_LIB_DIR})
+                                                              INTERFACE_LINK_DIRECTORIES ${PnetCDF_LIB_DIR}
+                                                              INTERFACE_LINK_LIBRARIES pnetcdf)
     if(PnetCDF_MODULE_DIR AND NOT PnetCDF_MODULE_DIR STREQUAL PnetCDF_INCLUDE_DIR )
         set_property(TARGET PnetCDF::PnetCDF_Fortran APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${PnetCDF_MODULE_DIR})
     endif()
     set(_new_components 1)
-    target_link_libraries(PnetCDF::PnetCDF_Fortran INTERFACE -lpnetcdf)
 endif()
 
 # PnetCDF::PnetCDF_C imported interface target
 if(PnetCDF_C_FOUND AND NOT TARGET PnetCDF::PnetCDF_C)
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindPnetCDF.cmake
```diff
@@ -133,27 +133,29 @@ set(_new_components)
 if(PnetCDF_Fortran_FOUND AND NOT TARGET PnetCDF::PnetCDF_Fortran)
     add_library(PnetCDF::PnetCDF_Fortran INTERFACE IMPORTED)
     set_target_properties(PnetCDF::PnetCDF_Fortran PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${PnetCDF_INCLUDE_DIR}
-                                                              INTERFACE_LINK_DIRECTORIES ${PnetCDF_LIB_DIR})
+                                                              INTERFACE_LINK_DIRECTORIES ${PnetCDF_LIB_DIR}
+                                                              INTERFACE_LINK_LIBRARIES "-L${PnetCDF_LIB_DIR} -lpnetcdf")
```
mgduda: Is `-L${PnetCDF_LIB_DIR}` needed in `INTERFACE_LINK_LIBRARIES` if `${PnetCDF_LIB_DIR}` is present in `INTERFACE_LINK_DIRECTORIES`?

amstokely: It is not. I removed it in my latest commit.

mgduda: Perhaps I should restate my original comment. Couldn't lines 137, 149, and 158 be reduced to
```
                                                              INTERFACE_LINK_LIBRARIES "-lpnetcdf")
```
?

mgduda (approved):

mgduda:
I've just approved this PR, but it would be good to hear from @jim-p-w whether the changes work in MPAS-JEDI.

jim-p-w:
I built mpas-bundle with these cmake changes. Both the single and double precision builds were fine. Double precision mpas-jedi ctests passed. A cylc workflow test (3dvar_OIE120km_ColdStart) passed.

jim-p-w (approved):

mgduda:
@amstokely Could you rework the commit history before we merge this PR? I think just a single commit is all that's needed.

amstokely:
> @amstokely Could you rework the commit history before we merge this PR? I think just a single commit is all that's needed.

Yes

fabiolrdiniz:
Hi @mgduda, I don't know what is the procedure for tagging the code, but would be possible to have this change in one of the tagged versions?

PR #1358, opened 2025-08-18
Update bl_ysu_gwdo parameterization scheme 

Songyou184:
Update of the YSU_GWD scheme (Choi and Hong 2015, JGR), and renamed it as KIM_GWD scheme.
Revisions include sub-grid orographic statistics and configuration of the drag stress and blocking stress, as 
described in Hong et al. (2025,WAF)

### DIFF
--- a/src/core_atmosphere/Externals.cfg
+++ b/src/core_atmosphere/Externals.cfg
@@ -1,8 +1,8 @@
 [MMM-physics]
 local_path = ./physics_mmm
 protocol = git
-repo_url = https://github.com/NCAR/MMM-physics.git
-tag = 20250616-MPASv8.3
+repo_url = https://github.com/Songyou184/MMM-physics.git
+branch=KIM_GWDO
 required = True
 
 [GSL_UGWP]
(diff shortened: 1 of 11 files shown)


byoung-joo:
FYI: Testing this PR requires a companion PR https://github.com/NCAR/MMM-physics/pull/7

Songyou184:
Yes, it is true... ! a few more points to be addressed.

1. The name of the scheme is "bl_kim_gwdo" instead of "bl_ysu_gwdo".
2. Two additional configures in "namelist.atmosphere"

&physics

    config_sst_update = false

    config_sstdiurn_update = false

    config_deepsoiltemp_update = false

    config_radtlw_interval = '00:30:00'

    config_radtsw_interval = '00:30:00'

    config_bucket_update = 'none'

    config_kim_tofd = true

    config_gwdo_nonhyd = true

    config_physics_suite = 'mesoscale_reference'

/

These two red configs are defined as "false" by default.
"config_kim_gwdo=true" should result in an overall decrease of surface
layer wind.
"config_gwdo_nonhyd=true" should result in an overall decrease of GWD
stress by less than 10 % over high mountains. A significant reduction
appears over the Antarctic.

Songyou

PR #1359, opened 2025-08-19
Refactoring duplicate definitions of halo_exchange_routine

abishekg7:
This PR introduces a new include file `mpas_halo_interface` under src/framework in order to remove the repeated definitions of the generic interface halo_exchange_routine and replace them with an include.

This is also preparatory work for https://github.com/MPAS-Dev/MPAS-Model/pull/1355

### DIFF
--- a/src/framework/mpas_halo_interface.inc
+++ b/src/framework/mpas_halo_interface.inc
@@ -1,0 +1,15 @@
+   !
+   ! Abstract interface for routine used to communicate halos of fields
+   ! in a named group
+   !
+   abstract interface
+      subroutine halo_exchange_routine(domain, halo_group, ierr)
+
+         use mpas_derived_types, only : domain_type
+
+         type (domain_type), intent(inout) :: domain
+         character(len=*), intent(in) :: halo_group
+         integer, intent(out), optional :: ierr
+
+      end subroutine halo_exchange_routine
+   end interface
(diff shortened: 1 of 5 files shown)


src/framework/mpas_halo_interface.inc
```diff
@@ -0,0 +1,15 @@
...
+      subroutine halo_exchange_routine(domain, halo_group, ierr)
+
+         use mpas_derived_types, only : domain_type
+
+         type (domain_type), intent(inout) :: domain
+         character(len=*), intent(in) :: halo_group
+         integer, intent(out), optional :: ierr
+
+      end subroutine halo_exchange_routine
+   end interface
```
mgduda: It looks like this file might be missing a final EOL character.

src/framework/mpas_halo_interface.inc
```diff
@@ -0,0 +1,15 @@
+!
```
mgduda: Let's add whitespace so the indentation matches the lines below.

src/framework/mpas_halo_interface.inc
abishekg7: I believe this is fixed now

src/framework/mpas_halo_interface.inc
abishekg7: done

mgduda (approved):
The code changes look good to me. Please feel free to squash the whitespace fixup commit into the original commit, and then I'll merge.

abishekg7:
Thanks! Squashed into a single commit.

PR #1360, opened 2025-08-25
Chemistry stub modules for MUSICA MICM

dwfncar:
- Added subdirs core_atmosphere/chemistry and core_atmosphere/chemistry/musica, with Makefiles. 
- Modified core_atmosphere/Makefile to build chemistry source. 
- Added MUSICA MICM stub modules chemistry/mpas_atm_chemistry.F and chemistry/musica/mpas_musica.F.

The module chemistry/mpas_atm_chemistry.F has non-implemented stub subroutines for
- `init_chemistry` (will call `musica_init` from **mpas_musica**)
- `step_chemistry` (will call `musica_step` from **mpas_musica**)
- `finalize_chemistry` (will call `musica_finalize` from **mpas_musica**)

chemistry/musica/mpas_musica.F will define the **mpas_musica** module with the `musica_init`, `musica_step`, and `musica_finalize` subroutines, also to be implemented.
It uses and directly links to the MUSIC MICM modules
- **musica_micm**
- **musica_state**
- **musica_util**

Also included is an error checking function `has_error_occurred`.

Note that the call to `music_init`  will take as an argument the chemistry species defined as a new mpas_var_array to be implemented in a new framework module, for instance **mpas_runtime_vars**.  The calling arguments will be further specified in subsequent PRs.

### DIFF
--- a/src/core_atmosphere/Makefile
+++ b/src/core_atmosphere/Makefile
@@ -2,17 +2,23 @@
 
 #
 # To build a dycore-only MPAS-Atmosphere model, comment-out or delete
-# the definition of PHYSICS, below
+# the definition of PHYSICS and CHEMISTRY, below
 #
-# If MPAS_CAM_DYCORE is found in CPPFLAGS, PHYSICS will become undefined automatically
+# If MPAS_CAM_DYCORE is found in CPPFLAGS, PHYSICS and CHEMISTRY will become undefined automatically
 #
 ifeq ($(findstring MPAS_CAM_DYCORE,$(CPPFLAGS)),)
     PHYSICS = -DDO_PHYSICS
+    CHEMISTRY = -DDO_CHEMISTRY
 endif
 
 ifdef PHYSICS
     PHYSCORE = physcore
     PHYS_OBJS = libphys/*.o
+endif
+
+ifdef CHEMISTRY
+    CHEMCORE = chemcore
+    CHEM_OBJS = libchem/*.o
 endif
 
 OBJS = mpas_atm_core.o \
@@ -21,10 +27,10 @@ OBJS = mpas_atm_core.o \
        mpas_atm_threading.o \
        mpas_atm_halos.o
 
-all: $(PHYSCORE) dycore diagcore atmcore utilities
+all: $(PHYSCORE) $(CHEMCORE) dycore diagcore atmcore utilities
 
 core_reg:
-	$(CPP) $(CPPFLAGS) $(CPPINCLUDES) $(PHYSICS) Registry.xml > Registry_processed.xml
+	$(CPP) $(CPPFLAGS) $(CPPINCLUDES) $(PHYSICS) $(CHEMISTRY) Registry.xml > Registry_processed.xml
 
 core_input_gen:
 	if [ ! -e default_inputs ]; then  mkdir default_inputs; fi
@@ -47,7 +53,11 @@ physcore: mpas_atm_dimensions.o
 	( cd ../..; ln -sf ./src/core_atmosphere/physics/physics_wrf/files/*DATA* .)
 	( cd ../..; ln -sf ./src/core_atmosphere/physics/physics_noahmp/parameters/*TBL .)
 
-dycore: mpas_atm_dimensions.o $(PHYSCORE)
+chemcore:
+	( cd chemistry; $(MAKE) all CHEMISTRY="$(CHEMISTRY)" )
+	( mkdir libchem; cd libchem; ar -x ../chemistry/libchem.a )
+
+dycore: mpas_atm_dimensions.o $(PHYSCORE) $(CHEMCORE)
 	( cd dynamics; $(MAKE) all PHYSICS="$(PHYSICS)" )
 
 diagcore: $(PHYSCORE) dycore
@@ -57,7 +67,7 @@ utilities: $(PHYSCORE)
 	( cd utils; $(MAKE) all PHYSICS="$(PHYSICS)" )
 
 atmcore: $(PHYSCORE) dycore diagcore $(OBJS)
-	ar -ru libdycore.a $(OBJS) dynamics/*.o $(PHYS_OBJS) diagnostics/*.o
+	ar -ru libdycore.a $(OBJS) dynamics/*.o $(PHYS_OBJS) $(CHEM_OBJS) diagnostics/*.o
 
 mpas_atm_core_interface.o: mpas_atm_core.o
 
@@ -67,6 +77,7 @@ mpas_atm_dimensions.o:
 
 clean:
 	( cd physics; $(MAKE) clean )
+	( cd chemistry; $(MAKE) clean )
 	( cd dynamics; $(MAKE) clean )
 	( cd diagnostics; $(MAKE) clean )
 	( cd utils; $(MAKE) clean )
@@ -83,8 +94,8 @@ clean:
 .F.o:
 	$(RM) $@ $*.mod
 ifeq "$(GEN_F90)" "true"
-	$(CPP) $(CPPFLAGS) $(PHYSICS) $(CPPINCLUDES) -I./inc $< > $*.f90
-	$(FC) $(FFLAGS) -c $*.f90 $(FCINCLUDES) -I../framework -I../operators -I./physics -I./dynamics -I./diagnostics -I./physics/physics_wrf -I./physics/physics_mmm -I./physics/physics_noaa/UGWP -I../external/esmf_time_f90
+	$(CPP) $(CPPFLAGS) $(PHYSICS) $(CHEMISTRY) $(CPPINCLUDES) -I./inc $< > $*.f90
+	$(FC) $(FFLAGS) -c $*.f90 $(FCINCLUDES) -I../framework -I../operators -I./physics -I./dynamics -I./diagnostics -I./physics/physics_wrf -I./physics/physics_mmm -I./physics/physics_noaa/UGWP -I../external/esmf_time_f90 -I./chemistry
 else
-	$(FC) $(CPPFLAGS) $(PHYSICS) $(FFLAGS) -c $*.F $(CPPINCLUDES) $(FCINCLUDES) -I./inc -I../framework -I../operators -I./physics -I./dynamics -I./diagnostics -I./physics/physics_wrf -I./physics/physics_mmm -I./physics/physics_noaa/UGWP -I../external/esmf_time_f90
+	$(FC) $(CPPFLAGS) $(PHYSICS) $(CHEMISTRY) $(FFLAGS) -c $*.F $(CPPINCLUDES) $(FCINCLUDES) -I./inc -I../framework -I../operators -I./physics -I./dynamics -I./diagnostics -I./physics/physics_wrf -I./physics/physics_mmm -I./physics/physics_noaa/UGWP -I../external/esmf_time_f90 -I./chemistry
 endif
--- a/src/core_atmosphere/chemistry/Makefile
+++ b/src/core_atmosphere/chemistry/Makefile
@@ -1,0 +1,47 @@
+.SUFFIXES: .F .o
+
+all:
+ifeq ($(MUSICA),true)
+	echo "****** building MUSICA ******"
+	$(MAKE) core_mpas_musica core_chemistry
+else
+	echo "****** not building MUSICA ******"
+	$(MAKE) core_chemistry
+endif
+
+dummy:
+	echo "****** compiling chemistry ******"
+
+OBJS = \
+	mpas_atm_chemistry.o
+
+ifeq ($(MUSICA),true)
+core_mpas_musica:
+	echo "****** building CORE MPAS MUSICA ******"
+	(cd musica; $(MAKE) all)
+endif
+
+core_chemistry: $(if $(filter true,$(MUSICA)),core_mpas_musica)
+	($(MAKE) chem_interface)
+	ar -ru libchem.a $(OBJS)
+ifeq ($(MUSICA),true)
+	($(MAKE) -C ./musica mpas_musica_lib)
+endif
+
+chem_interface: $(OBJS)
+
+clean:
+	$(RM) *.o *.mod *.f90 libchem.a
+	( cd musica; $(MAKE) clean )
+	@# Certain systems with intel compilers generate *.i files
+	@# This removes them during the clean process
+	$(RM) *.i
+
+.F.o:
+	$(RM) $@ $*.mod
+ifeq "$(GEN_F90)" "true"
+	$(CPP) $(CPPFLAGS) $(CPPINCLUDES) $< > $*.f90
+	$(FC) $(FFLAGS) -c $*.f90 $(FCINCLUDES) -I./musica -I.. -I../../framework -I../../../external/esmf_time_f90
+else
+	$(FC) $(CPPFLAGS) $(FFLAGS) -c $*.F $(CPPINCLUDES) $(FCINCLUDES) -I./musica -I.. -I../../framework -I../../../external/esmf_time_f90
+endif
--- a/src/core_atmosphere/chemistry/mpas_atm_chemistry.F
+++ b/src/core_atmosphere/chemistry/mpas_atm_chemistry.F
@@ -1,0 +1,126 @@
+! Copyright (c) 2025 The University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
+! Additional copyright and license information can be found in the LICENSE file
+! distributed with this code, or at https://mpas-dev.github.io/license.html .
+!
+!-----------------------------------------------------------------------
+!  mpas_atm_chemistry
+!
+!> \brief Manages interactions with chemistry packages
+!> \author CheMPAS-A Developers
+!> \date   20 August 2025
+!> \details
+!>  This module manages the interactions with chemistry packages,
+!>  including initialization, time-stepping, and finalization.
+!>  It provides a framework for integrating various chemistry models
+!>  into the CheMPAS-A system.
+!
+!-------------------------------------------------------------------------
+module mpas_atm_chemistry
+
+    implicit none
+
+    private
+
+    public :: chemistry_init, chemistry_step, chemistry_finalize
+
+    contains
+
+    !------------------------------------------------------------------------
+    !  routine chemistry_init
+    !
+    !> \brief Initializes the chemistry packages
+    !> \author CheMPAS-A Developers
+    !> \date   20 August 2025
+    !> \details
+    !>  This routine initializes the chemistry packages, setting up
+    !>  necessary parameters and data structures for the simulation.
+    !------------------------------------------------------------------------
+    subroutine chemistry_init(configs, dimensions)
+
+#ifdef MPAS_USE_MUSICA
+        use mpas_musica, only: musica_init
+#endif
+        use mpas_log, only : mpas_log_write
+        use mpas_derived_types, only: mpas_pool_type
+        use mpas_kind_types, only: StrKIND
+        use mpas_pool_routines, only: mpas_pool_get_config, mpas_pool_get_dimension 
+
+        type (mpas_pool_type), intent(in) :: configs
+        type (mpas_pool_type), intent(in) :: dimensions
+
+#ifdef MPAS_USE_MUSICA
+        integer                       :: error_code
+        character(len=:), allocatable :: error_message
+        integer                       :: nVertLevels
+        integer, pointer              :: nVertLevels_ptr
+        ! MUSICA will get the MICM JSON config from a namelist
+        ! hardcode filepath for now
+        character(len=StrKIND) :: filepath = 'chapman.json'
+#endif
+
+        call mpas_log_write('Initializing chemistry packages...')
+
+#ifdef MPAS_USE_MUSICA
+        call mpas_pool_get_dimension(dimensions, 'nVertLevels', nVertLevels_ptr)
+        nVertLevels = nVertLevels_ptr
+
+        call musica_init(filepath, nVertLevels, error_code, error_message)
+
+        ! TODO check error_code and generate MPAS error log message
+#endif
+
+    end subroutine chemistry_init
+
+
+    !------------------------------------------------------------------------
+    !  routine chemistry_step
+    !
+    !> \brief Steps the chemistry packages
+    !> \author CheMPAS-A Developers
+    !> \date   20 August 2025
+    !> \details
+    !>  This routine steps the chemistry packages, updating their state
+    !>  based on the current simulation time and conditions.
+    !------------------------------------------------------------------------
+    subroutine chemistry_step()
+
+#ifdef MPAS_USE_MUSICA
+        use mpas_musica, only: musica_step
+#endif
+        use mpas_log, only : mpas_log_write
+
+#ifdef MPAS_USE_MUSICA
+        call mpas_log_write('Stepping chemistry packages...')
+        ! call musica_step()
+#endif
+
+    end subroutine chemistry_step
+
+
+    !------------------------------------------------------------------------
+    !  routine chemistry_finalize
+    !
+    !> \brief Finalizes the chemistry packages
+    !> \author CheMPAS-A Developers
+    !> \date   20 August 2025
+    !> \details
+    !>  This routine finalizes the chemistry packages, cleaning up
+    !>  any resources and data structures used during the simulation.
+    !------------------------------------------------------------------------
+    subroutine chemistry_finalize()
+
+#ifdef MPAS_USE_MUSICA
+        use mpas_musica, only: musica_finalize
+#endif
+        use mpas_log, only : mpas_log_write
+
+#ifdef MPAS_USE_MUSICA
+        call mpas_log_write('Finalizing chemistry packages...')
+        call musica_finalize()
+#endif
+
+    end subroutine chemistry_finalize
+
+end module mpas_atm_chemistry
--- a/src/core_atmosphere/chemistry/musica/Makefile
+++ b/src/core_atmosphere/chemistry/musica/Makefile
@@ -1,0 +1,31 @@
+.SUFFIXES: .F .o
+
+.PHONY: mpas_musica mpas_musica_lib
+
+all: dummy mpas_musica
+
+dummy:
+	echo "****** compiling mpas_musica ******"
+
+OBJS = \
+  mpas_musica.o
+
+mpas_musica: $(OBJS)
+
+mpas_musica_lib:
+	ar -ru ./../libchem.a $(OBJS)
+
+clean:
+	$(RM) *.f90 *.o *.mod
+	@# Certain systems with intel compilers generate *.i files
+	@# This removes them during the clean process
+	$(RM) *.i
+
+.F.o:
+ifeq "$(GEN_F90)" "true"
+	$(CPP) $(CPPFLAGS) $(COREDEF) $(CPPINCLUDES) $< > $*.f90
+	$(FC) $(FFLAGS) -c $*.f90 $(FCINCLUDES) -I.. -I../../../framework -I../../../external/esmf_time_f90
+else
+	$(FC) $(CPPFLAGS) $(COREDEF) $(FFLAGS) -c $*.F $(CPPINCLUDES) $(FCINCLUDES) -I.. -I../../../framework -I../../../external/esmf_time_f90
+endif
+
--- a/src/core_atmosphere/chemistry/musica/mpas_musica.F
+++ b/src/core_atmosphere/chemistry/musica/mpas_musica.F
@@ -1,0 +1,161 @@
+! Copyright (c) 2025 The University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
+! Additional copyright and license information can be found in the LICENSE file
+! distributed with this code, or at https://mpas-dev.github.io/license.html .
+!
+!-----------------------------------------------------------------------
+!  mpas_musica
+!
+!> \brief Manages interactions with the MUSICA chemistry package
+!> \author CheMPAS-A Developers
+!> \date   20 August 2025
+!> \details
+!>  This module manages the interactions with the MUSICA chemistry package,
+!>  including initialization, time-stepping, and finalization for
+!>  MICM (ODE solver) and TUV-x (photolysis rate constant calculator).
+!--------------------------------------------------------------------------
+module mpas_musica
+
+    use musica_micm,  only: micm_t, get_micm_version
+    ! get_micm_version is here to avoid an ICE when within musica_init
+    use musica_state, only: state_t
+
+    implicit none
+
+    private
+
+    public :: musica_init, musica_step, musica_finalize
+
+    type(micm_t),  pointer :: micm  => null ( )  ! Pointer to the MICM ODE solver instance
+    type(state_t), pointer :: state => null ( )  ! Pointer to the state of the MICM solver
+
+    contains
+
+    !------------------------------------------------------------------------
+    !  subroutine musica_init
+    !
+    !> \brief Initializes the MUSICA chemistry package
+    !> \author CheMPAS-A Developers
+    !> \date   20 August 2025
+    !> \details
+    !>  This subroutine initializes the MUSICA chemistry package,
+    !>  setting up necessary parameters and data structures for the simulation.
+    !>  For now, we will load fixed configurations for MICM and TUV-x.
+    !>  Later, we will gradually replace the fixed configuration elements
+    !>  with runtime updates using the MUSICA API
+    !------------------------------------------------------------------------
+    subroutine musica_init(filename_of_micm_configuration, &
+            number_of_grid_cells, &
+            error_code, error_message)
+
+        use musica_micm, only : RosenbrockStandardOrder
+        use musica_util, only : error_t, string_t
+
+        use mpas_log, only : mpas_log_write
+
+        character(len=*),              intent(in)    :: filename_of_micm_configuration
+        integer,                       intent(in)    :: number_of_grid_cells
+        integer,                       intent(out)   :: error_code
+        character(len=:), allocatable, intent(out)   :: error_message
+
+        type(error_t)  :: error
+        type(string_t) :: micm_version
+
+        ! TEMPORARY: Hard-coded options for the MICM solver
+        integer :: solver_type = RosenbrockStandardOrder
+
+        micm_version = get_micm_version()
+
+        call mpas_log_write('Initializing MUSICA chemistry package...')
+        call mpas_log_write('MICM version: ' // micm_version%value_)
+        call mpas_log_write('MICM number of grid cells: $i', intArgs=[number_of_grid_cells])
+
+        micm => micm_t(trim(filename_of_micm_configuration), solver_type, error)
+        if (has_error_occurred(error, error_message, error_code)) return
+
+        state => micm%get_state(number_of_grid_cells, error)
+        if (has_error_occurred(error, error_message, error_code)) return
+
+    end subroutine musica_init
+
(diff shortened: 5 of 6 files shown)


K20shores (approved):

src/core_atmosphere/mpas_atm_core.F
```diff
@@ -48,6 +48,9 @@ function atm_core_init(domain, startTimeStamp) result(ierr)
       use mpas_attlist, only : mpas_modify_att
       use mpas_string_utils, only : mpas_string_replace
       use mpas_atm_halos, only: atm_build_halo_groups, exchange_halo_group
+#ifdef MPAS_USE_MUSICA
+      use mpas_atm_chemistry, only: chemistry_init
+#endif
```
mattldawson: I might be mis-remembering, but I thought we decided to use the `MPAS_USE_MUSICA` macro inside the `mpas_atm_chemistry` module. I think it would make more sense to use it to wrap calls to functions in `mpas_musica`. Then, inside functions like `chemistry_init()`, you could have something like:
```
#ifdef MPAS_USE_MUSICA
  call musica_init()
#endif
```

K20shores (approved):
I have no problems with this, but I do have the same question as matt

src/core_atmosphere/mpas_atm_core.F
dwfncar: Moved `#ifdef MPAS_USE_MUSICA` to module `mpas_atm_chemistry`.

src/core_atmosphere/chemistry/Makefile
```diff
@@ -0,0 +1,41 @@
+.SUFFIXES: .F .o
+
+ifeq ($(CORE),atmosphere)
+COREDEF = -Dmpas
+endif
+
+all:
+	./../tools/manage_externals/checkout_externals --externals ./../Externals.cfg
+	$(MAKE) core_mpas_musica core_chemistry
```
mattldawson, line 6: Have you tested with w/o including MUSICA library in the build? I think we should exclude the `mpas_musica.F` file in the build when MUSICA is not linked.

dwfncar: `# Add at the top:
ifdef MPAS_USE_MUSICA
BUILD_MUSICA = yes
else
BUILD_MUSICA = no
endif

...

# Modify the all and core_chemistry targets:

all:
	./../tools/manage_externals/checkout_externals --externals ./../Externals.cfg
ifeq ($(BUILD_MUSICA),yes)
	$(MAKE) core_mpas_musica core_chemistry
else
	$(MAKE) core_chemistry
endif

...

core_chemistry: $(if $(filter yes,$(BUILD_MUSICA)),core_mpas_musica)
	($(MAKE) chem_interface COREDEF="$(COREDEF)")
	ar -ru libchem.a $(OBJS)
ifeq ($(BUILD_MUSICA),yes)
	($(MAKE) -C ./musica mpas_musica_lib)
endif

...

# Only define the core_mpas_musica target if building musica:
ifeq ($(BUILD_MUSICA),yes)
core_mpas_musica:
	(cd musica; $(MAKE) all COREDEF="$(COREDEF)")
endif
`

dwfncar: Update chemistry/Makefile in PR to only build and link musica if MPAS_USE_MUSICA is defined.

dwfncar: `make -j8 gnu CORE=atmosphere MUSICA=true AUTOCLEAN=true`

`musica.build.log 
MPAS was built without OpenMP-offload GPU support.
MPAS was built without OpenACC accelerator support.
MPAS was linked with the MUSICA-Fortran library version 0.10.1.
Position-dependent code was generated.
The OK was cleaned and re-compiled.
MPAS was built with .F files.
The native timer interface is being used
Using the SMIOL library.`

dwfncar: all:
	./../tools/manage_externals/checkout_externals --externals ./../Externals.cfg
ifeq ($(BUILD_MUSICA),yes)
	$(MAKE) core_mpas_musica core_chemistry
else
	$(MAKE) core_chemistry
endif

dwfncar: core_chemistry: $(if $(filter yes,$(BUILD_MUSICA)),core_mpas_musica)
	($(MAKE) chem_interface COREDEF="$(COREDEF)")
	ar -ru libchem.a $(OBJS)
ifeq ($(BUILD_MUSICA),yes)
	($(MAKE) -C ./musica mpas_musica_lib)
endif

dwfncar: Updated chemistry/Makefile with 
`ifeq ($(MUSICA),true)
core_mpas_musica:
        (cd musica; $(MAKE) all COREDEF="$(COREDEF)")
endif`
etc ...

dwfncar: Test with

`make -j4 gnu CORE=atmosphere AUTOCLEAN=true > build.no_musica.log 2>&1`

MPAS was built with default single-precision reals.
Debugging is off.
Parallel version is on.
Using the mpi_f08 module.
Papi libraries are off.
TAU Hooks are off.
MPAS was built without OpenMP support.
MPAS was built without OpenMP-offload GPU support.
MPAS was built without OpenACC accelerator support.
MPAS was not linked with the MUSICA-Fortran library.
Position-dependent code was generated.
The OK was cleaned and re-compiled.
MPAS was built with .F files.
The native timer interface is being used
Using the PIO 2.x library.

dwfncar: Test with

`make -j4 gnu CORE=atmosphere MUSICA=true AUTOCLEAN=true > build.musica.log 2>&1`

MPAS was built with default single-precision reals.
Debugging is off.
Parallel version is on.
Using the mpi_f08 module.
Papi libraries are off.
TAU Hooks are off.
MPAS was built without OpenMP support.
MPAS was built without OpenMP-offload GPU support.
MPAS was built without OpenACC accelerator support.
MPAS was linked with the MUSICA-Fortran library version 0.12.2.
Position-dependent code was generated.
The OK was cleaned and re-compiled.
MPAS was built with .F files.
The native timer interface is being used
Using the PIO 2.x library.

src/core_atmosphere/chemistry/musica/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	@# This removes them during the clean process
+	$(RM) *.i
+
+.F.o:
+ifeq "$(GEN_F90)" "true"
+	$(CPP) $(CPPFLAGS) $(COREDEF) $(CPPINCLUDES) $< > $*.f90
+	$(FC) $(FFLAGS) -c $*.f90 $(FCINCLUDES) -I.. -I../../../framework -I../../../external/esmf_time_f90
+else
+	$(FC) $(CPPFLAGS) $(COREDEF) $(FFLAGS) -c $*.F $(CPPINCLUDES) $(FCINCLUDES) -I.. -I../../../framework -I../../../external/esmf_time_f90
+endif
```
mgduda: It looks like this file is missing a final EOL character.

src/core_atmosphere/chemistry/musica/mpas_musica.F
```diff
@@ -0,0 +1,161 @@
...
+
+        use musica_micm, only : get_micm_version, RosenbrockStandardOrder
+        use musica_util, only : error_t, string_t
+
+        use mpas_log, only : mpas_log_write
+
+        integer,                       intent(in)    :: number_of_grid_cells
+        integer,                       intent(out)   :: error_code
+        character(len=:), allocatable, intent(out)   :: error_message
+        character(len=*),              intent(in)    :: filename_of_micm_configuration
```
mgduda: Just curious if there was a particular reason for not declaring the dummy arguments in the order in which they appear in the argument list?

src/core_atmosphere/chemistry/musica/mpas_musica.F
```diff
@@ -0,0 +1,161 @@
...
+
+        call mpas_log_write('MICM version: ' // micm_version%value_)
+
+        micm => micm_t(trim(filename_of_micm_configuration), solver_type, error)
+        if (has_error_occurred(error, error_message, error_code)) return
+
+        state => micm%get_state(number_of_grid_cells, error)
+        if (has_error_occurred(error, error_message, error_code)) return
+
+
```
mgduda: For consistency with other routines, we could consider just one blank line here.

src/core_atmosphere/chemistry/musica/mpas_musica.F
```diff
@@ -0,0 +1,161 @@
...
+        call mpas_log_write('Finalizing MUSICA chemistry package...')
+
+        ! Here we would typically clean up resources, but for now we do nothing.
+
+    end subroutine musica_finalize
+
+    !-------------------------------------------------------------------------
+    !  function has_error_occurred
+    !
+    !> @brief Evaluate a MUSICA error for failure and convert to error data
```
mgduda: Would it be valuable to use the same style of documentation block as for other routines in this module?

src/core_atmosphere/chemistry/Makefile
```diff
@@ -0,0 +1,52 @@
+.SUFFIXES: .F .o
+
+ifeq ($(CORE),atmosphere)
+COREDEF = -Dmpas
+endif
```
mgduda: I think we may be able to omit this `COREDEF` logic. If I recall correctly, this was done in physics to accommodate differences between WRF and MPAS-A for shared physics schemes. But here, I would guess that all of the code that we're writing can be MPAS-specific.

src/core_atmosphere/chemistry/Makefile
```diff
@@ -0,0 +1,52 @@
...
+	$(RM) *.o *.mod *.f90 libchem.a
+	( cd musica; $(MAKE) clean )
+	@# Certain systems with intel compilers generate *.i files
+	@# This removes them during the clean process
+	$(RM) *.i
+
+.F.o:
+	$(RM) $@ $*.mod
+ifeq "$(GEN_F90)" "true"
+	$(CPP) $(CPPFLAGS) $(COREDEF) $(HYDROSTATIC) $(CPPINCLUDES) $< > $*.f90
```
mgduda: I don't think the `HYDROSTATIC` variable is necessary for chemistry.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
```diff
@@ -0,0 +1,121 @@
...
+    !  routine mpas_atm_chemistry_init
+    !
+    !> \brief Initializes the chemistry packages
+    !> \author CheMPAS-A Developers
+    !> \date   20 August 2025
+    !> \details
+    !>  This routine initializes the chemistry packages, setting up
+    !>  necessary parameters and data structures for the simulation.
+    !------------------------------------------------------------------------
+    ! subroutine chemistry_init(configs, number_of_grid_cells)
```
mgduda: I would be inclined to remove the commented-out code related to `configs` for now.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
```diff
@@ -0,0 +1,121 @@
...
+#ifdef MPAS_USE_MUSICA
+        call mpas_log_write('Stepping chemistry packages...')
+        ! call musica_step()
+#endif
+
+    end subroutine chemistry_step
+
+
+    !------------------------------------------------------------------------
+    !  routine mpas_atm_chemistry_finalize
```
mgduda: It looks like the routine is named `chemistry_finalize`.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
```diff
@@ -0,0 +1,121 @@
...
+        call musica_init(filepath, number_of_grid_cells, error_code, error_message)
+
+        ! TODO check error_code and generate MPAS error log message
+#endif
+
+    end subroutine chemistry_init
+
+
+    !------------------------------------------------------------------------
+    !  routine mpas_atm_chemistry_step
```
mgduda: The routine is actually named `chemistry_step`.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
```diff
@@ -0,0 +1,121 @@
...
+    implicit none
+
+    private
+
+    public :: chemistry_init, chemistry_step, chemistry_finalize
+
+    contains
+
+    !------------------------------------------------------------------------
+    !  routine mpas_atm_chemistry_init
```
mgduda: The actual routine name is `chemistry_init`.

src/core_atmosphere/Makefile
```diff
@@ -47,17 +53,21 @@ physcore: mpas_atm_dimensions.o
...
+chemcore: mpas_atm_dimensions.o
+	( cd chemistry; $(MAKE) all )
+	( mkdir libchem; cd libchem; ar -x ../chemistry/libchem.a )
+
+dycore: mpas_atm_dimensions.o $(PHYSCORE) $(CHEMCORE)
+	( cd dynamics; $(MAKE) all PHYSICS="$(PHYSICS)"; $(MAKE) all CHEMISTRY="$(CHEMISTRY)" )
 
-diagcore: $(PHYSCORE) dycore
-	( cd diagnostics; $(MAKE) all PHYSICS="$(PHYSICS)" )
+diagcore: $(PHYSCORE) $(CHEMCORE) dycore
```
mgduda: Why do we need to pass the definition of `CHEMCORE` to diagnostics? Similarly for utilities and atmcore.

src/core_atmosphere/Makefile
```diff
@@ -47,17 +53,21 @@ physcore: mpas_atm_dimensions.o
...
+	( cd diagnostics; $(MAKE) all PHYSICS="$(PHYSICS)"; $(MAKE) all CHEMISTRY="$(CHEMISTRY)" )
 
-utilities: $(PHYSCORE)
-	( cd utils; $(MAKE) all PHYSICS="$(PHYSICS)" )
+utilities: $(PHYSCORE) $(CHEMCORE)
+	( cd utils; $(MAKE) all PHYSICS="$(PHYSICS)"; $(MAKE) all CHEMISTRY="$(CHEMISTRY)" )
```
mgduda: Admittedly I haven't actually tried compiling yet, but do we really want to build the `all` target in the `utils` directory twice? Similarly for the diagnostics.

src/core_atmosphere/mpas_atm_core.F
```diff
@@ -295,6 +296,12 @@ function atm_core_init(domain, startTimeStamp) result(ierr)
       !
       call mpas_atm_dynamics_init(domain)
 
+      !
+      ! Initialize the chemistry package
+      ! Assume for now that the chemistry solver will work on columns
+      !
+      call chemistry_init(nVertLevels)
```
mgduda: Rather than passing a specific (vertical) dimension here for the initialization of all chemistry, I think it would be better to pass some higher-level data structures, perhaps the `configs` and `dimensions` pools (which can be given here as `domain % blocklist % configs` and `domain % blocklist % dimensions`). Then, within the `chemistry_init` routine, the musica initialization routine can get just `nVertLevels` if it chooses, while other chem/aerosol packages like GOCART-2G can obtain namelist options and/or dimensions.

mgduda (changes requested):
I've left just a few initial comments, and I'll take a closer look and probably offer a second round of comments.

mgduda:
On Derecho with the `gcc/12.4.0` module and the `musica/0.10.1` module from `/glade/work/mattdawson/modulefiles` I'm getting an ICE when compiling `mpas_musica.F`:
```
f951: internal compiler error: in matching_typebound_op, at fortran/interface.cc:4479
```
Which modules / compilers are you testing with on Derecho, @dwfncar @K20shores @mattldawson ?

dwfncar:
> On Derecho with the `gcc/12.4.0` module and the `musica/0.10.1` module from `/glade/work/mattdawson/modulefiles` I'm getting an ICE when compiling `mpas_musica.F`:
> 
> ```
> f951: internal compiler error: in matching_typebound_op, at fortran/interface.cc:4479
> ```
> 
> Which modules / compilers are you testing with on Derecho, @dwfncar @K20shores @mattldawson ?

I setup my environment with the following commands:

module --force purge
module load ncarenv/24.12
module load craype/2.7.31
module load gcc/12.4.0
module load ncarcompilers/1.0.0
module load cray-mpich/8.1.29
module load parallel-netcdf/1.14.0
module load netcdf/4.9.2
source /glade/campaign/mmm/wmr/mpas_tutorial/mpas_env/bin/activate
module load ncview
export PATH=/glade/campaign/mmm/wmr/mpas_tutorial/metis/bin:${PATH}
module use /glade/work/mattdawson/modulefiles
module load musica

Looks like same gcc and musica versions.
Let me try a clean build on derecho ...

dwfncar:
I replicated the compile error on derecho.
It is due to line 70 in mpas_musica.F
`micm_version = get_micm_version()`
Introduced in this commit from yesterday
https://github.com/MPAS-Dev/MPAS-Model/commit/2c001d868931a4f140483c1cdad23fdcc6327131

Prior to the commit I did successfully build and run with that call in place on my Ubuntu desktop ...
assumed would be OK on derecho ...

dwfncar:
The simple Ubuntu build environment was just setup via the conda package manager:
```
conda create --name mpas
conda activate mpas
conda install -c conda-forge gcc gxx gfortran mpich libnetcdf libpnetcdf parallelio
export NETCDF=$HOME/miniconda3/envs/mpas
(same for PNETCDF and PIO),
```
and building both MUSICA and MPAS with that compiler set (gcc v15).

dwfncar:
Looking back at the hackathon version of mpas_musica.F,
I see a `micm_t` object was instantiated and `micm%get_state` method was called,
but not `get_version`, ...

The MUSICA GitHub actions do test builds with gcc/gfortran v12 - v13 and the unit tests exercise all Fortran interface functions. I think I need to do a MUSICA build and check the unit tests on derecho.

dwfncar:
MUSICA MICM unit tests with our derecho build environment are all OK.

`(mpas_env) fillmore@derecho7:~/MUSICA/build> make test`

Running tests...
Test project /glade/u/home/fillmore/MUSICA/build
    Start 1: component_versions
1/9 Test #1: component_versions ...............   Passed    0.02 sec
    Start 2: util
2/9 Test #2: util .............................   Passed    0.01 sec
    Start 3: micm_wrapper
3/9 Test #3: micm_wrapper .....................   Passed    0.01 sec
    Start 4: micm_c_api
4/9 Test #4: micm_c_api .......................   Passed    0.02 sec
    Start 5: parser
5/9 Test #5: parser ...........................   Passed    0.15 sec
    Start 6: micm_fortran_api
6/9 Test #6: micm_fortran_api .................   Passed    0.01 sec
    Start 7: get_micm_version
7/9 Test #7: get_micm_version .................   Passed    0.01 sec
    Start 8: micm_box_model
8/9 Test #8: micm_box_model ...................   Passed    0.01 sec
    Start 9: demo_fortran
9/9 Test #9: demo_fortran .....................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 9

dwfncar:
Compiler issue on derecho persists with custom MUSICA build.

dwfncar:
ccmake for MUSICA on derecho:
Got the same BLAS not found error with default settings.
But config OK if

```
MUSICA_ENABLE_CARMA OFF
MUSICA_ENABLE_TUVX OFF
```
also will want
`MUSICA_BUILD_FORTRAN_INTERFACE ON`
And set
`CMAKE_INSTALL_PREFIX <my_install_path>`

Usually from the MUSICA root dir I do:
```
mkdir build
cd build
ccmake ..
make
make test
make install

```
Also need
`export PKG_CONFIG_PATH=<my_install_path>/lib64/pkgconfig`

Also in general a good idea to
`conda deactivate`
(including the base environment)
before cmake configuration.

mgduda:
@dwfncar With the following at the beginning of the `mpas_musica` module
```
    use musica_micm,  only: micm_t, get_micm_version
```
and modifying the `use` of `musica_micm` in the `musica_init` routine to
```
        use musica_micm, only : RosenbrockStandardOrder
```
, the ICE disappears for me. It's not ideal, but perhaps we can use a work-around like this so that we can build CheMPAS-A on Derecho with the GNU 12.4.0 compilers? In the commit message, and even with a code comment if we'd like, we can explain why we're importing `get_micm_version` at the module level rather than within the `musica_init` routine.

dwfncar:
> @dwfncar With the following at the beginning of the `mpas_musica` module
> 
> ```
>     use musica_micm,  only: micm_t, get_micm_version
> ```
> 
> and modifying the `use` of `musica_micm` in the `musica_init` routine to
> 
> ```
>         use musica_micm, only : RosenbrockStandardOrder
> ```
> 
> , the ICE disappears for me. It's not ideal, but perhaps we can use a work-around like this so that we can build CheMPAS-A on Derecho with the GNU 12.4.0 compilers? In the commit message, and even with a code comment if we'd like, we can explain why we're importing `get_micm_version` at the module level rather than within the `musica_init` routine.

I'll go ahead and make that update with a comment and commit the changes ...

src/core_atmosphere/chemistry/musica/mpas_musica.F
dwfncar: Reordered argument declarations for consistency.

src/core_atmosphere/chemistry/musica/Makefile
dwfncar: Added final EOL to Makefile.

src/core_atmosphere/chemistry/musica/mpas_musica.F
dwfncar: Improved formatting.

src/core_atmosphere/chemistry/musica/mpas_musica.F
dwfncar: Reformatted doc comments for consistency.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
dwfncar: Modified doc comments for consistency.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
dwfncar: Modified doc comments for consistency.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
dwfncar: Modified doc comments for consistency.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
dwfncar: Removed commented chemistry_init call.

src/core_atmosphere/chemistry/Makefile
dwfncar: Removed COREDEF logic.

src/core_atmosphere/chemistry/Makefile
dwfncar: Removed HYDROSTATIC variable.

src/core_atmosphere/Makefile
dwfncar: Removed redundant MAKE all in dynamics, diagnostics, utils.

src/core_atmosphere/Makefile
dwfncar: Removed CHEMCORE definition from diagcore, utilities, and atmcore dependencies.

src/core_atmosphere/mpas_atm_core.F
dwfncar: In mpas_atm_core.F, changed to `chemistry_init(domain % blocklist % configs, domain % blocklist % dimensions)`.

K20shores: Good idea, @mgduda.

This makes me think, is this init called per MPI rank? Or once for the entire model?

dwfncar: It appears once for the entire model.
It looks like `domain % blocklist` is initialized by `mpas_atm_threading_init`.
My understanding is that each MPI rank can be assigned multiple blocks (`config_number_of_blocks`).
```
atm_core_init(domain, ...)
    ...
    mpas_atm_threading_init(domain % blocklist, ...)
    ...
    chemistry_init(domain % blocklist % configs, domain % blocklist % dimensions)
```

dwfncar: In `atm_core_init` there are various sections that iterate over blocks:

```
block => domain % blocklist
do while (associated(block))
    ...
    atm_mpas_init_block(..., block, ...)
    ...
    block => block % next
end do
```

mgduda: The `chemistry_init` routine will be called exactly once by each MPI task.

Regarding multiple blocks, long ago we had the idea of allowing each MPI task to be assigned multiple blocks, but in practice I don't think we've found any significant benefits to doing so. Now, there are definitely parts of MPAS-A that would not work correctly except with a single block per MPI task, and so I don't think we need to worry about looping over blocks for newer code.

src/core_atmosphere/chemistry/Makefile
```diff
@@ -0,0 +1,48 @@
+.SUFFIXES: .F .o
+
+all:
+	./../tools/manage_externals/checkout_externals --externals ./../Externals.cfg
```
mgduda: Right now, the physics build invokes `checkout_externals` to pull in a couple of external physics repositories, but since we aren't using any chemistry from external repositories (that need to be compiled as part of the MPAS-A build), I think it would be good to remove this line. In future, if we need to pull chemistry code in from elsewhere, we can rework the Makefiles so that this happens once for physics and chemistry at the main `core_atmosphere` directory.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
```diff
@@ -0,0 +1,127 @@
...
+#ifdef MPAS_USE_MUSICA
+        ! MUSICA will get the MICM JSON config from a namelist
+        ! hardcode filepath for now
+        character(len=StrKIND) :: filepath = 'chapman.json'
+#endif
+
+        call mpas_log_write('Initializing chemistry packages...')
+
+        call mpas_pool_get_dimension(dimensions, 'nVertLevels', nVertLevels_ptr)
+        nVertLevels = nVertLevels_ptr
```
mgduda, line 67: The above two lines for retrieving `nVertLevels` are only needed by MUSICA, so we could consider moving the `#ifdef MPAS_USE_MUSICA` directive below so that it is above these two lines.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
```diff
@@ -0,0 +1,127 @@
...
+        use mpas_musica, only: musica_init
+#endif
+        use mpas_log, only : mpas_log_write
+        use mpas_derived_types, only: mpas_pool_type
+        use mpas_kind_types, only: StrKIND
+        use mpas_pool_routines, only: mpas_pool_get_config, mpas_pool_get_dimension 
+
+        type (mpas_pool_type), intent(in) :: configs
+        type (mpas_pool_type), intent(in) :: dimensions
+        integer                           :: error_code
```
mgduda: The local variables `error_code`, `error_message`, `nVertLevels`, and `nVertLevels_ptr` are only use by MUSICA, so we could consider enclosing them in the same `#ifdef MPAS_USE_MUSICA` directive that is used for the `filepath` variable below.

src/core_atmosphere/Makefile
```diff
@@ -47,17 +53,21 @@ physcore: mpas_atm_dimensions.o
 	( cd ../..; ln -sf ./src/core_atmosphere/physics/physics_wrf/files/*DATA* .)
 	( cd ../..; ln -sf ./src/core_atmosphere/physics/physics_noahmp/parameters/*TBL .)
 
-dycore: mpas_atm_dimensions.o $(PHYSCORE)
-	( cd dynamics; $(MAKE) all PHYSICS="$(PHYSICS)" )
+chemcore: mpas_atm_dimensions.o
```
mgduda: It doesn't look like anything in chemistry currently depends on `mpas_atm_dimensions.o`, so we could remove this from the list of dependencies.

src/core_atmosphere/Makefile
```diff
@@ -47,17 +53,21 @@ physcore: mpas_atm_dimensions.o
...
+	( cd chemistry; $(MAKE) all )
+	( mkdir libchem; cd libchem; ar -x ../chemistry/libchem.a )
+
+dycore: mpas_atm_dimensions.o $(PHYSCORE) $(CHEMCORE)
+	( cd dynamics; $(MAKE) all PHYSICS="$(PHYSICS)" CHEMISTRY="$(CHEMISTRY)" )
```
mgduda: Since nothing in the dynamics currently depends on whether we're building with chemistry, it might help to keep the `Makefile` simpler for now to omit `CHEMISTRY="$(CHEMISTRY)"` when building dynamics. Similarly for diagnostics and utilities, below.

src/core_atmosphere/mpas_atm_core.F
```diff
@@ -48,6 +48,7 @@ function atm_core_init(domain, startTimeStamp) result(ierr)
       use mpas_attlist, only : mpas_modify_att
       use mpas_string_utils, only : mpas_string_replace
       use mpas_atm_halos, only: atm_build_halo_groups, exchange_halo_group
+      use mpas_atm_chemistry, only: chemistry_init
```
mgduda, line 39: If the definition of `CHEMISTRY` on line 11 of `src/core_atmosphere/Makefile` is commented out, this use statement (and other code in this module that assumes an `mpas_atm_chemistry` module will trigger a compilation failure. We could surround this and other chemistry code in the `mpas_atm_core.F` file with `#ifdef DO_CHEMISTRY` macros to avoid this issue.

mgduda (changes requested):

src/core_atmosphere/chemistry/Makefile
dwfncar: Removed checkout_externals from chemistry/Makefile.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
dwfncar: Moved get nVertLevels inside MPAS_USE_MUSICA.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
dwfncar: Moved MUSICA local variables inside MPAS_USE_MUSICA.

src/core_atmosphere/Makefile
dwfncar: Removed  mpas_atm_dimensions.o from chemcore dependencies.

src/core_atmosphere/Makefile
dwfncar: Removed CHEMISTRY=... from dycore, diagnostics, utilities.

src/core_atmosphere/mpas_atm_core.F
dwfncar: In mpas_atm_core.F, embedded chemistry use and calls inside of `#ifdef DO_CHEMISTRY` blocks.  Also updated core_atmosphere/Makefile with `$(CHEMISTRY)` added to the chemcore target and the `$(FC)` lines.

mgduda (approved):
This looks good to me -- nice work!

Since we don't use "squash and merge" in this repository, it might be nice to rework the commit history so that it contains a small number of commits (or even a single commit) with detailed commit messages before we merge this PR.

mgduda:
@dwfncar Since there's a lot that's being done in commit dcd68493, I think it would help to describe at a slightly higher level (in paragraphs) what's being done, the practical impact of the changes, and what the changes mean for the user. I can help with this if you'd like to create a draft commit message in a document somewhere.

Regarding format for commit messages, we've tended to use the following:
- One-line summary, 80 characters or less, describing as clearly as possible what is a commit does
- Blank line
- One or more paragraphs providing further details about the code changes in the commit, wrapping lines at 80 characters.

Here are a few recent commit messages to help give an idea of the level of detail that we aim for: a55b1c5b 9b9e5a5c 91bb07bd (the first of these is a merge commit, and so the one-line summary has a different format).

dwfncar:
I've updated the commit message via:
`git commit --amend -F changes.log`
`git push --force-with-lease`
Contents of changes.log:

    Chemistry hooks integrated with MUSICA stubs and guarded build flags.
    
    The chemistry infrastructure now lives in a dedicated
    `core_atmosphere/chemistry` hierarchy, which includes a `musica`
    subdirectory and its Makefiles. Within this structure we added the MUSICA
    MICM-facing modules `mpas_atm_chemistry.F` and `musica/mpas_musica.F`,
    providing the stubs that expose the chemistry hooks needed by the
    atmospheric core.
    
    `mpas_atm_chemistry.F` now establishes the MICM configuration path,
    performs initialization at the end of `atm_core_init`, advances the
    chemistry tendencies through `chemistry_step` inside `atm_do_timestep`,
    and finalizes the subsystem during shutdown. These calls are wrapped in
    `#ifdef DO_CHEMISTRY` and coordinated with the new `#ifdef
    MPAS_USE_MUSICA` guards so that chemistry support is compiled and
    executed only when explicitly enabled. The interface routines also
    accept the MPAS pool, configuration, and dimension objects, including
    forwarding `nVertLevels` to `musica_init`.
    
    Inside `musica/mpas_musica.F`, the initialization sequence now queries
    the MICM version via `get_micm_version`, records both the version and the
    number of grid cells in the log, and honors the compile-time flag to keep
    MUSICA linkage optional. Matching updates to the atmospheric and
    chemistry Makefiles add the new sources, apply the `CHEMISTRY` flags to
    the compiler command line, and ensure the MUSICA library is only built
    and linked when `MPAS_USE_MUSICA` is defined.

mgduda:
@dwfncar It looks like you've pushed a new merge commit that may have brought in the old commits prior to squashing them into a single commit (dcd68493). GitHub is showing 40 commits in this PR. I think the git author and commit name may have also reverted to your personal email address?

I've reworked the commit history so that there's just a single commit with your latest commit message. Can you take a look at the [musica_micm_stubs](https://github.com/mgduda/MPAS-Model/commits/musica_micm_stubs/) branch in my fork and see whether that commit history (really, just the last commit) looks fine to you? If so, could you force-push that branch to your fork? Or I could do that for you, assuming I have permissions.

mgduda:
@amstokely @jim-p-w Could you see whether MPAS-JEDI still builds after including the changes from this PR?

jim-p-w:
> @amstokely @jim-p-w Could you see whether MPAS-JEDI still builds after including the changes from this PR?

@mgduda This doesn't build with cmake. It needs to be rebased from the develop branch of MPAS-Dev/MPAS-Model in order to pick up `Merge branch 'atmosphere/cmake_framework' into develop (PR #1366)`

mgduda:
@jim-p-w Does MPAS-JEDI compile if you locally merge this PR branch onto the `develop` branch? Or is it not easy to test MPAS-JEDI with a local branch?

jim-p-w:
> @jim-p-w Does MPAS-JEDI compile if you locally merge this PR branch onto the `develop` branch? Or is it not easy to test MPAS-JEDI with a local branch?

@mgduda MPAS-Jedi build is currently broken. However, after rebasing this PR branch against the develop branch I was able to build MPAS-Model with cmake.

mgduda:
@jim-p-w Thanks for the additional test after merging this PR branch to `develop`. I think we can proceed to merge this PR, and whenever the issues with the MPAS-JEDI build have been resolved, we can address any remaining issues that turn up.

PR #1364, opened 2025-09-24
Online distributed graph partitioning using PT-Scotch

abishekg7:
This PR enables the use of MPAS atmosphere core without requiring any graph decomposition files specified as input, if the atmosphere core has been built with the Scotch graph partitioning library (built separately). 

This PR supersedes https://github.com/MPAS-Dev/MPAS-Model/pull/1348 to implement a Fortran interface to the PT-Scotch C library within MPAS. 

### Building MPAS with Scotch

Instructions to build Scotch are provided later in the description. The Scotch version must be at least v7.0.8. Once Scotch has been installed, the MPAS atmosphere core can be leveraged to use online graph partitioning by setting the path to Scotch prior to building MPAS.

```
export SCOTCH=/path/to/scotch/installation
make nvhpc CORE=atmosphere
```

If MPAS is built/linked with Scotch successfully, you should see a message stating that at the end of the build output. 

### Usage and run-time behavior

After building MPAS with Scotch, you can still choose whether or not to use online graph partitioning by setting appropriate values for the `config_block_decomp_file_prefix` namelist option.  

- If `config_block_decomp_file_prefix`+`mpi_tasks` points to a valid graph decomp file that already exists in the run directory, then it is used to proceed with the model run without invoking Scotch. 

- If `config_block_decomp_file_prefix==''` or `config_block_decomp_file_prefix`+`mpi_tasks` does not match any valid graph decomp file in the run directory, then Scotch graph partitioning is invoked. During this process, the generated partition is saved as a graph decomp file to the run directory so that it may be reused in the following runs without needing to invoke Scotch again. 

If MPAS has not been built with scotch, any code paths relating to Scotch will not be taken. And an incorrect specification of `config_block_decomp_file_prefix`+`mpi_tasks` should lead to the model halting.

### Downloading and building Scotch

Scotch may be obtained from its gitlab repository or as tarballs from this [link](https://gitlab.inria.fr/scotch/scotch/-/releases). The minimum Scotch version must be at least v7.0.8. 
```
git clone https://gitlab.inria.fr/scotch/scotch.git && cd scotch 
git checkout v7.0.8
```

Build instructions are on this [page](https://gitlab.inria.fr/scotch/scotch#installation)

It can be as simple as 
```
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=prefix/scotch/install -DMPI_HOME=/your/mpi/path  ..
make -j5
make install
```

To be able to use the distributed graph partitioning provided by PT-Scotch, you will need to pass an appropriate MPI installation path during the Scotch build. 

Other system pre-requisites are Bison and Flex. I have observed that an older Flex version (v2.6.1) on Eris caused the Scotch build to fail. After pulling in a more recent version of Flex [v2.6.4](https://github.com/westes/flex/releases) it seemed to build fine.  Scotch documentation also references the requirement for a Bison version > 3.4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -759,6 +759,15 @@ endif
 	LIBS += $(NCLIB)
 endif
 
+ifneq "$(SCOTCH)" ""
+	SCOTCH_INCLUDES += -I$(SCOTCH)/include
+	SCOTCH_LIBS += -L$(SCOTCH)/lib64 -lptscotch -lscotch  -lptscotcherr -lm
+	SCOTCH_FLAGS = -DMPAS_SCOTCH
+
+	CPPINCLUDES += $(SCOTCH_INCLUDES)
+	LIBS += $(SCOTCH_LIBS)
+	override CPPFLAGS += $(SCOTCH_FLAGS)
+endif
 
 ifneq "$(PNETCDF)" ""
 ifneq ($(wildcard $(PNETCDF)/lib/libpnetcdf.*), )
@@ -1415,6 +1424,37 @@ musica_fortran_test:
 	$(eval MUSICA_FORTRAN_VERSION := $(shell pkg-config --modversion musica-fortran))
 	$(if $(findstring 1,$(MUSICA_FORTRAN_TEST)), $(info Built a simple test program with MUSICA-Fortran version $(MUSICA_FORTRAN_VERSION)), )
 
+scotch_c_test:
+	@#
+	@# Create a C test program and try to build against the PT-SCOTCH library
+	@#
+	$(info Checking for a working Scotch library...)
+	$(eval SCOTCH_C_TEST := $(shell $\
+	    printf "#include <stdio.h>\n\
+			&#include \"mpi.h\"\n\
+			&#include \"ptscotch.h\"\n\
+			&int main(){\n\
+			&    int err;\n\
+			&    SCOTCH_Dgraph *dgraph;\n\
+			&    err = SCOTCH_dgraphInit(dgraph, MPI_COMM_WORLD);\n\
+			&    SCOTCH_dgraphExit(dgraph);\n\
+			&    return err;\n\
+			&}\n" | sed 's/&/ /' > ptscotch_c_test.c; $\
+		$\
+		$(CC) $(CPPINCLUDES) $(CFLAGS) $(LDFLAGS) ptscotch_c_test.c -o ptscotch_c_test.x $(SCOTCH_LIBS) > ptscotch_c_test.log 2>&1; $\
+		scotch_c_status=$$?; $\
+		if [ $$scotch_c_status -eq 0 ]; then $\
+			printf "1"; $\
+			rm -f ptscotch_c_test.c ptscotch_c_test.x ptscotch_c_test.log; $\
+		else $\
+			printf "0"; $\
+		fi $\
+	))
+	$(if $(findstring 0,$(SCOTCH_C_TEST)), $(error Could not build a simple C program with Scotch. $\
+		Test program ptscotch_c_test.c and output ptscotch_c_test.log have been left $\
+	    in the top-level MPAS directory for further debugging ))
+	$(if $(findstring 1,$(SCOTCH_C_TEST)), $(info Built a simple C program with Scotch ))
+
 pnetcdf_test:
 	@#
 	@# Create test C programs that look for PNetCDF header file and some symbols in it
@@ -1471,6 +1511,13 @@ else
 MUSICA_MESSAGE = "MPAS was not linked with the MUSICA-Fortran library."
 endif
 
+ifneq "$(SCOTCH)" ""
+MAIN_DEPS += scotch_c_test
+SCOTCH_MESSAGE = "MPAS has been linked with the Scotch graph partitioning library."
+else
+SCOTCH_MESSAGE = "MPAS was NOT linked with the Scotch graph partitioning library."
+endif
+
 mpas_main: $(MAIN_DEPS)
 	cd src; $(MAKE) FC="$(FC)" \
                  CC="$(CC)" \
@@ -1508,6 +1555,7 @@ mpas_main: $(MAIN_DEPS)
 	@echo $(OPENMP_OFFLOAD_MESSAGE)
 	@echo $(OPENACC_MESSAGE)
 	@echo $(MUSICA_MESSAGE)
+	@echo $(SCOTCH_MESSAGE)
 	@echo $(SHAREDLIB_MESSAGE)
 ifeq "$(AUTOCLEAN)" "true"
 	@echo $(AUTOCLEAN_MESSAGE)
--- a/src/framework/Makefile
+++ b/src/framework/Makefile
@@ -87,7 +87,7 @@ mpas_timekeeping.o: mpas_string_utils.o mpas_kind_types.o mpas_derived_types.o m
 
 mpas_timer.o: mpas_kind_types.o mpas_dmpar.o mpas_threading.o mpas_log.o
 
-mpas_block_decomp.o: mpas_derived_types.o mpas_hash.o mpas_io_units.o mpas_dmpar.o mpas_ptscotch_interface.o
+mpas_block_decomp.o: mpas_derived_types.o mpas_hash.o mpas_io_units.o mpas_dmpar.o mpas_timer.o mpas_ptscotch_interface.o
 
 mpas_block_creator.o: mpas_dmpar.o mpas_hash.o mpas_sort.o mpas_io_units.o mpas_block_decomp.o mpas_stream_manager.o mpas_decomp.o mpas_abort.o $(DEPS)
 
--- a/src/framework/Makefile
+++ b/src/framework/Makefile
@@ -36,6 +36,8 @@ OBJS = mpas_kind_types.o \
        mpas_log.o \
        mpas_halo.o \
        mpas_string_utils.o \
+       mpas_ptscotch_interface.o \
+       ptscotch_interface.o \
        mpas_stream_inquiry.o \
        stream_inquiry.o
 
@@ -111,6 +113,8 @@ xml_stream_parser.o: xml_stream_parser.c
 	$(CC) $(CFLAGS) $(CPPFLAGS) $(CPPINCLUDES) -I../external/ezxml -c xml_stream_parser.c
 
 mpas_halo.o: mpas_derived_types.o mpas_pool_routines.o mpas_log.o
+
+mpas_ptscotch_interface.o : mpas_derived_types.o mpas_log.o
 
 mpas_stream_inquiry.o : mpas_derived_types.o mpas_log.o mpas_c_interfacing.o
 
--- a/src/framework/Makefile
+++ b/src/framework/Makefile
@@ -87,7 +87,7 @@ mpas_timekeeping.o: mpas_string_utils.o mpas_kind_types.o mpas_derived_types.o m
 
 mpas_timer.o: mpas_kind_types.o mpas_dmpar.o mpas_threading.o mpas_log.o
 
-mpas_block_decomp.o: mpas_derived_types.o mpas_hash.o mpas_io_units.o mpas_dmpar.o
+mpas_block_decomp.o: mpas_derived_types.o mpas_hash.o mpas_io_units.o mpas_dmpar.o mpas_ptscotch_interface.o
 
 mpas_block_creator.o: mpas_dmpar.o mpas_hash.o mpas_sort.o mpas_io_units.o mpas_block_decomp.o mpas_stream_manager.o mpas_decomp.o mpas_abort.o $(DEPS)
 
--- a/src/framework/mpas_block_decomp.F
+++ b/src/framework/mpas_block_decomp.F
@@ -80,6 +80,7 @@ module mpas_block_decomp
       character (len=StrKIND) :: filename
 
       logical :: no_blocks
+      logical :: useScotch
 
       no_blocks = .false.
 
@@ -98,53 +99,85 @@ module mpas_block_decomp
         allocate(local_nvertices(dminfo % nprocs))
         allocate(global_start(dminfo % nprocs))
         allocate(global_list(partial_global_graph_info % nVerticesTotal))
+ 
+         if (dminfo % my_proc_id == IO_NODE) then
 
-        if (dminfo % my_proc_id == IO_NODE) then
+            if (dminfo % total_blocks < 10) then
+               write(filename,'(a,i1)') trim(blockFilePrefix), dminfo % total_blocks
+            else if (dminfo % total_blocks < 100) then
+               write(filename,'(a,i2)') trim(blockFilePrefix), dminfo % total_blocks
+            else if (dminfo % total_blocks < 1000) then
+               write(filename,'(a,i3)') trim(blockFilePrefix), dminfo % total_blocks
+            else if (dminfo % total_blocks < 10000) then
+               write(filename,'(a,i4)') trim(blockFilePrefix), dminfo % total_blocks
+            else if (dminfo % total_blocks < 100000) then
+               write(filename,'(a,i5)') trim(blockFilePrefix), dminfo % total_blocks
+            else if (dminfo % total_blocks < 1000000) then
+               write(filename,'(a,i6)') trim(blockFilePrefix), dminfo % total_blocks
+            else if (dminfo % total_blocks < 10000000) then
+               write(filename,'(a,i7)') trim(blockFilePrefix), dminfo % total_blocks
+            else if (dminfo % total_blocks < 100000000) then
+               write(filename,'(a,i8)') trim(blockFilePrefix), dminfo % total_blocks
+            end if
 
-           if (dminfo % total_blocks < 10) then
-              write(filename,'(a,i1)') trim(blockFilePrefix), dminfo % total_blocks
-           else if (dminfo % total_blocks < 100) then
-              write(filename,'(a,i2)') trim(blockFilePrefix), dminfo % total_blocks
-           else if (dminfo % total_blocks < 1000) then
-              write(filename,'(a,i3)') trim(blockFilePrefix), dminfo % total_blocks
-           else if (dminfo % total_blocks < 10000) then
-              write(filename,'(a,i4)') trim(blockFilePrefix), dminfo % total_blocks
-           else if (dminfo % total_blocks < 100000) then
-              write(filename,'(a,i5)') trim(blockFilePrefix), dminfo % total_blocks
-           else if (dminfo % total_blocks < 1000000) then
-              write(filename,'(a,i6)') trim(blockFilePrefix), dminfo % total_blocks
-           else if (dminfo % total_blocks < 10000000) then
-              write(filename,'(a,i7)') trim(blockFilePrefix), dminfo % total_blocks
-           else if (dminfo % total_blocks < 100000000) then
-              write(filename,'(a,i8)') trim(blockFilePrefix), dminfo % total_blocks
-           end if
+            call mpas_new_unit(iunit)
+            open(unit=iunit, file=trim(filename), form='formatted', status='old', iostat=istatus)
+            
+            if (istatus /= 0) then
+#ifdef MPAS_SCOTCH
+               useScotch = .true.
+#else
+               call mpas_log_write('Could not open block decomposition file for $i blocks.', MPAS_LOG_ERR, intArgs=(/dminfo % total_blocks/) )
+               call mpas_log_write('Filename: '//trim(filename), MPAS_LOG_CRIT)
+#endif
+            else
+               useScotch = .false.
+            end if
+         end if
 
-           call mpas_new_unit(iunit)
-           open(unit=iunit, file=trim(filename), form='formatted', status='old', iostat=istatus)
+#ifdef MPAS_SCOTCH
+         call mpas_dmpar_bcast_logical(dminfo, useScotch)
 
-           if (istatus /= 0) then
-              call mpas_log_write('Could not open block decomposition file for $i blocks.', MPAS_LOG_ERR, intArgs=(/dminfo % total_blocks/) )
-              call mpas_log_write('Filename: '//trim(filename), MPAS_LOG_CRIT)
-           end if
+         if (useScotch) then ! Using PT-Scotch across all MPI ranks
 
-           local_nvertices(:) = 0
-           do i=1,partial_global_graph_info % nVerticesTotal
-              read(unit=iunit, fmt=*, iostat=err) global_block_id
+            ! Pre-emptively blocking this untested code path. 
+            if (numBlocks /= 0) then
+               call mpas_log_write('Scotch partitioning not available when config_number_of_blocks != 0 ', MPAS_LOG_CRIT)
+            end if
 
-              if ( err .ne. 0 ) then
-                 call mpas_log_write('Decomoposition file: ' // trim(filename) // ' contains less than $i cells', &
-                                      MPAS_LOG_CRIT, intArgs=(/partial_global_graph_info % nVerticesTotal/) )
-              end if
-              call mpas_get_owning_proc(dminfo, global_block_id, owning_proc)
-              local_nvertices(owning_proc+1) = local_nvertices(owning_proc+1) + 1
-           end do
+            call mpas_log_write('No existing block decomposition file found, invoking Scotch.')
+            call mpas_block_decomp_scotch(dminfo, partial_global_graph_info, blockFilePrefix, filename)
 
-           read(unit=iunit, fmt=*, iostat=err)
+            if (dminfo % my_proc_id == IO_NODE) then
+               open(unit=iunit, file=trim(filename), form='formatted', status='old', iostat=istatus)
+               if (istatus /= 0) then
+                  call mpas_log_write('Could not open block decomposition file for $i blocks.', MPAS_LOG_ERR, intArgs=(/dminfo % total_blocks/) )
+                  call mpas_log_write('Filename: '//trim(filename), MPAS_LOG_CRIT)
+               end if
+            end if
+         end if
+#endif
 
-           if ( err == 0 ) then
-              call mpas_log_write('Decomposition file: ' // trim(filename) // ' contains more than $i cells', &
-                                   MPAS_LOG_CRIT, intArgs=(/partial_global_graph_info % nVerticesTotal/) )
-           end if
+         if (dminfo % my_proc_id == IO_NODE) then
+
+               local_nvertices(:) = 0
+               do i=1,partial_global_graph_info % nVerticesTotal
+                  read(unit=iunit, fmt=*, iostat=err) global_block_id
+
+                  if ( err .ne. 0 ) then
+                     call mpas_log_write('Decomoposition file: ' // trim(filename) // ' contains less than $i cells', &
+                                          MPAS_LOG_CRIT, intArgs=(/partial_global_graph_info % nVerticesTotal/) )
+                  end if
+                  call mpas_get_owning_proc(dminfo, global_block_id, owning_proc)
+                  local_nvertices(owning_proc+1) = local_nvertices(owning_proc+1) + 1
+               end do
+
+               read(unit=iunit, fmt=*, iostat=err)
+
+               if ( err == 0 ) then
+                  call mpas_log_write('Decomposition file: ' // trim(filename) // ' contains more than $i cells', &
+                                       MPAS_LOG_CRIT, intArgs=(/partial_global_graph_info % nVerticesTotal/) )    
+               end if
 
            global_start(1) = 1
            do i=2,dminfo % nprocs
--- a/src/framework/mpas_block_decomp.F
+++ b/src/framework/mpas_block_decomp.F
@@ -54,6 +54,8 @@ module mpas_block_decomp
    subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, local_cell_list, block_id, block_start, &
                                                block_count, numBlocks, explicitProcDecomp, blockFilePrefix, procFilePrefix)!{{{
 
+      use mpas_timer, only : mpas_timer_start, mpas_timer_stop
+
       implicit none
 
       type (dm_info), intent(inout) :: dminfo !< Input: domain information
@@ -83,6 +85,8 @@ module mpas_block_decomp
       logical :: useScotch
 
       no_blocks = .false.
+
+      call mpas_timer_start('mpas_block_decomp_cells_for_proc')
 
       if (numBlocks == 0) then
         dminfo % total_blocks = dminfo % nProcs
@@ -321,6 +325,8 @@ module mpas_block_decomp
         end if
       end if
 
+      call mpas_timer_stop('mpas_block_decomp_cells_for_proc')
+
    end subroutine mpas_block_decomp_cells_for_proc!}}}
 
 !***********************************************************************
@@ -341,6 +347,7 @@ module mpas_block_decomp
 #ifdef MPAS_SCOTCH
    subroutine mpas_block_decomp_scotch(dminfo, partial_global_graph_info, blockFilePrefix, blockFilename)!{{{
 
+      use mpas_timer, only : mpas_timer_start, mpas_timer_stop
 #ifdef MPAS_USE_MPI_F08
       use mpi_f08, only : MPI_Comm, MPI_INTEGER, MPI_Comm_dup, MPI_Comm_free, MPI_Gather, MPI_Gatherv
 #else
@@ -385,6 +392,8 @@ module mpas_block_decomp
       allocate(global_start(dminfo % nprocs))
       allocate(global_block_id_arr(partial_global_graph_info % nVerticesTotal))
       allocate(local_block_id_arr(partial_global_graph_info % nVertices))
+
+      call mpas_timer_start('scotch_total')
 
       ! Count the number of edges (including to ghost cells) in the portion of graph
       ! owned by the current rank. Each edge is counted twice, once for each vertex,
@@ -451,9 +460,12 @@ module mpas_block_decomp
       ! Initialize the strategy data structure
       call scotch_stratinit (stradat)
 
+      call mpas_timer_start('scotch_graph_partitioning')
       ! Partition the distributed graph and save the result in local_block_id_arr
       npart = dminfo % nProcs
       call scotch_dgraphpart (scotchdgraph, npart, stradat, local_block_id_arr)
+
+      call mpas_timer_stop('scotch_graph_partitioning')
 
       ! After the paritioning above, each processor would not necessarily have information about all of the
       ! vertices it owns. To obtain this information, Scotch provides a convenience function to redistribute the graph
@@ -532,6 +544,8 @@ module mpas_block_decomp
       deallocate(local_block_id_arr)
 
       call MPI_Comm_free(localcomm, mpi_ierr)
+
+      call mpas_timer_stop('scotch_total')
 
    end subroutine mpas_block_decomp_scotch
 #endif
--- a/src/framework/mpas_block_decomp.F
+++ b/src/framework/mpas_block_decomp.F
@@ -25,6 +25,9 @@ module mpas_block_decomp
    use mpas_derived_types
    use mpas_io_units
    use mpas_log
+#ifdef MPAS_SCOTCH
+   use mpas_ptscotch_interface
+#endif
 
    type graph
       integer :: nVerticesTotal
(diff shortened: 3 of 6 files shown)


src/core_atmosphere/Registry.xml
```diff
@@ -319,7 +319,7 @@
         </nml_record>
 
         <nml_record name="decomposition" in_defaults="true">
-                <nml_option name="config_block_decomp_file_prefix" type="character" default_value="x1.40962.graph.info.part."
+                <nml_option name="config_block_decomp_file_prefix" type="character" default_value=""
```
mgduda: At least initially, I would guess that most people would not be linking with the PT-Scotch library, in which case it might be best to leave a the placeholder prefix "x1.40962.graph.info.part." here as a remind to set the prefix of an existing graph partition file. Those who read the documentation and understand how the online partitioning works can simply delete the placeholder prefix.

Makefile
```diff
@@ -1415,6 +1425,33 @@ musica_fortran_test:
 	$(eval MUSICA_FORTRAN_VERSION := $(shell pkg-config --modversion musica-fortran))
 	$(if $(findstring 1,$(MUSICA_FORTRAN_TEST)), $(info Built a simple test program with MUSICA-Fortran version $(MUSICA_FORTRAN_VERSION)), )
 
+scotch_fortran_test:
+	@#
+	@# Create a Fortran test program that will link against the SCOTCH library
+	@#
+	$(info Checking for a working MUSICA-Fortran library...)
```
mgduda: I think "MUSICA-Fortran" should be "Scotch" (or "PT-Scotch").

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,304 @@
...
+module mpas_ptscotch_interface
+    use iso_c_binding, only : c_int, c_double
+#include "ptscotchf.h"
+    public :: scotch_dgraphinit, scotch_dgraphbuild
+
+    
+
+contains
+
+    function scotch_dgraphinit(dgraph, comm) result(ierr)
```
mgduda: Eventually, I think we'll want to have documentation blocks for each new function or subroutine.

src/framework/mpas_bootstrapping.F
```diff
@@ -430,6 +432,8 @@ subroutine mpas_bootstrap_framework_phase1(domain, mesh_filename, mesh_iotype, p
       deallocate(block_count)
       deallocate(readingBlock)
 
+      call mpas_timer_stop('bootstrap_framework_phase1')
```
mgduda: Would it make sense to omit these timers as part of this PR, and instead, if we need more detailed timing of the bootstrapping, we could make a separate PR to add timers throughout?

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,304 @@
...
+    function scotch_dgraphbuild(dgraph, nVertices, vertloctab, nLocEdgesGraph, edgelocsiz, adjncy) result(ierr)
+
+        use iso_c_binding, only : c_ptr, c_loc
+        use mpas_log, only : mpas_log_write
+        use mpas_derived_types, only : MPAS_LOG_CRIT
+
+        implicit none
+
+        doubleprecision, target, intent(in) :: dgraph (SCOTCH_DGRAPHDIM)
+        integer(SCOTCH_NUMSIZE), intent(in) :: nVertices
```
mgduda: I'm getting a compilation error related to `SCOTCH_NUMSIZE`:
```
mpas_ptscotch_interface.F(53): error #6683: A kind type parameter must be a compile-time constant.   [SCOTCH_NUMSIZE]
        integer(SCOTCH_NUMSIZE), intent(in) :: nVertices
----------------^
```
As far as I can tell, I've installed Scotch correctly (and I was able to use this library to test earlier versions of this PR). Is `SCOTCH_NUMSIZE` supposed to be defined in `ptscotchf.h`?

mgduda: Or is there an option that I need to activate when compiling Scotch?

abishekg7: It seems like `SCOTCH_NUMSIZE` may have been added to `ptscotchf.h` in v7.0.8. Prior to that it was only present in `scotchf.h`. 

I could just go ahead and remove the reference to `SCOTCH_NUMSIZE`  as it's not technically needed in our case. As we don't plan on supporting Scotch builds with 64-bit integers. 

Alternatively, we could enforce Scotch >= 7.0.8.

mgduda: I've updated to Scotch 7.0.10, and compilation is now successful!

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,412 @@
+#ifdef MPAS_SCOTCH
```
mgduda, line 7: It would be good to add a copyright statement at the beginning of this file.

src/framework/mpas_block_decomp.F
```diff
@@ -195,7 +232,7 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
 
            close(unit=iunit)
            call mpas_release_unit(iunit)
-
+
```
mgduda: Let's undo this whitespace addition.

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +322,239 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+         do j=1,partial_global_graph_info % nAdjacent(i)
+            if (partial_global_graph_info % adjacencyList(j,i) == 0) cycle
+            edgeloctab(k) = partial_global_graph_info % adjacencyList(j,i)
+            k = k + 1
+         end do
+      end do 
+      vertloctab(partial_global_graph_info % nVertices+1) = nLocEdgesGraph + 1
+
+      ! Duplicate the communicator to be used by Scotch
+      call MPI_Comm_dup(MPI_COMM_WORLD, localcomm, mpi_ierr)
```
mgduda: In general, we can't assume that MPAS owns the entire `MPI_COMM_WORLD` communicator (as might be the case, for example, when MPAS-Atmosphere is operating as a dynamical core in CAM within CESM).
Instead of `MPI_COMM_WORLD`, we can use the communicator found in `domain % dminfo`.

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +322,239 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+      end do
+
+      ! Gather all the local block ids to global_block_id_arr so IO_NODE can write out the partitioning data
+      call MPI_Gatherv( local_block_id_arr, partial_global_graph_info % nVertices, MPI_INTEGER, global_block_id_arr, &
+                              local_nvertices, global_start, MPI_INTEGER, 0, localcomm, ierr)
+      ! Write out the paritioning data to a file from IO_NODE
+      if (dminfo % my_proc_id == IO_NODE) then
+         partitionFilePrefix=trim(blockFilePrefix)
+         if (trim(partitionFilePrefix) == '') then
+            write(partitionFilePrefix,'(a,i0,a)') 'x1.',partial_global_graph_info%nVerticesTotal,'.graph.info.part.'
```
mgduda: I think we had decided not to use the `x1.` prefix here.

mgduda (changes requested):

src/framework/mpas_ptscotch_interface.F
abishekg7: Added some basic documentation blocks, but will double check the documentation.

src/framework/mpas_bootstrapping.F
abishekg7: Makes sense. I've removed these timers.

src/framework/mpas_block_decomp.F
abishekg7: Makes sense. Thanks for the catch.

src/framework/mpas_ptscotch_interface.F
abishekg7: Fixed

Makefile
abishekg7: Fixed

src/framework/mpas_block_decomp.F
abishekg7: Fixed

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,238 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+!>  local block ids (for each MPI rank) to the IO_NODE to write out to a partition file with the
+!>  specified prefix.
+!
+!-----------------------------------------------------------------------
+#ifdef MPAS_SCOTCH
+   subroutine mpas_block_decomp_scotch(dminfo, partial_global_graph_info, blockFilePrefix, blockFilename)!{{{
+
+      use mpas_timer, only : mpas_timer_start, mpas_timer_stop
+#ifdef MPAS_USE_MPI_F08
+      use mpi_f08, only : MPI_Comm, MPI_COMM_WORLD, MPI_INTEGER, MPI_Comm_dup, MPI_Comm_free, MPI_Gather, MPI_Gatherv
```
mgduda: It doesn't look like `MPI_COMM_WORLD` is needed in this `use` statement.

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,238 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+      type (graph), intent(in) :: partial_global_graph_info !< Input: Global graph information
+      character (len=*), intent(in) :: blockFilePrefix !< Input: File prefix for block decomposition
+      character (len=*), intent(out) :: blockFilename !< Output: Block decomposition file name
+
+      integer, dimension(:), pointer :: global_start
+      integer, dimension(:), allocatable :: local_cell_list
+      integer, dimension(:), allocatable :: local_block_list
+
+      integer, dimension(:), allocatable ::global_block_id_arr, local_block_id_arr
+      integer :: i, global_block_id, local_block_id, iunit, ounit, istatus, ostatus, j, k
```
mgduda: With the `-Wall` flag, the gfortran compiler notes some unused variables:
```
mpas_block_decomp.F:375:35:

  375 |       integer :: i, global_block_id, local_block_id, iunit, ounit, istatus, ostatus, j, k
Warning: Unused variable 'global_block_id' declared at (1) [-Wunused-variable]
mpas_block_decomp.F:375:74:

  375 |       integer :: i, global_block_id, local_block_id, iunit, ounit, istatus, ostatus, j, k
      |                                                                          1
Warning: Unused variable 'istatus' declared at (1) [-Wunused-variable]
mpas_block_decomp.F:375:58:

  375 |       integer :: i, global_block_id, local_block_id, iunit, ounit, istatus, ostatus, j, k
      |                                                          1
Warning: Unused variable 'iunit' declared at (1) [-Wunused-variable]
mpas_block_decomp.F:375:51:

  375 |       integer :: i, global_block_id, local_block_id, iunit, ounit, istatus, ostatus, j, k
      |                                                   1
Warning: Unused variable 'local_block_id' declared at (1) [-Wunused-variable]
mpas_block_decomp.F:371:59:

  371 |       integer, dimension(:), allocatable :: local_cell_list
      |                                                           1
Warning: Unused variable 'local_cell_list' declared at (1) [-Wunused-variable]
mpas_block_decomp.F:379:36:

  379 |       character (len=StrKIND) :: msg
      |                                    1
Warning: Unused variable 'msg' declared at (1) [-Wunused-variable]
mpas_block_decomp.F:385:54:

  385 |       doubleprecision :: scotchgraph (SCOTCH_GRAPHDIM)
      |                                                      1
Warning: Unused variable 'scotchgraph' declared at (1) [-Wunused-variable]
```

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,285 @@
...
+ *
+ * Returns:
+ *   nothing (wraps `SCOTCH_stratExit`).
+ *
+ ********************************************************************************/
+void scotchm_stratexit(void *strat_ptr)
+{
+	SCOTCH_stratExit((SCOTCH_Strat *)strat_ptr);
+}
+#endif
```
mgduda: It looks like this file is missing and EOF character.

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,418 @@
...
+                integer(c_int) :: cell_list(*)
+            end subroutine scotchfdgraphdata
+        end interface
+
+        call scotchfdgraphdata(c_loc(dgraph), local_cell_list)
+
+    end subroutine scotch_dgraphdata
+
+end module mpas_ptscotch_interface
+#endif
```
mgduda: Could you add an EOF character to this file?

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,418 @@
...
+        use mpas_log, only : mpas_log_write
+        use iso_c_binding, only : c_ptr, c_loc
+
+        implicit none
+
+        doubleprecision, target, intent(in) :: dgraph (SCOTCH_DGRAPHDIM)
+        integer(SCOTCH_NUMSIZE), intent(out) :: local_cell_list(*)
+
+        interface
+            subroutine scotchfdgraphdata(dgraph_ptr, cell_list) bind(C, name='scotchm_dgraphdata')
```
mgduda: Here and elsewhere in this file, would it make the code easier to follow if we declared these interface names to match the C names? Or is there a reason for using interface names that differ from the names of the C routines that we're calling?

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,238 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+            if (partial_global_graph_info % adjacencyList(j,i) == 0) cycle
+            edgeloctab(k) = partial_global_graph_info % adjacencyList(j,i)
+            k = k + 1
+         end do
+      end do 
+      vertloctab(partial_global_graph_info % nVertices+1) = nLocEdgesGraph + 1
+
+      ! Duplicate the communicator to be used by Scotch
+      call MPI_Comm_dup(dminfo % comm, localcomm, mpi_ierr)
+      if (mpi_ierr .ne. 0) then
```
mgduda: Just for consistency with other code in MPAS, could you change `.ne.` to `/=`?

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,238 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+            call mpas_log_write('Cannot duplicate communicator')
+      endif
+      ! Initialize the Scotch graph data structure, and an extra one to hold the re-distributed graph
+#ifdef MPAS_USE_MPI_F08
+      call scotch_dgraphinit(scotchdgraph(1), localcomm% mpi_val)
```
mgduda: Here and in other calls, it would be more in line with the rest of MPAS to pass array arguments simply using their names rather than by passing their first element.

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,418 @@
...
+    !
+    !-----------------------------------------------------------------------
+    subroutine scotch_dgraphinit(dgraph, comm)
+        use iso_c_binding, only : c_ptr, c_loc
+        use mpas_log, only : mpas_log_write
+        use mpas_derived_types, only : MPAS_LOG_CRIT
+
+        implicit none
+        ! Arguments
+        doubleprecision, target, intent(in) :: dgraph (SCOTCH_DGRAPHDIM)
```
mgduda: Stylistically, I think it would be preferable to declare arrays using the `dimension` attribute. E.g.,
```fortran
doubleprecision, dimension(SCOTCH_DGRAPHDIM), target, intent(in) :: dgraph
```

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,285 @@
...
+ *   cell_list  - output array to receive vertex labels for local vertices
+ *
+ * Returns:
+ *   integer error code (currently returns the local `err` variable; 0 on success).
+ *
+ ********************************************************************************/
+void scotchm_dgraphdata(void *ptr, SCOTCH_Num *cell_list)
+{
+	
+	int err;
```
mgduda: The `err` variable doesn't appear to be used.

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,285 @@
...
+	SCOTCH_Num edgelocnbr = nLocEdgesGraph;
+	SCOTCH_Num edgelocsiz = edgelocsiz_1;
+	SCOTCH_Num *edgegsttab = NULL; /* Optional array holding the local and ghost indices */
+	SCOTCH_Num *edloloctab = NULL; /* Optional array of integer loads for each local edge */
+
+	SCOTCH_Num *vertloctab = (SCOTCH_Num *)vertloctab_1;
+	SCOTCH_Num *vendloctab = vertloctab_1 + 1;
+	SCOTCH_Num *edgeloctab = (SCOTCH_Num *)adjncy;
+
+	int i, err;
```
mgduda: The `i` variable doesn't appear to be used anywhere in this function.

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,285 @@
...
+ *
+ * Returns:
+ *   integer error code returned by `SCOTCH_dgraphInit` (0 on success).
+ *
+ ********************************************************************************/
+int scotchm_dgraphinit(void *ptr, int localcomm)
+{
+	MPI_Comm comm;
+	MPI_Comm comm2;
+	int size, rank, err;
```
mgduda: With the `-Wall` flag, the gcc compiler reports several unused variables:
```
ptscotch_interface.c: In function 'scotchm_dgraphinit':
ptscotch_interface.c:37:19: warning: unused variable 'rank' [-Wunused-variable]
   37 |         int size, rank, err;
      |                   ^~~~
ptscotch_interface.c:37:13: warning: unused variable 'size' [-Wunused-variable]
   37 |         int size, rank, err;
      |             ^~~~
ptscotch_interface.c:36:18: warning: unused variable 'comm2' [-Wunused-variable]
   36 |         MPI_Comm comm2;
      |                  ^~~~~
```

mgduda (changes requested):

src/framework/mpas_block_decomp.F
abishekg7: Done

src/framework/mpas_block_decomp.F
abishekg7: Thanks for the catch. Fixed.

src/framework/ptscotch_interface.c
abishekg7: Done.

src/framework/mpas_ptscotch_interface.F
abishekg7: Done.

src/framework/mpas_block_decomp.F
abishekg7: Done.

src/framework/mpas_block_decomp.F
abishekg7: Done

src/framework/mpas_ptscotch_interface.F
abishekg7: Fixed.

src/framework/ptscotch_interface.c
abishekg7: Fixed.

src/framework/ptscotch_interface.c
abishekg7: Fixed.

src/framework/ptscotch_interface.c
abishekg7: Fixed.

src/framework/mpas_ptscotch_interface.F
abishekg7: I wasn't entirely sure if that was allowed. But I've fixed this now.

src/framework/Makefile
```diff
@@ -85,7 +87,7 @@ mpas_timekeeping.o: mpas_string_utils.o mpas_kind_types.o mpas_derived_types.o m
 
 mpas_timer.o: mpas_kind_types.o mpas_dmpar.o mpas_threading.o mpas_log.o
 
-mpas_block_decomp.o: mpas_derived_types.o mpas_hash.o mpas_io_units.o mpas_dmpar.o
+mpas_block_decomp.o: mpas_derived_types.o mpas_hash.o mpas_io_units.o mpas_dmpar.o mpas_ptscotch_interface.o
```
mgduda: I think `mpas_block_decomp.o` now depends on `mpas_timer.o` as well.

src/framework/Makefile
```diff
@@ -112,6 +114,8 @@ xml_stream_parser.o: xml_stream_parser.c
 
 mpas_halo.o: mpas_derived_types.o mpas_pool_routines.o mpas_log.o
 
+mpas_ptscotch_interface.o : mpas_derived_types.o mpas_dmpar.o mpas_log.o ptscotch_interface.o
```
mgduda: I doesn't look like the current version of the `mpas_ptsctoch_interface` depends on `mpas_dmpar`.

Also, although `mpas_ptscotch_interface` makes calls to routines that are implemented in `ptscotch_interface`, I think we can drop the dependency on `ptscotch_interface.o` since compilation of `ptscotch_interface` doesn't produce any artifacts (e.g., a `.mod` file) that are needed to compile `mpas_ptscotch_interface`.

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,240 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+            edgeloctab(k) = partial_global_graph_info % adjacencyList(j,i)
+            k = k + 1
+         end do
+      end do 
+      vertloctab(partial_global_graph_info % nVertices+1) = nLocEdgesGraph + 1
+
+      ! Duplicate the communicator to be used by Scotch
+      call MPI_Comm_dup(dminfo % comm, localcomm, mpi_ierr)
+      if (mpi_ierr /= 0) then
+            call mpas_log_write('Cannot duplicate communicator')
```
mgduda: Would it be better to make this a critical error? It seems like not being able to duplicate `dminfo % comm` as `localcomm` would break the rest of the code in this routine that depends on `localcomm`.

Also, it might be good to stick with the same indentation as elsewhere in this routine.

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,240 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+      ! num_local_vertices is the number of vertices that this processor owns, determined by the
+      ! Scotch paritioning. Whereas artial_global_graph_info % nVertices is the number of vertices
+      ! resident in the partial graph read by this processor. The latter is the correct size of the
+      ! local_block_id_arr. 
+      call MPI_Gather( partial_global_graph_info % nVertices, 1, MPI_INTEGER, local_nvertices, &
+                           1, MPI_INTEGER, 0, localcomm, ierr)
+      
+      ! Compute the displacements for gathering all the local_block_id_arr to global_block_id_arr
+      global_start(1) = 0
+         do i=2,dminfo % nprocs
```
mgduda: The indentation seems off here.

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,240 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+                              local_nvertices, global_start, MPI_INTEGER, 0, localcomm, ierr)
+      ! Write out the paritioning data to a file from IO_NODE
+      if (dminfo % my_proc_id == IO_NODE) then
+         partitionFilePrefix=trim(blockFilePrefix)
+         if (trim(partitionFilePrefix) == '') then
+            write(partitionFilePrefix,'(i0,a)') partial_global_graph_info%nVerticesTotal,'.graph.info.part.'
+         end if
+         write(blockFilename,'(a,i0)') trim(partitionFilePrefix), dminfo % nProcs
+         
+         call mpas_log_write('Writing out Scotch Graph paritioning data to '//trim(blockFilename))
```
mgduda: `paritioning` -> `partitioning`

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,465 @@
...
+! distributed with this code, or at https://mpas-dev.github.io/license.html .
+!
+#ifdef MPAS_SCOTCH
+module mpas_ptscotch_interface
+    use iso_c_binding, only : c_ptr
+    public :: scotch_dgraphinit, scotch_dgraphbuild
+
+    ! Dummy type declaration for SCOTCH distributed graph
+    ! Member ptr is a pointer to the SCOTCH_Dgraph C structure
+    type, bind(c) :: scotchm_dgraph
```
mgduda: I don't think this `type` needs to be `bind(c)` since instances of this type are never directly passed to C code; only the `type(c_ptr)` member is passed to C code as far as I can tell, and a `type(c_ptr)` is by definition interoperable with a C data pointer.

Similarly for the `scotchm_strat` derived type.

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,465 @@
...
+    !>  Logs an error message if there is a mismatch.
+    !-----------------------------------------------------------------------
+    subroutine scotch_checkintsize()
+        use iso_c_binding, only : c_int, c_size_t, c_sizeof
+        use mpas_log, only : mpas_log_write
+        use mpas_derived_types, only : MPAS_LOG_CRIT
+
+        implicit none
+        ! Arguments
+
```
mgduda, line 42: Was this blank line intended to go above the `! Arguments` comment line?

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,295 @@
...
+ * distributed with this code, or at http://mpas-dev.github.com/license.html
+ */
+#ifdef MPAS_SCOTCH
+#include <stddef.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <string.h>
+#include <stdio.h>
+#include <mpi.h>
```
mgduda: From the perspective of the C compiler, MPI is a non-system library just like PT-Scotch, so I think we've typically used quotes when including `mpi.h`; i.e.,
```c
#include "mpi.h"
```

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,295 @@
...
+ * Unless noted otherwise source code is licensed under the BSD license.
+ * Additional copyright and license information can be found in the LICENSE file
+ * distributed with this code, or at http://mpas-dev.github.com/license.html
+ */
+#ifdef MPAS_SCOTCH
+#include <stddef.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <string.h>
```
mgduda: You might want to double check, but it looks like the following headers don't need to be included:
- `stddef.h`
- `sys/stat.h`
- `unistd.h`
- `string.h`

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,295 @@
...
+int scotchm_dgraphinit(SCOTCH_Dgraph **dgraph_ptr, int localcomm)
+{
+	MPI_Comm comm;
+	int err;
+
+	comm = MPI_Comm_f2c((MPI_Fint)localcomm);
+
+	*dgraph_ptr = (SCOTCH_Dgraph *) malloc(sizeof (SCOTCH_Dgraph));
+
+	err = SCOTCH_dgraphInit(*dgraph_ptr, comm);
```
mgduda: Just a suggestion (and it looks like you've done this elsewhere), if you'd like to avoid having to declare the local variable `err`, you could directly return the result of the call to `SCOTCH_dgraphInit`.

src/framework/ptscotch_interface.c
```diff
@@ -0,0 +1,295 @@
...
+							 vertlocnbr,
+							 vertloctab,
+							 vendloctab,
+							 veloloctab,
+							 vlblloctab,
+							 edgelocnbr,
+							 edgelocsiz,
+							 edgeloctab,
+							 edgegsttab,
+							 edloloctab);
```
mgduda: Out of curiosity, was there a reason for using one line for each argument here? Other function calls in this file list multiple arguments on the same line.

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,240 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
+
+      implicit none
+
+      type (dm_info), intent(inout) :: dminfo !< Input: domain information
+      type (graph), intent(in) :: partial_global_graph_info !< Input: Global graph information
+      character (len=*), intent(in) :: blockFilePrefix !< Input: File prefix for block decomposition
+      character (len=*), intent(out) :: blockFilename !< Output: Block decomposition file name
+
+      integer, dimension(:), pointer :: global_start
+      integer, dimension(:), allocatable :: local_cell_list
```
mgduda: The `gfortran` compiler with `-Wall` reports `local_cell_list` as being declared but never used.

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,465 @@
...
+        integer(c_int) :: dummy_int
+        integer(c_size_t) :: dummy_size
```
mgduda: It doesn't look like `dummy_size` is used in this routine.

src/framework/mpas_ptscotch_interface.F
```diff
@@ -0,0 +1,465 @@
...
+        interface
+            function scotchm_get_intsize() bind(C, name='scotchm_get_intsize') result(intsize)
+                use iso_c_binding, only : c_size_t
+                integer(c_size_t) :: intsize
+            end function scotchm_get_intsize
+        end interface
+
+       if (scotchm_get_intsize() /= c_sizeof(dummy_int)) then
+          call mpas_log_write("Error: Scotch SCOTCH_Num size does not match MPAS integer size \n &
+                                Please build Scotch with 32 bit integers", MPAS_LOG_CRIT)
```
mgduda: The `gfortran` compiler warns about this continued character constant:
```
mpas_ptscotch_interface.F:56:33:

   56 |                                 Please build Scotch with 32 bit integers", MPAS_LOG_CRIT)
      |                                 1
Warning: Missing '&' in continued character constant at (1) [-Wampersand]
```

mgduda (changes requested):
I think we're getting really close, and I've just left a few more clean-up comments.

src/framework/ptscotch_interface.c
abishekg7: I might have originally wanted a comment after each argument, but it wasn't necessary. I can change this.

src/framework/ptscotch_interface.c
abishekg7: Got it. I was iterating through some designs, and perhaps it can be removed now.

src/framework/mpas_ptscotch_interface.F
abishekg7: Thanks for the catch!

src/framework/mpas_block_decomp.F
abishekg7: I can comment this declaration line out as `local_cell_list ` could possibly be needed at a later point if we decide to use the partition data directly from the redistributed graph, instead of writing out to a partition file and then reading from it.

src/framework/ptscotch_interface.c
abishekg7: Checked that they can be removed.

abishekg7:
@mgduda thanks for the comments! I've addressed most of them, except for the `local_cell_list` declaration.

src/framework/mpas_block_decomp.F
mgduda: In this case, it might improve readability to add a space between the `!` character and the beginning of the variable declaration.

src/framework/mpas_block_decomp.F
```diff
@@ -95,53 +103,86 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
-           if (istatus /= 0) then
+            call mpas_new_unit(iunit)
+            open(unit=iunit, file=trim(filename), form='formatted', status='old', iostat=istatus)
+            
+            if (istatus /= 0) then
+#ifdef MPAS_SCOTCH
+               useScotch = .true.
+#else
               call mpas_log_write('Could not open block decomposition file for $i blocks.', MPAS_LOG_ERR, intArgs=(/dminfo % total_blocks/) )
               call mpas_log_write('Filename: '//trim(filename), MPAS_LOG_CRIT)
```
mgduda: Could we add one space of indentation for these calls to `mpas_log_write` so their indentation matches the lines around them?

src/framework/mpas_block_decomp.F
```diff
@@ -285,8 +326,240 @@ subroutine mpas_block_decomp_cells_for_proc(dminfo, partial_global_graph_info, l
...
 
+      call mpas_log_write('mpas_block_decomp_cells_for_proc successful ')
+
+      call mpas_timer_stop('mpas_block_decomp_cells_for_proc')
+
    end subroutine mpas_block_decomp_cells_for_proc!}}}
 
+
+
+
```
mgduda: It looks like the rest of the routines in this module use just one blank line between routines. For consistency, should we do the same here?

Makefile
```diff
@@ -759,6 +759,16 @@ endif
...
+ifneq "$(SCOTCH)" ""
+	SCOTCH_FCINCLUDES += -I$(SCOTCH)/include
+	SCOTCH_LIBS += -L$(SCOTCH)/lib64 -lptscotch -lscotch  -lptscotcherr -lm
+	SCOTCH_FFLAGS = -DMPAS_SCOTCH
+
+	FCINCLUDES += $(SCOTCH_FCINCLUDES)
+	CPPINCLUDES += $(SCOTCH_FCINCLUDES)
+	LIBS += $(SCOTCH_LIBS)
+	override CPPFLAGS += $(SCOTCH_FFLAGS)
+endif
```
mgduda, line 770: Since it's only C code that needs to interface directly with the PT-Scotch library, we might be able to simplify this block as:
```make
ifneq "$(SCOTCH)" ""
        SCOTCH_INCLUDES += -I$(SCOTCH)/include
        SCOTCH_LIBS += -L$(SCOTCH)/lib64 -lptscotch -lscotch -lptscotcherr -lm
        SCOTCH_FLAGS = -DMPAS_SCOTCH

        CPPINCLUDES += $(SCOTCH_INCLUDES)
        LIBS += $(SCOTCH_LIBS)
        override CPPFLAGS += $(SCOTCH_FLAGS)
endif
```
(Note that the `FC` in `FCINCLUDES` was meant -- as far as I recall -- to refer to "Fortran Compiler"; similarly, `FFLAGS` was "Fortran Flags". So, the suggestion above removes `FC` and `F` prefixes).

src/framework/mpas_block_decomp.F
abishekg7: Done

abishekg7:
Addressed the remaining review comments, and following the C-indentation style found in `smiol.c`

mgduda:
There are currently quite a few messages that are written to the log files when PT-Scotch is used to create a new graph partitioning:
```
 No existing block decomposition file found, invoking Scotch.
 Scotch SCOTCH_Num size is compatible with MPAS integer size
 Successfully initialized distributed Scotch graph
 Successfully initialized distributed Scotch graph
 Before SCOTCH distributed graph build. nLocVertices =2729  nLocEdges= 16359
 Successfully built distributed Scotch graph
 Successfully checked distributed Scotch graph
 Successfully initialized Scotch strategy
 Successfully partitioned distributed Scotch graph
 Successfully redistributed Scotch graph
 Writing out Scotch Graph partitioning data to colorado.graph.info.part.256
 Scotch partition successful
 After Scotch decomposition, attempting to read block decomposition file: colorado.graph.info.part.256
 mpas_block_decomp_cells_for_proc successful
```
I suspect some of these were primarily added to aid in development, and now that the functionality is in good shape, they could be removed.

@abishekg7 Could you go through and remove some of these log messages?

mgduda:
Near the beginning of the log file, we write out a summary of compile-time options:
```
 Compile-time options:
   Build target: gnu
   OpenMP support: no
   OpenACC support: no
   Default real precision: single
   Compiler flags: optimize
   I/O layer: SMIOL
   MUSICA support: no
```
As we've done for the MUSICA library, I think it would be helpful to add a similar message for the PT-Scotch library. This could be done in the [`mpas_framework_report_settings`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.3.1/src/framework/mpas_framework.F#L190-L201) routine in `mpas_framework.F`,

abishekg7:
> There are currently quite a few messages that are written to the log files when PT-Scotch is used to create a new graph partitioning:
> 
> ```
>  No existing block decomposition file found, invoking Scotch.
>  Scotch SCOTCH_Num size is compatible with MPAS integer size
>  Successfully initialized distributed Scotch graph
>  Successfully initialized distributed Scotch graph
>  Before SCOTCH distributed graph build. nLocVertices =2729  nLocEdges= 16359
>  Successfully built distributed Scotch graph
>  Successfully checked distributed Scotch graph
>  Successfully initialized Scotch strategy
>  Successfully partitioned distributed Scotch graph
>  Successfully redistributed Scotch graph
>  Writing out Scotch Graph partitioning data to colorado.graph.info.part.256
>  Scotch partition successful
>  After Scotch decomposition, attempting to read block decomposition file: colorado.graph.info.part.256
>  mpas_block_decomp_cells_for_proc successful
> ```
> 
> I suspect some of these were primarily added to aid in development, and now that the functionality is in good shape, they could be removed.
> 
> @abishekg7 Could you go through and remove some of these log messages?

I can retain these messages for the successful case, and remove the rest. 

```
No existing block decomposition file found, invoking Scotch.
Successfully built distributed Scotch graph
Successfully partitioned distributed Scotch graph
Writing out Scotch Graph partitioning data to colorado.graph.info.part.256
```

mgduda (approved):
Nice work -- this looks good to me!

PR #1365, opened 2025-10-01
Add support for active alarm windows in MPAS timekeeping

amstokely:
This PR introduces user-defined active windows for recurring alarms in the MPAS timekeeping module. Previously, alarms could only be defined with an anchor time and a recurrence interval. Once the anchor was reached, the alarm would ring repeatedly until the end of the clock. There was no way to restrict alarms to a bounded time range.

### Previous behavior

Alarms were defined only by anchor time and interval:

```
A ---------------- B ---------------------- E
^                 ^                        ^
Clock start    Alarm anchor              Clock stop
```

Once the anchor was reached, alarms rang every interval until the clock ended. There was no mechanism to disable alarms outside of a chosen window.

### New behavior

With this PR, all alarms now have an active window. If the user does not explicitly provide a start or stop time, the window defaults to the clock start and stop times, so existing behavior is preserved. This allows the same logic to be applied consistently to both alarms with user-defined windows and alarms without them.

Timeline with user-defined window:

```
A                B          C            D                E
|----------------|----------|------------|----------------|
^                ^          ^            ^                ^
Clock start   Alarm anchor  Window     Window stop     Clock stop
                           start
```

Alarms do not ring before the window opens, ring normally inside it at the specified interval, and stop ringing after the window closes, even if the clock continues. Resets and direction changes interact correctly with these boundaries.

### Internal structure

Several internal helpers were added to make the logic clearer and easier to maintain:

* `mpas_is_alarm_active` checks whether the current clock time is inside an alarm’s active window.
* `mpas_prev_ring_in_window` verifies whether the alarm’s previous ring time occurred strictly inside the window, using open-interval semantics `(start, stop)`.
* `mpas_time_in_interval` performs the low-level interval membership check, supporting both closed `[start, end]` and open `(start, end)` intervals.

These helpers are not tested directly; instead, their correctness is validated indirectly through behavioral tests of alarms. This ensures that the implementation can be refactored without breaking tests, as long as the observable alarm behavior remains consistent.

### Tests

A new test fixture (`alarm_fixture_t`) and suite (`test_window_alarm`) validate alarm behavior across 17 scenarios. These cover cases before and after the anchor, ringing at the window boundaries, ringing inside the window, leaving the window, delayed anchor times, resets, and clock direction changes.

`mpas_reset_clock_alarm` is invoked in several cases to confirm that the previous ring time is updated correctly. It is called at both window boundaries and inside the active window to exercise different internal bound checks that could fail in one case but not the other.

The tests focus strictly on behavior: they verify only whether alarms ring when expected, not how that behavior is implemented. This design allows the internal code to evolve without destabilizing the suite. At the same time, the tests serve as executable documentation of the alarm ringing contract, which was previously implicit. They will make future development in MPAS timekeeping more efficient, reliable, and safe.

### DIFF
--- a/src/core_test/mpas_test_core_timekeeping_tests.F
+++ b/src/core_test/mpas_test_core_timekeeping_tests.F
@@ -141,8 +141,8 @@ contains
    !> \date   10/28/2025
    !> \details
    !>   Verifies that the provided logical expression evaluates to `.true.`.
-   !>   If the condition is true, a PASS message is logged; otherwise, a FAIL
-   !>   message is logged and the status is set to 1.
+   !>   If the condition is false, a FAIL message is logged and the status
+   !    is set to 1.
    !>
    !>   This routine is used throughout test cases to simplify repetitive
    !>   PASS/FAIL logic and ensure consistent logging and error tracking.
@@ -174,8 +174,8 @@ contains
    !> \date   10/28/2025
    !> \details
    !>   Verifies that the provided logical expression evaluates to `.false.`.
-   !>   If the condition is false, a PASS message is logged; otherwise, a FAIL
-   !>   message is logged and the status is set to 1.
+   !>   If the condition is true, a FAIL message is logged and the status
+   !>   is set to 1.
    !>
    !>   This routine complements `assert_true` and is used when the expected
    !>   behavior requires a logical expression to remain false.
@@ -572,6 +572,7 @@ contains
       integer :: i, ierr_local
 
       ierr = 0
+      call mpas_log_write('Running 18 window alarm tests')
       do i = 1, 18
          ierr_local = 0
          call test_window_alarm(i, ierr_local)
--- a/src/core_test/mpas_test_core_timekeeping_tests.F
+++ b/src/core_test/mpas_test_core_timekeeping_tests.F
@@ -158,7 +158,6 @@ contains
       integer, intent(out) :: status
 
       if (condition) then
-         call mpas_log_write('PASS: ' // trim(message))
          status = 0
       else
          call mpas_log_write('FAIL: ' // trim(message))
@@ -192,7 +191,6 @@ contains
       integer, intent(out) :: status
 
       if (.not. condition) then
-         call mpas_log_write('PASS: ' // trim(message))
          status = 0
       else
          call mpas_log_write('FAIL: ' // trim(message))
--- a/src/core_test/mpas_test_core_timekeeping_tests.F
+++ b/src/core_test/mpas_test_core_timekeeping_tests.F
@@ -23,11 +23,549 @@ module test_core_timekeeping_tests
    private
 
    public :: test_core_test_intervals, &
-             mpas_adjust_alarm_tests
+             mpas_adjust_alarm_tests, &
+             mpas_window_alarm_tests
 
-   contains
+   !-----------------------------------------------------------------------
+   !  type alarm_fixture_t
+   !
+   !> \brief Fixture type for testing alarm ringing behavior with windows.
+   !> \author Andy Stokely
+   !> \date   10/02/2025
+   !> \details
+   !>   This type provides all clocks, alarms, and precomputed iteration
+   !>   steps needed to exercise windowed alarm test cases.
+   !>
+   !>   The test timeline is organized as:
+   !>
+   !>       A                B          C            D                E
+   !>       |----------------|----------|------------|----------------|
+   !>       ^                ^          ^            ^                ^
+   !>   Clock start      Alarm anchor   Window    Window stop     Clock stop
+   !>                                    start
+   !>
+   !>   Fields map directly to these points and derived step counts.
+   !-----------------------------------------------------------------------
+   type alarm_fixture_t
+      !> Clock start time (A on the timeline)
+      type(MPAS_Time_type)         :: clock_start_time
 
-   !***********************************************************************
+      !> Clock stop time (E on the timeline)
+      type(MPAS_Time_type)         :: clock_stop_time
+
+      !> Alarm anchor time (B on the timeline)
+      type(MPAS_Time_type)         :: alarm_time
+
+      !> Alarm window start time (C on the timeline)
+      type(MPAS_Time_type)         :: window_start_time
+
+      !> Alarm window stop time (D on the timeline)
+      type(MPAS_Time_type)         :: window_stop_time
+
+      !> Clock step interval (time advanced per tick)
+      type(MPAS_TimeInterval_type) :: clock_time_step
+
+      !> Alarm recurrence interval (ring spacing)
+      type(MPAS_TimeInterval_type) :: alarm_interval
+
+      !> Clock under test
+      type(MPAS_Clock_type)        :: clock
+
+      !> Alarm with a user-defined active window
+      type(MPAS_Alarm_type), pointer :: alarm
+
+      !> Identifier string for the alarm
+      character(len=:), allocatable  :: alarm_id
+
+      !> Total number of steps the clock will run
+      integer :: num_clock_steps        = 120
+
+      !> Number of substeps per interval (clock resolution)
+      integer :: num_steps_per_interval = 6
+
+      !> Steps from clock start to alarm anchor (A→B)
+      integer :: num_steps_before_anchor = 12
+
+      !> Steps from anchor to window start (B→C)
+      integer :: num_steps_before_window = 36
+
+      !> Steps covering full window duration (C→D)
+      integer :: num_steps_window        = 36
+
+      !> Steps from window end to clock end (D→E)
+      integer :: num_steps_post_window   = 36
+
+      !> Absolute step index at anchor (B)
+      integer :: steps_to_anchor
+
+      !> Absolute step index at window start (C)
+      integer :: steps_to_window_start
+
+      !> Absolute step index at window midpoint
+      integer :: steps_to_window_midpoint
+
+      !> Absolute step index at window end (D)
+      integer :: steps_to_window_end
+
+      !> Absolute step index in post-window region
+      integer :: steps_to_post_window
+   end type alarm_fixture_t
+
+contains
+
+
+   !-----------------------------------------------------------------------
+   !  subroutine advance_clock_n_times
+   !
+   !> \brief Advance a clock forward by a specified number of steps.
+   !> \author Andy Stokely
+   !> \date   10/02/2025
+   !> \details This routine repeatedly advances the provided clock
+   !>   forward by calling mpas_advance_clock `n` times.
+   !-----------------------------------------------------------------------
+   subroutine advance_clock_n_times(clock, n)
+      use mpas_derived_types, only : MPAS_Clock_type
+      implicit none
+      type(MPAS_Clock_type), intent(inout) :: clock
+      integer, intent(in)                  :: n
+      integer                              :: i
+
+      do i = 1, n
+         call mpas_advance_clock(clock)
+      end do
+   end subroutine advance_clock_n_times
+
+
+   !-----------------------------------------------------------------------
+   !  subroutine setup_alarm_fixture
+   !
+   !> \brief Initialize a test fixture with a clock and alarms.
+   !> \author Andy Stokely
+   !> \date   10/02/2025
+   !> \details This routine allocates and initializes an alarm_fixture_t
+   !>   structure.
+   !-----------------------------------------------------------------------
+   subroutine setup_alarm_fixture(fixture)
+      implicit none
+      type(alarm_fixture_t), pointer :: fixture
+      integer                        :: ierr
+
+      allocate(fixture)
+
+      ! Alarm ID
+      fixture%alarm_id        = 'alarm'
+
+      ! Clock setup
+      call mpas_set_time(fixture%clock_start_time, YYYY=2000, MM=01, DD=01, H=0,  &
+            M=0, S=0, S_n=0, S_d=0, ierr=ierr)
+      call mpas_set_time(fixture%clock_stop_time,  YYYY=2000, MM=01, DD=01, H=20, &
+            M=0, S=0, S_n=0, S_d=0, ierr=ierr)
+
+      call mpas_set_timeInterval(fixture%clock_time_step, dt=600.0_RKIND, ierr=ierr)
+
+      call mpas_create_clock(fixture%clock, fixture%clock_start_time, &
+            fixture%clock_time_step, fixture%clock_stop_time, ierr=ierr)
+
+      ! Alarm anchor and interval
+      call mpas_set_time(fixture%alarm_time, YYYY=2000, MM=01, DD=01, H=2, M=0, S=0, S_n=0, S_d=0, ierr=ierr)
+      call mpas_set_timeInterval(fixture%alarm_interval, dt=3600.0_RKIND, ierr=ierr)
+
+      ! Alarm window
+      call mpas_set_time(fixture%window_start_time, YYYY=2000, MM=01, DD=01, H=8, M=0, S=0, S_n=0, S_d=0, ierr=ierr)
+      call mpas_set_time(fixture%window_stop_time,  YYYY=2000, MM=01, DD=01, H=14, M=0, S=0, S_n=0, S_d=0, ierr=ierr)
+
+      ! Add alarms to the clock
+      call mpas_add_clock_alarm(fixture%clock, fixture%alarm_id, fixture%alarm_time, &
+            alarmTimeInterval = fixture%alarm_interval, &
+            alarmStartTime    = fixture%window_start_time, &
+            alarmStopTime     = fixture%window_stop_time)
+      fixture%alarm => fixture%clock%alarmListHead
+
+      ! Derived iteration steps
+      fixture%steps_to_anchor          = fixture%num_steps_before_anchor
+      fixture%steps_to_window_start    = fixture%num_steps_before_anchor + fixture%num_steps_before_window
+      fixture%steps_to_window_midpoint = fixture%steps_to_window_start + fixture%num_steps_window/2
+      fixture%steps_to_window_end      = fixture%steps_to_window_start + fixture%num_steps_window
+      fixture%steps_to_post_window     = fixture%steps_to_window_end + fixture%num_steps_post_window/2
+   end subroutine setup_alarm_fixture
+
+
+   !-----------------------------------------------------------------------
+   !  subroutine teardown_alarm_fixture
+   !
+   !> \brief Finalize and deallocate a test alarm fixture.
+   !> \author Andy Stokely
+   !> \date   10/02/2025
+   !> \details This routine removes alarms from the test clock, destroys
+   !>   the clock, nullifies alarm pointers, and deallocates the fixture.
+   !>   It ensures a clean state for subsequent tests.
+   !-----------------------------------------------------------------------
+   subroutine teardown_alarm_fixture(fixture)
+      implicit none
+      type(alarm_fixture_t), pointer :: fixture
+      integer                        :: ierr
+
+      call mpas_remove_clock_alarm(fixture%clock, fixture%alarm_id, ierr=ierr)
+
+      call mpas_destroy_clock(fixture%clock, ierr=ierr)
+
+      nullify(fixture%alarm)
+      deallocate(fixture)
+   end subroutine teardown_alarm_fixture
+
+
+   !-----------------------------------------------------------------------
+   !  subroutine test_window_alarm
+   !
+   !> \brief Unit test for alarm ringing behavior with user-defined windows.
+   !> \author Andy Stokely
+   !> \date   10/02/2025
+   !> \details This test suite verifies that an alarm rings only when expected
+   !>   relative to the defined timeline:
+   !>
+   !>   A = Clock start time
+   !>   B = Alarm anchor time
+   !>   C = Window start time
+   !>   D = Window end time
+   !>   E = Clock end time
+   !>
+   !>   Timeline:
+   !>      A ---------------- B ----- C ---------------- D --------- E
+   !>
+   !>   The test covers 16 cases, including behavior:
+   !>   - Before anchor (A → B)
+   !>   - Between anchor and window start (B → C)
+   !>   - At window boundaries (C, D)
+   !>   - Inside the window (C → D)
+   !>   - After leaving the window (D → E)
+   !>   - After reset operations at various points
+   !>   - With anchor times shifted into the window
+   !>   - When the clock direction changes (forward/backward)
+   !>
+   !>   Each case checks whether the alarm rings at the correct times and
+   !>   uses mpas_log_write to log PASS/FAIL outcomes.
+   !-----------------------------------------------------------------------
+   subroutine test_window_alarm(case_idx, ierr)
+      implicit none
+      integer, intent(in) :: case_idx
+      integer, intent(inout) :: ierr
+      type(alarm_fixture_t), pointer :: f
+      logical :: ringing
+
+      call setup_alarm_fixture(f)
+      ierr = 0
+
+      select case (case_idx)
+
+         !-----------------------------------------------------------------------
+         ! Case 1: Before anchor (A → B)
+         ! Alarm should not ring before the anchor time.
+         !-----------------------------------------------------------------------
+      case(1)
+         ringing = mpas_is_alarm_ringing(f%clock, f%alarm_id)
+         if (ringing) then
+            call mpas_log_write('FAIL: Alarm is ringing before anchor time')
+            ierr = ierr + 1
+         else
+            call mpas_log_write('PASS: Alarm is not ringing before anchor time')
+         end if
+
+         !-----------------------------------------------------------------------
+         ! Case 2: Before anchor after reset (A → B)
+         ! Reset should not cause false ringing before anchor.
+         !-----------------------------------------------------------------------
+      case(2)
+         call mpas_reset_clock_alarm(f%clock, f%alarm_id)
+         ringing = mpas_is_alarm_ringing(f%clock, f%alarm_id)
+         if (ringing) then
+            call mpas_log_write('FAIL: Alarm is ringing before anchor time after reset')
+            ierr = ierr + 1
+         else
+            call mpas_log_write('PASS: Alarm is not ringing before anchor time after reset')
+         end if
+
+         !-----------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/framework/mpas_timekeeping.F
```diff
@@ -810,7 +824,89 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+      logical function mpas_time_in_interval(start_time, end_time, time, open)
+         implicit none
+         type(MPAS_Time_type), intent(in) :: start_time
+         type(MPAS_Time_type), intent(in) :: end_time
+         type(MPAS_Time_type), intent(in) :: time
+         logical, intent(in), optional    :: open
+
+         mpas_time_in_interval = .false.
+
+         if (present(open) .and. open) then
```
mgduda: Fortran doesn't guarantee short-circuit evaluation, if I recall correctly, so we may need to split this into nested conditionals.

amstokely: Also open is a keyword in Fortran, so I need to rename that argument as well.

src/framework/mpas_timekeeping.F
```diff
@@ -810,7 +822,94 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+         else if (.not. open_interval) then
+            if (time == start_time .or. time == end_time) then
+               mpas_time_in_interval = .true.
+            end if
+         end if
+
+      end function mpas_time_in_interval
+
+
+      logical function mpas_is_alarm_ringing(clock, alarmID, interval, ierr)
```
mgduda: It looks from the GitHub view that all of this new code should be indented by three fewer spaces.

src/framework/mpas_timekeeping.F
```diff
@@ -810,7 +822,94 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+      !>   The boundaries themselves are excluded. If the previous ring time
+      !>   lies inside this open interval, the function returns `.true.`.
+      !-----------------------------------------------------------------------
+      logical function mpas_prev_ring_in_window(alarm)
+         implicit none
+         type(MPAS_Alarm_type), pointer :: alarm
+
+         mpas_prev_ring_in_window = mpas_time_in_interval( &
+               alarm%activeStartTime, alarm%activeStopTime, &
+               alarm%prevRingTime, is_open_interval=.true.)
```
mgduda: While I can see some benefits to having a generic routine `mpas_time_in_interval` and to using that routine here, at some point the use of such a routine may obscure the essence of what's going on in my opinion. What happens when we eventually want half-open intervals? Which arguments to `mpas_time_in_interval` are the interval endpoints, and which is the time to be compared against those endpoints?

Instead, what about simply using code like this (adjust whitespace to taste) to replace the above three lines?
```fortran
      mpas_prev_ring_in_window = (alarm % activeStartTime < alarm % prevRingTime &
                            .and. alarm % prevRingTime < alarm % activeStopTime)
```

src/framework/mpas_timekeeping.F
```diff
@@ -810,7 +822,94 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+      !  function mpas_is_alarm_active
+      !
+      !> \brief Determine if an alarm is currently active.
+      !> \author Andy Stokely
+      !> \date   10/01/2025
+      !> \details This function checks whether the current clock time falls
+      !>   within the active start and stop times of the given alarm. If so,
+      !>   it returns `.true.`. An optional error code may also be returned.
+      !-----------------------------------------------------------------------
+      logical function mpas_is_alarm_active(clock, alarm, ierr)
```
mgduda: As an alternative to passing in the `clock` and always querying `MPAS_NOW` to determine whether an alarm is active, we could instead pass in an `MPAS_Time_type` instance representing the actual -- or hypothetical -- current time to be considered when determining whether the alarm is active. This would make the routine more general, and it potentially leads to less code.

jim-p-w:
What is the use case? 
Why would the window_stop time be before the clock_stop time? If the anchor occurs after the window_stop time the alarm will be missed.
Why would the window_start time be later than the clock_start time?

mgduda:
> What is the use case? Why would the window_stop time be before the clock_stop time? If the anchor occurs after the window_stop time the alarm will be missed. Why would the window_start time be later than the clock_start time?

@jim-p-w The original motivation for enabling alarms to ring only within a specified window was to allow us to create output streams that are written only during part of a model simulation (i.e., within a window of time). The output times for a stream are controlled by alarms, and so starting with a change to the fundamental alarm functionality made sense. Generally when running MPAS, the starting and stopping times for the simulation clock -- within which alarms are instantiated -- are the starting and stopping times for the simulation.

amstokely:
> What is the use case? Why would the window_stop time be before the clock_stop time? If the anchor occurs after the window_stop time the alarm will be missed. Why would the window_start time be later than the clock_start time?

@jim-p-w Here’s a snippet from a `streams.atmosphere` file that demonstrates how this feature is used:

```xml
<immutable_stream name="da_state"
                  type="output"
                  precision="single"
                  clobber_mode="truncate"
                  filename_template="mpasout.$Y-$M-$D_$h.$m.$s.nc"
                  packages="jedi_da"
                  io_type="pnetcdf,cdf5"
                  output_interval="0:40:00"
                  stop_time="0_03:20:00"/>

<stream name="da_state_low_freq"
        type="output"
        precision="single"
        clobber_mode="truncate"
        filename_template="mpasout_low_freq.$Y-$M-$D_$h.$m.$s.nc"
        packages="jedi_da"
        io_type="pnetcdf,cdf5"
        output_interval="1:00:00"
        start_time="0_04:20:00">
    <stream name="da_state"/>
</stream>
```

In this example, the `da_state` stream has an `activeStartTime` equal to the clock’s start time and an `activeEndTime` 3 hours and 20 minutes later, which is earlier than the clock’s end time. The `da_state_low_freq` stream begins 4 hours and 20 minutes after the start of the calculation and remains active until the clock’s end time.

The `da_state` stream writes output more frequently than `da_state_low_freq`. Although they are defined as separate streams, they write the same data and differ only in `stream_name`, `filename_template`, `active window`, and `output frequency`. This setup is useful in `MPAS-JEDI`, where users may want to vary the output frequency of the `da_state`  stream during different phases of a model run.

src/framework/mpas_timekeeping.F
```diff
@@ -541,7 +543,16 @@ subroutine mpas_add_clock_alarm(clock, alarmID, alarmTime, alarmTimeInterval, ie
 
          alarmPtr % isSet = .true.
          alarmPtr % ringTime = alarmTime
-         
+         if (present(alarmStartTime)) then
+            alarmPtr % activeStartTime = alarmStartTime
```
jim-p-w, line 549: consider adding check for alarmStartTime < alarmStopTime if it exists, and alarmStartTime < MPAS_STOP_TIME
log an error if out of range.

similar checks for alarmStopTime

src/framework/mpas_timekeeping.F
```diff
@@ -810,6 +821,56 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+   !  function mpas_prev_ring_in_window
+   !
+   !> \brief Check if the alarm’s previous ring was inside its window.
+   !> \author Andy Stokely
+   !> \date   10/01/2025
+   !> \details This function tests whether the alarm’s `prevRingTime`
+   !>   occurred strictly within the defined start and stop times of the
+   !>   alarm’s active window. The check uses an open interval:
+   !>
+   !>       (start, stop)
```
jim-p-w, line 861: why does mpas_prev_ring_in_window use an open interval, while mpas_is_alarm_active uses a closed interval?

jim-p-w:
> > What is the use case? Why would the window_stop time be before the clock_stop time? If the anchor occurs after the window_stop time the alarm will be missed. Why would the window_start time be later than the clock_start time?
> 
> 
> In this example, the `da_state` stream has an `activeStartTime` equal to the clock’s start time and an `activeEndTime` 3 hours and 20 minutes later, which is earlier than the clock’s end time. The `da_state_low_freq` stream begins 4 hours and 20 minutes after the start of the calculation and remains active until the clock’s end time.
> 
> The `da_state` stream writes output more frequently than `da_state_low_freq`. Although they are defined as separate streams, they write the same data and differ only in `stream_name`, `filename_template`, `active window`, and `output frequency`. This setup is useful in `MPAS-JEDI`, where users may want to vary the output frequency of the `da_state` stream during different phases of a model run.

So `start_time` in the yaml is actually `activeStartTime` and `stop_time` is actually `activeStopTime`?

And what is the `anchor` time?

src/framework/mpas_timekeeping.F
mgduda: There are potential uses for an alarm whose `activeStartTime` is before the start time of a clock, for example, in a future implementation of digital filtering initialization in which we integrate backward in time at the beginning of a simulation. So I'd suggest we don't check that an alarm's active window is completely contained within the clock's [`startTime`, `stopTime`] window.

src/framework/mpas_timekeeping.F
```diff
@@ -882,10 +942,12 @@ logical function mpas_in_ringing_envelope(clock, alarmPtr, interval, ierr)
       integer, intent(out), optional :: ierr
       
       type (MPAS_Time_type) :: alarmNow
+      type (MPAS_Time_type) :: currentTime
       type (MPAS_Time_type) :: alarmThreshold
 
-      alarmNow = mpas_get_clock_time(clock, MPAS_NOW, ierr)
-      alarmThreshold = alarmPtr % ringTime 
+      currentTime = mpas_get_clock_time(clock, MPAS_NOW, ierr)
+      alarmNow = currentTime
```
mgduda, line 953: What's the value in having both `alarmNow` and `currentTime` if both are set to contain the current time on the clock? Could we get by with just `alarmNow`? If so, that might eliminate one potential source of confusion.

src/framework/mpas_timekeeping.F
```diff
@@ -900,7 +962,10 @@ logical function mpas_in_ringing_envelope(clock, alarmPtr, interval, ierr)
          end if
 
          if (alarmThreshold <= alarmNow) then
-            mpas_in_ringing_envelope = .true.
+            if (mpas_is_alarm_active(alarmPtr, currentTime) &
+                  .or. mpas_prev_ring_in_window(alarmPtr)) then
```
jim-p-w, line 970: It looks like this could be true even though the "current time" is not in the alarm window.
Consider the interval is 1 hr, the alarmTime is start+4, the alarmStartTime is start+2, alarmStopTime is start+5, the duration is 12 hours. 
Then the prevRingTime is initialized to be start+3: `prevRingTime = alarmTime - alarmTimeInterval`, and `mpas_prev_ring_in_window` is always true.
I guess I don't understand the semantics of `prevRingTime`.

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -23,11 +23,549 @@ module test_core_timekeeping_tests
...
+   !>
+   !>   Timeline:
+   !>      A ---------------- B ----- C ---------------- D --------- E
+   !>
+   !>   The test covers 16 cases, including behavior:
+   !>   - Before anchor (A → B)
+   !>   - Between anchor and window start (B → C)
+   !>   - At window boundaries (C, D)
+   !>   - Inside the window (C → D)
+   !>   - After leaving the window (D → E)
```
mgduda: While I would agree that it's unlikely to be a problem in practice, I'd suggest using only ASCII characters in Fortran code just to avoid any potential issues in future (either with compilers, or with other tools that we might use to operate on our source code).

src/framework/mpas_timekeeping.F
amstokely: I just added this check before seeing your comment @mgduda. I'll go ahead and remove it.

amstokely: @mgduda Is there any way to check those cases besides comparing activeStartTime and activeStopTime? I agree with @jim-p-w that this is an important validation for all current stream use cases.

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -23,11 +23,588 @@ module test_core_timekeeping_tests
...
+      type(MPAS_Clock_type)        :: clock
+
+      !> Alarm with a user-defined active window
+      type(MPAS_Alarm_type), pointer :: alarm
+
+      !> Identifier string for the alarm
+      character(len=:), allocatable  :: alarm_id
+
+      !> Total number of steps the clock will run
+      integer :: num_clock_steps        = 120
```
jim-p-w: `num_clock_steps` isn't used.

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -23,11 +23,588 @@ module test_core_timekeeping_tests
...
+   !
+   !> \brief Initialize a test fixture with a clock and alarms.
+   !> \author Andy Stokely
+   !> \date   10/02/2025
+   !> \details This routine allocates and initializes an alarm_fixture_t
+   !>   structure.
+   !-----------------------------------------------------------------------
+   subroutine setup_alarm_fixture(fixture)
+      implicit none
+      type(alarm_fixture_t), pointer :: fixture
```
jim-p-w: add `intent(out)` to `fixture` ?

src/framework/mpas_timekeeping.F
amstokely: Great question! @jim-p-w 
Say you have an alarm (**&**) with an active window whose start is denoted **B** and whose end is denoted **C**. The overall clock runs from **A** (the clock start time) to **D** (the clock stop time), and the alarm anchor time is equal to the clock start time (**A**). The alarm is currently inside its active window, and its previous ring time is also within that window, though not at its bounds.

```
A                  B        &        C                  D
|------------------|--------|--------|------------------|
```

The alarm then advances outside of its active window:

```
A                  B                   C       &         D
|------------------|-------------------|-------|---------|
```

In this scenario, the alarm should ring until `mpas_reset_clock_alarm` is called. However, if the alarm advances exactly to its `activeStopTime` and `mpas_reset_clock_alarm` is called:

```
                                       &
A                  B                   C                  D
|------------------|-------------------|------------------|
```

and then advances beyond its active window:

```
A                  B                   C       &         D
|------------------|-------------------|-------|---------|
```

it should **not** be ringing.

By having `mpas_prev_ring_time` use an open interval, an alarm whose previous ring time is equal to either of its bounds will not register as ringing. Meanwhile, `mpas_is_alarm_active` uses a closed interval, which allows alarms to ring at their window bounds. The one exception to this behavior occurs when the clock’s direction changes.

If an alarm advances outside its window:

```
A                  B                   C       &         D
|------------------|-------------------|-------|---------|
```

and the clock direction reverses (in this case, backward), the alarm moves to its window’s end point and `mpas_reset_clock_alarm` is called:

```
                                        &
A                  B                   C                  D
|------------------|-------------------|------------------|
```

Then, if the direction changes again at that end point and the alarm advances outside its window:

```
A                  B                   C       &         D
|------------------|-------------------|-------|---------|
```

the alarm would still be ringing despite having been reset at the end point. This happens because a direction change updates all alarm previous ring times to reflect the new direction, creating this subtle edge case.

src/framework/mpas_timekeeping.F
amstokely: @mgduda If an interval is provided, `alarmNow` is updated as `alarmNow = alarmNow + interval`. In this case, the check for whether an alarm is within its active window could return false even if the current clock time is still before the window’s end time (e.g., when `currentTime <= activeStopTime` but `currentTime + interval > activeStopTime`). Is this the intended behavior?

src/framework/mpas_timekeeping.F
amstokely: @jim-p-w That makes two of us lol. Let me think about this.

amstokely: I think this [test case](https://github.com/MPAS-Dev/MPAS-Model/blob/fbc8a239300659c9698bad86a1b9108af81b4024/src/core_test/mpas_test_core_timekeeping_tests.F#L441) tests the scenario you're describing?

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -23,11 +23,588 @@ module test_core_timekeeping_tests
...
+      select case (case_idx)
+
+         !-----------------------------------------------------------------------
+         ! Case 1: Before anchor (A → B)
+         ! Alarm should not ring before the anchor time.
+         !-----------------------------------------------------------------------
+      case(1)
+         ringing = mpas_is_alarm_ringing(f%clock, f%alarm_id)
+         if (ringing) then
+            call mpas_log_write('FAIL: Alarm is ringing before anchor time')
```
jim-p-w: consider adding a procedure or macro which adds the PASS/FAIL string to the message, e.g.
write_fail('Alarm is ringing before anchor time')
and
write_pass('Alarm is not ringing before anchor time')
for two reasons,
1. all the other tests append the pass/fail string to the end of the message
2. some test suites use 'SUCCESS' and some use 'PASSED'

If there's a function you only need to change it to change prepend/append or the word used to indicate pass or failure

amstokely:
> > > What is the use case? Why would the window_stop time be before the clock_stop time? If the anchor occurs after the window_stop time the alarm will be missed. Why would the window_start time be later than the clock_start time?
> > 
> > 
> > In this example, the `da_state` stream has an `activeStartTime` equal to the clock’s start time and an `activeEndTime` 3 hours and 20 minutes later, which is earlier than the clock’s end time. The `da_state_low_freq` stream begins 4 hours and 20 minutes after the start of the calculation and remains active until the clock’s end time.
> > The `da_state` stream writes output more frequently than `da_state_low_freq`. Although they are defined as separate streams, they write the same data and differ only in `stream_name`, `filename_template`, `active window`, and `output frequency`. This setup is useful in `MPAS-JEDI`, where users may want to vary the output frequency of the `da_state` stream during different phases of a model run.
> 
> So `start_time` in the yaml is actually `activeStartTime` and `stop_time` is actually `activeStopTime`?
> 
> And what is the `anchor` time?

@jim-p-w The anchor time is an internally defined variable that serves as the reference point for the alarm’s time field. You’re also correct about the start_time and stop_time attributes. I find those names concise and intuitive for users. Internally, though, I avoided them in the timekeeping APIs because it’s clearer to think of an alarm as entering or exiting an active window rather than starting or stopping. The active window terminology makes this distinction explicit, since the start and stop values define when an alarm is active rather than when it begins or ends execution.

src/core_test/mpas_test_core_timekeeping_tests.F
amstokely: I’ll make this change before merging since I don’t want to disrupt the diagram formatting while the code is under review.

src/core_test/mpas_test_core_timekeeping_tests.F
amstokely: I replaced the PASS/FAIL conditionals with assert utility subroutines to centralize test status handling and eliminate repetition. This achieves the same outcome as your suggestion above.

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -551,30 +541,19 @@ subroutine test_window_alarm(case_idx, ierr)
          ! Start time after stop time rejected
          call mpas_remove_clock_alarm(f%clock, f%alarm_id)
          call mpas_add_clock_alarm(f%clock, f%alarm_id, f%alarm_time, &
-               alarmTimeInterval = f%alarm_interval, &
-               alarmStartTime    = f%window_stop_time, &
-               alarmStopTime     = f%window_start_time, ierr=local_ierr)
+                 alarmTimeInterval = f%alarm_interval, &
+                 alarmStartTime    = f%window_stop_time, &
+                 alarmStopTime     = f%window_start_time, ierr=local_ierr)
+         call assert_true(local_ierr /= 0, 'Alarm with start time after stop time did not return error', local_ierr)
```
jim-p-w, line 546: All of these messages read oddly in the passing case, e.g.
`PASS: Alarm with start time after stop time did not return error`
I don't know if it's worth providing both pass and fail messages to assert_true / assert_false.

amstokely: Woops, I didn't mean to include messages with passing cases.

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -23,11 +23,564 @@ module test_core_timekeeping_tests
...
+   !> \date   10/02/2025
+   !> \details This driver function executes the suite of test cases
+   !>   implemented in test_window_alarm, aggregating any errors encountered.
+   !-----------------------------------------------------------------------
+   integer function mpas_window_alarm_tests() result(ierr)
+
+      integer :: i, ierr_local
+
+      ierr = 0
+      do i = 1, 18
```
jim-p-w: Perhaps output a message here saying something like `Running 18 window alarm tests`?

src/core_test/mpas_test_core_timekeeping_tests.F
```diff
@@ -23,11 +23,564 @@ module test_core_timekeeping_tests
...
+
+   !-----------------------------------------------------------------------
+   !  subroutine assert_true
+   !
+   !> \brief Assert that a condition is true; logs PASS/FAIL accordingly.
+   !> \author Andy Stokely
+   !> \date   10/28/2025
+   !> \details
+   !>   Verifies that the provided logical expression evaluates to `.true.`.
+   !>   If the condition is true, a PASS message is logged; otherwise, a FAIL
```
jim-p-w: This comment is no longer true.
Same for the comment at line 177.

jim-p-w (approved):

src/framework/mpas_timekeeping.F
mgduda: It's hard for me to say whether this was the intended behavior, but given that this behavior has been present for so long, I think it has become the accepted behavior.

My original question came up because I only looked at the GitHub diff and didn't see the lines where `alarmNow` is modified. I agree that both `alarmNow` and `currentTime` are needed.

src/framework/mpas_timekeeping.F
mgduda: I think comparing `activeStartTime` and `activeStopTime` is the most reasonable approach.

src/framework/mpas_timekeeping.F
```diff
@@ -486,20 +486,24 @@ type (MPAS_Time_type) function mpas_get_clock_time(clock, whichTime, ierr)
...
       type (MPAS_TimeInterval_type), intent(in), optional :: alarmTimeInterval
-      integer, intent(out), optional :: ierr
+      type (MPAS_Time_type), intent(in), optional :: alarmStartTime
+      type (MPAS_Time_type), intent(in), optional :: alarmStopTime
+      integer, intent(out),  optional :: ierr
 
       type (MPAS_Alarm_type), pointer :: alarmPtr
       integer :: threadNum
 
+      if (present(ierr)) ierr = ESMF_SUCCESS
```
mgduda: Since existing code compares the value of `ierr` returned by this subroutine with 0 to determine success, I'd suggest we set `ierr` to 0 rather than to `ESMF_SUCCESS`.

src/framework/mpas_timekeeping.F
```diff
@@ -486,20 +486,24 @@ type (MPAS_Time_type) function mpas_get_clock_time(clock, whichTime, ierr)
...
       implicit none
 
       type (MPAS_Clock_type), intent(inout) :: clock
       character (len=*), intent(in) :: alarmID
       type (MPAS_Time_type), intent(in) :: alarmTime
       type (MPAS_TimeInterval_type), intent(in), optional :: alarmTimeInterval
-      integer, intent(out), optional :: ierr
+      type (MPAS_Time_type), intent(in), optional :: alarmStartTime
+      type (MPAS_Time_type), intent(in), optional :: alarmStopTime
+      integer, intent(out),  optional :: ierr
```
mgduda: Let's use just one space between `intent(out),` and `optional`.

src/framework/mpas_timekeeping.F
```diff
@@ -810,6 +825,56 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+   !> \date   10/01/2025
+   !> \details This function checks whether the provided time falls
+   !>   within the active start and stop times of the given alarm. If so,
+   !>   it returns `.true.`.
+   !-----------------------------------------------------------------------
+   logical function mpas_is_alarm_active(alarm, time)
+
+      implicit none
+
+      type(MPAS_Alarm_type), pointer :: alarm
```
mgduda: Does the `alarm` argument need to be a `pointer` here?

src/framework/mpas_timekeeping.F
```diff
@@ -810,6 +825,56 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+   !>       (start, stop)
+   !>
+   !>   The boundaries themselves are excluded. If the previous ring time
+   !>   lies inside this open interval, the function returns `.true.`.
+   !-----------------------------------------------------------------------
+   logical function mpas_prev_ring_in_window(alarm)
+
+      implicit none
+
+      type(MPAS_Alarm_type), pointer :: alarm
```
mgduda: Does `alarm` need to be a pointer?

src/framework/mpas_timekeeping.F
```diff
@@ -824,7 +889,6 @@ logical function mpas_is_alarm_ringing(clock, alarmID, interval, ierr)
       if (present(ierr)) ierr = 0
 
       mpas_is_alarm_ringing = .false.
-
```
mgduda, line 827: Let's leave this blank line in place.

src/framework/mpas_timekeeping.F
```diff
@@ -810,6 +825,56 @@ subroutine mpas_print_alarm(clock, alarmID, ierr)
...
+      mpas_is_alarm_active = (alarm % activeStartTime <= time &
+              .and. time <= alarm % activeStopTime)
+
+   end function mpas_is_alarm_active
+
+
+   !-----------------------------------------------------------------------
+   !  function mpas_prev_ring_in_window
+   !
+   !> \brief Check if the alarm’s previous ring was inside its window.
```
mgduda, line 854: Here (and maybe elsewhere), can you check that we're using an ASCII single-quote rather than some other Unicode character?

mgduda (changes requested):

PR #1366, opened 2025-10-06
Update the core_atmosphere cmake file to include src/framework

jim-p-w:
This change adds an include of the src /framework directory when building core-atmosphere.

This is required due to a dependency added in https://github.com/MPAS-Dev/MPAS-Model/pull/1359.

This fix was tested by building MPAS-Model core_atmosphere via cmake, using the gnu compiler.
This was also tested by building the mpas-bundle suite via cmake using the gnu compiler.

### DIFF
--- a/src/framework/CMakeLists.txt
+++ b/src/framework/CMakeLists.txt
@@ -69,6 +69,9 @@ if (MPAS_PROFILE)
     list(APPEND FRAMEWORK_LINK_LIBRARIES GPTL::GPTL)
 endif ()
 target_link_libraries(framework PUBLIC ${FRAMEWORK_LINK_LIBRARIES})
+target_include_directories(framework INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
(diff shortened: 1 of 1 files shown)


amstokely (changes requested):
Consider adding the following to the framework CMakeLists.txt:
```

target_include_directories(framework
    INTERFACE
        ${CMAKE_CURRENT_SOURCE_DIR}
)

```

This will automatically append the framework source directory to the core_atmosphere include path, achieving the same result without relying on relative paths.

jim-p-w:
> Consider adding the following to the framework CMakeLists.txt:
> target_include_directories(framework
>     INTERFACE
>         ${CMAKE_CURRENT_SOURCE_DIR}

Putting the dependency in the framework cmake file sounds much cleaner. I'll make an update.

jim-p-w:
The src/framework directory is now exported by the framework CMakeLists.txt, as suggested by @amstokely.

abishekg7 (approved):
Looks good to me. Checked that it builds correctly.

amstokely (approved):

mgduda (approved):

Issue #1350, opened 2025-07-23
Check for input buffer size when reading character array variables.


jim-p-w:
Character array variables are read into a 512 byte buffer. If the value of the variable exceeds 512 bytes a buffer overflow will occur.
The size of the variable value should be checked before reading it into the buffer, and an error returned if the value exceeds the size of the provided buffer.
Alternatively, the read could be limited to the size of the buffer.

Here's the call sequence (Line numbers are from origin/release-v8.3.0)
In src/framework/mpas_io_streams.F

MPAS_readStream() reads strings into 512 byte buffer 
```
 Line 3204   MPAS_io_get_var(stream % fileHandle, field_cursor % char0dField % fieldName, 
                                                field_cursor % char0dField % scalar, io_err)
  		scaler is a 512 byte buffer
Calls MPAS_io_get_var_char0d in src/framework/mpas_io.F via MPAS_io_get_var interface
    line 2995 Calls MPAS_io_get_var_generic in mpas_io.F
      line 1988 Calls SMIOLf_get_var in smiolf.F90 
        Calls SMIOLf_get_var_0d_char in src/external/SMIOL/smiolf_put_get_var.inc
          line 188 Calls SMIOL_get_var in src/external/SMIOL/smiol.c.
		   line 1515 calls read_chunk_pnetcdf
		       line 2840 calls ncmpi_get_vara_all
```

jim-p-w:
A potential solution is call `MPAS_io_inq_var` immediately prior to the call to `SMIOLf_get_var` in `MPAS_io_get_var_generic`
If the returned dimension of the variable exceeds the size of the buffer passed to `MPAS_io_get_var_generic` set an error,
otherwise make the call to `SMIOLf_get_var`

**Potential drawbacks**

*If using SMIOL*
A potential drawback is `MPAS_io_inq_var` makes the following SMIOL calls:
- `SMIOLf_inquire_var`
- `SMIOLf_inquire_dim`

This is mitigated by caching the results for each variable, so multiple reads of the same variable won't result in multiple SMIOL_inquire calls.

jim-p-w:
This issue has been addressed by https://github.com/MPAS-Dev/MPAS-Model/pull/1367.

PR #1367, opened 2025-10-14
Ensure the buffer provided to MPAS_io_get_var_generic is large enough.

jim-p-w:
This PR fixes a potential buffer overrun when reading string variables from a netcdf file.

A fixed size array is provided as an output buffer when reading a 0d-char character variable. 
Call MPAS_io_inq_var prior to the read to get the size of the variable, and only proceed with the read if the size of the variable will fit in the provided array.
Return an error code if the variable value is larger than the provided output buffer.

A unit test is included to verify:
1. An attempt to read a string variable into a buffer which is too small to hold the string value is detected
2. When a buffer which is too small is detected, the read won't occur and an error code is returned
3. When the test is run via valgrind, valgrind detects no memory errors or corruption
4. If the code which detects the too small buffer is commented out the test crashes and valgrind reports memory corruption

Fixes issue https://github.com/MPAS-Dev/MPAS-Model/issues/1350

### Note ###
When building with PIO, if the charArray (or the charArray1d) value exceeds the size of the tempchar buffer provided to the call to `PIO_get_var`, the value will be truncated to the size of the provided tempchar buffer (lines 2018, 2023, and 2042 in src/framework/mpas_io.F).

### DIFF
--- a/src/core_test/Makefile
+++ b/src/core_test/Makefile
@@ -12,7 +12,8 @@ OBJS = mpas_test_core.o \
        mpas_test_core_dmpar.o \
        mpas_test_core_stream_inquiry.o \
        mpas_test_openacc.o \
-       mpas_test_core_stream_list.o
+       mpas_test_core_stream_list.o \
+       mpas_test_core_io.o
 
 all: core_test
 
@@ -44,7 +45,7 @@ mpas_test_core.o: mpas_test_core_halo_exch.o mpas_test_core_streams.o \
                   mpas_test_core_sorting.o mpas_halo_testing.o \
                   mpas_test_core_string_utils.o mpas_test_core_dmpar.o \
                   mpas_test_core_stream_inquiry.o mpas_test_openacc.o \
-                  mpas_test_core_stream_list.o
+                  mpas_test_core_stream_list.o mpas_test_core_io.o
 
 mpas_test_core_halo_exch.o:
 
--- a/src/core_test/mpas_test_core.F
+++ b/src/core_test/mpas_test_core.F
@@ -97,6 +97,7 @@ module test_core
       use test_core_string_utils, only : mpas_test_string_utils
       use mpas_test_core_dmpar, only : mpas_test_dmpar
       use mpas_test_core_stream_inquiry, only : mpas_test_stream_inquiry
+      use test_core_io, only : test_core_io_test
       use mpas_test_core_openacc, only : mpas_test_openacc
 
       implicit none
@@ -223,6 +224,18 @@ module test_core
       end do
 
       call mpas_stream_mgr_write(domain % streamManager, forceWriteNow=.true.)
+
+      !
+      ! Run io tests
+      !
+      call mpas_log_write('')
+      call test_core_io_test(domain, iErr)
+      if (iErr == 0) then
+          call mpas_log_write('All tests PASSED')
+      else
+          call mpas_log_write('$i tests FAILED', intArgs=[iErr])
+      end if
+      call mpas_log_write('')
 
       !
       ! Run mpas_test_openacc
--- a/src/core_test/mpas_test_core_io.F
+++ b/src/core_test/mpas_test_core_io.F
@@ -1,0 +1,200 @@
+! Copyright (c) 2025 The University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
+! Additional copyright and license information can be found in the LICENSE file
+! distributed with this code, or at https://mpas-dev.github.io/license.html .
+!
+module test_core_io
+
+#define ERROR_WRITE(M) call mpas_log_write( M , messageType=MPAS_LOG_ERR)
+#define ERROR_WRITE_ARGS(M, ARGS) call mpas_log_write( M , ARGS, messageType=MPAS_LOG_ERR)
+   use mpas_log
+   use mpas_io
+
+   implicit none
+   private
+   public :: test_core_io_test 
+
+   contains
+
+   !***********************************************************************
+   !
+   !  routine close_file_with_message
+   !
+   !> \brief   closes the provided file handle and writes an error message.
+   !-----------------------------------------------------------------------
+   subroutine close_file_with_message(fileHandle, message, args)
+      type(MPAS_IO_Handle_type), intent(inout) :: fileHandle
+      character (len=*), intent(in), optional :: message
+      integer, dimension(:), intent(in), optional :: args
+
+      integer :: local_ierr
+
+      ! log an error message
+      if (present(message)) then
+         ERROR_WRITE_ARGS(message, intArgs=args)
+      end if
+
+      ! close the provided file
+      call MPAS_io_close(fileHandle, local_ierr)
+      if (local_ierr /= MPAS_IO_NOERR) then
+         ERROR_WRITE_ARGS('MPAS_io_close failed with error code:$i', intArgs=(/local_ierr/))
+         return
+      endif
+
+   end subroutine close_file_with_message
+
+   !***********************************************************************
+   !
+   !  routine test_read_string_buffer_check
+   !
+   !> \brief   verifies attempts to read strings into buffers which are too small
+   !>          to hold the value fails safely.
+   !> \details 
+   !>  Run these tests with valgrind to ensure there are no buffer overflows when
+   !>  attempting to read strings into undersized buffers.
+   !-----------------------------------------------------------------------
+   subroutine test_read_string_buffer_check(domain, ierr)
+
+      type (domain_type), intent(inout) :: domain
+      integer, intent(out) :: ierr
+
+      integer :: local_ierr, i
+      type(MPAS_IO_Handle_type) :: fileHandle
+      character (len=StrKIND), dimension(1), parameter :: dimNamesString = ['StrLen']
+      character (len=StrKIND), dimension(2), parameter :: dimNamesStringTime = &
+         [character(len=StrKIND) :: 'StrLen', 'Time']
+      character (len=32), parameter :: varName1 = 'stringVar'
+      character (len=32), parameter :: varName2 = 'stringTimeVar'
+      character (len=*), parameter :: varValue1 = 'This is a string'
+      character (len=32), dimension(2), parameter :: varNames = [varName1, varName2]
+      integer, parameter :: bufferSize=128
+      integer, parameter :: smallBufferSize=bufferSize/2
+      character (len=bufferSize) :: buffer
+      character (len=smallBufferSize) :: smallBuffer
+      character (len=*), parameter :: filename = 'char_data.nc'
+
+      ierr = 0
+
+      ! open a file to write char variables to
+      fileHandle = MPAS_io_open(filename, MPAS_IO_WRITE, MPAS_IO_NETCDF, domain % ioContext, &
+                                  clobber_file=.true., truncate_file=.true., ierr=local_ierr)
+      if (local_ierr /= MPAS_IO_NOERR) then
+         ierr = 1
+         ERROR_WRITE('Error opening file ' // trim(filename))
+         return
+      end if
+
+      ! define dimensions and char variables
+      call MPAS_io_def_dim(fileHandle, dimNamesStringTime(1), bufferSize, local_ierr)
+      if (local_ierr /= MPAS_IO_NOERR) then
+         ierr = 1
+         call close_file_with_message(fileHandle, 'Error defining '//trim(dimNamesStringTime(1))//', error=$i', (/local_ierr/))
+         return
+      end if
+      call MPAS_io_def_dim(fileHandle, dimNamesStringTime(2), MPAS_IO_UNLIMITED_DIM, local_ierr)
+      if (local_ierr /= MPAS_IO_NOERR) then
+         ierr = 1
+         call close_file_with_message(fileHandle, 'Error defining '//trim(dimNamesStringTime(2))//', error=$i', (/local_ierr/))
+         return
+      end if
+      call MPAS_io_def_var(fileHandle, varNames(1), MPAS_IO_CHAR, dimNamesString, ierr=local_ierr)
+      if (local_ierr /= MPAS_IO_NOERR) then
+         ierr = 1
+         call close_file_with_message(fileHandle, 'Error defining var "'//trim(varNames(1))//'" error=$i', (/local_ierr/))
+         return
+      end if
+      call MPAS_io_def_var(fileHandle, varNames(2), MPAS_IO_CHAR, dimNamesStringTime, ierr=local_ierr)
+      if (local_ierr /= MPAS_IO_NOERR) then
+         ierr = 1
+         call close_file_with_message(fileHandle, 'Error defining var "'//trim(varNames(2))//'" error=$i', (/local_ierr/))
+         return
+      end if
+
+      ! write the string values
+      do i=1,size(varNames)
+         call MPAS_io_put_var_char0d(fileHandle, varNames(i), varValue1, local_ierr)
+         if (local_ierr /= MPAS_IO_NOERR) then
+            ierr = 1
+            call close_file_with_message(fileHandle, 'Error writing "'//trim(varNames(i))// &
+                                       '", error=$i', (/local_ierr/))
+            return
+         end if
+
+         ! verify the strings are read into buffers which are large enough for the string values
+         call MPAS_io_get_var_char0d(fileHandle, varNames(i), buffer, local_ierr)
+         if (local_ierr /= MPAS_IO_NOERR) then
+            ierr = 1
+            call close_file_with_message(fileHandle, 'Error reading "'//trim(varNames(i))// &
+                                       '", error=$i', (/local_ierr/))
+            return
+         end if
+      end do
+
+      ! verify attempts to read strings into buffers which are too small generates an error
+      call mpas_log_write(' ')
+      call mpas_log_write('Expect to see the following error:')
+      call MPAS_io_err_mesg(domain % ioContext, MPAS_IO_ERR_INSUFFICIENT_BUF, .false.)
+      call mpas_log_write(' ')
+      do i=1,size(varNames)
+         ! this should return an error
+         call MPAS_io_get_var_char0d(fileHandle, varNames(i), smallBuffer, local_ierr)
+         call mpas_log_write(' ')
+
+         if (local_ierr /= MPAS_IO_ERR_INSUFFICIENT_BUF) then
+            ierr = 1
+            if (local_ierr == MPAS_IO_NOERR) then
+               call close_file_with_message(fileHandle, 'Expected MPAS_IO_ERR_INSUFFICIENT_BUF ($i)'&
+                //' but recieved no error reading "'//trim(varName1), (/local_ierr/))
+            else
+               call close_file_with_message(fileHandle, 'Expected MPAS_IO_ERR_INSUFFICIENT_BUF ($i)'&
+                                        //' but recieved error $i reading "'//trim(varName1)//'"', &
+                                        (/MPAS_IO_ERR_INSUFFICIENT_BUF, local_ierr/))
+            end if
+            return
+         end if
+      end do
+      call close_file_with_message(fileHandle)
+
+   end subroutine test_read_string_buffer_check
+
+
+   !***********************************************************************
+   !  Subroutine test_core_io_test
+   !
+   !> \brief   Core test suite for I/O
+   !>
+   !> \details This subroutine tests mpas_io features.
+   !>          It calls individual tests for I/O operations.
+   !>          See the subroutine body for details.
+   !>          The results of each test are logged with a success or failure message.
+   !>
+   !> \param domain      The domain object that contains the I/O context
+   !> \param ierr        The error code that indicates the result of the test.
+   !
+   !-----------------------------------------------------------------------
+   subroutine test_core_io_test(domain, ierr)
+
+      use mpas_log
+
+      type (domain_type), intent(inout) :: domain
+      integer, intent(out) :: ierr
+
+      integer :: test_status
+
+      ierr = 0
+      test_status = 0
+
+      call mpas_log_write('Testing char-0 buffer reads')
+      call test_read_string_buffer_check(domain, test_status)
+      if (test_status == 0) then
+         call mpas_log_write('char-0 buffer tests: SUCCESS')
+      else
+         call mpas_log_write('char-0 buffer tests: FAILURE', MPAS_LOG_ERR)
+         ierr = ierr + abs(test_status)
+      end if
+
+
+   end subroutine test_core_io_test
+
+end module test_core_io
--- a/src/framework/mpas_io.F
+++ b/src/framework/mpas_io.F
@@ -7,6 +7,9 @@
 !
 module mpas_io
 
+#define IO_DEBUG_WRITE(M, ARGS) !call mpas_log_write(M, ARGS)
+#define IO_ERROR_WRITE(M, ARGS) call mpas_log_write( M, ARGS, messageType=MPAS_LOG_ERR)
+
    use mpas_derived_types
    use mpas_attlist
    use mpas_dmpar
@@ -1847,6 +1850,13 @@ module mpas_io
       character (len=:), pointer :: charVal_p
       character (len=:), dimension(:), pointer :: charArray1d_p
 
+      ! local variables returned from MPAS_io_inq_var
+      integer :: fieldtype
+      integer :: ndims
+      integer, dimension(:), pointer :: dimsizes
+      character (len=StrKIND), dimension(:), pointer :: dimnames
+      character (len=StrKIND) :: message
+
 #ifdef MPAS_SMIOL_SUPPORT
       type (SMIOLf_decomp), pointer :: null_decomp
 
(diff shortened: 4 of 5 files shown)


amstokely (approved):

src/core_test/Makefile
```diff
@@ -12,7 +12,8 @@ OBJS = mpas_test_core.o \
        mpas_test_core_dmpar.o \
        mpas_test_core_stream_inquiry.o \
        mpas_test_openacc.o \
-       mpas_test_core_stream_list.o
+       mpas_test_core_stream_list.o \
+       mpas_test_core_io.o \
```
mgduda: Although it's not incorrect, for the sake of consistency and to avoid problems if someone were to inadvertently delete the blank line below, I'd suggest not using a trailing `\` for `mpas_test_core_io.o`.

src/core_test/mpas_test_core.F
```diff
@@ -224,6 +225,17 @@ function test_core_run(domain) result(iErr)!{{{
...
+      ! Run io tests
+      !
+      call mpas_log_write('')
+      call test_core_io_test(domain, threadErrs, iErr)
+      if (iErr == 0) then
+          call mpas_log_write('All tests PASSED')
+      else
+          call mpas_log_write('$i tests FAILED', intArgs=[iErr])
+      end if
+      call mpas_log_write('')
```
mgduda, line 238: I'd suggest adding a blank line below this one for consistency of style.

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,208 @@
+! Copyright (c) 2013,  Los Alamos National Security, LLC (LANS)
```
mgduda: At a minimum, we should update the year to 2025. But you can also follow the copyright statement in, e.g., [mpas_test_core_stream_list.F](https://github.com/MPAS-Dev/MPAS-Model/blob/develop/src/core_test/mpas_test_core_stream_list.F).

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,208 @@
...
+   subroutine close_file_with_message(fileHandle, message, args)
+      type(MPAS_IO_Handle_type), intent(inout) :: fileHandle
+      character (len=*), intent(in), optional :: message
+      integer, dimension(:), intent(in), optional :: args
+
+      integer :: local_ierr
+
+      ! log an error message
+      if (present(message)) then
+        if (present(args)) then
```
mgduda: Some places in this file use two spaces of indentation and others use three. I'd suggest sticking with a consistent indentation (with three or four spaces being the most common throughout MPAS, as far as I'm aware).

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,208 @@
...
+      do i=1,2
+        call MPAS_io_put_var_char0d(fileHandle, varNames(i), varValue1, local_ierr)
+        if (local_ierr /= MPAS_IO_NOERR) then
+           ierr = 1
+           call close_file_with_message(fileHandle, 'Error writing "'//trim(varNames(i))// &
+                                       '", error=$i', (/local_ierr/))
+           return
+        end if
+
+        ! verify the strings are read into buffers which are large enough for the strin values
```
mgduda: "strin" -> "string"

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,208 @@
...
+      end if
+      call MPAS_io_def_var(fileHandle, varNames(2), MPAS_IO_CHAR, dimNamesStringTime, ierr=local_ierr)
+      if (local_ierr /= MPAS_IO_NOERR) then
+         ierr = 1
+         call close_file_with_message(fileHandle, 'Error defining var "'//trim(varNames(2))//'" error=$i', (/local_ierr/))
+         return
+      end if
+
+      ! write the string values
+      do i=1,2
```
mgduda: We could replace `2` with `size(varNames)`.

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,208 @@
...
+           return
+        end if
+      end do
+
+      ! verify attempts to read strings into buffers which are too small generates an error
+      call mpas_log_write(' ')
+      call mpas_log_write('Expect to see the following error:')
+      call MPAS_io_err_mesg(domain % ioContext, MPAS_IO_ERR_INSUFFICIENT_ARG, .false.)
+      call mpas_log_write(' ')
+      do i=1,2
```
mgduda: Similarly, we could use `size(varNames)` here rather than `2`.

src/framework/mpas_io.F
```diff
@@ -1984,22 +1993,40 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
 !         call mpas_log_write('  value is char')
 
          charVal_p => charVal
+
+         ! get the dimension of the char variable to ensure the provided output buffer is large enough
+         call MPAS_io_inq_var(handle, fieldname, fieldtype, ndims, dimnames, dimsizes, local_ierr)
+         do i = 1, ndims
+           message = '  MPAS_io_get_var_generic len(charVal):$i var "'//trim(fieldname)// &
```
mgduda: I'd suggest following the surrounding code and using three spaces of indentation.

mgduda (changes requested):

src/core_test/Makefile
jim-p-w: I addressed this, plus added the new file to the `mpas_test_core.` dependencies.

src/core_test/mpas_test_core_io.F
jim-p-w: I used the copyright from `mpas_test_core_stream_list.F`

src/core_test/mpas_test_core_io.F
jim-p-w: I think I've got everything uniformly using 3 spaces per indent level.

src/core_test/mpas_test_core_io.F
jim-p-w: Fixed the typo.

src/core_test/mpas_test_core_io.F
jim-p-w: Now using the `size()` function for look upper bound.

src/framework/mpas_io.F
jim-p-w: Think I've got everything using 3 spaces uniformly.

src/framework/mpas_io.F
```diff
@@ -6498,6 +6531,10 @@ subroutine MPAS_io_err_mesg(ioContext, ierr, fatal)
             call mpas_log_write('MPAS IO Error: Would clobber existing file', MPAS_LOG_ERR)
          case (MPAS_IO_ERR_NOEXIST_READ)
             call mpas_log_write('MPAS IO Error: Attempting to read a file which does not exist.', MPAS_LOG_ERR)
+         case (MPAS_IO_ERR_MISSING_DIM)
+            call mpas_log_write('MPAS IO Error: Attempting to read a dimension which does not exist.', MPAS_LOG_ERR)
+         case (MPAS_IO_ERR_INSUFFICIENT_ARG)
+            call mpas_log_write('MPAS IO Error: Attempting to read a string into a buffer which is too small.', MPAS_LOG_ERR)
```
mgduda: Just a thought, but what if we named this error code `MPAS_IO_ERR_INSUFFICIENT_BUF`, and generalized the error string to something like: `MPAS IO Error: Attempting to read a variable into a buffer of insufficient size.` This might allow us to re-use this error condition more generally throughout the `mpas_io` module when we detect that there is too much data for the buffer that was provided, rather than having the error be specific to strings.

src/framework/mpas_io.F
```diff
@@ -1847,6 +1849,13 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
       character (len=:), pointer :: charVal_p
       character (len=:), dimension(:), pointer :: charArray1d_p
 
+      ! local variables returned from MPAS_io_inq_var
+      integer :: fieldtype
+      integer :: ndims
+      integer, dimension(:), pointer :: dimsizes
+      character (len=StrKIND), dimension(:), pointer :: dimnames
+      character (len=StrKind) :: message
```
mgduda: For consistency with the line above it and elsewhere in MPAS, let's change `StrKind` to `StrKIND` on this line.

src/framework/mpas_io.F
```diff
@@ -1984,22 +1993,40 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
...
-         else
-            start1(1) = 1
-            count1(1) = field_cursor % fieldhandle % dims(1) % dimsize
-            pio_ierr = PIO_get_var(handle % pio_file, field_cursor % fieldhandle % fieldid, start1, count1, tempchar)
-            charVal(1:count1(1)) = tempchar(1)(1:count1(1))
-         end if
+            if (field_cursor % fieldhandle % has_unlimited_dim) then
+               count2(1) = field_cursor % fieldhandle % dims(1) % dimsize
+               pio_ierr = PIO_get_var(handle % pio_file, field_cursor % fieldhandle % fieldid, start2, count2, tempchar)
+              charVal(1:count2(1)) = tempchar(1)(1:count2(1))
```
mgduda: I think one more space of indentation is needed here.

src/framework/mpas_io.F
```diff
@@ -1984,22 +1993,40 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
...
+         call MPAS_io_inq_var(handle, fieldname, fieldtype, ndims, dimnames, dimsizes, local_ierr)
+         do i = 1, ndims
+            message = '  MPAS_io_get_var_generic len(charVal):$i var "'//trim(fieldname)// &
+                      '" type is $i dim $i is '// trim(dimnames(i))//' size is $i'
+            IO_DEBUG_WRITE(message , intArgs=(/len(charVal), fieldtype, i, dimsizes(i)/))
+         end do
+         ! because charVal is provided, assume dimension 1 is the string length
+         if (dimsizes(1) > len(charVal)) then
+            local_ierr = MPAS_IO_ERR_INSUFFICIENT_ARG
+            message = '  MPAS_io_get_var_generic var "'//trim(fieldname)//  &
```
mgduda: Are the leading two spaces in this string intentional? They lead to messages that look like this:
```
ERROR:   MPAS_io_get_var_generic var "xtime" len too big, len=128 buflen=64
ERROR: MPAS IO Error: Attempting to read a string into a buffer which is too small.
```

src/framework/mpas_io.F
jim-p-w: Good idea; implemented.

src/framework/mpas_io.F
jim-p-w: Capitalization is amended.

src/framework/mpas_io.F
jim-p-w: A vi search string of `/^\(   \)\+` (note 3 spaces between the parens) was quite helpful!

src/framework/mpas_io.F
jim-p-w: whitespace is modified.

src/framework/mpas_io.F
```diff
@@ -2765,28 +2792,34 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
       end if
 
 !      call mpas_log_write('Checking for error')
+      if (local_ierr == MPAS_IO_ERR_INSUFFICIENT_BUF) then
+         call MPAS_io_err_mesg(handle % ioContext, local_ierr, .false.)
+         io_global_err = local_ierr
+         if (present(ierr)) ierr = local_ierr
```
mgduda, line 2801: In case of an `MPAS_IO_ERR_INSUFFICIENT_BUF` error, this will be the last statement executed in this subroutine, and there may be an opportunity to minimize the changes to the code below this block. Rather than logic that is structured like
```fortran
      if (local_ierr == MPAS_IO_ERR_INSUFFICIENT_ARG) then
         ... code ...
      else
#ifdef MPAS_PIO_SUPPORT
        if (pio_ierr /= PIO_noerr) then
           ... code ...
           return
        end if
#endif

#ifdef MPAS_SMIOL_SUPPORT
        if (local_ierr /= SMIOL_SUCCESS) then
           ... code ...
           return
        end if
#endif
      end if
```
what if we used something like this:
```fortran
      if (local_ierr == MPAS_IO_ERR_INSUFFICIENT_ARG) then
         ... code ...
         return
      end if

#ifdef MPAS_PIO_SUPPORT
      if (pio_ierr /= PIO_noerr) then
         ... code ...
         return
      end if
#endif

#ifdef MPAS_SMIOL_SUPPORT
      if (local_ierr /= SMIOL_SUCCESS) then
         ... code ...
         return
      end if
#endif
```
The idea here is that everything from the `#ifdef MPAS_PIO_SUPPORT` line downward is completely unchanged from the base code.

src/framework/mpas_io.F
```diff
@@ -7,6 +7,8 @@
 !
 module mpas_io
 
+#define IO_DEBUG_WRITE(M, ARGS) !call mpas_log_write(M, ARGS)
+#define IO_ERROR_WRITE(M, ARGS) call mpas_log_write( M, ARGS, messageType=MPAS_LOG_ERR)
```
mgduda: So that we don't need to make up an unused argument in cases where we don't have any arguments to the `mpas_log_write` call, for example:
```
IO_ERROR_WRITE('Here is a message', intArgs=[0])
```
we could instead define this macro as:
```
#define IO_ERROR_WRITE(M) call mpas_log_write(M, messageType=MPAS_LOG_ERR)
```
In order to allow for the passing of other arguments to `mpas_log_write`, we can also define the macro `COMMA` as:
```
#define COMMA ,
```
Then, an error message with arguments would look something like:
```
IO_ERROR_WRITE('The answer is $i' COMMA intArgs=[42])
```

src/framework/mpas_io.F
```diff
@@ -1984,22 +1993,40 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
...
+         do i = 1, ndims
+            message = '  MPAS_io_get_var_generic len(charVal):$i var "'//trim(fieldname)// &
+                      '" type is $i dim $i is '// trim(dimnames(i))//' size is $i'
+            IO_DEBUG_WRITE(message , intArgs=(/len(charVal), fieldtype, i, dimsizes(i)/))
+         end do
+         ! because charVal is provided, assume dimension 1 is the string length
+         if (dimsizes(1) > len(charVal)) then
+            local_ierr = MPAS_IO_ERR_INSUFFICIENT_BUF
+            message = 'MPAS_io_get_var_generic var "'//trim(fieldname)//  &
+                      '" len too big, len=$i buflen=$i'
```
mgduda: It might make this message more helpful to users if we used something like:
```fortran
            message = 'Length of string variable "'//trim(fieldname)//'" in file "'//trim(handle % filename)//'"'
            IO_ERROR_WRITE(message, intArgs=(/0/))
            message = '    exceeds buffer size: len('//trim(fieldname)//')=$i, len(buffer)=$i'
            IO_ERROR_WRITE(message, intArgs=(/dimsizes(1), len(charVal)/)) 
```
This would generate an error log entry something like:
```
ERROR: Length of string variable "xtime" in file "x1.10242.init.nc"
ERROR:     exceeds buffer size: len(xtime)=128, len(buffer)=64
```

src/framework/mpas_io.F
jim-p-w: Nice observation. Done.

src/framework/mpas_io.F
jim-p-w: Clever! I had wondered about the definition of `COMMA` in other files. Done.

src/framework/mpas_io.F
jim-p-w: Good idea. I love printing out the filename which contains the variable. Done.

src/framework/mpas_io.F
```diff
@@ -2765,6 +2794,12 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
       end if
 
 !      call mpas_log_write('Checking for error')
+      if (local_ierr == MPAS_IO_ERR_INSUFFICIENT_BUF) then
+         call MPAS_io_err_mesg(handle % ioContext, local_ierr, .false.)
+         io_global_err = local_ierr
+         if (present(ierr)) ierr = local_ierr
+         return
+      endif
```
mgduda, line 2803: A blank line below this `endif` might be helpful in providing a visual indicator that the block below is independent of this block that is being terminated here.

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,207 @@
...
+   !
+   !  routine test_read_string_buffer_check
+   !
+   !> \brief   verifies attempts to read strings into buffers which are too small
+   !>          to hold the value fails safely.
+   !> \details 
+   !>  Run these tests with valgrind to ensure there are no buffer overflows when
+   !>  attempting to read strings into undersized buffers.
+   !-----------------------------------------------------------------------
+   subroutine test_read_string_buffer_check(domain, threadErrs, ierr)
```
mgduda: With the GCC compilers and using the `-Wall` flag, there's a warning that `threadErrs` is never set:
```
mpas_test_core_io.F:61:62:

   61 |    subroutine test_read_string_buffer_check(domain, threadErrs, ierr)
      |                                                              1
Warning: Dummy argument 'threaderrs' at (1) was declared INTENT(OUT) but was not set [-Wunused-dummy-argument]
```
We could either ensure that this `intent(out)` argument is set, or -- since it doesn't appear to be used by any calling code -- we could just eliminate this argument.

src/framework/mpas_io.F
```diff
@@ -1984,22 +1994,41 @@ subroutine MPAS_io_get_var_generic(handle, fieldname, intVal, intArray1d, intArr
...
+                      '" type is $i dim is $i '// trim(dimnames(i))//' size is $i'
+            IO_DEBUG_WRITE(message COMMA intArgs=(/len(charVal), fieldtype, i, dimsizes(i)/))
+         end do
+         ! because charVal is provided, assume dimension 1 is the string length
+         if (dimsizes(1) > len(charVal)) then
+            local_ierr = MPAS_IO_ERR_INSUFFICIENT_BUF
+            message = 'Length of string variable "'//trim(fieldname)//'" in file "'//trim(handle % filename)//'"'
+            IO_ERROR_WRITE(message)
+            message = '    exceeds buffer size: len('//trim(fieldname)//')=$i, len(buffer)=$i'
+            IO_ERROR_WRITE(message COMMA intArgs=(/dimsizes(1), len(charVal)/))
```
mgduda, line 2012: With the GCC 12.4.0 compilers, this macro seems to work as expected, and I get a message like
```
ERROR: Length of string variable "xtime" in file "x1.10242.init.nc"
ERROR:     exceeds buffer size: len(xtime)=128, len(buffer)=64
```
in the log file for a test that I've been using.

However, with the Intel oneAPI 2025.1.0 compilers, I get
```
ERROR: Length of string variable "xtime" in file "x1.10242.init.nc"
ERROR:     exceeds buffer size: len(xtime)=**, len(buffer)=**
```

The problem seems to be in the macro expansion with the `ifx` compiler. Using the `-preprocess-only` option, it appears the `ifx` compiler is expanding the macro as:
```fortran
            call mpas_log_write( message, messageType=MPAS_LOG_ERR)
```

Could you investigate whether there's a workaround that we can employ so that this works with the `ifx` compiler?

mgduda (changes requested):

src/framework/mpas_io.F
jim-p-w: Blank line has been added.

src/core_test/mpas_test_core_io.F
jim-p-w: threaderrs argument has been removed from test_core_io_test and test_read_string_buffer_check

src/framework/mpas_io.F
jim-p-w: I have changed the log macros to use two explicit arguments while I investigate a one argument solution which works with the intel compiler.

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,203 @@
...
+   subroutine close_file_with_message(fileHandle, message, args)
+      type(MPAS_IO_Handle_type), intent(inout) :: fileHandle
+      character (len=*), intent(in), optional :: message
+      integer, dimension(:), intent(in), optional :: args
+
+      integer :: local_ierr
+
+      ! log an error message
+      if (present(message)) then
+         if (present(args)) then
```
mgduda: I think optional arguments can be passed through to other subroutines whose corresponding dummy arguments are also optional without any need to check presence. So it may be possible to simplify this block of code as:
```fortran
      if (present(message)) then
         ERROR_WRITE_ARGS(message, intArgs=args)
      end if
```

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,203 @@
...
+   !-----------------------------------------------------------------------
+   subroutine test_read_string_buffer_check(domain, ierr)
+
+      type (domain_type), intent(inout) :: domain
+      integer, intent(out) :: ierr
+
+      integer :: local_ierr, i
+      type(MPAS_IO_Handle_type) :: fileHandle
+      character (len=StrKIND), dimension(1), parameter :: dimNamesString = ['StrLen']
+      character (len=StrKIND), dimension(2), parameter :: dimNamesStringTime = ['StrLen', 'Time  ']
```
mgduda: Although it apparently does no harm, are the two spaces in the `'Time  '` dimension intentional?

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,203 @@
...
+      character (len=32), parameter :: varName1 = 'stringVar'
+      character (len=32), parameter :: varName2 = 'stringTimeVar'
```
mgduda: Out of curiosity, is there a reason for specifying a length of `32` for these strings and a length of `StrKIND` for the strings on lines 68 - 69?

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,203 @@
...
+      integer, parameter :: bufferSize=128
+      integer, parameter :: smallBufferSize=bufferSize/2
+      character (len=bufferSize) :: buffer
+      character (len=smallBufferSize) :: smallBuffer
+      character (len=*), parameter :: filename = 'char_data.nc'
+
+      ierr = 0
+
+      ! open a file to write char variables to
+      fileHandle = MPAS_io_open(filename, MPAS_IO_WRITE, MPAS_IO_NETCDF4, domain % ioContext, &
```
mgduda: Since we're not writing any large variables (and so parallel I/O isn't essential), it might increase compatibility if we used `MPAS_IO_NETCDF` for the format here rather than `MPAS_IO_NETCDF4`.

src/core_test/mpas_test_core_io.F
mgduda: Ok, I see there's a compilation error if these strings don't have the same length:
```
mpas_test_core_io.F:65:84:

   65 |       character (len=StrKIND), dimension(2), parameter :: dimNamesStringTime = ['StrLen', 'Time']
      |                                                                                    1
Error: Different CHARACTER lengths (6/4) in array constructor at (1)
```

It should work to declare `dimNamesStringTime` as:
```fortran
character (len=StrKIND), dimension(2), parameter :: dimNamesStringTime = [character(len=StrKIND) :: 'StrLen', 'Time']
```

mgduda (changes requested):

src/core_test/mpas_test_core_io.F
jim-p-w: Thanks, I wouldn't have figured that out!

src/core_test/mpas_test_core_io.F
jim-p-w: OK, but that didn't change the file type. ncdump -k reports `cdf5` for both filetypes of `MPAS_IO_NETCDF` and `MPAS_IO_NETCDF4`

src/core_test/mpas_test_core_io.F
jim-p-w: I see.

src/core_test/mpas_test_core_io.F
jim-p-w: Thanks, I wouldn't have figured that out!

src/core_test/mpas_test_core_io.F
jim-p-w: OK. It looks like that parameter isn't used with SMIOL? ncdump -k reports `cdf5` when using `MPAS_IO_NETCDF4` and when using `MPAS_IO_NETCDF4`.

mgduda: Right -- SMIOL ignores `ioformat` at present, but the PIO library does make use of `ioformat` when creating new files.

mgduda (approved):
@jim-p-w I've verified that the changes to `mpas_io.F` catch cases where we would otherwise have tried to read a string into a buffer that was too small. If you could rework the commit history so that it contains just a couple of commits -- perhaps one to modify the framework mpas_io code, and another to add unit tests -- I think we'll be ready to merge this.

jim-p-w:
@mgduda Is the initial commit message in the PR suitable for the framework commit? Or do you have suggestions/edits you would prefer to see?

src/core_test/mpas_test_core_io.F
```diff
@@ -0,0 +1,200 @@
+! Copyright (c) 2025 The University Corporation for Atmospheric Research (UCAR).
+!
+! Unless noted otherwise source code is licensed under the BSD license.
+! Additional copyright and license information can be found in the LICENSE file
+! distributed with this code, or at https://mpas-dev.github.io/license.html .
+!
+module test_core_io
+
+#define ERROR_WRITE(M) call mpas_log_write( M , messageType=MPAS_LOG_ERR)
+#define ERROR_WRITE_ARGS(M, ARGS) call mpas_log_write( M , ARGS, messageType=MPAS_LOG_ERR)
```
mgduda, line 10: Rather than two macros here for writing error messages, I suspect we could use the same `COMMA` macro approach as in `mpas_io.F` and have just a single `ERROR_WRITE` macro.

mgduda (changes requested):

src/core_test/mpas_test_core_io.F
mgduda: On second thought, since the code is correct as written, let's just keep the macros as they are. Better to get this PR merged, in my opinion!

mgduda (approved):

PR #1368, opened 2025-10-22
add a new stream property 'gattr_update' to control whether to update global attributes 

guoqing-noaa:
When we do coldstart data assimilation on `init.nc` using JEDI, we will get a crash in MPAS IO with the following error:
```
ERROR: MPAS IO Error: PIO error -38: NetCDF: Operation not allowed in data mode
```
Per discussions with NOAA @samuelTrahanNOAA and NCAR Tao Sun, we identified that the problem came from `src/framework/mpas_stream_manager.F` where `mpas_writeStreamAtt` tries to write not-defined global attributes to `init.nc`. Since we don't expect JEDI to update any global attributes (at least for `init.nc`), we can skip the relevant `mpas_writeStreamAtt` part when doing JEDI data assimilation.

This PR is to address this issue by adding a new stream property `gattr_update`. It takes values of `yes` or `no` and defaults to `yes`. So it does not affect any existing MPAS-Model capabilities but allow coldstart JEDI DA to skip the relevant `mpas_writeStreamAtt` part to avoid crash, by setting `gattr_update=no` in the `analysis` stream.

### DIFF
--- a/src/framework/mpas_stream_list_types.inc
+++ b/src/framework/mpas_stream_list_types.inc
@@ -22,6 +22,7 @@
         integer :: nRecords
         integer :: precision = MPAS_IO_NATIVE_PRECISION
         integer :: clobber_mode
+        integer :: gattr_update = 1
         integer :: io_type
         type (MPAS_TimeInterval_type), pointer :: recordInterval => null()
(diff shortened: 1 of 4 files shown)


guoqing-noaa:
Tag @byoung-joo @liujake for awareness.

PR #1371, opened 2025-10-26
Add CMake support for building and setting up the MPAS test core

amstokely:
This PR introduces full CMake integration for building and running the MPAS test core. It adds new CMake utility functions that automate downloading, extracting, and linking the test input data required to run the core.

The following helper functions were added:

* `download_mpas_test_data`
* `untar_tarball`
* `mpas_link_file_force`
* `mpas_link_directory`
* `setup_mpas_test_core`

When the test core is built with CMake, these utilities automatically download the `mpas_test_data.tar.gz` archive (if not already present), extract it, and link the necessary files into the CMake binary directory. This ensures the test core can be built and executed directly through CMake without any manual setup steps.

No existing build logic or model behavior was modified. These changes only add support for building and preparing the test core environment through CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@
 cmake_minimum_required(VERSION 3.12)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Functions/MPAS_Functions.cmake)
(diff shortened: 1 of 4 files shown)


jim-p-w:
Consider adding instructions for building and running the test to the commit message. Something like
```
To build the test, cmake -DMPAS_CORES=test .. followed by make.
Then cd to the test directory and run ./mpas_test
```

PR #1372, opened 2025-10-27
Nullify pointer arguments in all pool accessor routines for consistency

amstokely:
This PR updates all pool accessor routines to nullify their pointer arguments at the start of each subroutine. The affected routines are:

* `mpas_pool_get_config_real`
* `mpas_pool_get_config_int`
* `mpas_pool_get_config_char`
* `mpas_pool_get_config_logical`
* `mpas_pool_get_subpool`
* `mpas_pool_get_package`

Previously, these pointer arguments (`value`, `subPool`, and `package`) were not initialized before assignment. This could lead to undefined behavior if callers used `associated()` to check whether a config key, subpool, or package existed in the pool.

By explicitly nullifying these pointers, they are guaranteed to be in a defined state before assignment, allowing callers to safely perform `associated()` checks. This change also makes the behavior of these accessor routines consistent with the existing behavior of `mpas_pool_get_field_*`, `mpas_pool_get_array_*`, and `mpas_pool_get_dimension_*`.

### DIFF
--- a/src/framework/mpas_pool_routines.F
+++ b/src/framework/mpas_pool_routines.F
@@ -4770,6 +4770,8 @@ module mpas_pool_routines
 
       type (mpas_pool_data_type), pointer :: mem
 
+      nullify(value)
+
       if ( present(record) ) then
          call mpas_pool_get_subpool(inPool, record, recordPool)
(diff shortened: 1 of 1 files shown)


abishekg7 (approved):
Looks good to me!

mgduda (approved):
Looks good to me.

PR #1376, opened 2025-11-10
MPAS MUSICA namelist option for MICM configuration

dwfncar:
**MICM configuration is now driven by a MUSICA namelist option with added logging.**

Registry.xml: Added a MUSICA namelist record gated by MPAS_USE_MUSICA with the `config_micm_file` option so the MICM JSON path can be provided through the standard configuration system.
mpas_atm_chemistry.F: Removed the hardcoded `chapman.json`, pull the MICM file path from the configs pool, then propagate errors from `musica_init` via `mpas_log_write` to fail when initialization breaks. mpas_musica.F: Track the species description pointer and log each MICM species name from `state%species_ordering` so users can verify the runtime mapping.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -394,6 +394,15 @@
                      possible_values="`mpas_dmpar', `mpas_halo'"/>
         </nml_record>
 
+#ifdef MPAS_USE_MUSICA
+        <nml_record name="musica" in_defaults="true">
+                <nml_option name="config_micm_file" type="character" default_value=""
+                     units="-"
+                     description="MICM configuration file name"
+                     possible_values="Any valid filename"/>
+        </nml_record>
+#endif
+
 <!-- **************************************************************************************** -->
 <!-- **************************************  Packages  ************************************** -->
 <!-- **************************************************************************************** -->
--- a/src/core_atmosphere/chemistry/mpas_atm_chemistry.F
+++ b/src/core_atmosphere/chemistry/mpas_atm_chemistry.F
@@ -43,7 +43,7 @@ module mpas_atm_chemistry
         use mpas_musica, only: musica_init
 #endif
         use mpas_log, only : mpas_log_write
-        use mpas_derived_types, only: mpas_pool_type
+        use mpas_derived_types, only: mpas_pool_type, MPAS_LOG_CRIT
         use mpas_kind_types, only: StrKIND
         use mpas_pool_routines, only: mpas_pool_get_config, mpas_pool_get_dimension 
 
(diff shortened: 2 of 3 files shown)


src/core_atmosphere/chemistry/musica/mpas_musica.F
```diff
@@ -61,9 +61,10 @@ subroutine musica_init(filename_of_micm_configuration, &
 
         type(error_t)  :: error
         type(string_t) :: micm_version
-
+        type(string_t) :: description
```
mgduda: It doesn't look like the `description` variable is used anywhere in this routine.

src/core_atmosphere/chemistry/musica/mpas_musica.F
```diff
@@ -77,6 +78,12 @@ subroutine musica_init(filename_of_micm_configuration, &
         state => micm%get_state(number_of_grid_cells, error)
         if (has_error_occurred(error, error_message, error_code)) return
 
+        associate(map => state%species_ordering)
```
mgduda: Although it's completely standard Fortran, it might be worth avoiding an `associate` block here for a couple of reasons:
1. With Noah-MP, we hit an internal compiler error with the GCC 13 compilers due to `associate` blocks in that code; so there is evidence that some compiler releases don't handle `associate` very well.
2. With an `associate` block, it's not possible to tell the type of `map`, and one would have to look elsewhere to find, e.g., that `name` is a valid member of an instance of whatever type `map` has.

So it might prevent compilation issues and lead to clearer code if we provide an explicit declaration of `map` as a local variable in the `musica_init` routine.

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
```diff
@@ -43,21 +43,19 @@ subroutine chemistry_init(configs, dimensions)
         use mpas_musica, only: musica_init
 #endif
         use mpas_log, only : mpas_log_write
-        use mpas_derived_types, only: mpas_pool_type
+        use mpas_derived_types
```
mgduda: Was there a specific reason for eliminating the `only` clause for this `use` statement?

mgduda (changes requested):

src/core_atmosphere/chemistry/mpas_atm_chemistry.F
mgduda: I see now that this was probably to get access to `MPAS_LOG_CRIT`. In that case, perhaps it would be better to add `MPAS_LOG_CRIT` to the `use, only` statement?

dwfncar: Restored use mpas_derived_types, only: mpas_pool_type, added MPAS_LOG_CRIT.

src/core_atmosphere/chemistry/musica/mpas_musica.F
dwfncar: Removed associate block in musica_init.

src/core_atmosphere/Registry.xml
```diff
@@ -394,6 +394,15 @@
                      possible_values="`mpas_dmpar', `mpas_halo'"/>
         </nml_record>
 
+#ifdef MPAS_USE_MUSICA
+        <nml_record name="musica" in_defaults="true">
+                <nml_option name="config_micm_file" type="character" default_value=""
```
mgduda, line 399: While I agree that we should let the default value be an empty string, should we also add logic in `musica_init` to not try to initialize MUSICA if `config_micm_file` is an empty string? Otherwise, would there be another way to not call MUSICA in a particular simulation when MPAS was compiled with `MUSICA=true`?

mgduda (changes requested):

src/core_atmosphere/Registry.xml
dwfncar: 1. Added musica_is_initialized flag (line 32)
  2. Added early return when config file is empty, with log message (lines 69-75)
  3. Added logging of the config file path (line 80)
  4. Set musica_is_initialized = .true. after successful init (line 93)
  5. Added guard in musica_step (line 114)
  6. Added guard in musica_finalize (line 137)

mgduda (approved):

PR #1378, opened 2025-11-13
Fix units designations of GW drag diagnostic variables dusfcg, dtaux2d, etc.

mdtoyNOAA:
This PR corrects the units designations of the gravity wave drag diagnostics variables dusfcg, dvsfcg, dtaux3d and dtauy3d as follows:

The units for the “dusfcg” and “dvsfcg” variables are currently “Pa m s^{-1}” when they actually should simply be “Pa”.  These variables represent the vertical integral of “d(tau)/dz”, which turns out to be the GWD at the surface, whose units are “Pa”.

The units for the “dtaux” and “dtauy” variables are currently “m s^{-1}” when they actually should be “m s^{-2}”. These variables represent the vertical divergence of the horizontal GWD momentum flux ( (-1/rho) * d(tau)/dz ), where tau is the GWD momentum flux or “wave stress” (units are Pascal).

This code has not been tested as the changes are extremely minor and they do not change the model results.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -2948,10 +2948,10 @@
                 <!-- ... PARAMETERIZATION OF GRAVITY WAVE DRAG OVER OROGRAPHY:                                          -->
                 <!-- ================================================================================================== -->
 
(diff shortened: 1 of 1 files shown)


mgduda (approved):
Thanks for these corrections, and apologies that it's taken so long to review this PR!

PR #1384, opened 2025-12-05
fix typos in Registry_noahmp for acc output variables in restart file

CharlesZheZhang:
This PR is to fix a few typos in the Registry_noahmp.xml file.
Essentially it is for outputing a few accumulated water fluxes per soil timestep in the restart file
correcting typo, for example, from **acc_ecan** to **acc_ecanxy**, for the correct variable name.

### DIFF
--- a/src/core_atmosphere/physics/Registry_noahmp.xml
+++ b/src/core_atmosphere/physics/Registry_noahmp.xml
@@ -319,31 +319,31 @@
       <var name="snliqxy" type="real" dimensions="nSnowLevels nCells Time" units="mm"
                 description="snow layer liquid" missing_value="-9999.0"/>
 
-      <var name="acc_ssoil" type="real" dimensions="nCells Time" units="W m^{-2}"
+      <var name="acc_ssoilxy" type="real" dimensions="nCells Time" units="W m^{-2}"
                 description="accumulated ssoil between dt_soil"/>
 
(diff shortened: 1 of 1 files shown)


cenlinhe:
This bug fix looks good to me. It does not affect model run by default because MPAS uses the same main timestep as Noah-MP soil timestep.

mgduda:
@CharlesZheZhang @cenlinhe With the changes in this PR, I'm seeing the following compilation error:
```
ERROR: Field acc_ssoilxy is not unique.
ERROR: Structures and Fields are required to have unique names for I/O reasons.
       Please fix duplicates in the Registry.xml file.
       You may use the name_in_code attribute to give them the same name inside the model,
       but the name attribute is required to be unique.
```
Looking through the `Registry_noahmp.xml` file, it does appear that there is already an entry for `acc_ssoilxy` before the changes in this PR rename `acc_ssoil` to `acc_ssoilxy`.

cenlinhe:
@mgduda a quick question: it seems that the existing "acc_ssoilxy" is in the "output_noahmp" structure part while the one fixed in this PR is in the "diag_physics_noahmp" structure, both in the sf_noahmp_in package. So what is the difference between "diag_physics_noahmp" and "output_noahmp"? Variables cannot share the same name in these two different structures? Originally, I thought "diag_physics_noahmp" variables are for restart file while "output_noahmp" are for model output files, but it seems my understanding is wrong?

Issue #1385, opened 2025-12-12
MPAS I/O performance degradation during Parallel NetCDF header reallocation


amstokely:
If MPAS encounters an I/O scenario that requires Parallel NetCDF to reallocate the file header (for example, when `ncmpi_enddef` is called after the header has grown beyond its preallocated padding while in define mode), a significant I/O performance degradation may be observed, which can appear as an I/O hang. This behavior was the root cause of [issue 384](https://github.com/NCAR/MPAS-Workflow/issues/384) in MPAS-Workflow. In that case, the perceived hang occurred when overwriting the `config_block_decomp_file_prefix` global attribute in the analysis NetCDF file with a longer string. Using Forge DDT, the slowdown was traced to the call to `ncmpi_enddef` following this attribute update.

byoung-joo:
This issue seems to be resolved with #1386.

PR #1386, opened 2025-12-12
Increase default PnetCDF header size 

amstokely:
This PR increases the default Parallel NetCDF (PnetCDF) header size to 128 KB to reduce the likelihood of header reallocation during MPAS I/O. In certain situations, when MPAS overwrites existing string attributes or variables with larger values, the NetCDF header can grow beyond its preallocated padding, requiring PnetCDF to reallocate the header during `ncmpi_enddef`. This can significantly decrease I/O performance and emulate an I/O hang. This behavior was identified as the root cause of the hang reported in [MPAS-Workflow issue #384](https://github.com/NCAR/MPAS-Workflow/issues/384). By increasing the default header size, this PR decreases the likelihood that header reallocation is required when string attributes or variables are overwritten with larger values. Preliminary testing of the calculation that previously triggered the hang indicates that this change resolves the issue without impacting calculation results or I/O performance. 

Fixes https://github.com/MPAS-Dev/MPAS-Model/issues/1385

### DIFF
--- a/src/external/SMIOL/smiol.c
+++ b/src/external/SMIOL/smiol.c
@@ -12,6 +12,7 @@
 #define PNETCDF_DEFINE_MODE 0
 #define PNETCDF_DATA_MODE 1
 #define MAX_REQS 256
+#define PNETCDF_HEADER_ALIGN_SIZE_STR "131072"
 #endif
 
 #define START_COUNT_READ 0
@@ -324,6 +325,7 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
 			/*
 			 * Convert fformat to a PNetCDF file creation mode
 			 */
+			MPI_Info info = MPI_INFO_NULL;
 			int filecmode;
 			if (fformat == SMIOL_FORMAT_CDF2) {
 				filecmode = NC_64BIT_OFFSET;
@@ -336,11 +338,14 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
 				MPI_Comm_free(&io_group_comm);
 				return SMIOL_INVALID_FORMAT;
 			}
-
+			MPI_Info_create(&info);
+			MPI_Info_set(info, "nc_header_align_size",
+			             PNETCDF_HEADER_ALIGN_SIZE_STR);
 			ierr = ncmpi_create(io_file_comm, filename,
 			                    (filecmode | NC_CLOBBER),
-			                    MPI_INFO_NULL,
+			                    info,
 			                    &((*file)->ncidp));
+			MPI_Info_free(&info);
 		}
 		(*file)->state = PNETCDF_DEFINE_MODE;
 #endif


src/external/SMIOL/smiol.c
```diff
@@ -336,10 +336,12 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
 				MPI_Comm_free(&io_group_comm);
 				return SMIOL_INVALID_FORMAT;
 			}
-
+            MPI_Info info;
```
mgduda: I'd suggest following the convention in SMIOL of declaring variables at the beginning of code blocks (beginning on line 323 in this case).

src/external/SMIOL/smiol.c
```diff
@@ -336,10 +336,12 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
...
+            MPI_Info_create(&info);
+            MPI_Info_set(info, "nc_header_align_size", "131072");
```
mgduda: So that it's easier to find and modify in future, I'd suggest declaring the string `"131072"` as a macro at the top of the file where we also define `MAX_REQS`.

src/external/SMIOL/smiol.c
```diff
@@ -336,10 +336,12 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
 				MPI_Comm_free(&io_group_comm);
 				return SMIOL_INVALID_FORMAT;
 			}
-
+            MPI_Info info;
+            MPI_Info_create(&info);
```
mgduda: Do we need to free `info` after the call to `ncmpi_create`?

mgduda (changes requested):

src/external/SMIOL/smiol.c
amstokely: Yes, we do. I'll fix this.

wkliao:
(Sorry to jump into the discussion.)
There is another way to do this without creating an MPI_Info object, which is to call [`ncmpi__enddef()`](https://parallel-netcdf.github.io/doc/c-reference/pnetcdf-c/ncmpi_005f_005fenddef.html#ncmpi_005f_005fenddef), i.e.
```
ierr = ncmpi__enddef(file->ncidp, 0, 131072, 0, 0);
```
You can also set argument `h_minfree`, which ensures at least this amount of free space in the header section.

FYI. When adding new data objects into an existing file which causes the file header section to grow, PnetCDF must move the data section to a place with a higher file offset, which can be expensive, especially when the size of exiting file is large. The application program most likely just ran very slowly, but not hanging.

mgduda:
@wkliao Thanks very much for the suggestion to consider `ncmpi__enddef` as an alternative to passing a hint to `ncmpi_create`. I do like that this approach would require fewer lines of new code.

If we were to open an existing file for writing and call `ncmpi___enddef` (even if we didn't define any new dimensions or variables), would a header reallocation always be forced in case the original header was less than the size of the `v_align` argument to `nmcpi__enddef`?

amstokely:
> (Sorry to jump into the discussion.) There is another way to do this without creating an MPI_Info object, which is to call [`ncmpi__enddef()`](https://parallel-netcdf.github.io/doc/c-reference/pnetcdf-c/ncmpi_005f_005fenddef.html#ncmpi_005f_005fenddef), i.e.
> 
> ```
> ierr = ncmpi__enddef(file->ncidp, 0, 131072, 0, 0);
> ```
> 
> You can also set argument `h_minfree`, which ensures at least this amount of free space in the header section.
> 
> FYI. When adding new data objects into an existing file which causes the file header section to grow, PnetCDF must move the data section to a place with a higher file offset, which can be expensive, especially when the size of exiting file is large. The application program most likely just ran very slowly, but not hanging.

@wkliao This issue only surfaced when running MPAS through MPAS-JEDI under very specific conditions, which made it difficult to track down. Is there a check you’re aware of that we could have put in place to make this easier to catch?

wkliao:
Let me use two PnetCDF terminologies to help explain.
* "header extent" is referred to as the size of file header section and
* "header size" is the true size of header (i.e. metadata). 

Subtracting the two gives you the free space available in the header section.

> If we were to open an existing file for writing and call `ncmpi___enddef` (even if we didn't define any new dimensions or variables), would a header reallocation always be forced in case the original header was less than the size of the `v_align` argument to `nmcpi__enddef`?

In this case, PnetCDF will check if the file extent of the existing file is aligned with `v_align`. If it is not, PnetCDF will move the data section to a high file offset that is aligned with `v_align`.

Please note that `v_align` is not the size of file extent. It is used to align the header extent to a multiple of `v_align`. For example, say the file header size is 2000 bytes and `v_align` is 1024, then the file extent will become 2048.

wkliao:
> Is there a check you’re aware of that we could have put in place to make this easier to catch?

Two PnetCDF APIs can be used to query the `header extent`  and `header size` of an existing file. They can also be called after `ncmpi_enddef()`:
* [ncmpi_inq_header_extent()](https://parallel-netcdf.github.io/doc/c-reference/pnetcdf-c/ncmpi_005finq_005fheader_005fextent.html)
* [ncmpi_inq_header_size()](https://parallel-netcdf.github.io/doc/c-reference/pnetcdf-c/ncmpi_005finq_005fheader_005fsize.html)

If the free space is sufficiently large, then just call `ncmpi_enddef()`.
Otherwise, moving the data section is inevitable.

src/external/SMIOL/smiol.c
```diff
@@ -243,6 +244,7 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
 	int io_group;
 	MPI_Comm io_file_comm;
 	MPI_Comm io_group_comm;
+	MPI_Info info = MPI_INFO_NULL;
```
mgduda: I'm not sure that there's any benefit to declaring `info` in this scope and using it in all calls to `ncmpi_open`. I'd suggest again that we declare this variable in the block beginning on line 323 (325 in this PR) and apply `info` only when creating new files.

mgduda (changes requested):

src/external/SMIOL/smiol.c
amstokely: My reasoning here was to ensure that the `info` variable is always in a valid state (which I think is important when possible) and to improve code consistency, so that all calls to `ncmpi_open` use the same `MPI_Info` variable. This also makes it easier to pass a non-null `MPI_Info` object to other `ncmpi_open` calls in the future, if needed. That said, I don’t feel strongly about this change and am happy to revert it if you prefer the previous approach to calling `ncmpi_open`.

mgduda: Since our intent is only to pass a hint to `ncmpi_create`, and not to change the semantics of the calls to `ncmpi_open` for writing an existing file or reading a file, I think it would be much cleaner and clearer if we only use `info` within the scope of the block containing the call to `ncmpi_create`.

If in future we decide to pass a non-null `info` to either of the `ncmpi_open` calls, we can decide at that time how to handle the `MPI_Info` instance.

mgduda (changes requested):

src/external/SMIOL/smiol.c
amstokely: Are you okay with initializing info to `MPI_INFO_NULL` at declaration? I’ll revert the other `ncmpi_open` changes that aren’t directly related to this feature.

mgduda: I think it's fine to initialize `info` at declaration.

src/external/SMIOL/smiol.c
mgduda, line 324: Can you please move the declaration of `info` to this block here, since that variable doesn't need broader scope?

mgduda (changes requested):

src/external/SMIOL/smiol.c
amstokely: done

src/external/SMIOL/smiol.c
```diff
@@ -320,6 +321,7 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
 
 	if (mode & SMIOL_FILE_CREATE) {
 #ifdef SMIOL_PNETCDF
+		MPI_Info info = MPI_INFO_NULL;
```
mgduda: Can you please move the declaration of `info` inside the block beginning on line 325 (in this PR branch; line 323 in the original code)?

src/external/SMIOL/smiol.c
```diff
@@ -336,11 +338,13 @@ int SMIOL_open_file(struct SMIOL_context *context, const char *filename,
 				MPI_Comm_free(&io_group_comm);
 				return SMIOL_INVALID_FORMAT;
 			}
-
+            MPI_Info_create(&info);
+            MPI_Info_set(info, "nc_header_align_size", PNETCDF_HEADER_ALIGN_SIZE_STR);
```
mgduda: It looks like this line and the one above it uses spaces for indentation, rather than tabs.

mgduda (changes requested):

src/external/SMIOL/smiol.c
mgduda: I should also mention that we've tried to keep lines wrapped at 80 characters unless doing so would significantly compromise readability.

mgduda (approved):
The code changes look fine to me. Could you rework the commit history so that it contains just a single commit?

jim-p-w:
Has anyone checked to see how much larger output files are with this change?

jim-p-w:
I have verified that this fixes the MPAS-Workflow hang reported in https://github.com/NCAR/MPAS-Workflow/issues/384

I extracted the sources from the 8/16/25 build of mpas-bundle, and rebuilt.
I verified the hang reported above was hit.
I then applied the changes from this PR and rebuilt.
I verified that the hang reported above was resolved.

jim-p-w (approved):

mgduda:
> Has anyone checked to see how much larger output files are with this change?

I think we can expect that output files will increase in size by less than 128 KiB with the changes in this PR. In a relative sense, this seems minor to me. Consider that the smallest global mesh that we provide on the MPAS mesh download page is a 480-km mesh with 2562 cells, and that mesh file has a size of 3,508,132 bytes (approximately 26x larger than the maximum increase file size due to the 128 KiB header alignment).

mgduda (approved):

PR #1392, opened 2026-01-07
Bugfix/intel fpp fix

jim-p-w:
This PR addresses a problem with the intel fortran preprocessor (fpp).

In particular, the default preprocessor ordering results in a failure to expand a macro properly when it is used as an argument to another macro.
Given the following definitions:
```
#define COMMA ,
#define STREAM_DEBUG_WRITE(M) call mpas_log_write(M)
```
followed by the following invocation:
```
STREAM_DEBUG_WRITE(' -- Setting record to: $i' COMMA intArgs=(/stream % nRecords/))
```
results in the COMMA and everything after it being eliminated from the final expansion:
```
call mpas_log_write(' -- Setting record to: $i')
```

There is a switch you can provide to fpp (via the compiler) which changes the order of nexted macro expansion and gives the desire result.

I tested this by running test_model and verifying correct printouts, using both an intel make build and an intl cmake build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -663,11 +663,11 @@ intel:   # BUILDTARGET Intel oneAPI Fortran, C, and C++ compiler suite
 	"CC_SERIAL = icx" \
 	"CXX_SERIAL = icpx" \
 	"FFLAGS_PROMOTION = -real-size 64" \
-	"FFLAGS_OPT = -O3 -convert big_endian -free -align array64byte" \
+	"FFLAGS_OPT = -O3 -convert big_endian -free -align array64byte -Qoption,fpp,-macro_expand=vc" \
 	"CFLAGS_OPT = -O3" \
 	"CXXFLAGS_OPT = -O3" \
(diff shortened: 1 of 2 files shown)


pb475:
Can confirm, experienced the same issue with intel oneapi compiler.

My hack of a fix was to edit the lined using the COMMA expansion, but yours is a much better fix.

I grabbed your fix and re-compiled and ran a quick test run on my machine too. All looks good.

fmahebert:
This PR fixes my build using intel 2025.3, as reported in #1397

mgduda (approved):
This looks good to me.

@pb475 @fmahebert Thanks for your feedback on this PR!

mgduda:
@jim-p-w Before we merge this, could you modify the commit history so that your two commits have unique one-line summaries? Currently, running `git log --oneline` on the PR branch shows:
```
f476e1c63 Change the intel fortran preprocessing order.
d5b1f7d28 Change the intel fortran preprocessing order.
```

PR #1393, opened 2026-01-09
TEMPO Microphysics for NCAR MPAS

AndersJensen-NOAA:
This PR connects TEMPO microphysics to MPAS.

RELEASE NOTE:
TEMPO is the Thompson-Eidhammer Microphysics Parameterization for Operations. TEMPO lives here: https://github.com/NCAR/TEMPO and technical documentation lives here: https://ncar.github.io/TEMPO/.

TEMPO specific release notes are here: https://github.com/NCAR/TEMPO/releases/tag/tempo_v3.1.0
TEMPO version 3.1.0 is the current release version.

TEMPO is being tested in the ufs-community fork of MPAS for next-generation convection-allowing forecast systems (https://github.com/ufs-community/MPAS-Model). Thus, TEMPO is being continually evaluated for improvement.

In MPAS, TEMPO has its own registry. By default, TEMPO will be run with both aerosol- and hail-aware options set to true, which is equivalent to:
```
&physics_mp_tempo
    config_tempo_aerosolaware = .true.
    config_tempo_hailaware = .true.
```

TEMPO needs several lookups table to run:
ccn_activate.bin
 freeze_water_data_tempo_v3
 qr_acr_qg_data_tempo_v3
 qr_acr_qs_data_tempo_v3

These can be downloaded (contact the main developer), build from an executable `build_tempo_tables`


src/core_init_atmosphere/Registry.xml
```diff
@@ -386,6 +375,7 @@
                 <package name="met_stage_out" description="Active only if meteorological fields are being interpolated"/>
                 <package name="first_guess_field" description="3-d atmospheric or land-surface fields on first-guess levels"/>
                 <package name="mp_thompson_aers_in" description="initialization of GOCART-based Thompson water- and ice-friendly aerosols"/>
+                <package name="tempo_aerosolaware_in" description="initialization of GOCART-based TEMPO water- and ice-friendly aerosols"/>
```
mgduda: It looks like `tempo_aerosolaware_in` is active if and only if `mp_thompson_aers_in` is active. If this is the case, perhaps we don't need to add a new package for TEMPO? Or, if you'd prefer, we could generalize the name of the existing `mp_thompson_aers_in` package?

mgduda:
When building with Make using the `gnu` build target, there are a couple of compilation errors (perhaps because we include the `-std=f2008` flag):
```
src/module_mp_tempo_diags.F90:119:28:

  119 |       dbz(k) = max(-35._wp, 10._wp*log10((ze_rain(k)+ze_snow(k)+ze_graupel(k))*1.e18_dp))
      |                            1
Error: GNU Extension: Different type kinds at (1)
src/module_mp_tempo_diags.F90:53:59:

   53 |         re_qs(k) = max(5.01e-6_wp, min(0.5_wp*(smoc/smob), 999.e-6_wp))
      |                                                           1
Error: GNU Extension: Different type kinds at (1)
```

mgduda:
When building with Make, after the physics have all been compiled, we symbolically link physics look-up tables into the top-level MPAS-Model directory (see https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Makefile#L49-L54). Should we do this for the `ccn_activate.bin` file as well?

mgduda:
@AndersJensen-NOAA From where should users obtain the file `freeze_water_data_tempo_v3`? This doesn't seem to be present in the [`TEMPO/tables`](https://github.com/NCAR/TEMPO/tree/tempo_v3.0.4/tables) directory.

Here's the runtime error I'm seeing without this file in my run directory:
```
ERROR STOP --- file "freeze_water_data_tempo_v3" needed for TEMPO microphysics was not found.
```

src/core_atmosphere/physics/mpas_atmphys_interface.F
mgduda, line 919: When running with the `convection_permitting` suite (no TEMPO microphysics enabled), debug builds point to an out-of-bounds array access on this line:
```
At line 870 of file mpas_atmphys_interface.F
Fortran runtime error: Index '1' of dimension 2 of array 'ni' below lower bound of 22364884
```
This doesn't seem to occur with the v8.3.1 tag, and so the error seems to be specific to this PR branch.

AndersJensen-NOAA:
> @AndersJensen-NOAA From where should users obtain the file `freeze_water_data_tempo_v3`? This doesn't seem to be present in the [`TEMPO/tables`](https://github.com/NCAR/TEMPO/tree/tempo_v3.0.4/tables) directory.
> 
> Here's the runtime error I'm seeing without this file in my run directory:
> 
> ```
> ERROR STOP --- file "freeze_water_data_tempo_v3" needed for TEMPO microphysics was not found.
> ```

@mgduda you need three tables that are large-ish so they aren't stored on git at the moment.
I keep a copy of the table on noaa servers:
`wget https://gsl.noaa.gov/thredds/fileServer/retro/jensen/qr_acr_qg_data_tempo_v3`
`wget https://gsl.noaa.gov/thredds/fileServer/retro/jensen/qr_acr_qs_data_tempo_v3`
`wget https://gsl.noaa.gov/thredds/fileServer/retro/jensen/freeze_water_data_tempo_v3`

If you need to build the tables, you can compile/run the standalone build_tables program. If you are using gfortran to build the tables you will want to modify the default optimization or use MPI, otherwise the build will take a long time.
https://ncar.github.io/TEMPO/program/build_tables.html

I'm open to suggestions for better ways to deal the lookup tables. Maybe they can be automatically downoaded/linked?

src/core_atmosphere/physics/mpas_atmphys_interface.F
mgduda: Perhaps the new block of code beginning at line 705 is causing the code beginning at line 784 to no longer be executed?

src/core_init_atmosphere/Registry.xml
AndersJensen-NOAA: > It looks like `tempo_aerosolaware_in` is active if and only if `mp_thompson_aers_in` is active. If this is the case, perhaps we don't need to add a new package for TEMPO? Or, if you'd prefer, we could generalize the name of the existing `mp_thompson_aers_in` package?

I think that generalizing the name should be the way to go for now.

AndersJensen-NOAA:
> When building with Make, after the physics have all been compiled, we symbolically link physics look-up tables into the top-level MPAS-Model directory (see https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Makefile#L49-L54). Should we do this for the `ccn_activate.bin` file as well?

Yes.

mgduda:
> I'm open to suggestions for better ways to deal the lookup tables. Maybe they can be automatically downoaded/linked?

Would it be feasible to incorporate your stand-alone program for computing these tables into MPAS-A's existing [`src/core_atmosphere/utils`](https://github.com/MPAS-Dev/MPAS-Model/tree/v8.4.0/src/core_atmosphere/utils) directory? If the code is more or less static, we could just place the source file directly in the MPAS-Model repository; otherwise, we could fetch the code from a repository with `manage_externals`.

src/core_init_atmosphere/Registry.xml
mgduda: Ok, this sounds like something I'd be able to take care of with a separate PR. Do you have any suggestions for the name of the package? Maybe `microphysics_aerosols` with a description of `Initialization of GOCART-based water- and ice-friendly aerosols`?

mgduda: (Maybe extending the description to `Initialization of GOCART-based water- and ice-friendly aerosols for Thompson and TEMPO microphysics`?)

AndersJensen-NOAA:
> > I'm open to suggestions for better ways to deal the lookup tables. Maybe they can be automatically downoaded/linked?
> 
> Would it be feasible to incorporate your stand-alone program for computing these tables into MPAS-A's existing [`src/core_atmosphere/utils`](https://github.com/MPAS-Dev/MPAS-Model/tree/v8.4.0/src/core_atmosphere/utils) directory? If the code is more or less static, we could just place the source file directly in the MPAS-Model repository; otherwise, we could fetch the code from a repository with `manage_externals`.

The lookup tables are now the same for WRF/MPAS/CCPP, and I image that 99% of users won't want to change or rebuild the tables. Even though the build code is mostly static at this point, I'd like to keep the executable in the TEMPO repository so I don't have manage the program outside of TEMPO. I think the best path forward is to to either pull the lookup tables from NOAA servers or store them as git artifacts or in git LFS.

src/core_atmosphere/physics/mpas_atmphys_interface.F
AndersJensen-NOAA: @mgduda Ya, I likely messed something up with the case statements. I'll take a look.

AndersJensen-NOAA:
> When building with Make using the `gnu` build target, there are a couple of compilation errors (perhaps because we include the `-std=f2008` flag):
> 
> ```
> src/module_mp_tempo_diags.F90:119:28:
> 
>   119 |       dbz(k) = max(-35._wp, 10._wp*log10((ze_rain(k)+ze_snow(k)+ze_graupel(k))*1.e18_dp))
>       |                            1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_diags.F90:53:59:
> 
>    53 |         re_qs(k) = max(5.01e-6_wp, min(0.5_wp*(smoc/smob), 999.e-6_wp))
>       |                                                           1
> Error: GNU Extension: Different type kinds at (1)
> ```

This commit should fix the issue:
https://github.com/NCAR/TEMPO/commit/8e30fc0321fa67553bc820b8e1bf47a2ff705fb3

src/core_atmosphere/physics/mpas_atmphys_interface.F
AndersJensen-NOAA: I'm not a huge fan of the nested case-select statements. 

I rearranged the logic to call `case("mp_tempo")` after the nested part of the code so that my case would not interfere with the nested logic.

mgduda:
The changes in https://github.com/NCAR/TEMPO/commit/8e30fc0321fa67553bc820b8e1bf47a2ff705fb3 seem to fix the original compilation issues I was seeing (with `-std=f2008`), though I'm now getting a different set of errors:
```
src/module_mp_tempo_main.F90:3085:63:

 3085 |             melt_f = max(0.05_wp, min(tend%prr_gml(k)*dt/rg(k),1._wp))
      |                                                               1
Error: GNU Extension: Different type kinds at (1)
src/module_mp_tempo_main.F90:2418:53:

 2418 |               snow_dens_frac = min(1._wp, max(0._wp, rs(k)*odt / &
      |                                                     1
Error: GNU Extension: Different type kinds at (1)
src/module_mp_tempo_main.F90:1092:10:

 1092 |           min(cig(1)*oig2*qi1d(k)/am_i*lami**bm_i, max_ni/rho(k)))
      |          1
Error: GNU Extension: Different type kinds at (1)
src/module_mp_tempo_main.F90:945:64:

  945 |             min(ccg(1,nu_c)*ocg2(nu_c)*qc1d(k)/am_r*lamc**bm_r, nt_c_max/rho(k)))
      |                                                                1
Error: GNU Extension: Different type kinds at (1)
src/module_mp_tempo_main.F90:956:66:

  956 |         mvd_c(k) = max(min((3.0_wp + nu_c + 0.672_wp) * ilamc(k), d0r), d0c)
      |                                                                  1
Error: GNU Extension: Different type kinds at (1)
```

mgduda:
> > > I'm open to suggestions for better ways to deal the lookup tables. Maybe they can be automatically downoaded/linked?
> > 
> > 
> > Would it be feasible to incorporate your stand-alone program for computing these tables into MPAS-A's existing [`src/core_atmosphere/utils`](https://github.com/MPAS-Dev/MPAS-Model/tree/v8.4.0/src/core_atmosphere/utils) directory? If the code is more or less static, we could just place the source file directly in the MPAS-Model repository; otherwise, we could fetch the code from a repository with `manage_externals`.
> 
> The lookup tables are now the same for WRF/MPAS/CCPP, and I image that 99% of users won't want to change or rebuild the tables. Even though the build code is mostly static at this point, I'd like to keep the executable in the TEMPO repository so I don't have manage the program outside of TEMPO. I think the best path forward is to to either pull the lookup tables from NOAA servers or store them as git artifacts or in git LFS.

I can help with changes to download the tables. However, I think we might want to consider how to version these. What happens, for example, if the tables on the NOAA server are updated to support a newer version of TEMPO, but a user is still running an older release of MPAS-Atmosphere (with an older version of TEMPO)?

src/core_atmosphere/physics/mpas_atmphys_interface.F
mgduda: I think some further fixes may be needed. When I attempt to run with the `convection_permitting` suite (and after compiling with `DEBUG=true`), I'm seeing the following runtime error:
```
#0  0x8173c7 in __mpas_atmphys_interface_MOD_microphysics_from_mpas
	at /glade/derecho/scratch/duda/pull/MPAS-Model/src/core_atmosphere/physics/mpas_atmphys_interface.F:716
```
I think the issue may be that the `microp2_select` `case` statement in `allocate_microphysics` in `mpas_atmphys_driver_microphysics.F` is contained within the `mp_tempo` case in the `microp_select` `case` statement further up in the routine.

src/core_init_atmosphere/Registry.xml
mgduda: I've just opened PR #1449 with these changes, and I'll add you as a reviewer. One that PR has been merged, we can rebase this PR onto the HEAD of the `develop` branch.

AndersJensen-NOAA:
> The changes in [NCAR/TEMPO@8e30fc0](https://github.com/NCAR/TEMPO/commit/8e30fc0321fa67553bc820b8e1bf47a2ff705fb3) seem to fix the original compilation issues I was seeing (with `-std=f2008`), though I'm now getting a different set of errors:
> 
> ```
> src/module_mp_tempo_main.F90:3085:63:
> 
>  3085 |             melt_f = max(0.05_wp, min(tend%prr_gml(k)*dt/rg(k),1._wp))
>       |                                                               1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:2418:53:
> 
>  2418 |               snow_dens_frac = min(1._wp, max(0._wp, rs(k)*odt / &
>       |                                                     1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:1092:10:
> 
>  1092 |           min(cig(1)*oig2*qi1d(k)/am_i*lami**bm_i, max_ni/rho(k)))
>       |          1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:945:64:
> 
>   945 |             min(ccg(1,nu_c)*ocg2(nu_c)*qc1d(k)/am_r*lamc**bm_r, nt_c_max/rho(k)))
>       |                                                                1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:956:66:
> 
>   956 |         mvd_c(k) = max(min((3.0_wp + nu_c + 0.672_wp) * ilamc(k), d0r), d0c)
>       |                                                                  1
> Error: GNU Extension: Different type kinds at (1)
> ```

@mgduda I'm working out final issues with gnu debug, and came upon this issue when opening lookup tables to read:

src/module_mp_tempo_driver.F90:624:43:

  624 |       action='read', iostat=istat, convert='big_endian')
      |                                           1
Error: GNU Extension: CONVERT tag at (1)

It seems like the convert option in the open statement is not supported by f2008 (but otherwise okay). Our computers are down today, but I'll confirm later this week that this is in fact the case. Do you have any thoughts on how to get around this? I think that `convert` is only needed for the ccpp since WRF and MPAS force big_endian at compile time.

mgduda:
> @mgduda I'm working out final issues with gnu debug, and came upon this issue when opening lookup tables to read:
> 
> src/module_mp_tempo_driver.F90:624:43:
> 
> 624 | action='read', iostat=istat, convert='big_endian') | 1 Error: GNU Extension: CONVERT tag at (1)
> 
> It seems like the convert option in the open statement is not supported by f2008 (but otherwise okay). Our computers are down today, but I'll confirm later this week that this is in fact the case. Do you have any thoughts on how to get around this? I think that `convert` is only needed for the ccpp since WRF and MPAS force big_endian at compile time.

I don't think the Fortran standard provides a way to force I/O to assume a particular byte order, and so this is something that must be handled in a compiler-specific way.

There are probably better ways of going about this, but one option might be to add the `convert='big_endian'` keyword argument conditionally through pre-processing, e.g.,
```fortran
    open(unit=mp_unit, file=filename, form='unformatted', status='old', &
         action='read', iostat=istat &
#ifdef TEMPO_USE_CONVERT_ARGUMENT
         , convert='big_endian' &
#endif
         )
```
For host models whose build systems don't provide compiler-specific flags to assume big-endian I/O, you could define the `TEMPO_USE_CONVERT_ARGUMENT` macro during compilation (recognizing that this would be incompatible with strict standards compliance).

src/core_init_atmosphere/Registry.xml
mgduda: With PR #1449 now merged, we could rebase this PR branch onto the HEAD of the `develop` branch and eliminate the changes related to the `tempo_aerosolaware_in` package. This would also be a chance to resolve some of the merge conflicts in the current PR branch. If you'd like some assistance with the rebase, just let me know.

AndersJensen-NOAA: Rebase is complete. There are a few formatting changes that cropped up when I fixed merge conflicts. I will fix those formatting changes.

src/core_atmosphere/physics/mpas_atmphys_interface.F
AndersJensen-NOAA: I _think_ I have this fixed in the driver and interface. I haven't tried to run convection_permitting yet.

AndersJensen-NOAA:
> > @mgduda I'm working out final issues with gnu debug, and came upon this issue when opening lookup tables to read:
> > src/module_mp_tempo_driver.F90:624:43:
> > 624 | action='read', iostat=istat, convert='big_endian') | 1 Error: GNU Extension: CONVERT tag at (1)
> > It seems like the convert option in the open statement is not supported by f2008 (but otherwise okay). Our computers are down today, but I'll confirm later this week that this is in fact the case. Do you have any thoughts on how to get around this? I think that `convert` is only needed for the ccpp since WRF and MPAS force big_endian at compile time.
> 
> I don't think the Fortran standard provides a way to force I/O to assume a particular byte order, and so this is something that must be handled in a compiler-specific way.
> 
> There are probably better ways of going about this, but one option might be to add the `convert='big_endian'` keyword argument conditionally through pre-processing, e.g.,
> 
> ```fortran
>     open(unit=mp_unit, file=filename, form='unformatted', status='old', &
>          action='read', iostat=istat &
> #ifdef TEMPO_USE_CONVERT_ARGUMENT
>          , convert='big_endian' &
> #endif
>          )
> ```
> 
> For host models whose build systems don't provide compiler-specific flags to assume big-endian I/O, you could define the `TEMPO_USE_CONVERT_ARGUMENT` macro during compilation (recognizing that this would be incompatible with strict standards compliance).

I let AI write a block of code for the MPAS TEMPO makefile, which will use the convert= argument unless gnu/f2008 is being use: 
```
# Check if the compiler is GNU Fortran by querying its version string
IS_GNU := $(shell $(FC) --version 2>/dev/null | grep -i "GNU Fortran")

# Evaluate the logic and print statements
ifneq ($(IS_GNU),)
    # Check if -std=f2008 is present in FFLAGS
    ifneq (,$(findstring -std=f2008,$(FFLAGS)))
        TEMPO_FFLAGS += -DTEMPO_IGNORE_CONVERT_ARG
        # Prevent duplication: check if big-endian is ALREADY in FFLAGS
        ifeq (,$(findstring -fconvert=big-endian,$(FFLAGS)))
            TEMPO_FFLAGS += -fconvert=big-endian
        endif
    endif
endif
```

AndersJensen-NOAA:
> The changes in [NCAR/TEMPO@8e30fc0](https://github.com/NCAR/TEMPO/commit/8e30fc0321fa67553bc820b8e1bf47a2ff705fb3) seem to fix the original compilation issues I was seeing (with `-std=f2008`), though I'm now getting a different set of errors:
> 
> ```
> src/module_mp_tempo_main.F90:3085:63:
> 
>  3085 |             melt_f = max(0.05_wp, min(tend%prr_gml(k)*dt/rg(k),1._wp))
>       |                                                               1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:2418:53:
> 
>  2418 |               snow_dens_frac = min(1._wp, max(0._wp, rs(k)*odt / &
>       |                                                     1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:1092:10:
> 
>  1092 |           min(cig(1)*oig2*qi1d(k)/am_i*lami**bm_i, max_ni/rho(k)))
>       |          1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:945:64:
> 
>   945 |             min(ccg(1,nu_c)*ocg2(nu_c)*qc1d(k)/am_r*lamc**bm_r, nt_c_max/rho(k)))
>       |                                                                1
> Error: GNU Extension: Different type kinds at (1)
> src/module_mp_tempo_main.F90:956:66:
> 
>   956 |         mvd_c(k) = max(min((3.0_wp + nu_c + 0.672_wp) * ilamc(k), d0r), d0c)
>       |                                                                  1
> Error: GNU Extension: Different type kinds at (1)
> ```

I fixed these issues and a few more: I can now compile on our machines with `gnu DEBUG=true`

src/core_init_atmosphere/mpas_init_atm_core_interface.F
```diff
@@ -170,6 +170,9 @@ function init_atm_setup_packages(configs, streamInfo, packages, iocontext) resul
       nullify(microphysics_aerosols)
       call mpas_pool_get_package(packages, 'microphysics_aerosolsActive', microphysics_aerosols)
 
+      nullify(tempo_aerosolaware_in)
+      call mpas_pool_get_package(packages, 'tempo_aerosolaware_inActive', tempo_aerosolaware_in)
```
mgduda: I think it would be safe now to revert all changes in the `src/core_init_atmosphere/mpas_init_atm_core_interface.F` and `src/core_init_atmosphere/Registry.xml` files.

mgduda: At present, the `init_atmosphere` core fails to compile with the following error:
```
mpas_init_atm_core_interface.F:173:35:

  173 |       nullify(tempo_aerosolaware_in)
Error: Symbol 'tempo_aerosolaware_in' at (1) has no IMPLICIT type
```

AndersJensen-NOAA: Changes have been reverted -- init_atmosphere now compiles.

mgduda:
> > > @mgduda I'm working out final issues with gnu debug, and came upon this issue when opening lookup tables to read:
> > > src/module_mp_tempo_driver.F90:624:43:
> > > 624 | action='read', iostat=istat, convert='big_endian') | 1 Error: GNU Extension: CONVERT tag at (1)
> > > It seems like the convert option in the open statement is not supported by f2008 (but otherwise okay). Our computers are down today, but I'll confirm later this week that this is in fact the case. Do you have any thoughts on how to get around this? I think that `convert` is only needed for the ccpp since WRF and MPAS force big_endian at compile time.
> > 
> > 
> > I don't think the Fortran standard provides a way to force I/O to assume a particular byte order, and so this is something that must be handled in a compiler-specific way.
> > There are probably better ways of going about this, but one option might be to add the `convert='big_endian'` keyword argument conditionally through pre-processing, e.g.,
> > ```fortran
> >     open(unit=mp_unit, file=filename, form='unformatted', status='old', &
> >          action='read', iostat=istat &
> > #ifdef TEMPO_USE_CONVERT_ARGUMENT
> >          , convert='big_endian' &
> > #endif
> >          )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For host models whose build systems don't provide compiler-specific flags to assume big-endian I/O, you could define the `TEMPO_USE_CONVERT_ARGUMENT` macro during compilation (recognizing that this would be incompatible with strict standards compliance).
> 
> I let AI write a block of code for the MPAS TEMPO makefile, which will use the convert= argument unless gnu/f2008 is being use:
> 
> ```
> # Check if the compiler is GNU Fortran by querying its version string
> IS_GNU := $(shell $(FC) --version 2>/dev/null | grep -i "GNU Fortran")
> 
> # Evaluate the logic and print statements
> ifneq ($(IS_GNU),)
>     # Check if -std=f2008 is present in FFLAGS
>     ifneq (,$(findstring -std=f2008,$(FFLAGS)))
>         TEMPO_FFLAGS += -DTEMPO_IGNORE_CONVERT_ARG
>         # Prevent duplication: check if big-endian is ALREADY in FFLAGS
>         ifeq (,$(findstring -fconvert=big-endian,$(FFLAGS)))
>             TEMPO_FFLAGS += -fconvert=big-endian
>         endif
>     endif
> endif
> ```

Since the MPAS build system ensures that big-endian / byteswap flags are used for all supported compilers, it seems like we could just always define `TEMPO_IGNORE_CONVERT_ARG` in the `Makefile.mpas` (assuming this file is only used by MPAS). The AI-generated logic could be replaced with something like
```
#
# The MPAS build system ensures that compiler-specific flags are used to
# set big-endian byte order for unformatted file I/O, so the non-standard
# convert='big_endian' argument to open(...) is not necessary
#
TEMPO_FFLAGS += -DTEMPO_IGNORE_CONVERT_ARG
```

src/core_atmosphere/physics/mpas_atmphys_interface.F
mgduda: Thanks for the updates -- it looks like the `convection_permitting` suite is working now.

AndersJensen-NOAA:
> > > > I'm open to suggestions for better ways to deal the lookup tables. Maybe they can be automatically downoaded/linked?
> > > 
> > > 
> > > Would it be feasible to incorporate your stand-alone program for computing these tables into MPAS-A's existing [`src/core_atmosphere/utils`](https://github.com/MPAS-Dev/MPAS-Model/tree/v8.4.0/src/core_atmosphere/utils) directory? If the code is more or less static, we could just place the source file directly in the MPAS-Model repository; otherwise, we could fetch the code from a repository with `manage_externals`.
> > 
> > 
> > The lookup tables are now the same for WRF/MPAS/CCPP, and I image that 99% of users won't want to change or rebuild the tables. Even though the build code is mostly static at this point, I'd like to keep the executable in the TEMPO repository so I don't have manage the program outside of TEMPO. I think the best path forward is to to either pull the lookup tables from NOAA servers or store them as git artifacts or in git LFS.
> 
> I can help with changes to download the tables. However, I think we might want to consider how to version these. What happens, for example, if the tables on the NOAA server are updated to support a newer version of TEMPO, but a user is still running an older release of MPAS-Atmosphere (with an older version of TEMPO)?



> > > > I'm open to suggestions for better ways to deal the lookup tables. Maybe they can be automatically downoaded/linked?
> > > 
> > > 
> > > Would it be feasible to incorporate your stand-alone program for computing these tables into MPAS-A's existing [`src/core_atmosphere/utils`](https://github.com/MPAS-Dev/MPAS-Model/tree/v8.4.0/src/core_atmosphere/utils) directory? If the code is more or less static, we could just place the source file directly in the MPAS-Model repository; otherwise, we could fetch the code from a repository with `manage_externals`.
> > 
> > 
> > The lookup tables are now the same for WRF/MPAS/CCPP, and I image that 99% of users won't want to change or rebuild the tables. Even though the build code is mostly static at this point, I'd like to keep the executable in the TEMPO repository so I don't have manage the program outside of TEMPO. I think the best path forward is to to either pull the lookup tables from NOAA servers or store them as git artifacts or in git LFS.
> 
> I can help with changes to download the tables. However, I think we might want to consider how to version these. What happens, for example, if the tables on the NOAA server are updated to support a newer version of TEMPO, but a user is still running an older release of MPAS-Atmosphere (with an older version of TEMPO)?

I don't have a good way to version tables off the top of my head. Currently the tables all have "_v3" at the end of the filenames, and my plan is to ensure that these tables will always work with Tempo v3.x.x. If tables are updated but the user is still running an older release then either the code will crash if the table sizes are different from what the code expects or results will be answer changing. We can keep old tables on servers, but we'll still need a way to version/communicate changes to users.

mgduda:
@AndersJensen-NOAA Apologies for the delay in returning to this PR. From my notes, here are some of the items that we may still need to address:

- [x] We need to link the `ccn_activate.bin` file into the main MPAS-Model directory after compilation (as we do with other physics tables), and we should update the [`setup_run_dir.py`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/testing_and_setup/atmosphere/setup_run_dir.py) script to handle `ccn_activate.bin` as well. **I can help with this, if you don't mind me pushing commits to your PR branch.**
- [ ] We need to download the TEMPO tables and add them to the main MPAS-Model directory as well. As with the `ccn_activate.bin` file, we should update the `setup_run_dir.py` script to link these downloaded tables into any run directories that the user creates. **I can help with this as well.**
- [x] For the `nvfortran` compiler, we need to add `-I./physics/physics_noaa/TEMPO` to the compilation rules in `src/core_atmosphere/Makefile`. **I can help with this as well.**
- [x] Also for the `nvfortran` compiler, we need to add `-I../physics/physics_noaa/TEMPO` to the compilation rules in the `src/core_atmosphere/dynamics/Makefile` file. **I can help with this as well.**
- [ ] I think it would be worth considering changes to TEMPO so that its code doesn't directly abort if look-up tables are not found. As it is, no error messages are written to MPAS-A log files, and this may make it harder for users to determine why their runs crash if tables are not present.
- [ ] Would you consider a switch to turn off any writes to stdout/stderr? We typically try to avoid this behavior in MPAS.

mgduda:
@AndersJensen-NOAA Would you like me to proceed to help with the first four items in the list from my previous comment? I'd need to push changes to your PR branch, so I'd like to first make sure you're fine with me doing so.

AndersJensen-NOAA:
> @AndersJensen-NOAA Would you like me to proceed to help with the first four items in the list from my previous comment? I'd need to push changes to your PR branch, so I'd like to first make sure you're fine with me doing so.

@mgduda go for it.

For planning purposes, I'll eventually merge these changes into the tempo 3.1.0 release branch that will be the final version for MPAS and CCPP (and WRF if they allow another small TEMPO update).

https://github.com/NCAR/TEMPO/tree/tempo_3.1.0

mgduda:
> > @AndersJensen-NOAA Would you like me to proceed to help with the first four items in the list from my previous comment? I'd need to push changes to your PR branch, so I'd like to first make sure you're fine with me doing so.
> 
> @mgduda go for it.
> 
It looks like I don't have permission to push to your fork:
```
ERROR: Permission to AndersJensen-NOAA/MPAS-Model.git denied to mgduda.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
I think you may need to add me as a collaborator on your fork?

AndersJensen-NOAA:
> > > @AndersJensen-NOAA Would you like me to proceed to help with the first four items in the list from my previous comment? I'd need to push changes to your PR branch, so I'd like to first make sure you're fine with me doing so.
> > 
> > 
> > @mgduda go for it.
> 
> It looks like I don't have permission to push to your fork:
> 
> ```
> ERROR: Permission to AndersJensen-NOAA/MPAS-Model.git denied to mgduda.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.
> ```
> 
> I think you may need to add me as a collaborator on your fork?

Now it should work (after you accept the invitation).

mgduda:
> Now it should work (after you accept the invitation).

Success!

mgduda:
The table files `freeze_water_data_tempo_v3`, `qr_acr_qs_data_tempo_v3`, and `qr_acr_qg_data_tempo_v3` can take a while to download (I recorded 6m 41s on a Derecho login node this afternoon), and so downloading these when MPAS-A is compiled might not be the best option.

@AndersJensen-NOAA What do you think about providing an error message in the MPAS log file when these files are not found suggesting that the user either (1) download them manually (perhaps using a shell script that would could provide) or (2) compute them using the `tempo_build_tables` tool?

As a follow-on, what do you think about think about the possibility of compiling an executable named `tempo_build_tables` using the code in `TEMPO/tools` as part of the regular MPAS-A build process? We do this already for a `build_tables` tool for the Thompson microphysics.

AndersJensen-NOAA:
@mgduda I'm okay with adding an error message. The largest table (rain-graupel collection) is about 650 MB, so yes it can take a few minutes to download. 

The only concern with building the tempo tables is that they can take a long time to build (~1 hr) unless compiled with mpi. And I have only tested the build with intel and gnu compilers. 

But, I'm not opposed to letting MPAS control compilation of the table build script. Could you help with setting up that code?

mgduda:
> @mgduda I'm okay with adding an error message. The largest table (rain-graupel collection) is about 650 MB, so yes it can take a few minutes to download.
> 
> The only concern with building the tempo tables is that they can take a long time to build (~1 hr) unless compiled with mpi. And I have only tested the build with intel and gnu compilers.
> 
> But, I'm not opposed to letting MPAS control compilation of the table build script. Could you help with setting up that code?

Sure, I can push some changes to build the `tempo_build_tables` tool. I can also test the code (at least, see if it runs) with the `nvhpc` compilers.

mgduda:
Would it be possible to add the `TEMPO_IGNORE_CONVERT_ARG` logic to `module_mp_tempo_tables.F90` to avoid compilation errors with the `gfortran` compiler with `-std=f2008`?
```
../physics/physics_noaa/TEMPO/src/module_mp_tempo_tables.F90:414:28:

  414 |       iostat=istat, convert='big_endian')
      |                            1
Error: GNU Extension: CONVERT tag at (1)
../physics/physics_noaa/TEMPO/src/module_mp_tempo_tables.F90:281:28:

  281 |       iostat=istat, convert='big_endian')
      |                            1
Error: GNU Extension: CONVERT tag at (1)
../physics/physics_noaa/TEMPO/src/module_mp_tempo_tables.F90:112:28:

  112 |       iostat=istat, convert='big_endian')
      |                            1
Error: GNU Extension: CONVERT tag at (1)
```

AndersJensen-NOAA:
@mgduda changes added here: https://github.com/NCAR/TEMPO/tree/tempo_3.1.0

mgduda:
@AndersJensen-NOAA I think the code changes are in generally good shape. Whenever you've got a TEMPO 3.1.0 tag, could you update the `Externals.cfg` file to point to that tag?

It also looks like some recent changes in the `develop` branch have resulted in merge conflicts with this PR branch. One option might be to resolve those conflicts when merging the HEAD of the `develop` branch into this PR branch. Alternatively (and my preference), we could rebase this PR branch on the HEAD of the `develop` branch and resolve conflicts in the process, and we could also take the opportunity during the rebase to clean up the commit history.

Finally, could you add some details to the PR description? What new namelist options are available? What new settings for existing namelist options are available? What new input fields or files are needed when running with the TEMPO microphysics?

AndersJensen-NOAA:
@mgduda I'll plan on having a 3.1.0 tag next week, and then I can update Externals? What's the ETA for the 8.5 release?

I can rebase the code. There may be a bit of work since 8.4 has a feature to set the top-level used for microphysics.

And yes, I'll add to the PR description and also have some information in the Tempo repo as well related to the release tag. Thanks for the help getting this in shape.

mgduda:
> @mgduda I'll plan on having a 3.1.0 tag next week, and then I can update Externals? What's the ETA for the 8.5 release?

Perhaps naively, I'm still hoping to get v8.5.0 out by the MPAS/WRF workshop, given that the focus is really just on the inclusion of several new physics schemes.

AndersJensen-NOAA:
> @AndersJensen-NOAA Apologies for the delay in returning to this PR. From my notes, here are some of the items that we may still need to address:
> 
> * [x]  We need to link the `ccn_activate.bin` file into the main MPAS-Model directory after compilation (as we do with other physics tables), and we should update the [`setup_run_dir.py`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/testing_and_setup/atmosphere/setup_run_dir.py) script to handle `ccn_activate.bin` as well. **I can help with this, if you don't mind me pushing commits to your PR branch.**
> * [ ]  We need to download the TEMPO tables and add them to the main MPAS-Model directory as well. As with the `ccn_activate.bin` file, we should update the `setup_run_dir.py` script to link these downloaded tables into any run directories that the user creates. **I can help with this as well.**
> * [x]  For the `nvfortran` compiler, we need to add `-I./physics/physics_noaa/TEMPO` to the compilation rules in `src/core_atmosphere/Makefile`. **I can help with this as well.**
> * [x]  Also for the `nvfortran` compiler, we need to add `-I../physics/physics_noaa/TEMPO` to the compilation rules in the `src/core_atmosphere/dynamics/Makefile` file. **I can help with this as well.**
> * [ ]  I think it would be worth considering changes to TEMPO so that its code doesn't directly abort if look-up tables are not found. As it is, no error messages are written to MPAS-A log files, and this may make it harder for users to determine why their runs crash if tables are not present.
> * [ ]  Would you consider a switch to turn off any writes to stdout/stderr? We typically try to avoid this behavior in MPAS.

@mgduda 
I updated to TEMPO v3.1.0 today and updated Externals

As part of TEMPO v3.1.0, the error stops are gone and the stdout write statements are wrapped around a `verbose` logical that is set to false by default.

I also rebased onto develop

The code no longer stops when the lookup tables are not present. 

I added the hookup to reflectivity

What else needs needs to be done to get this across the finish line?

mgduda:
If the `ccn_activate.bin` file is not in the run directory, the TEMPO code crashes on [line 645 of module_mp_tempo_driver.F90](https://github.com/NCAR/TEMPO/blob/tempo_v3.1.0/src/module_mp_tempo_driver.F90#L645). I think it would be great if we could find a way to notify users of missing tables through error logs -- preferably with calls to `mpas_log_write(..., messageType=MPAS_LOG_ERR)` so that `log.atmosphere.XXXX.err` files are created.

src/core_atmosphere/Registry.xml
```diff
@@ -1704,43 +1707,51 @@
...
-                             packages="mp_thompson_in;mp_thompson_aers_in"/>
+                             packages="mp_thompson_in;mp_thompson_aers_in;mp_tempo_in"/>
+
+                        <var name="ng" array_group="number" units="nb kg^{-1}"
+                             description="Graupel number concentration"
+                             packages="tempo_hailaware_in"/>
+
+                        <var name="volg" array_group="volume" units="m^{3} kg^{-1} or L kg^{-1}"
+                             description="Graupel volume"
+                             packages="tempo_hailaware_in"/>
```
mgduda: Rather than placing `ng` (with `array_group="number"`) and `volg` (with `array_group="volume"`) in the middle of the list of existing number concentrations, I would propose placing `ng` after `nc` (i.e., at the end of the sequence of hydrometeor number concentrations) and placing `volg` just before `tke` (since `volg` is in a different `array_group`). The same order would be needed in the `scalars_tend` and `lbc_scalars` variable arrays.

mgduda (changes requested):

src/core_atmosphere/physics/Registry_tempo.xml
```diff
@@ -0,0 +1,17 @@
...
+<!-- CONFIGURATION OPTIONS AND ARRAYS SPECIFIC TO TEMPO:                                                       -->
+<!-- ========================================================================================================= -->
+
+  <nml_record name="physics_mp_tempo" in_defaults="true">
+    <nml_option name="config_tempo_aerosolaware" type="logical" default_value="true" in_defaults="false"
+      units="-"
+      description="Logical flag to turn on/off prognostic cloud droplet and aerosol number concentrations"
+      possible_values=".true. or .false."/>
+
+    <nml_option name="config_tempo_hailaware" type="logical" default_value="true" in_defaults="false"
```
mgduda, line 11: The PR description states that both `config_tempo_aerosolaware` and `config_tempo_hailaware` are `true` by default; however, they are given a default value of `false` here. I'd suggest either updating the PR description, or updating the default values for these options in `Registry_tempo.xml`.

mgduda (changes requested):

src/core_atmosphere/physics/Registry_tempo.xml
mgduda: Sorry... I was looking at the `in_defaults` attribute here. Everything seems to be in agreement.

src/core_atmosphere/Registry.xml
AndersJensen-NOAA: @mgduda This should be fixed now

AndersJensen-NOAA:
> If the `ccn_activate.bin` file is not in the run directory, the TEMPO code crashes on [line 645 of module_mp_tempo_driver.F90](https://github.com/NCAR/TEMPO/blob/tempo_v3.1.0/src/module_mp_tempo_driver.F90#L645). I think it would be great if we could find a way to notify users of missing tables through error logs -- preferably with calls to `mpas_log_write(..., messageType=MPAS_LOG_ERR)` so that `log.atmosphere.XXXX.err` files are created.

@mgduda Generalized logging is on the to-do list. Is this a need-to-have before this gets approved?

mgduda:
> > If the `ccn_activate.bin` file is not in the run directory, the TEMPO code crashes on [line 645 of module_mp_tempo_driver.F90](https://github.com/NCAR/TEMPO/blob/tempo_v3.1.0/src/module_mp_tempo_driver.F90#L645). I think it would be great if we could find a way to notify users of missing tables through error logs -- preferably with calls to `mpas_log_write(..., messageType=MPAS_LOG_ERR)` so that `log.atmosphere.XXXX.err` files are created.
> 
> @mgduda Generalized logging is on the to-do list. Is this a need-to-have before this gets approved?

I think it's more or less guaranteed that there will be people who will forget to link the `ccn_activate.bin` and other tables into their run directories, and so having something in the job output to let them know why their run failed would be important. I appreciate that generalized logging isn't an easy task, so for now, perhaps it's better to turn the stdout/stderr message back on in case tables are missing?

Alternatively, as a potential option of intermediate difficulty, would it be easy enough to add an optional return status code to the argument lists of the `read_table_*` routines, and to propagate that status code out through another optional argument in the `tempo_init` routine? We could then check that in the MPAS-A `init_microphysics` routine and write a more general error message to the log files stating that at least one of the TEMPO tables `ccn_activate.bin`, etc. could not be read.

AndersJensen-NOAA:
> > > If the `ccn_activate.bin` file is not in the run directory, the TEMPO code crashes on [line 645 of module_mp_tempo_driver.F90](https://github.com/NCAR/TEMPO/blob/tempo_v3.1.0/src/module_mp_tempo_driver.F90#L645). I think it would be great if we could find a way to notify users of missing tables through error logs -- preferably with calls to `mpas_log_write(..., messageType=MPAS_LOG_ERR)` so that `log.atmosphere.XXXX.err` files are created.
> > 
> > 
> > @mgduda Generalized logging is on the to-do list. Is this a need-to-have before this gets approved?
> 
> I think it's more or less guaranteed that there will be people who will forget to link the `ccn_activate.bin` and other tables into their run directories, and so having something in the job output to let them know why their run failed would be important. I appreciate that generalized logging isn't an easy task, so for now, perhaps it's better to turn the stdout/stderr message back on in case tables are missing?
> 
> Alternatively, as a potential option of intermediate difficulty, would it be easy enough to add an optional return status code to the argument lists of the `read_table_*` routines, and to propagate that status code out through another optional argument in the `tempo_init` routine? We could then check that in the MPAS-A `init_microphysics` routine and write a more general error message to the log files stating that at least one of the TEMPO tables `ccn_activate.bin`, etc. could not be read.

@mgduda Please take a peek at this PR. This was coded by AI. This should propagate missing file errors out of tempo init. We'll need to add the connection to the microphysics driver in mpas. 
https://github.com/NCAR/TEMPO/pull/55/changes

PR #1394, opened 2026-01-09
mpas_init_atmosphere mesh connectivity out-of-bounds bug

pb475:
Found an issue when running ./mpas_init_atmosphere, getting a SIGSEGV when processing the x1.40962 mesh, I'd previously been using x1.2562 mesh for development and it was fine. This is using the official meshes provided https://www2.mmm.ucar.edu/projects/mpas/site/downloads/meshes.html. 

Narrowed it down to the subroutine atm_initialize_advection_rk / deformation-weight initialization). The mesh itself is valid, but some connectivity arrays (e.g., edgesOnCell/cellsOnCell) are padded with sentinel zeros by design, and under some MPI decompositions/halo configurations a halo cell can present incomplete connectivity where the code path assumes valid edge/vertex IDs. When a 0 or out-of-range ID is used as a 1-based Fortran index (e.g., indexing cellsOnEdge(:,iEdge), verticesOnEdge(:,iEdge), or xVertex(iv)), the model can dereference out of bounds and crash.

As this is using the official mesh provided, it's probably a unique compiler/machine issue which must not be triggering on other machines, which would explain why this hasn't been picked up on before. Thought it was simply a memory issue initially, but logs I added during testing are consistent with out-of-bounds access.

The fix does the following: 
- add bounds checks immediately before dereferencing edge and vertex connectivity in src/core_init_atmosphere/mpas_atm_advection.F 
  - Eseentially a check which validates iEdge = edgesOnCell(i,iCell) is within [1,nEdges] and that verticesOnEdge(:,iEdge) yields vertex IDs within [1,nVertices] before indexing xVertex/yVertex/zVertex or cellsOnEdge. If any required connectivity is invalid for a cell, we skip computing that cell’s stencil/weights. This matches the existing intent of these initialization routines (already avoiding forming polynomial fits when stencil reaches outside of the available halo via do_the_cell gating). 

After re-compiling, segfault no longer occurs and I'm able to run ./mpas_init_atmosphere successfully with sensible results.

### DIFF
--- a/src/core_init_atmosphere/mpas_atm_advection.F
+++ b/src/core_init_atmosphere/mpas_atm_advection.F
@@ -51,6 +51,7 @@ module atm_advection
       real (kind=RKIND) :: thetae_tmp
       real (kind=RKIND) :: xv1, xv2, yv1, yv2, zv1, zv2
       integer :: i, j, k, ip1, ip2, n
+      integer :: iv1, iv2
       integer :: iCell, iEdge
       real (kind=RKIND) :: pii
       real (kind=RKIND), dimension(25) :: xp, yp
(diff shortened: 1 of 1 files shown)


pb475:
the mesh I was using turned out to be the issue 🙃. Hence the closure.

Issue #1395, opened 2026-01-12
mpas_atmphys_interface.F indexing issue for evapprod_p assignment


jamesopinto:
There appears to be an inadvertent duplication of the k index in the assignment of evapprod_p on line 652:
                         rainprod_p(i,k,j) = rainprod(k,i)
         line 652:  evapprod_p(i,k,j) = evapprod(k,k)
                         recloud_p(i,k,j)  = re_cloud(k,i)
                         reice_p(i,k,j)    = re_ice(k,i)
                         resnow_p(i,k,j)   = re_snow(k,i)

jihyeonjang:
Thank you for catching this. evapprod(k,k) looks like a typo and should be evapprod(k,i). 

I'd like to note that evapprod is a diagnostic/output variable, so this should not affect the results. 

I will open a PR with the one-line fix. Thank you!

jihyeonjang:
Closed by merge of #1396

PR #1396, opened 2026-01-12
Fix evapprod indexing typo

jihyeonjang:
This PR fixes a typo in src/core_atmosphere/physics/mpas_atmphys_interface.F where evapprod(k,k) was used instead of evapprod(k,i) (introduced in [8e0057689d](https://github.com/MPAS-Dev/MPAS-Model/commit/8e0057689d)). 
This issue was reported in [Issue #1395](https://github.com/MPAS-Dev/MPAS-Model/issues/1395).

This corrects the diagnostic output and should not affect model results. 

I based the fix on the commit where the typo was introduced, so it can be applied to older MPAS versions without bringing in unrelated changes.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_interface.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_interface.F
@@ -573,7 +573,7 @@
           qg_p(i,k,j) = qg(k,i)
 
           rainprod_p(i,k,j) = rainprod(k,i)
-          evapprod_p(i,k,j) = evapprod(k,k)
+          evapprod_p(i,k,j) = evapprod(k,i)
           recloud_p(i,k,j)  = re_cloud(k,i)
           reice_p(i,k,j)    = re_ice(k,i)
           resnow_p(i,k,j)   = re_snow(k,i)


mgduda:
@jihyeonjang The fix looks good to me. Would you be willing to add more detail to the commit message (the PR description is fine)? I think it's helpful when looking back through the commit log years after commits were made to be able to get a better idea of exactly what was changed in a commit based on the log entry. The information from the PR description would probably work well if integrated into the commit message. If you have any questions about amending a commit, just let me know and I'll be glad to help.

jihyeonjang:
@mgduda Thank you for reviewing this PR. I added the commit message, similar to the PR description. Please let me know if anything should be revised. Thanks!

mgduda (approved):
Thanks so much for reworking the commit message -- this looks great!

mgduda:
@jihyeonjang Actually, although the commit message suggests the fix is based on the commit where the typo was introduced (i.e., 8e0057689d), it looks like this PR branch is actually based on 447de92d4e . If you don't object, I can rebase this PR branch onto 8e0057689d before merging.

jihyeonjang:
Sure. That would be better. Thank you for checking the details.

mgduda:
> Sure. That would be better. Thank you for checking the details.

After rebasing the changes in this branch onto 8e0057689d, I think the commit message should still be correct.

PR #1402, opened 2026-01-29
Enable temperature output in history files

jihyeonjang:
This PR enables the "temperature" variable in the "history" output.

By adding "temperature" to stream_list.atmosphere.output, the history files will include temperature(Time, nCells, nVertLevels).

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -1767,8 +1767,7 @@
                      description="Relative humidity"/>
 
                 <var name="spechum" type="real" dimensions="nVertLevels nCells Time" units="kg kg^{-1}"
-                     description="Specific humidity"
-                     packages="jedi_da"/>
+                     description="Specific humidity"/>
 
                 <var name="v" type="real" dimensions="nVertLevels nEdges Time" units="m s^{-1}"
                      description="Horizontal tangential velocity at edges"/>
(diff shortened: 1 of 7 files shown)


byoung-joo:
This should be very useful, which can provide a direct access to the "temperature" variable.
I wonder if we can add the same thing to `subroutine atm_compute_output_diagnostics` for `spechum`, but with a `jedi_da` package maybe.

jihyeonjang:
@mgduda -- what do you think? I think this is doable, but we should confirm the variables are updated each time step so the output reflects the correct values.  

@byoung-joo -- for DA runs, is it OK to remove "packages="jedi_da"" from the "temperature" variable definition in Registry.xml?

byoung-joo:
>  I think this is doable, but we should confirm the variables are updated each time step so the output reflects the correct values.

Yes, the added computation at each time step is a side effect. I wonder if we can add `spechum` diagnostic in other place... perhaps somewhere from the diagnostic directory (again with `jedi_da` package)? 
 
> for DA runs, is it OK to remove "packages="jedi_da"" from the "temperature" variable definition in Registry.xml?

Then, the variable `temperature` would be always available, so it should not affect the DA application (I think). Let me know if I misunderstand about this.

jihyeonjang:
> Yes, the added computation at each time step is a side effect. I wonder if we can add spechum diagnostic in other place... perhaps somewhere from the diagnostic directory (again with jedi_da package)?

I agree that computing these every time step is unnecessary for output-only variables. I removed the changes from subroutine `atm_compute_output_diagnostics` and instead compute `temperature` and `spechum` in the diagnostics module. So they are only computed when written to output (no timestep overhead). 

Also, `temperature` and `spechum` are no longer with `packages="jedi_da"`, so they are available for diagnostics output.  

Thus, with this PR, `temperature` and `spechum` will be written to output when added to the stream_list file.

PR #1404, opened 2026-02-03
Initial LES options for MPAS-Atmosphere

mgduda:
This PR introduces an initial Large-Eddy Simulation (LES) capability with two sub-grid turbulence models:
 - a diagnostic TKE scheme based on a 3-d Smagorinsky formulation
 - a prognostic 1.5-order TKE scheme

The LES capability is controlled by the new namelist option `config_les_model`, which may be set to either `'none'` (default), `'3d_smagorinsky'`, or `'prognostic_1.5_order'`.

Note that if an LES option is enabled, a PBL parameterization scheme should not be used -- that is, `config_pbl_scheme` should be set to `off` in the `&physics` namelist group.

Other namelist parameter associated with the LES models include:
 - `config_les_surface` (default value: `none`):
   - `'specified'` uses fixed values from `config_surface_*` options
   - `'varying'` uses inputs from physics surface heat flux, moisture flux and friction velocity
 - `config_surface_heat_flux` (K m s-1; default value: 0.0) specifies w'θ' at surface
 - `config_surface_moisture_flux` (kg/kg m s-1; default value: 0.0) specifies w'q' at surface
 - `config_surface_drag_coefficient` (unitless; default value: 0.0) Cd defined from lowest level V such that surface stress = rho*Cd*V^2

Application of LES mixing, `2d_smagorinsky` horizontal mixing, and the background 4th-order filter can be enabled for scalar variables (qv, etc.) by setting
```
config_mix_scalars = true (default value: false)
```
in the `&nhyd_model` namelist group. Previously, the `2d_smagorinsky` horizontal mixing and the background 4th-order filtering were not applied to scalar variables.

As of this PR, the default configuration for filtering is unchanged: `config_horiz_mixing = '2d_smagorinsky'` and a 4th-order horizontal background filter are active for the dry dynamics variables (`u`, `w`, `theta_m`).

Along with the refactoring of the dynamics to more cleanly support the new sub-grid turbulence models, five new static fields have been introduced:
 - `deformation_coef_c2`
 - `deformation_coef_s2`
 - `deformation_coef_cs`
 - `deformation_coef_c`
 - `deformation_coef_s`

These new fields must be present in "static", "invariant", initial conditions, and restart files. As such, new input files will generally need to be created, regardless of whether any LES options are used in a simulation.


src/core_init_atmosphere/mpas_atm_advection.F
```diff
@@ -772,19 +777,32 @@ subroutine atm_initialize_deformation_weights( mesh, nCells, on_a_sphere, sphere
       integer :: iCell
       real (kind=RKIND) :: pii
       real (kind=RKIND), dimension(25) :: xp, yp
+      real (kind=RKIND) :: xe, ye
+
```
abishekg7: Is the extra newline needed here?

src/core_init_atmosphere/mpas_atm_advection.F
```diff
@@ -927,15 +992,242 @@ subroutine atm_initialize_deformation_weights( mesh, nCells, on_a_sphere, sphere
...
+                                x_period, y_period,                       &
+                                xEdge, yEdge, zEdge,                      &
+                                xCell, yCell, zCell, nCells,              &
+                                angleEdge, nEdgesOnCell, edgesOnCell )
+
+
    end subroutine atm_initialize_deformation_weights
 
- end module atm_advection
+!-----------------------
```
abishekg7: This line could be removed

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -6184,153 +6231,76 @@ subroutine init_atm_case_lbc(timestamp, block, mesh, nCells, nEdges, nVertLevels
...
-   !
-   !-----------------------------------------------------------------------
-   subroutine init_atm_case_cam_mpas(stream_manager, dminfo, block, mesh, &
-                                     dims, configs, nVertLevels)
-
-      use mpas_dmpar, only : mpas_dmpar_exch_halo_field, mpas_dmpar_min_real, mpas_dmpar_max_real
-      use mpas_stream_manager, only : MPAS_stream_mgr_stream_exists, MPAS_stream_mgr_read
-      use mpas_derived_types, only : MPAS_STREAM_MGR_NOERR
-      use mpas_vector_operations, only : mpas_initialize_vectors
+   subroutine init_atm_case_les(dminfo, mesh, fg, nCells, nVertLevels, state, diag, test_case, configs)
```
abishekg7, line 6210: This subroutine might need a comment block above to make it consistent with the remaining subroutines in the file

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -6345,55 +6315,810 @@ subroutine init_atm_case_cam_mpas(stream_manager, dminfo, block, mesh, &
...
+        else
+          atm_get_sounding = max(7.8/1000. - (height-442.5)*0.004/1000.,0.0)
+        end if
+
+      end if 
+
+   end function atm_get_sounding
+
+
+   real (kind=RKIND) function atm_get_sounding_1( variable, height )
```
abishekg7, line 6827: Wondering if there could be more descriptive names for these three `atm_get_sounding` functions, or perhaps some comments that briefly describe it.

src/core_atmosphere/dynamics/mpas_atm_dissipation_models.F
```diff
@@ -0,0 +1,1643 @@
...
+
+      if (trim(config_les_surface) == 'none') then
+         les_surface_opt = LES_SURFACE_NONE
+      else if (trim(config_les_surface) == 'specified') then
+         les_surface_opt = LES_SURFACE_SPECIFIED
+      else if (trim(config_les_surface) == 'varying') then
+         les_surface_opt = LES_SURFACE_VARYING
+      else
+         ! Error
+      end if
```
abishekg7: Both of these if clauses seem to be repeated 3-4 times in this file. Wondering if we could move them to a function, and call it once from an appropriate place in `mpas_atm_time_integration.F`

src/core_init_atmosphere/mpas_atm_advection.F
mgduda: I've removed this newline.

src/core_init_atmosphere/mpas_atm_advection.F
mgduda: I've removed this comment line.

src/core_atmosphere/dynamics/mpas_atm_dissipation_models.F
mgduda: I've pushed changes to factor all of this logic out into two new functions. Namelist strings are now converted to integer parameters in `atm_compute_dyn_tend` using these new functions, and all logic further down in the call tree uses integer logic for the LES sub-grid model and surface options.

src/core_init_atmosphere/mpas_init_atm_cases.F
mgduda: This isn't totally inconsistent with the subroutines to handle the setup of other cases; see, e.g., [init_atm_case_gfs](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.3.1/src/core_init_atmosphere/mpas_init_atm_cases.F#L2633-L2636) . I'd suggest we keep the comment for this subroutine as is for this PR.

src/core_init_atmosphere/mpas_init_atm_cases.F
mgduda: I think we'll need input from either @skamaroc or @dudhia regarding these LES sounding routines.

abishekg7 (approved):
Looks good to me. Tested that it's bit identical (with appropriate flags) between CPU and GPU runs for various LES code paths as well as non-LES code path. 

In addition, checked that the non-LES code paths don't differ from the hotfix-v8.3.1 branch, provided that the updated initial conditions are used in both runs.

dudhia:
My LES test of the current PR code looks fine.

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -1859,6 +1889,18 @@ subroutine init_atm_case_squall_line(dminfo, mesh, nCells, nVertLevels, state, d
...
+!
+!     call mpas_log_write('index_qv $i ', intArgs=(/index_qv/))
+!     call mpas_log_write('index_tke $i ', intArgs=(/index_tke/))
+      scalars(index_tke,:,:) = 0.
+
+      do k = 1,nz1  
+          do i=1,nCells
+             scalars(index_tke,k,i) = 0.1
+          end do
+      end do
```
mgduda: @dudhia This PR contains several changes to the squall line and mountain wave test cases. Should these changes be reverted (i.e., should this PR not modify the setup for non-LES test cases)?

src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -1572,7 +1601,8 @@ subroutine init_atm_case_squall_line(dminfo, mesh, nCells, nVertLevels, state, d
       !     metrics for hybrid coordinate and vertical stretching
 
       str = 1.0
-      zt = 20000.
+!      zt = 20000.
+      zt = config_ztop
```
dudhia: This one removes the hardwired value and forces it to use the namelist. I think it is OK.

mgduda: This is just one of the various changes that have been added to the squall line test case setup, and in fact, this could modify the behavior when running this test case with our prepared tar files (where I think we may omit `config_ztop` from the namelist because we assumed that the test case would force the use of a specific value).

I would propose we make these sort of changes in one or more separate PRs.

src/core_init_atmosphere/mpas_init_atm_cases.F
dudhia: This is needed in case someone uses a tke turbulence option, because with no seed, turbulence won't develop. Perhaps it would be case independent and I am not sure what happens if a non-tke option is chosen here.

src/core_init_atmosphere/mpas_init_atm_cases.F
dudhia: I think the actual namelist had 30000 which confused me at first.

dudhia: correction: the registry default is 30000. The code hardwires 20000. It's a fix that can be done separately from this PR.

mgduda: For this and other changes to the squall-line and mountain-wave test cases, I think we'll plan to make separate PRs. That way, this PR can focus on only what is essential to the initial LES capability and a supporting test case (case 10).

mgduda:
@dudhia Would you be able to take a look at the PR description, which is an abbreviated version of the draft user's guide material?

dudhia:
> @dudhia Would you be able to take a look at the PR description, which is an abbreviated version of the draft user's guide material?

This looks OK. You might mention there should be no PBL option chosen to enable LES. I am not sure whether the code prevents that anyway. In WRF when the PBL option is zero, vertical diffusion is called instead.

abishekg7:
@mgduda Could you also add a note to the description about the new fields introduced to the init file, and the need to re-generate the init/input files.

src/core_atmosphere/dynamics/Makefile
```diff
@@ -1,11 +1,12 @@
 .SUFFIXES: .F .o
 
 OBJS = mpas_atm_time_integration.o \
-       mpas_atm_boundaries.o
+       mpas_atm_boundaries.o \
+       mpas_atm_dissipation_models.o
```
jim-p-w, line 5: We will need a corresponding change to [src/core_atmosphere/CMakelists.txt](https://github.com/MPAS-Dev/MPAS-Model/blob/b9090a1434b8ceb2a92fc985f4d97e7947cd95a6/src/core_atmosphere/CMakeLists.txt#L337)
Add `mpas_atm_dissipation_models.F` to `ATMOSPHERE_CORE_DYNAMICS_SOURCES`

I was able to build mpas-bundle with that change.

mgduda: @jim-p-w Thanks for testing out the CMake build! I'll update the `CMakeLists.txt` in `src/core_atmosphere`.

PR #1417, opened 2026-03-10
Add MPAS Atmosphere NUOPC build

danrosen25:
# Overview
The MPAS Atmosphere NUOPC cap infrastructure is designed to build an MPAS NUOPC library and Fortran module file for use within a NUOPC driven system. The initial work is an MPAS Atmosphere NUOPC cap "stub", which does not execute the MPAS Atmosphere. The stub provides entry point for initializing, running, and finalizing the future MPAS Atmosphere NUOPC model. Follow on work will include execution of the MPAS Atmosphere model from within the NUOPC cap along with adding coupling fields to the import and export states.

Testing infrastructure is implemented on a separate branch: https://github.com/NCAR/MPAS-Model-NUOPC-dev/tree/nuopc/testing

## Configuration options
**MPAS_PREFIX**: Installation prefix used to determine location for `lib` and `mod` installation directories
**MPAS_LIBDIR**: Full path to the `lib` installation directory - location for installation MPAS NUOPC library
**MPAS_MODDIR**: Full path to the `mod` installation directory - location for installation of MPAS NUOPC Fortran module files
**NUOPC**: Generate NUOPC library for the desired CORE. Options are `true` | `false` (default)

## Testing
Checkout the testing work from this branch
https://github.com/NCAR/MPAS-Model-NUOPC-dev/tree/nuopc/testing

Testing requires a system and environment capable of building and running MPAS-A with these additional requirements

- cmake
- an MPI implementation with mpirun included in users PATH
- ESMF v8.5+ (which includes ESMX building infrastructure)
- python

```bash
cd testing_and_setup/nuopc/atmosphere
./mpas_nuopc_atm_test.py
```

For detailed instructions see the README
https://github.com/NCAR/MPAS-Model-NUOPC-dev/tree/nuopc/testing/testing_and_setup/nuopc/atmosphere


danrosen25:
@mgduda @scrasmussen @uturuncoglu @anntsay

I've drafted a PR for the NUOPC cap infrastructure. I'd like for you to take a look before I mark this "ready" for review, in case you have any major revisions. I've rebased the work so that it builds on the MPAS_ESMF="external" work recently merged. I've also split the work into two commits, one for the NUOPC build infrastructure and stub cap, and a second one for the regression testing infrastructure. Please let me know if this work is on the right track.

The quicker we move forward the faster we can get to executing the running standalone (uncoupled) MPAS-Atmosphere cap, followed by coupled MPAS-Atmosphere caps. All follow on NUOPC work should include appropriate tests in `testing_and_setup/nuopc/atmosphere/all_tests.ini`

scrasmussen:
This PR worked for me for Dan's testcase. 

The [mpas-hydro-develop](https://github.com/scrasmussen/MPAS-Model/tree/mpas-hydro-develop) branch I've been working on builds and I'm debugging things so it produces identical results at runtime with an MPAS idealized testcase. Once that is fixed, I'll rebase off this branch and use this testing framework to make sure it builds and runs. This looks good to me right now! And I'll be able to give better feedback once I've used it to add a testcase.

danrosen25:
@mgduda @scrasmussen @uturuncoglu
[78c501e](https://github.com/MPAS-Dev/MPAS-Model/pull/1417/commits/78c501e13b505c3c5a7b0fe9c7c29cc2c5a67318) enhances the NUOPC build and test infrastructure and will better position this work for the MPAS Atmosphere standalone cap. It utilizes nccmp or cdo (two most popular netcdf comparison utilities) to compare netcdf files. The revision also fixes a few things like https://github.com/NCAR/MPAS-Model-NUOPC-dev/issues/14

danrosen25:
@mgduda @scrasmussen @scrasmussen @anntsay
Hi team,
I'm simplifying this PR through moving the NUOPC cap infrastructure to a new branch, to reduce the number of line changes in this commit. Most of the line changes are in Makefiles and they throw an error if you try to build a NUOPC library for a CORE other than atmosphere. The NUOPC cap in this PR is just a stub, it does not call into the model but it tests the NUOPC build infrastructure.

The testing infrastructure can be cherry-picked from this branch
https://github.com/NCAR/MPAS-Model-NUOPC-dev/tree/nuopc/testing
da3baa95a71fcbf00f31c33fe928b847988c8a1a

We'll rebase our following iterations once this work is merged.

Makefile
```diff
@@ -1537,6 +1552,8 @@ SCOTCH_MESSAGE = "MPAS was NOT linked with the Scotch graph partitioning library
 endif
 
 mpas_main: $(MAIN_DEPS)
+	if [ ! -d $(MPAS_LIBDIR) ]; then mkdir $(MPAS_LIBDIR); fi
+	if [ ! -d $(MPAS_MODDIR) ]; then mkdir $(MPAS_MODDIR); fi
```
mgduda: If `MPAS_LIBDIR` and `MPAS_MODDIR` are subdirectories of a directory that doesn't yet exist, I think these two `mkdir` commands will fail. For example, if I compile with
```
make -j 6 gnu CORE=atmosphere NUOPC=true MPAS_PREFIX=/glade/derecho/scratch/duda/nuopc_pr/mpas_nuopc
```
I get the following error
```
mkdir: cannot create directory ‘/glade/derecho/scratch/duda/nuopc_pr/mpas_nuopc/lib’: No such file or directory
```
(in my test, the `/glade/derecho/scratch/duda/nuopc_pr` directory already exists, but the `mpas_nuopc` sub-directory does not exist).

Makefile
```diff
@@ -684,6 +684,21 @@ intel:   # BUILDTARGET Intel oneAPI Fortran, C, and C++ compiler suite
 CPPINCLUDES =
 FCINCLUDES =
 LIBS =
+MPAS_PREFIX ?= $(CURDIR)
+MPAS_LIBDIR ?= $(MPAS_PREFIX)/lib
+MPAS_MODDIR ?= $(MPAS_PREFIX)/mod
```
mgduda: These names (`MPAS_PREFIX`, `MPAS_LIBDIR`, and `MPAS_MODDIR`) are fairly generic, though they're only used when building a NUOPC cap/library. Would you consider adding `NUOPC` to these names? E.g., `MPAS_PREFIX` becomes `MPAS_NUOPC_PREFIX`?

mgduda (changes requested):

Makefile
mgduda: Also, if `NUOPC` is `false` or not specified, it looks like empty `lib` and `mod` directories will be created in the top-level `MPAS-Model` directory. Could we create these directories only if compiling with `NUOPC=true`?

mgduda (changes requested):

Makefile
danrosen25: - changed to mkdir -p
- added NUOPC=true condition

Makefile
danrosen25: changed to
- MPAS_NUOPC_PREFIX
- MPAS_NUOPC_LIBDIR
- MPAS_NUOPC_MODDIR

danrosen25:
@mgduda 
See changes in 8b9b1e9
Note that I made a fix so that MPAS_NUOPC_LIBDIR and MPAS_NUOPC_MODDIR can only be deleted if they are empty and the clean command will only remove libmpas_nuopc.a and mpas_nuopc_*.mod. this is to prevent deleting non mpas_nuopc files and folders.

src/Makefile
```diff
@@ -64,3 +90,12 @@ endif
 	( cd framework; $(MAKE) clean )
 	( cd operators; $(MAKE) clean )
 	( cd driver; $(MAKE) clean )
+
+clean_nuopc:
```
danrosen25: @mgduda 
What are your thoughts on cleaning the MPAS_NUOPC_LIBDIR and MPAS_NUOPC_MODDIR? I wanted the ability to install it in user defined directories but cleaning it from user defined directories can be unsafe.  Say the MPAS_NUOPC_PREFIX is /usr, MPAS_NUOPC_LIBDIR is /lib, and MPAS_NUOPC_MODDIR is /include.  Should I remove the clean functionality or do something different? Do we want to be able to delete the <MPAS_DIR>/lib and <MPAS_DIR>/mod directories when cleaning?

mgduda: Installing files outside of the main `MPAS-Model` directory isn't something that we've had to think much about until now, so I'm still not sure what the best approach would be. Perhaps if we could determine that the installation `mod` and `lib` directories resided within the `MPAS-Model` directory, we could clean them, and otherwise, running `make clean` could print a message informing the user that they should manually remove the installation directories (which we could list in the message)?

src/core_init_atmosphere/Makefile
```diff
@@ -42,6 +42,13 @@ core_input_gen:
 	( cd default_inputs; $(NL_GEN) ../Registry_processed.xml namelist.init_atmosphere in_defaults=true )
 	( cd default_inputs; $(ST_GEN) ../Registry_processed.xml streams.init_atmosphere stream_list.init_atmosphere. listed )
 
+nuopc:
+	@echo "************ ERROR ************"
+	@echo "NUOPC build does not exist for the init_atmosphere core. Quitting."
+	@echo "************ ERROR ************"
+	@echo ""
+	exit 1
```
mgduda: Rather than modifying the `Makefile` files under each core directory (besides `core_atmosphere`) to indicate that a NUOPC build is not supported, we could simplify matters by making a NUOPC build an "opt-in" feature. Here's an example of an alternative approach that would avoiding any modifications to cores that don't support NUOPC: https://github.com/mgduda/MPAS-Model/commit/5c4563a50b822dbe6fc3ac1e3437ae9532ba075b .

mgduda (changes requested):

src/Makefile
```diff
@@ -6,7 +6,12 @@ include Makefile.in.$(ESM)
 
 else
 
-all: mpas
+BUILD_ALL = mpas
+ifeq "$(NUOPC)" "true"
+BUILD_ALL = nuopc_lib
```
mgduda: Should `=` be `+=` here? Otherwise, compiling the `atmosphere` core with `NUOPC=true` will not produce an `atmosphere_model` executable.

src/Makefile
```diff
@@ -35,8 +40,29 @@ dycore: $(AUTOCLEAN_DEPS) build_tools externals frame ops
...
 
+nuopc_lib: $(AUTOCLEAN_DEPS) externals frame ops dycore drver
+	( cd core_$(CORE); $(MAKE) CPPFLAGS="$(CPPFLAGS)" CPPINCLUDES="$(CPPINCLUDES)" FFLAGS="$(FFLAGS)" FCINCLUDES="$(FCINCLUDES)" nuopc )
+	ar rcs $(MPAS_NUOPC_LIBDIR)/libmpas_nuopc.a \
+		core_$(CORE)/nuopc/*.o \
+		core_$(CORE)/libphys/*.o \
+		core_$(CORE)/*.o \
+		core_$(CORE)/diagnostics/*.o \
+		core_$(CORE)/libchem/*.o \
+		core_$(CORE)/dynamics/*.o \
```
mgduda: Some of these directories are specific to the `atmosphere` core and so I don't think they should be hard-wired in the `src/Makefile` for the `nuopc_lib` target. An alternative may be to let a core's `build_options.mk` file specify all core-specific files and directories that are needed to support its NUOPC build.

src/Makefile
danrosen25: @mgduda 
We discussed this back in April 2026 https://github.com/NCAR/MPAS-Model-NUOPC-dev/issues/14. The guidance I was given was to utilize an existing preprocessor directive `MPAS_NO_ESMF_INIT` during NUOPC builds. This makes the `atmosphere_model` unusable because ESMF will never be initialized. For what it's worth I mentioned that I don't like using preprocessor directives for this type of behavior. In order to avoid this we'd need to add a flag that tracks whether ESMF was initialized by the MPAS code or by an external driver.

src/Makefile
danrosen25: Added check to MPAS_NUOPC_* directory location before removing in [5d1cc74](https://github.com/MPAS-Dev/MPAS-Model/commit/5d1cc74b9c150d7a388c5adcb4c9746b8ae9d16f)

src/core_init_atmosphere/Makefile
danrosen25: Moved (core) build check to build_options.mk in [5d1cc74](https://github.com/MPAS-Dev/MPAS-Model/commit/5d1cc74b9c150d7a388c5adcb4c9746b8ae9d16f)

src/Makefile
danrosen25: Moved MPAS atm nuopc library build into core directory. I don't know if there's a clean way to assign a list of *.o files in the build_options.mk file so I didn't want to do this. The list of specific *.o files needed by the MPAS atm nuopc cap will evolve as the cap is expanded to support new features of MPAS atm.

src/Makefile
danrosen25: @mgduda [b8102f4](https://github.com/MPAS-Dev/MPAS-Model/pull/1417/commits/b8102f49b6761723fb1c7558e84b8d722b81e510) Removes the MPAS_NO_ESMF_INIT directive from NUOPC builds which enables the MPAS core executable. This will have to be dealt with in the standalone PR.

PR #1419, opened 2026-03-12
Compute, rather than read from input, `deformation_coef_*` fields at model startup

mgduda:
This PR modifies both the `init_atmosphere` and `atmosphere` cores so that the `deformation_coef_{c2,s2,cs,c,s}` fields are computed by the model during its initialization rather than being read from a netCDF input file (either an invariant file, an initial conditions file, or a restart file).

The changes in this PR may be logically viewed as occurring in three steps:
1. Add code to the `atmosphere` core to compute the `deformation_coef_{c2,s2,cs,c,s}` fields at model start up using identical code to that which is found in the `init_atmosphere` core
2. Remove the `deformation_coef_{c2,s2,cs,c,s}` fields from the `atmosphere` core's 'invariant' stream (effectively removing these fields from the 'input' and 'restart' streams as well) so the fields are no longer read from netCDF input files.
3. Remove the `deformation_coef_{c2,s2,cs,c,s}` fields and their computation from the `init_atmosphere` core.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core.F
+++ b/src/core_atmosphere/mpas_atm_core.F
@@ -419,6 +419,10 @@ module atm_core
 
       logical, pointer :: config_do_restart, config_do_DAcycling
 
+      logical, pointer :: on_a_sphere
+      real (kind=RKIND), pointer :: sphere_radius
+
+
       call atm_compute_signs(mesh)
    
       call mpas_pool_get_subpool(block % structs, 'diag', diag)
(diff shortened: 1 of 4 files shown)


mgduda:
Note that the code in `mpas_atm_advection.F` has essentially been reverted to its state before the computation of the `deformation_coef_*` fields was introduced by PR #1404 .

src/core_atmosphere/mpas_atm_core.F
```diff
@@ -419,6 +419,10 @@ subroutine atm_mpas_init_block(dminfo, stream_manager, block, mesh, dt)
 
       logical, pointer :: config_do_restart, config_do_DAcycling
 
+      logical, pointer :: on_a_sphere
+      real (kind=RKIND), pointer :: sphere_radius
+
+
```
abishekg7, line 425: Is the extra newline here required?

mgduda: Using two blank lines after local variable declarations is something that we do in other parts of the code. See, e.g., lines 1658 - 1659 in this same file.

mgduda: For other examples, see `mpas_halo.F` and the subroutines therein.

abishekg7:
I've tested the PR with a real global case and also the LES case - and get bit identical results with the current develop HEAD. Code changes look good - checked the current PR against the develop branch prior to the merge of the LES PR.

abishekg7 (approved):
Looks good.

Issue #1420, opened 2026-03-17
Recent change in MMM-physics breaks MPAS-Model


eap:
https://github.com/NCAR/MMM-physics/pull/16 was submitted several hours ago breaking jedi-bundle which depends on MPAS-Model 8.2.1. I'm not sure if the change breaks all existing tags of MPAS-Model or just 8.2.1, although the nature of the issue makes me suspect that all builders of MPAS-model are going to encounter the issue.

## Build error
```
 Building Fortran object mpas/src/core_atmosphere/CMakeFiles/core_atmosphere.dir/physics/physics_wrf/module_bl_gwdo.F.o
/workdir/bundle/mpas/src/core_atmosphere/physics/physics_wrf/module_bl_gwdo.F:215:49:

  215 |                     ,errmsg=errmsg,errflg=errflg)
      |                                                 1
Error: Missing actual argument for argument 'omax' at (1)
make[2]: *** [mpas/src/core_atmosphere/CMakeFiles/core_atmosphere.dir/build.make:2769: mpas/src/core_atmosphere/CMakeFiles/core_atmosphere.dir/physics/physics_wrf/module_bl_gwdo.F.o] Error 1
```

## Suspected root cause

I found this line of code in [core_atmosphere/CMakeLists.txt:102-108](https://github.com/MPAS-Dev/MPAS-Model/blob/b9090a1434b8ceb2a92fc985f4d97e7947cd95a6/src/core_atmosphere/CMakeLists.txt#L102-L108) which references NCAR/MMM-Physics without a git tag or hash. So any new changes that break API compatibility will end up breaking core_atmosphere.

```
if(NOT EXISTS ${ATMOSPHERE_CORE_PHYSICS_MMM_DIR})
    set(PHYSICS_MMM_REPO_URL "https://github.com/NCAR/MMM-physics")
    execute_process(COMMAND git clone ${PHYSICS_MMM_REPO_URL} ${ATMOSPHERE_CORE_PHYSICS_MMM_DIR}
                    RESULT_VARIABLE GIT_CLONE_RESULT
                    OUTPUT_VARIABLE GIT_CLONE_OUTPUT
                    ERROR_VARIABLE GIT_CLONE_ERROR)
    if(NOT GIT_CLONE_RESULT EQUAL 0)
```

mgduda:
[...]

PR #1421, opened 2026-03-19
Use the checkout_externals utility to checkout physics submodules with cmake

jim-p-w:
core_atmosphere requires specific versions of the physics code from https://github.com/NCAR/MMM-physics.git and https://github.com/NOAA-GSL/UGWP.git

The checkout_externals utility uses the versions of those repositories specified in src/core_atmosphere/Externals.cfg.
The gnu make build system uses checkout_externals; this change modifies the cmake build files to use the same mechanism as the make build system.

This PR fixes [issue 1420](https://github.com/MPAS-Dev/MPAS-Model/issues/1420)

This was tested by building MPAS-Model code with cmake. It was also tested by building [mpas-bundle](https://github.com/JCSDA-internal/mpas-bundle).

This change was copied from [wrf-model](https://github.com/wrf-model/WRF/blob/f15568ccc1447780e3bd664b9f0196edd784bf33/CMakeLists.txt#L332-L351) where it has been in use for about a year.

### DIFF
--- a/src/core_atmosphere/CMakeLists.txt
+++ b/src/core_atmosphere/CMakeLists.txt
@@ -97,20 +97,16 @@ set(ATMOSPHERE_CORE_PHYSICS_WRF_SOURCES
 list(TRANSFORM ATMOSPHERE_CORE_PHYSICS_WRF_SOURCES PREPEND physics/physics_wrf/)
 
 
-set(ATMOSPHERE_CORE_PHYSICS_MMM_DIR ${CMAKE_CURRENT_SOURCE_DIR}/physics/physics_mmm)
-
-if(NOT EXISTS ${ATMOSPHERE_CORE_PHYSICS_MMM_DIR})
-    set(PHYSICS_MMM_REPO_URL "https://github.com/NCAR/MMM-physics")
-    execute_process(COMMAND git clone ${PHYSICS_MMM_REPO_URL} ${ATMOSPHERE_CORE_PHYSICS_MMM_DIR}
-                    RESULT_VARIABLE GIT_CLONE_RESULT
-                    OUTPUT_VARIABLE GIT_CLONE_OUTPUT
-                    ERROR_VARIABLE GIT_CLONE_ERROR)
-    if(NOT GIT_CLONE_RESULT EQUAL 0)
-        message(FATAL_ERROR "Git clone failed with error: ${GIT_CLONE_ERROR}")
-    endif()
-
+set(CHECKOUT ${CMAKE_CURRENT_SOURCE_DIR}/tools/manage_externals/checkout_externals )
+execute_process( COMMAND
+                 ${CHECKOUT} --externals ${CMAKE_CURRENT_SOURCE_DIR}/Externals.cfg
+                 WORKING_DIRECTORY       ${CMAKE_CURRENT_SOURCE_DIR}/physics/
+                 RESULT_VARIABLE         MANAGE_EXTERNALS_STATUS
+                )
+if ( ${MANAGE_EXTERNALS_STATUS} AND NOT ${MANAGE_EXTERNALS_STATUS} EQUAL 0 )
+    message( FATAL_ERROR "Failed to checkout external repos via manage_externals" )
 else()
-    message(STATUS "Directory ${DIR_TO_CHECK} already exists, skipping clone")
+    message( STATUS "Finished checking out external repos via manage_externals" )
 endif()
 
 set(ATMOSPHERE_CORE_PHYSICS_MMM_SOURCES
@@ -129,22 +125,6 @@ set(ATMOSPHERE_CORE_PHYSICS_MMM_SOURCES
 )
 
 list(TRANSFORM ATMOSPHERE_CORE_PHYSICS_MMM_SOURCES PREPEND physics/physics_mmm/)
-
-set(ATMOSPHERE_CORE_PHYSICS_NOAA_DIR ${CMAKE_CURRENT_SOURCE_DIR}/physics/physics_noaa/UGWP)
-
-if(NOT EXISTS ${ATMOSPHERE_CORE_PHYSICS_NOAA_DIR})
-    set(PHYSICS_NOAA_REPO_URL "https://github.com/NOAA-GSL/UGWP.git")
-    execute_process(COMMAND git clone ${PHYSICS_NOAA_REPO_URL} ${ATMOSPHERE_CORE_PHYSICS_NOAA_DIR}
-                    RESULT_VARIABLE GIT_CLONE_RESULT
-                    OUTPUT_VARIABLE GIT_CLONE_OUTPUT
-                    ERROR_VARIABLE GIT_CLONE_ERROR)
-    if(NOT GIT_CLONE_RESULT EQUAL 0)
-        message(FATAL_ERROR "Git clone failed with error: ${GIT_CLONE_ERROR}")
-    endif()
-
-else()
-    message(STATUS "Directory ${DIR_TO_CHECK} already exists, skipping clone")
-endif()
 
 set(ATMOSPHERE_CORE_PHYSICS_NOAA_SOURCES
     bl_ugwp.F


mgduda:
@jim-p-w Could you base your branch on the v8.3.0 tag?

jim-p-w:
> @jim-p-w Could you base your branch on the v8.3.0 tag?

Will do.

guoqing-noaa:
FYI, we are addressing the same issue in different routes:
https://github.com/ufs-community/MPAS-Model/pull/233 tries to use git submodules universally.
https://github.com/ufs-community/MPAS-Model/pull/231 tries to apply a bandage to fix the cmake compiling failure.

I would advocate we adopt the git submodules universally.

src/core_atmosphere/CMakeLists.txt
```diff
@@ -97,20 +97,16 @@ set(ATMOSPHERE_CORE_PHYSICS_WRF_SOURCES
...
+execute_process( COMMAND
+                 ${CHECKOUT} --externals ${CMAKE_CURRENT_SOURCE_DIR}/Externals.cfg
+                 WORKING_DIRECTORY       ${CMAKE_CURRENT_SOURCE_DIR}/physics/
+                 RESULT_VARIABLE         MANAGE_EXTERNALS_STATUS
+                )
+if ( ${MANAGE_EXTERNALS_STATUS} AND NOT ${MANAGE_EXTERNALS_STATUS} EQUAL 0 )
+  message( FATAL_ERROR "Failed to checkout external repos via manage_externals" )
 else()
-    message(STATUS "Directory ${DIR_TO_CHECK} already exists, skipping clone")
+  message( STATUS "Finished checking out external repos via manage_externals" )
```
guoqing-noaa: It will be great to add a logic if `physics_mmm` and `UGWP` exist, skip running `checkout_externals`. Thanks!

mgduda: Although `physics_mmm` and `UGWP` may already exist, they may be out of sync with the tag specified in the `Externals.cfg` file.

guoqing-noaa: I think an "existence" check  is consistent with all other "git lone" in this CMakeLists.txt.
Folks may want to make modifications inside `physics_mmm` or `UGWP` and don't want them to get overwritten by `checkout_externals`. This check does not break anything if one do a clean clone of `MPAS-Model`

jim-p-w:
I'm making this a draft until the target branch gets updated to be v8.3.0

src/core_atmosphere/CMakeLists.txt
islas: `checkout_externals` will not overwrite dirty checkouts.

I think keeping the logic as one-to-one with the `make` build where possible helps reduce issues such as this one where extra or divergent code may lead to unexpected difference in behavior.

That said, it would be nice to wholly use submodules (in my opinion).

mgduda:
> I'm making this a draft until the target branch gets updated to be v8.3.0

What I'm suggesting is that your PR branch should begin at the `v8.3.0` tag but still target the `develop` branch, adding a single commit to address the issue with fetching the `HEAD` of the `MMM-physics` `main` branch.

src/core_atmosphere/CMakeLists.txt
```diff
@@ -97,20 +97,16 @@ set(ATMOSPHERE_CORE_PHYSICS_WRF_SOURCES
...
-    endif()
-
+set(CHECKOUT ${CMAKE_CURRENT_SOURCE_DIR}/tools/manage_externals/checkout_externals )
+execute_process( COMMAND
+                 ${CHECKOUT} --externals ${CMAKE_CURRENT_SOURCE_DIR}/Externals.cfg
+                 WORKING_DIRECTORY       ${CMAKE_CURRENT_SOURCE_DIR}/physics/
+                 RESULT_VARIABLE         MANAGE_EXTERNALS_STATUS
+                )
+if ( ${MANAGE_EXTERNALS_STATUS} AND NOT ${MANAGE_EXTERNALS_STATUS} EQUAL 0 )
+  message( FATAL_ERROR "Failed to checkout external repos via manage_externals" )
```
mgduda: It looks like the rest of this file uses four spaces of indentation. Should we do the same here?

src/core_atmosphere/CMakeLists.txt
guoqing-noaa: @islas Thanks for the discussion!

Will we get any issue if we add an existence check? 
If `physics_mmm` or `UGWP` is out of sync, it is mostly due to some corner cases and in those situations, developers/users are expected to handle it manually.

It is nice that if we may use git submodules for all subcomponents. 
GSL is transitioning to do that.
Adding an `existence` check here will help a little bit since we use `git clone --recursive` to get all codes and don't want `check_externals` to do duplicate or unexpected work in that situation. Thanks!

jim-p-w: Because an existence check of directories is not sufficient to know the correct code has been checked out my inclination is to always run `checkout_externals`. For example, if the main branch is checked out, as currently implemented the correct tag will get checked out. 

Also, in the event a different tag is needed, editing `src/core_atmosphere/Externals.cfg` and rerunning cmake will update things properly without requiring manual deletion of directories.

guoqing-noaa: @jim-p-w Thanks for the discussion! FYI, I would like to provide a few alternate scenarios:
1. One makes modifications under `physics_mmm` or `UGWP`, commits them to personal fork/branch. So these directories are clean (not dirty). If one wants to rebuild the MPAS-Model, `checkout_externals` will "overwrites" (to be accurate, switches the top reference) all changes. This is NOT what we expect.
2. If one add anything under `physics_mmm` or `UGWP`, even touching a file, these directories will be dirty and `checkout_externals` will no longer take effect (per previous @islas's comment), right?
3. I think we can only be able to make sure checking out specified frozen checkouts from **a clean clone**. For other situations, if one want to update hash/tag or get back the default frozen checkout,  one can run `checkout_externals` or `git checkout` manually at the command line instead of our forcing it in the CMake runtime.
4. This PR changes the current behavior:
```
if(NOT EXISTS ${ATMOSPHERE_CORE_PHYSICS_MMM_DIR})
......
if(NOT EXISTS ${ATMOSPHERE_CORE_PHYSICS_NOAA_DIR})
......
```
and it may break downstream applications.

mgduda: @guoqing-noaa I think we can all agree that the existing `CMakeLists.txt` file is not correct. If I've understood the changes proposed in https://github.com/ufs-community/MPAS-Model/pull/231 , it seems that there are a couple of potential drawbacks:
1. If we create a fresh clone of MPAS-Model, compile, change one of the tags in the `Externals.cfg` file, and re-compile, the build will not use the updated physics repository tags (because, e.g., the `physics_mmm` directory already exists). This is also a problem if we locally merge a branch that switches external physics repository tags and makes corresponding MPAS-Model code changes to work with the new external tag: compiling will make use of the modified MPAS-Model code without the corresponding updates to the external physics.
2. The behavior of the CMake build is inconsistent with that of the GNU Make build.

While it requires a change in expectations (i.e., the external physics used by the build will always be those specified in the `Externals.cfg` file), it looks to me like the changes proposed in this PR should address both of the above drawbacks.

I'm not sure if we'll reach an agreement in this PR discussion thread. If you intend to update the `ufs-community` fork in the near future with changes to the MPAS-Dev/MPAS-Model `develop` branch, perhaps that can be followed by a PR to the `ufs-community` fork to enforce whatever behavior you'd prefer, and subsequent merges of MPAS-Dev/MPAS-Model `develop` would hopefully leave those changes in place.

mgduda:
@jim-p-w I think it would be safe to squash the changes from [8ff5e97](https://github.com/MPAS-Dev/MPAS-Model/pull/1421/commits/8ff5e97975f0bf3b027ac3224d8a738ee142b2e7) onto the first commit in this PR branch.

src/core_atmosphere/CMakeLists.txt
guoqing-noaa: @mgduda Thanks for the discussions! We have different use situations and preferences. That's fine and expected. We can make a little bit tweaks in the ufs-community fork after this PR is merged. Thanks!

mgduda (approved):

islas (approved):

PR #1405, opened 2026-02-04
Add MPAS_ESMF for external ESMF linking

danrosen25:
This commit replaces hardcoded links to `external/esmf_time_f90` in Makefiles with `MPAS_ESMF_INC` and `MPAS_ESMF_LIBS`. `MPAS_ESMF_INC` and `MPAS_ESMF_LIBS` are calculated in the root `Makefile` depending on `MPAS_ESMF=embedded` (default) or `MPAS_ESMF=external`. `MPAS_ESMF=embedded` will use the `external/esmf_time_f90` folder and `MPAS_ESMF=external` will read INC and LIBS variables in a `ESMFMKFILE` file. These variables are then exported for all other Makefiles.

I first tried adding include paths and libraries to `FCINCLUDES` and `LIBS` but this failed while checking PIO when executing `MPAS_ESMF=embedded` because `external/esmf_time_f90` isn't build until after the PIO check.

As I'm new to MPAS and the MPAS build system please let me know if I'm on the right track.

### Testing
Testing within a Docker container that includes mpich, netcdf-c, netcdf-fortran, parallel-netcdf, parallelio, and esmf (libraries installed using spack).
`make clean gnu CORE="atmosphere"`
`make clean gnu CORE="atmosphere" MPAS_ESMF=embedded`
`make clean gnu CORE="atmosphere" MPAS_ESMF=external`
- tested with and without `ESMFMKFILE` environment variable

**Before change**
clean build ~2m24s
rebuild.    ~0m18s

**MPAS_ESMF=embedded**
clean build ~2m25s
rebuild     ~0m18s

**MPAS_ESMF=external**
clean build ~3m38s
rebuild     ~0m35s

PR #1422, opened 2026-03-19
Remove defc_a and defc_b from the init_atmosphere and atmosphere cores

mgduda:
This PR removes the `defc_a` and `defc_b` fields from the `init_atmosphere` and `atmosphere` cores.

With the introduction of the `deformation_coef_*` fields in merge commit 63f2d449 (PR #1405), the `defc_a` and `defc_b` fields are no longer needed in either the `init_atmosphere` core or the `atmosphere` core. Accordingly, this PR removes the `defc_a` and `defc_b` fields from the `Registry.xml` files as well as from the source code of both of these cores.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -552,8 +552,6 @@
 			<var name="zb3"/>
 			<var name="dss"/>
 			<var name="deriv_two"/>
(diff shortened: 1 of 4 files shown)


abishekg7 (approved):
Tested the atmosphere and init_atmosphere cores. Looks good.

Issue #1414, opened 2026-02-14
output at flexible time levels


guoqing-noaa:
Currently, we can configure `output_interval` in a stream to be a specific interval, `none`, or `initial_only`  to output MPAS forecasts.

In operational workflows, we would like the output_interval to be more flexible.

For example, for 10 day global forecasts, we may want to output every 15 minutes in the first 3 hours, every 1 hour up to 3 days, every 3 hours to 7 days, and every 6 hours to 10 days.

Another example is that we make 84 hours regional forecasts. We output `history.nc` files every 1 hour. We would like to be able to only output `mpasout.nc` files at forecast hours f001, f002 and f003 to enable hourly cycling data assimilation (with a possible fallback).

It will be great if we can have another stream attribute, such as `output_timelevels`, and then we can enumerate the time levels to write forecast outputs based on that stream.

It may be something like:
```
<stream name="output"
        type="output"
        filename_template="history.$Y-$M-$D_$h.$m.$s.nc"
        io_type="pnetcdf,cdf5"
        clobber_mode="replace_files"
        output_timelevels="0-3-15m  4-72  75-168-3  174-240-6">
        <file name="stream_list/stream_list.atmosphere.output"/>
</stream>
```

byoung-joo:
There exists a similar effort in https://github.com/MPAS-Dev/MPAS-Model/pull/1365, but I don't know what the status is.

guoqing-noaa:
> There exists a similar effort in [#1365](https://github.com/MPAS-Dev/MPAS-Model/pull/1365), but I don't know what the status is.

@byoung-joo Thanks a lot for the information. It is good to be able to specify start_time and/or stop_time!
But that PR may not be able to provide flexible output every 15 minutes in the fist 3 hours, then hourly int he next 3 days, ..., etc.

guoqing-noaa:
We may use different streams for the first 3 hours, next 3 days and the final 4 days to achieve flexible outputs. But in this situation, we have to use different output file names for different streams, right?

byoung-joo:
Can you give some comment on the question above, @amstokely ? Thanks!

amstokely:
@byoung-joo @guoqing-noaa

For now, you’ll need to use different stream filename templates when using this feature. MPAS currently performs a duplicate stream check that requires each stream to have a unique filename template.

The MPAS-Dev team is aware of this limitation and plans to remove the requirement in the future, though we don’t have a timeline for that change yet.

[...]

PR #1423, opened 2026-03-19
enable the capabilty to output at flexible time levels

guoqing-noaa:
Introduce a new `output_timelevels` attribute for MPAS streams that enables variable output intervals. 

With this capability, we may outoput every 15 minutes in the first hour, every hour in the first 3 days, every 3 hours for the next 4 days, and every 6 hours in the last 3 days. 

We can also use this to only write out forecast files during a given period, such as: `output_timelevels="6-12"`  
Or only output mpasout files at a limited time levels `output_timelevels="0 1"` to facilitate cycling data assimilation

Resolve issue https://github.com/MPAS-Dev/MPAS-Model/issues/1414

### DIFF
--- a/src/framework/mpas_stream_list_types.inc
+++ b/src/framework/mpas_stream_list_types.inc
@@ -2,6 +2,18 @@
                           MPAS_STREAM_LIST_DUPLICATE = 1, &
                           MPAS_STREAM_LIST_NOT_FOUND = 2
 
+    ! Maximum number of segments in the output_timelevels specification
+    integer, parameter :: MAX_TIMELEVEL_SEGMENTS = 32
+
+    ! timelevel specification
+    type MPAS_timelevel_spec_type
+        integer :: n_segments = 0                                  ! Number of parsed segments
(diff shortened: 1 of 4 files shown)


guoqing-noaa:
As mentioned in Wednesday’s meeting, I created this PR for those who want to review the code changes in more detail. From my perspective, it is ready for review. That said, whether we proceed in this direction is still an open decision, and I’m fine with either outcome.

IMHO, [Line 3194 of mpas_stream_manager.F](https://github.com/guoqing-noaa/MPAS-Model/blob/944b976db8f1369d4e90f9495a3cd4e9795a2ffb/src/framework/mpas_stream_manager.F#L3194) clearly separates the logic between `output_interval` and `output_timelevels`, with no conflict between them.
```
if (stream % timelevel_spec % is_parsed) then
```
Therefore, both methods can coexist without issues, and users may choose whichever is most convenient for their specific applications.

@mgduda @skamaroc @liujake @byoung-joo

PR #1428, opened 2026-03-24
Remove PV diagnostics from default "output" stream in MPAS-Atmosphere

mgduda:
This PR removes the following PV diagnostic fields from the default "output" stream for MPAS-Atmosphere:

 - ertel_pv
 - u_pv
 - v_pv
 - theta_pv
 - vort_pv
 - iLev_DT
 - depv_dt_lw
 - depv_dt_sw
 - depv_dt_bl
 - depv_dt_cu
 - depv_dt_mix
 - dtheta_dt_mp
 - depv_dt_mp
 - depv_dt_diab
 - depv_dt_fric
 - depv_dt_diab_pv
 - depv_dt_fric_pv

These fields may not be of general interest, and so removing them saves space in output files (the "history" netCDF files are ~20% smaller based on testing with a global 60-km mesh).

Additionally, removing these fields from default output streams saves a small amount of computation time, as the logic in the `mpas_pv_diagnostics` module computes these diagnostics only if they are requested in an output stream.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -917,28 +917,7 @@
 			<var name="ke"/>
 			<var name="uReconstructZonal"/>
 			<var name="uReconstructMeridional"/>
-
-                        <!-- Begin Ertel PV diagnostics defined in diagnostics/Registry_pv.xml -->
-			<var name="ertel_pv"/>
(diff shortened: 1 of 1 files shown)


weiwangncar (approved):
Make sense.

PR #1432, opened 2026-03-27
Fix for ESMF link path

abishekg7:
This PR addresses the linking error that arises when building the MPAS atmosphere core with the `nvhpc` compiler in DEBUG mode. 

The incorrect make variable `MPAS_ESMF_LIBS` with `MPAS_ESMF_LIB`, which is the variable defined in the root Makefile.

### DIFF
--- a/src/core_atmosphere/utils/Makefile
+++ b/src/core_atmosphere/utils/Makefile
@@ -7,7 +7,7 @@ endif
 all: $(UTILS)
 
 build_tables: build_tables.o atmphys_build_tables_thompson.o
-	$(LINKER) $(LDFLAGS) -o build_tables build_tables.o atmphys_build_tables_thompson.o -L../../framework -L../physics -lphys -lframework $(LIBS) $(MPAS_ESMF_LIBS)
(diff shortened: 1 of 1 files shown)


danrosen25:
@abishekg7 Sorry for the typo! I looked through my work and I think that you caught the only place `MPAS_ESMF_LIBS` was used. NCAR CISL is working on some CI infrastructure so we'll see less of these mistakes in the future.
https://github.com/MPAS-Dev/MPAS-Model/commit/66517f6013fccc482805e4c98d5e6beb38fbafb6

mgduda (approved):

mgduda:
> @abishekg7 Sorry for the typo! I looked through my work and I think that you caught the only place `MPAS_ESMF_LIBS` was used. NCAR CISL is working on some CI infrastructure so we'll see less of these mistakes in the future. [66517f6](https://github.com/MPAS-Dev/MPAS-Model/commit/66517f6013fccc482805e4c98d5e6beb38fbafb6)

In principle, I don't think the `build_tables` executable should even need the ESMF library, but it seems that the NVHPC linker requires the ESMF library when compiling with certain debug options. So the typo in the name of the `MPAS_ESMF_LIB` variable is something that would probably have only tripped up the NVHPC debug builds, but otherwise doesn't impact the correctness of essentially any other build.

PR #1433, opened 2026-03-30
Define new package, `les`, and attach it to the `tke` constituent of var_arrays

mgduda:
This PR defines a new package, `les`, for the atmosphere core, and it attaches this package to the `tke` constituent of the `scalars`, `scalars_tend`, and `lbc_scalars` var_arrays in the atmosphere core's `Registry.xml` file.

The `tke` constituent is only required when an LES model has been selected at runtime (i.e., when `config_les_model = '3d_smagorinsky'` or `config_les_model = 'prognostic_1.5_order'`), and the `les` package is therefore active if and only if `config_les_model` is not `none`.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -468,6 +468,9 @@
                 <package name="ugwp_orog_stream" description="Input stream (and variables) for UGWP orography"/>
                 <package name="ugwp_ngw_stream"  description="Input stream (and variables) for UGWP NGW lookup table"/>
                 <package name="ugwp_diags_stream" description="Output stream (and variables) for UGWP output diagnostics"/>
+                <package name="les"
+                         description="Variables and options associated with LES models"
+                         active_when="trim(config_les_model) /= 'none'"/>
         </packages>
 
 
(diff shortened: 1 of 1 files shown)


byoung-joo (approved):
Thank you, @mgduda! This looks good. This has fixed the issue we had with the existing mpas-jedi test.

I might be wrong but interestingly, this `les` package seems to be a first usecase of `active_when`.

byoung-joo:
@jim-p-w,
FYI: This PR will make the MPAS-Workflow test work ( WarmStart scenarios ).

mgduda:
> Thank you, @mgduda! This looks good. This has fixed the issue we had with the existing mpas-jedi test.
> 
> I might be wrong but interestingly, this `les` package seems to be a first usecase of `active_when`.

Yes, this may be the first use of `active_when` (introduced by PR #1321) in the `develop` branch, though I think we've used the functionality in some other branches.

mgduda:
@dudhia Are you aware of any code besides the LES `3d_smagorinsky` and `prognostic_1.5_order` options that uses the `tke` constituent in the scalar arrays?

dudhia:
OK, tke was only added with the LES option and so that is the only use so far.

mgduda:
> OK, tke was only added with the LES option and so that is the only use so far.

Could I interpret this as an approval of the changes in this PR?

dudhia:
I approve.

Issue #1434, opened 2026-03-31
After merge of PR #1404, atmosphere core builds without physics fail


mgduda:
After the merge of PR #1404 (Initial LES options for MPAS-Atmosphere), builds of the atmosphere core without physics (i.e., after commenting-out the definition of the `PHYSICS` variable in `src/core_atmosphere/Makefile` fail because the `mpas_atmphys_constants` module needed by the `mpas_atm_dissipation_models` module is not available:
```
mpas_atm_dissipation_models.F:14:8:

   14 |    use mpas_atmphys_constants
      |        1
Fatal Error: Cannot open module file 'mpas_atmphys_constants.mod' for reading at (1): No such file or directory
compilation terminated.
```

mgduda:
It appears that the `mpas_atm_dissipation_models` module requires several physical constants from `mpas_atmphys_constants` for use in calculating the Brunt–Väisälä frequency.

There are several solutions that come immediately to mind:
1. Move the `mpas_atmphys_constants.F` file from the `src/core_atmosphere/physics` directory up to the `src/core_atmosphere` directory so that it can be shared by dynamics and physics. This has the downside of changing the location of the compiled `mpas_atmphys_constants.mod` file, and this may necessitate changes to the `Makefile` in one or more external physics repositories.
2. Modify the logic in `src/core_atmosphere/Makefile` so that the `mpas_atmphys_constants.F` file is always compiled in the `src/core_atmosphere/physics` directory, regardless of whether `PHYSICS` is defined. This has the downside of complicating the `Makefile` logic.
3. Add local definitions of the required physical constants in `mpas_atm_dissipation_models.F` to remove the dependency of `mpas_atm_dissipation_models` on `mpas_atmphys_constants`. This could be done only when the `DO_PHYSICS` macro is not defined, so that typical builds of MPAS-Atmosphere with physics would still use the definitions of constants from `mpas_atmphys_constants`, and only builds without physics would use the local definitions in `mpas_atm_dissipation_models`. This has the downside of introducing duplicating definitions of physical constants in two locations.

mgduda:
PR #1435 closes this issue, taking the third approach outlined above.

PR #1435, opened 2026-03-31
Fix build failure for mpas_atm_dissipation_models when DO_PHYSICS is not defined

mgduda:
This PR fixes a build failure for `mpas_atm_dissipation_models.F` when the `DO_PHYSICS` macro is not defined.

When `DO_PHYSICS` is not defined, the `mpas_atmphys_constants.F` code in `src/core_atmosphere/physics` is not compiled, leading to build failures for `mpas_atm_dissipation_models.F` in the `src/core_atmosphere/dynamics` directory, since the `calculate_n2` routine within the `mpas_atm_dissipation_models` module relies on the definitions of parameters from the `mpas_atmphys_constants` module.

The fix provided by this PR involves adding local definitions of the required physical constants in the `calculate_n2` routine only when the `DO_PHYSICS` macro is not defined. When `DO_PHYSICS` is defined, the definitions of these physical constants are imported from the `mpas_atmphys_constants` module as before.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_dissipation_models.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_dissipation_models.F
@@ -11,7 +11,6 @@
 module mpas_atm_dissipation_models
 
    use mpas_kind_types, only : RKIND
-   use mpas_atmphys_constants
    use mpas_constants
    use mpas_log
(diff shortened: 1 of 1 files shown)


mgduda:
This PR closes #1434.

abishekg7 (approved):
Looks good. Build completes successfully.

PR #1436, opened 2026-04-01
Fix crash in atm_compute_dyn_tend when DO_PHYSICS is not defined in build

mgduda:
This PR fixes a crash in the `atm_compute_dyn_tend` routine when `DO_PHYSICS` is not defined in a build of the `atmosphere` core.

When `DO_PHYSICS` is not defined, the `diag_physics` pool/var_struct is not defined in the `Registry.xml` file, leading to a crash in the `atm_compute_dyn_tend` routine when attempting to query `ustm`, `hfx`, or `qfx` from the `diag_physics` pool.

The simple fix in this PR is to guard the calls to `mpas_pool_get_array` for `ustm`, `hfx`, and `qfx` in the `atm_compute_dyn_tend` with preprocessor directives, so that the access of these arrays from the `diag_physics` pool occurs only when `DO_PHYSICS` is defined.

Note that the `atm_compute_dyn_tend` routine already contains logic to accommodate situations in which `ustm`, `hfx`, and `qfx` are not available (because they are allocated only when the either of the `bl_mynn_in` or `bl_ysu_in` packages are active), and so it's only the calls to set pointers to these arrays from the `diag_physics` pool that need to be guarded.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -5025,11 +5025,14 @@ module atm_time_integration
       call mpas_pool_get_array(tend_physics, 'rthdynten', rthdynten)
 
       nullify(ustm)
-      call mpas_pool_get_array(diag_physics,'ustm',ustm)
       nullify(hfx)
-      call mpas_pool_get_array(diag_physics,'hfx',hfx)
       nullify(qfx)
+
+#ifdef DO_PHYSICS
+      call mpas_pool_get_array(diag_physics,'ustm',ustm)
+      call mpas_pool_get_array(diag_physics,'hfx',hfx)
       call mpas_pool_get_array(diag_physics,'qfx',qfx)
+#endif
 
       nopbl = .false.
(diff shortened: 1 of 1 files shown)


mgduda:
Note that the changes in this PR branch are not based on the PR branch from #1435, and so testing will likely require merging this PR branch to the current HEAD of the `develop` branch, or otherwise incorporating the build fix from #1435.

abishekg7:
Can get this to build and run, but only after rebasing with the current HEAD of develop.

mgduda:
> Can get this to build and run, but only after rebasing with the current HEAD of develop.

As I noted in my comment above, this is expected. In fact, the base of this PR branch was intentionally chosen to not include the compilation fix from PR #1435. Others may have branches that fix the compilation problem in a different way (perhaps taking one of the other approaches mentioned in Issue #1434), an in such cases, they would still be able to merge this PR without also picking up changes from PR #1435.

mgduda:
@abishekg7 Did you have any other comments on this PR?

abishekg7 (approved):
Looks good.

PR #1437, opened 2026-04-01
Fix calls to mpas_halo_exch_group_full_halo_exch in mpas_halo_testing

abishekg7:
This PR introduces explicit argument names for the optional argument iErr in calls to `mpas_halo_exch_group_full_halo_exch` in `mpas_halo_testing`.

 The merge of PR #1355 to develop, which introduced an extra optional argument (`withGPUAwareMPI`) to the signature of `mpas_halo_exch_group_full_halo_exch`, broke the build of the test core. This PR fixes the issue by using named arguments when referencing the optional arguments in the function.

### DIFF
--- a/src/core_test/mpas_halo_testing.F
+++ b/src/core_test/mpas_halo_testing.F
@@ -168,7 +168,7 @@ module mpas_halo_testing
       end do
       end do
 
-      call mpas_halo_exch_group_full_halo_exch(domain, 'persistent_group', ierr_local)
+      call mpas_halo_exch_group_full_halo_exch(domain, 'persistent_group', iErr=ierr_local)
       ierr = ior(ierr, ierr_local)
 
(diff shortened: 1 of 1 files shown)


mgduda:
Although it's probably not critical in this particular case, in general it's probably best to base a bugfix branch on the commit that introduced the bug whenever feasible. Could you rebase this PR branch onto b0ce6301 ?

abishekg7:
> Although it's probably not critical in this particular case, in general it's probably best to base a bugfix branch on the commit that introduced the bug whenever feasible. Could you rebase this PR branch onto [b0ce630](https://github.com/MPAS-Dev/MPAS-Model/commit/b0ce63013ab05bd81c39a6488e0b610720084c64) ?

Have rebased it onto [b0ce630] now.

mgduda (approved):

PR #1438, opened 2026-04-01
Fix an issue in snow initialization

smileMchen:
This PR fixes the bug in snow initialization over Arctic Ocean and other areas covered by seaice.

Snow water equivalent (as well as snow depth and snow cover) in the initial conditions of MPAS is currently set to zero over sea ice areas. This leads to a significant warm bias of skintemp during model integration. When MPAS is initialized in winter, the simulated conditions over polar regions become physically unrealistic.

This PR corrects snow initialization, making it physically reasonable over Arctic Ocean and other areas covered by seaice in winter.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
@@ -606,6 +606,7 @@
  real(kind=RKIND),pointer:: tsk_seaice_threshold
  real(kind=RKIND),dimension(:),pointer  :: vegfra
  real(kind=RKIND),dimension(:),pointer  :: seaice,snoalb,xice
+ real(kind=RKIND),dimension(:),pointer  :: snow, snowh, snowc
  real(kind=RKIND),dimension(:),pointer  :: skintemp,tmn,xland
  real(kind=RKIND),dimension(:,:),pointer:: tslb,smois,sh2o,smcrel
 
@@ -637,6 +638,9 @@
 
  call mpas_pool_get_array(input, 'seaice', seaice)
  call mpas_pool_get_array(input, 'xice', xice)
+ call mpas_pool_get_array(input, 'snowc', snowc)
+ call mpas_pool_get_array(input, 'snowh', snowh)
+ call mpas_pool_get_array(input, 'snow', snow)
  call mpas_pool_get_array(input, 'vegfra', vegfra)
 
  call mpas_pool_get_array(input, 'skintemp', skintemp)
@@ -699,6 +703,11 @@
           call mpas_log_write('$i $r $r $r', intArgs=(/iCell/), &
                                              realArgs=(/real(landmask(iCell),kind=RKIND),xland(iCell),xice(iCell)/))
        xice(iCell) = 0._RKIND
+       if(landmask(iCell) .eq. 0) then
+          snowc(iCell) = 0._RKIND  ! snow = 0 over water points
+          snowh(iCell) = 0._RKIND
+          snow(iCell) = 0._RKIND
+       endif
     endif
 
  enddo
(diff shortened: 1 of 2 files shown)


smileMchen:
image

smileMchen:
The upper two panels show initial snow before and after the bug fix ( initial time is 2024-02-01_00)

src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
```diff
@@ -699,6 +703,9 @@ subroutine physics_init_seaice(mesh, input, dims, configs)
           call mpas_log_write('$i $r $r $r', intArgs=(/iCell/), &
                                              realArgs=(/real(landmask(iCell),kind=RKIND),xland(iCell),xice(iCell)/))
        xice(iCell) = 0._RKIND
+       snowc(iCell) = 0._RKIND  ! snow = 0 over water points
+       snowh(iCell) = 0._RKIND
```
mgduda: I'm seeing a runtime failure due to an out-of-bounds array access:
```
At line 707 of file mpas_atmphys_initialize_real.F
Fortran runtime error: Index '1' of dimension 1 of array 'snowh' below lower bound of 1584
```

src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
```diff
@@ -637,6 +638,9 @@ subroutine physics_init_seaice(mesh, input, dims, configs)
 
  call mpas_pool_get_array(input, 'seaice', seaice)
  call mpas_pool_get_array(input, 'xice', xice)
+ call mpas_pool_get_array(input, 'snowc', xice)
+ call mpas_pool_get_array(input, 'snowh', xice)
+ call mpas_pool_get_array(input, 'snow', xice)
```
mgduda: It looks like these three calls to `mpas_pool_get_array` are all setting the local variable `xice`. Should these calls be setting `snowc`, `snowh`, and `snow`?

mgduda (changes requested):

src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
smileMchen: yes you are right ! I apologize for the stupid error. I will change them.

smileMchen: @mgduda 
I have corrected the errors. Thanks !

mgduda: Thanks for the quick turnaround! I've squashed the corrections into the original commit based the PR branch on the v8.2.0 tag.

mgduda:
@smileMchen When testing the changes in this PR with a global 60-km mesh using ERA5 data valid on 2025-01-01, it looks like snow over land points is getting removed. In the screenshot below, the top panel is the `snowc` field from the current `release-v8.4.0` branch, and the bottom panel is the `snowc` field after locally merging the changes from this PR.
snowc_pr1438
Is this expected?

smileMchen:
@mgduda 

This is not what I expect. I will take a look and get back to you. snowc is a diagnostic variable derived from SWE and snowh so its initial value is not used. However, if SWE is correct, I expect snowc should be correct, too.

smileMchen:
image

smileMchen:
@mgduda 
I have corrected the error in my previous code. Please see the plot I posted above, which shows SWE and SNOWC before and after the fix. 

Can I issue the command "git push --force upstream initsnow" to commit the modified code? I am concerned that your previous work on this PR may be lost if I issue this command. 

Thank you.

mgduda:
@smileMchen Thanks for looking into this!

Regarding the push of the changes, I think it may depend on how the changes were committed. If you run
```
git push upstream initsnow:initsnow_fixed
```
that will push your corrected branch to a new branch named `initsnow_fixed`, and I can take a quick look at the commit history. Then, I could handle the overwriting of the `initsnow` branch as needed.

mgduda (approved):
Looks good to me -- thanks for the fix for this issue!

PR #1440, opened 2026-04-03
This PR fix the error in PR#1438 

smileMchen:
This PR fixes the issue in PR#1438 that all snow information, snow water equivalent(swe), snow depth(snowh), and snow cover(snowc),   over land areas  are removed mistakenly. 

In the initialization of snow over seaice areas, we use xice as a criteria to determine whether snow should exit or not. When xice at a grid cell is smaller than a specified threshold, this grid cell is treated as a water point (swe=0, snowh=0, snowc=0)

However, xice ==0  over land. When using the above approach to determine snow, we mistakenly  remove snow from land areas. 

The present PR  fixes the mistake and results are reasonable.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_initialize_real.F
@@ -606,6 +606,7 @@
  real(kind=RKIND),pointer:: tsk_seaice_threshold
  real(kind=RKIND),dimension(:),pointer  :: vegfra
  real(kind=RKIND),dimension(:),pointer  :: seaice,snoalb,xice
+ real(kind=RKIND),dimension(:),pointer  :: snow, snowh, snowc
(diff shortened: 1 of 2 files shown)


smileMchen:
@mgduda 
Please take a look at the PR and let me know if you have any concerns. Thanks a lot !

mgduda:
> @mgduda Please take a look at the PR and let me know if you have any concerns. Thanks a lot !

@smileMchen I've added the fix from this PR to set snow to zero only over water points to PR #1438. Could you check that the changes in 20750685 (currently a separate commit in PR #1438) are as you intended them?

If everything looks good for PR #1438, I'll suggest we close this PR and continue with #1438.

smileMchen:
@mgduda 

Yes  all looks fine in #1438. Please close this PR and continue with PR#1438.  Thanks a lot !

mgduda:
Great -- we'll proceed with PR #1438!

PR #1441, opened 2026-04-06
Fix premature assignment to optional argument before checking existence

islas:
This PR fixes an issue within `mpas_get_stream_filename()` where the optional argument `ierr` was being assigned to before checking if it was passed in.

The routine now checks if `ierr` is present, and if so initializes it to `MPAS_STREAM_MGR_NOERR`.

### DIFF
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -4073,7 +4073,9 @@ module mpas_stream_manager
         type (MPAS_TimeInterval_type) :: filename_interval
         type (MPAS_Time_type) :: now_time
 
-        ierr = 0
+        if ( present(ierr) ) then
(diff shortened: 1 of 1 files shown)


src/framework/mpas_stream_manager.F
```diff
@@ -4107,6 +4107,10 @@ subroutine mpas_get_stream_filename(manager, streamID, when, blockID, filename,
             end if
```
islas, line 4109: This `if( present() )` statement above (line 4104) makes use of `ierr` and assigns to it instead of `ierr_local` - @mgduda I think it may make sense to also modify the assignment so that the final `ierr` assignment has the correct error code.

mgduda (approved):
Thanks for the fix, and for spotting this issue!

PR #1442, opened 2026-04-08
Update config_init_case Registry.xml description to include LES case

mgduda:
This PR updates the `Registry.xml` entry for the `config_init_case` namelist option in the `init_atmosphere` core so that its `description` and `possible_values` attributes reflect the addition of the new LES initialization case (case 10).


abishekg7 (approved):
Looks good.

PR #1443, opened 2026-04-09
Feature ci nsys

cenamiller:
Manual-only GPU workflow that builds MPAS-A with NVHPC + OpenACC in the CUDA hpcdev image, 
runs a very short atmosphere case under `nsys profile`. Default resolution 120km, `config_run_duration` `0_00:12:00`
so the run covers one dynamical timestep. Dispatch inputs allow a different resolution or duration when needed.
 
Helper script .github/scripts/run-nsys-profile.sh wraps mpirun with the same MPI environment pattern as run-mpas


cenamiller:
Sorry went to MPAS-Dev instead of my fork!

PR #1445, opened 2026-04-15
Fix for parallel reproducibility issue with `3d_smagorinsky`

abishekg7:
This PR fixes parallel reproducibility issues (For ex, runs with 32 and 64 MPI tasks) when `config_les_model= 3d_smagorinsky`. This issue affects both CPU and GPU runs. This issue does not seem to affect parallel reproducibility when `config_les_model= prognostic_1.5_order`. 

This PR firstly introduces a new halo exchange group in `mpas_atm_halos` in order to perform halo exchanges for the two eddy viscosity fields relevant to LES model runs. This group is added to both `mpas_dmpar` and `mpas_halo` halo exchange methods in `mpas_atm_halos`. Then a halo exchange is performed using this group immediately after the call to `les_models` in `atm_compute_dyn_tend`.

### Background

With trial and error, it was observed that performing halo exchanges of the `eddy_visc_horz` and `eddy_visc_vert` fields, immediately after the call to `les_models` seemed to fix this error. The exact mechanism of this error propagation has not been fully understood yet and would need to be revisited. 

However, the issue originating with `eddy_visc_horz` might involve this [line](https://github.com/MPAS-Dev/MPAS-Model/blob/c0f839d525eaecf8be7ed59cd7f20bf98e716637/src/core_atmosphere/dynamics/mpas_atm_dissipation_models.F#L715). Both serialbox and simpler print debugging point to issues originating from this loop. The max(cell1,cell2) values in the loop seems to exceed `cellEnd`. 

It is also not yet understood why these discrepancies are not observed with `config_les_model= prognostic_1.5_order`

### DIFF
--- a/src/core_atmosphere/mpas_atm_halos.F
+++ b/src/core_atmosphere/mpas_atm_halos.F
@@ -122,6 +122,9 @@ module mpas_atm_halos
          call mpas_dmpar_exch_group_add_field(domain, 'dynamics:theta_m,pressure_p,rtheta_p', 'rtheta_p', &
                                               timeLevel=1, haloLayers=(/1,2/))
 
+         call mpas_dmpar_exch_group_create(domain, 'dynamics:eddy_visc_horz,eddy_visc_vert')
+         call mpas_dmpar_exch_group_add_field(domain, 'dynamics:eddy_visc_horz,eddy_visc_vert', 'eddy_visc_horz', timeLevel=1, haloLayers=(/1,2/))
+         call mpas_dmpar_exch_group_add_field(domain, 'dynamics:eddy_visc_horz,eddy_visc_vert', 'eddy_visc_vert', timeLevel=1, haloLayers=(/1,2/))
 
          call mpas_dmpar_exch_group_create(domain, 'dynamics:exner')
          call mpas_dmpar_exch_group_add_field(domain, 'dynamics:exner', 'exner', timeLevel=1, haloLayers=(/1,2/))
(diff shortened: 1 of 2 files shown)


src/core_atmosphere/mpas_atm_halos.F
```diff
@@ -122,6 +122,8 @@ subroutine atm_build_halo_groups(domain, ierr)
          call mpas_dmpar_exch_group_add_field(domain, 'dynamics:theta_m,pressure_p,rtheta_p', 'rtheta_p', &
                                               timeLevel=1, haloLayers=(/1,2/))
 
+         call mpas_dmpar_exch_group_create(domain, 'dynamics:eddy_visc_horz,eddy_visc_vert')
+         call mpas_dmpar_exch_group_add_field(domain, 'dynamics:eddy_visc_horz', 'eddy_visc_horz,eddy_visc_vert', timeLevel=1, haloLayers=(/1,2/))
```
mgduda: I haven't tested, but it looks from inspection that the second argument should match the name of the halo exchange group created in the line above this one, and the third argument should be the name of the field to add (not a comma-separated list). @abishekg7 Could you double-check this code?

mgduda (changes requested):

src/core_atmosphere/mpas_atm_halos.F
abishekg7: Thanks! I've fixed this and also tested that `mpas_dmpar` results match `mpas_halo`

PR #1448, opened 2026-04-29
Fix crash in dynamics due to Fortran standard violation

kuanchihwang:
At [line 1970](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/dynamics/mpas_atm_time_integration.F#L1969-L1971) inside the `atm_srk3` time stepping subroutine, if the `DO_PHYSICS` macro is undefined, the `diag_physics` pointer will not be initialized by the call to `mpas_pool_get_subpool`, and therefore its pointer association status will remain undefined.

However, at [line 2250](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/dynamics/mpas_atm_time_integration.F#L2250-L2257), this pointer is used as an actual argument to call `atm_compute_dyn_tend`, which constitutes a Fortran standard violation.

Quoted from Fortran 2023,

> 15.5.2.4 Argument association
> Except in references to intrinsic inquiry functions, a pointer actual argument that corresponds to a nonoptional nonpointer dummy argument shall be pointer associated with a target.

This bug can lead to a runtime crash for models that use MPAS as a dynamical core (e.g., CAM, CAM-SIMA).

Fix this issue by adding the `pointer` attribute to the `diag_physics` dummy argument for the `atm_compute_dyn_tend` subroutine.

In addition, two one-line changes have also been introduced to better conform to the Fortran best practices.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -5809,7 +5809,7 @@ module atm_time_integration
       ! Dummy arguments
       !
       type (mpas_pool_type), intent(inout) :: tend
-      type (mpas_pool_type), pointer :: tend_physics
+      type (mpas_pool_type), pointer, intent(in) :: tend_physics
       type (mpas_pool_type), intent(in) :: state
       type (mpas_pool_type), intent(in) :: diag
       type (mpas_pool_type), intent(in) :: mesh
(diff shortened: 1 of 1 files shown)


mgduda:
@kuanchihwang Could you rebase this PR branch off of the `v8.4.0` tag? Also, would you see if there's additional details that could be included in the one-line summary of each commit message (while staying under 80 characters)? This is helpful, e.g., when scanning through the commit log with `git log --oneline`. Adding variable, function, or file names, etc. helps to make this output easier to sift through with, e.g., `grep`.

kuanchihwang:
I rebased the PR branch against the `v8.4.0` tag. I also added clues about subroutine names to the commit titles.

mgduda (approved):
Thanks for uncovering this issue, and thanks for the fix!

mgduda (approved):

PR #1451, opened 2026-05-07
Fix small memory leaks

DusanJovic-NOAA:
This PR fixes small memory leaks in various parts of the model, mostly within the framework. I ran the model compiled with the address sanitizer compiler flag and noticed hundreds of memory leak reports. All of them are located in the code executed during initialization, and they do not contribute to ever increasing memory consumption during model execution. So fixing them is not critical. Major benefit is cleaning up the log file to make it easier to spot potential real memory leaks.

Here is the initial sanitizer log file, showing all leaks originating in the model code. I suppressed leaks from the MPI library and several leaks originating in the ezxml library.

[dev_leak.0.txt](https://github.com/user-attachments/files/27493550/dev_leak.0.txt)

### DIFF
--- a/src/external/ezxml/ezxml.c
+++ b/src/external/ezxml/ezxml.c
@@ -481,6 +481,7 @@ ezxml_t ezxml_parse_str(char *s, size_t len)
     int l, i, j;
 
     root->m = s;
+    root->len = -1; // so we know to free s in ezxml_free()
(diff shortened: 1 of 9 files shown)


mgduda (approved):
Thanks so much for tracking down and fixing these leaks, and thanks especially for separating each of the fixes into its own commit with a thorough explanation!

PR #1453, opened 2026-05-07
Provide default initialization for members of the met_data derived type

mgduda:
This PR adds default initialization for the members of the `met_data` derived type in the `init_atm_read_met` module. Without well-defined values for some members, unpredictable floating-point exceptions were possible, e.g.,
```

  Backtrace for this error:
  #0  0x15494ec288ff in ???
  #1  0x59ddc0 in __init_atm_read_met_MOD_read_next_met_field
        at src/core_init_atmosphere/mpas_init_atm_read_met.F:388
  #2  0x4ccb7a in __init_atm_cases_MOD_init_atm_case_lbc
        at src/core_init_atmosphere/mpas_init_atm_cases.F:6218
  #3  0x58a373 in __init_atm_cases_MOD_init_atm_setup_case
        at src/core_init_atmosphere/mpas_init_atm_cases.F:327
  #4  0x497807 in __init_atm_core_MOD_init_atm_core_run
        at src/core_init_atmosphere/mpas_init_atm_core.F:92
  #5  0x406084 in __mpas_subdriver_MOD_mpas_run
        at src/driver/mpas_subdriver.F:416
  #6  0x4046da in mpas
        at src/driver/mpas.F:20
  #7  0x404745 in main
        at src/driver/mpas.F:10
```
By initializing the members of the met_data type, these unpredictable errors can be avoided.

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_read_met.F
+++ b/src/core_init_atmosphere/mpas_init_atm_read_met.F
@@ -25,17 +25,37 @@ module init_atm_read_met
 
    ! Derived types
    type met_data
-      integer                       :: version, nx, ny, iproj
-      real (kind=real32)                 :: xfcst, xlvl, startlat, startlon, starti, startj, &
-                                       deltalat, deltalon, dx, dy, xlonc, &
-                                       truelat1, truelat2, earth_radius
-      real (kind=real32), pointer, dimension(:,:) :: slab
-      logical                       :: is_wind_grid_rel
-      character (len=9)             :: field
-      character (len=24)            :: hdate
-      character (len=25)            :: units
-      character (len=32)            :: map_source
-      character (len=46)            :: desc
+      integer :: version = 5, &         ! Format version (must =5 for WPS format)
+                 nx      = 0, &         ! First (x) dimension of 2-d array 'slab'
+                 ny      = 0, &         ! Second (y) dimension of 2-d array 'slab'
+                 iproj   = PROJ_LATLON  ! Code for projection of data in array
+
+      real (kind=real32) :: xfcst        = 0.0_real32, &  ! Forecast hour of data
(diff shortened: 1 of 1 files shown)


src/core_init_atmosphere/mpas_init_atm_read_met.F
```diff
@@ -25,17 +25,37 @@ module init_atm_read_met
...
-                                       deltalat, deltalon, dx, dy, xlonc, &
-                                       truelat1, truelat2, earth_radius
-      real (kind=real32), pointer, dimension(:,:) :: slab
-      logical                       :: is_wind_grid_rel
-      character (len=9)             :: field
-      character (len=24)            :: hdate
-      character (len=25)            :: units
-      character (len=32)            :: map_source
-      character (len=46)            :: desc
+      integer :: version = 5, &         ! Format version (must =5 for WPS format)
```
islas, line 28: Given the specification of `real32` for variables to use standardized format size, would it be worth using `int32` as well for integers?

I know that's not quite in the scope of this PR to set default values, but I thought it may be worth mentioning here, especially as it seems like the specification of variable size as it pertains to reading binary data from a fortran record.

mgduda: It's a longer story why we use `(kind=real32)` for `real` variables but use the default `integer` kind. In short, MPAS-A has the ability to switch between default 32-bit reals or default 64-bit reals at compile time, but no similar option exists to switch between different default integer sizes. So the `real` members of the `met_data` derived type needed to be pinned to a fixed size, but it was (probably?) assumed at the time we made that change that `integer` variables could safely be assumed to always be the same size.

Anyway, as you said, adding `(kind=int32)` is probably outside the scope of this PR, which simply aims to ensure that all members have a reasonable default initialization.

islas (approved):

PR #1454, opened 2026-05-15
Added SST field to mpas_init_atm_cases.F

calvinwallc:

This allows sea surface temperatures to be ingested in to
init files. Previously, SSTs were not read in from intermediate files. Therefore, all SSTs defaulted to the
skin temperature, which produced unphysically
cold SST values in the Arctic, where the skin
temperature above sea ice is well below freezing. Additionally, modified mpas_init_atm_surface.F such that SST update files must be made with SSTs, rather than either SSTs or skintemps.

A more in depth discussion of this issue and solution can be found in the mpas forum at https://forum.mmm.ucar.edu/threads/sst-bias-underneath-sea-ice.25524/page-2#post-61310

### DIFF
--- a/src/core_init_atmosphere/mpas_init_atm_cases.F
+++ b/src/core_init_atmosphere/mpas_init_atm_cases.F
@@ -3585,6 +3585,7 @@ module init_atm_cases
              trim(field % field) == 'PRESSURE' .or. &
              trim(field % field) == 'SNOW' .or. &
              trim(field % field) == 'SEAICE' .or. &
+             trim(field % field) == 'SST' .or. &
              trim(field % field) == 'SKINTEMP') then
 
             if (trim(field % field) == 'SM000010' .or. &
@@ -4560,6 +4561,13 @@ module init_atm_cases
                latPoints => latCell
                lonPoints => lonCell
                call mpas_pool_get_array(fg, 'xice', destField1d)
+               ndims = 1
+            else if (trim(field % field) == 'SST') then
+               call mpas_log_write('Interpolating SST')
+               nInterpPoints = nCells
+               latPoints => latCell
+               lonPoints => lonCell
+               call mpas_pool_get_array(fg, 'sst', destField1d)
                ndims = 1
             else if (trim(field % field) == 'SKINTEMP') then
                call mpas_log_write('Interpolating SKINTEMP')
--- a/src/core_init_atmosphere/mpas_init_atm_surface.F
+++ b/src/core_init_atmosphere/mpas_init_atm_surface.F
@@ -213,7 +213,7 @@
  do while (istatus == 0)
 
     !sea-surface data:
-    if((trim(field % field) == 'SKINTEMP') .or. (trim(field % field) ==  'SST')) then
+    if(trim(field % field) ==  'SST') then
 !      call mpas_log_write('... Processing SST:')
        sst(1:nCells) = 0.0_RKIND
        destField1d => sst


smileMchen:
SST has already been included in the MPAS initial condition.  SKINTEMP is used to replace SST only when SST is missing in the analysis/reanalysis products used to create initial condition for MPAS.   

ERA5 includes both SKINTEMP and SST, and the difference between SST and SKINTEMP does reflect the reasonable situation in winter.  However, **ERA5 is different to other reanalysis products  in that  it has no snow over seaice area in the winter, which is physically unreasonable.** When we use ERA5 to create initial condition for MPAS run in wintertime, snow missing in the initial condition would lead to unrealistic results, ---- in this case, I believe we need to be cautious to directly compare MPAS simulations of SKINTEMP with that from ERA5. 

Furthermore, areas covered by seaice is treated as land points in Noah. Therefore, it is not reasonable to use SST as SKINTEMP at such points.     

Attached are plots of SST, SKINTEMP and SNOW DEPTH from ERA5 in the winter of 2013-01-23. This is an example of ERA5 data in the winter. 

image

calvinwallc:
Hi Mchen! 

I may not understand the MPAS source code as well as I should, but if I'm not mistaken, if SST were to be included in the MPAS initial conditions for a real-data test case (ie config_init_case==7), it should be read in to the subroutine init_atm_case_gfs in the mpas_init_atm_cases.F file somewhere in the lines 3539 to 3588. It should then be interpolated to the MPAS mesh somewhere in the lines 3690 to 4571. The SST is not referenced anywhere in these lines.

However, the SST is set to 0 everywhere at line 3454 of mpas_init_atm_cases.F, and is not updated until line 4769, where it is set to be equal to the skintemp if it is not found in the input data, like you say. What I am contending is that it is not found in the input data, because it is not read from the input data. My fix simply reads the SST from the input intermediate files, such that the SST does not default to the skintemps on line 4769.

I agree with you that it is not reasonable to use skintemps as SSTs or vice versa over ice points. My fix prevents that from happening. In MPASv8.4, the SSTs are taken to be the skintemps for all cases where config_init_case = 7 because the SSTs are not read in from the intermediate files. This occurs even though ERA5 does include SST information, and this SST information is being included in the intermediate files generates using the era5_to_int.py utility. 

I hope this makes sense! I am happy to answer any further questions on this topic.

smileMchen:
Hi Calvin,

Thank you for the detailed description. You are correct that SST is never processed independently. We always use SKINTEMP to represent SST. My understanding is that SST and SKINTEMP are identical over open ocean points (without sea ice), but they differ significantly in areas covered by sea ice. We use SKINTEMP to represent SST everywhere because, as I mentioned previously, Noah and Noah-MP treat sea ice as “land.” Over sea-ice-covered regions, SST becomes invalid, while SKINTEMP is actively involved in the land surface physics. In this situation, SKINTEMP, rather than SST, is required as the initial condition. This is why in MPAS, SKINTEMP is always used to represent SST.  

Please let me know if you have more questions. Thanks. 

((PS: my previous message that "SKINTEMP is used to replace SST only when SST is missing" is not correct. Thank you for clarifying this issue. As we discussed in the MPAS Forum, I still believe that the poor results of your case, which is initialized by ERA5, may be attributed to unreasonable snow from ERA5. )

calvinwallc:
Hi Mchen!

I appreciate the clarifications here! I would contend, however, that the SST beneath the sea-ice does matter for the skin-temperatures. I don't know exactly where in the MPAS source code, but I believe that the SSTs are used by the Noah land model to determine the bottom "soil temperature" of sea-ice points. To illustrate the effect that SSTs underneath sea-ice have on the temperature in the Arctic, I performed two simulations, one using the fix I propose in the pull request here, and one without the fix, with no other changes to the simulations. Both simulations use MPASv8.4 at a uniform 60 kilometer resolution, with the default mesoscale-reference physics package, with config_sst_update turned on.

The point you make about the difference between MPAS and ERA5 definitions of the skin-temperature is well taken, so I attach here plots of the MPAS and ERA5 two meter temperature, which should have a uniform definition and be directly comparable. The bias in the 2m temperature in MPAS as compared to ERA5 without the fix I propose (the file t2m_bias_comp_nofix.png) is about -4K in the Arctic average (black curve), driven largely by a -7K bias in the area above sea ice (orange curve). However, if we ingest the SSTs properly in to MPAS via the edits in my pull request, (the file t2m_bias_comp.png) the Arctic average bias is reduced to between -2 and -2.5K, as the ice-area bias is reduced to -3.5K. Additionally, all t2m values return to near ERA5 values after about 10 days when my fix is implemented, which is not the case in the base model.

t2m_bias_comp_nofix
t2m_bias_comp

I would also like to clarify that my fix does not replace the skintemp initialization with SST initialization. There are two things that need to be initialized, the skintemp above the sea-ice, and the SSTs below the sea-ice. In the base model, both the skintemp above the ice and the SST below the ice are set to be the skintemp, which is unphysical. With my fix, the skintemp above the ice is still initialized correctly, but the SST below the sea-ice is also initialized correctly at near freezing values.

**To summarize:** In the base model, the SST below sea-ice is initialized as the skintemp above the ice, which is ~20 degrees below freezing. These SSTs are used in some way (I believe as soil temperature) by the Noah land model to calculate the temperature above the sea-ice. As these SSTs are much too cold, this results in a strong cold bias in the two meter temperatures above sea-ice in the Arctic. My fix correctly initializes the SSTs underneath sea-ice, while maintaining the correct initialization of skin temperature above the sea-ice, which ameliorates the cold bias in the Arctic.

smileMchen:
@calvinwallc 

Thanks for the update and the plots, which are really helpful. I agree with you that the model results are greatly improved with your fix, and your argument of SST impact on TSK does make sense, although  I don't know exactly how Noah handles SST beneath seaice.  

Thanks for raising an important issue on LSM physics over seaice areas. Previously we mostly focus on studies in mid- and lower latitudes, and didn't pay enough attention to seaice physics. However, for global MPAS implementation, I suppose LSM  physics over seaice is important, and we do need to fix any possible bugs related to seaice physics.        

I will copy our LSM expert and he can provide more in-depth information. 

@cenlinhe 
Cenlin, would you please take a look at Calvin's PR and the information he provided? His results are encouraging, although the reason behind the improvement remains unclear for me. I have a few questions regarding this issue: 
(1) In our original codes, we use SKINTEMP as initial condition over seaice areas. We treat seaice as 'land' and calculate TSK. In this situation, how does the model handle SST?  
(2) Calvin argued that we should separately initialize temperature below and above seaice using SST and SKINTEMP respectively, ---- I intend to agree with him because this is physically reasonable. Please let me know what you think. 
(3) I am still concerned of snow missing over seaice area in the initial condition. Without snow, the model may yield cold bias of tsk, is this correct?

cenlinhe:
I double checked the seaice_noah module code,
(1) The seaice_noah code only uses skintemp as input to drive the seaice surface processes. The deep seaice temperature is not interacting with SST beneath seaice but is prescribed to -1.8degC here: https://github.com/MPAS-Dev/MPAS-Model/blob/515e5203d8dc7b63192e50678284bd7da2ef6927/src/core_atmosphere/physics/physics_wrf/module_sf_noah_seaice_drv.F#L354-L355 . SST is not taken by the seaice_noah module. I do not understand why Calvin's results showed such a large difference in T2 simulations from the seaice code either.
(2) I agree that from the physical perspective, we should include Calvin's fix which is physically more reasonable and correct. However, a deeper code dig-out needs to be done to understand how the SST initialization update propagates into the seaice module.
(3) For snow on seaice, the seaice_noah module does include snow process over seaice depending on different SEAICE_SNOWDEPTH_OPT namelist options here: https://github.com/MPAS-Dev/MPAS-Model/blob/515e5203d8dc7b63192e50678284bd7da2ef6927/src/core_atmosphere/physics/physics_wrf/module_sf_noah_seaice_drv.F#L293-L336 . I guess users will need to choose the correct option to activate the snow on seaice.

weiwangncar:
Apparently the differences between SKINTEMP and SST depend on the source data. For GFS, there is only one surface temp in the analysis and forecast file, which we call it SKINTEMP. We believe the SST analysis (performed once per day at NCEP) has been merged with the field we call SKINTEMP since a few years ago. But it is different with ERA5. They are two fields from the archive, one is call 'skin temperature', and the other 'sea surface temperature'. And they are different over open water. When we run our models with constant SST, we think it is SST from ERA5 we should be using. In WRF, when there is valid SST field from the input file, we swap the SKINTEMP (named TSK) over water with SST. In MPAS init_atmosphere, the assumption is ok for GFS data, but probably not general enough to accommodate a separate SST input.

weiwangncar:
@cenlinhe Even though SST is not used in seaice module, it is used to compute skintemp over fractional seaice area, where skintemp is a combination of open water temp and temp over seaice. A warm SST over this area would contribute to warmer skintemp. This is done in mpas_atmphy_driver_seaice.F.

smileMchen:
@cenlinhe 

Thanks a lot for your comments and information.  

(1) If SST is specified to 271.36 below seaice and serves as the bottom boundary condition for sea ice physics (which I believe is reasonable), then  I suppose it is not necessary to separately initialize and update SST over sea ice area ( note that over open water areas, SST and SKINTEMP are identical). 

(2) I am still concerned of the snow issue in ERA5, ---- the permanently missing snow over sea ice area in ERA5 is unreasonable. Note that GFS always includes snow information. 

(3) The default option of SEAICE_SNOWDEPTH_OPT is always used in my cases, and it remains unclear to me how other options may affect the results. I will take a look and keep you updated about the results.

Thanks again.

PR #1455, opened 2026-05-18
Implementation of MYNN-EDMF and MYNN-SFC submodules

joeolson42:
This PR adds the official MYNN-EDMF and MYNN-SFC submodules to MPAS with via manage_externals. Some notes on the implementation are below:

1. The submodules were added simultaneously due to the common package (bl_mynnedmf_in) used for both schemes (following the previous implementations). 
2. Two additional packages are added to control the allocation of 10 optional 3d output arrays and 5 tke budget arrays.  
3. a new cloudiness option is added to handle the processing of the MYNN-EDMF subgrid clouds analogous to radiation driver in WRF.
4. The implementation was complicated by the earlier implementation of MYNN schemes, so renaming of key modules and subroutines was required to allow both submodules to work as alternatives to the other versions. The other schemes should still work but will hopefully be removed.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
@@ -16,6 +16,7 @@
 
  use bl_mynn,only: bl_mynn_init
  use module_bl_mynn,only: mynn_bl_driver
+ use module_bl_mynnedmf_driver,only: mynnedmf_driver
  use module_bl_ysu
 
  implicit none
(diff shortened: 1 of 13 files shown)


src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
```diff
@@ -957,16 +1339,148 @@ subroutine driver_pbl(itimestep,configs,mesh,sfc_input,diag_physics,tend_physics
...
+                  sm3d       = sm3d_p       , spp_pbl      = spp_pbl        , pattern_spp   = pattern_spp_pbl  , &
+                  do_restart         = config_do_restart   ,                                              &
+                  do_DAcycling       = config_do_DAcycling ,                                              &
+                  initflag           = initflag            ,                                              &
+                  bl_mynn_tkeadvect  = bl_mynn_tkeadvect   ,                                              &
+                  bl_mynn_tkebudget  = bl_mynn_tkebudget   ,                                              &
+                  bl_mynn_cloudpdf   = bl_mynn_cloudpdf    ,                                              &
+                  bl_mynn_mixlength  = bl_mynn_mixlength   ,                                              &
+                  bl_mynn_closure    = bl_mynn_closure     ,                                              &
+                  bl_mynn_stfunc     = bl_mynn_stfunc      ,                                              &
```
mgduda: When this PR branch is merged to the HEAD of the `develop` branch, I'm seeing a compilation failure with the GNU compilers:
```
mpas_atmphys_driver_pbl.F:1445:39:

 1445 |                   bl_mynn_stfunc     = bl_mynn_stfunc      ,                                              &
      |                                       1
Error: Keyword argument 'bl_mynn_stfunc' at (1) is not in the procedure; did you mean 'bl_mynn_output'?
```

joeolson42: bizarre... I must not have done a "git add mpas_atmphys_driver_pbl.F" before the commit. That line is not in my code. Will update shortly.

src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
```diff
@@ -788,12 +1155,27 @@ subroutine driver_pbl(itimestep,configs,mesh,sfc_input,diag_physics,tend_physics
...
                    bl_mynn_mixscalars,  &
+                   bl_mynn_mixaerosols, &
+                   bl_mynn_mixnumcon,   &
                    bl_mynn_cloudmix,    &
                    bl_mynn_mixqt,       &
-                   bl_mynn_tkebudget
+                   bl_mynn_ess,         &
+                   bl_mynn_tkebudget,   &
+                   icloud_bl,           &
+                   spp_pbl
```
mgduda: Declaring `icloud_bl` and `spp_pbl` here as local variables overrides the variables of the same name imported from the `mpas_atmphys_vars` module. Since these local variables are pointers, and since they're not set by a call to `mpas_pool_get_config` when using `config_pbl_scheme = 'bl_mynn'`, we'll get a segmentation fault (the apparent location of the segfault may depend on compiler).

One option to avoid this issues might be to choose different names for the local pointer variables `icloud_bl` and `spp_pbl`, and to use those different names only in MYNN-EDMF cases (leaving behavior with the MYNN scheme unchanged).

src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
mgduda: Thanks for the update. All looks good compilation-wise with the changes in [5a2cdc6](https://github.com/MPAS-Dev/MPAS-Model/pull/1455/commits/5a2cdc6edbc136bdb1b44092de773d770af2c428).

src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F
joeolson42: I think it's best to remove icloud_bl from the pbl driver entirely (at least for "bl_mynnedmf"). icloud_bl really only has meaning in WRF (package declaration and coupling to radiation). In MPAS, it has no functionality, even for the cloudiness driver. I'll update the MYNN-EDMF submodule, the WRF registry and pbl driver, and this PR, to remove (or lessen) the need for icloud_bl. As for spp_pbl, the spp development is almost finished, so it will be available soon and will not be hardwired in the mpas_atmphys_vars module inthe near future. For now, I can remove that line as well.

joeolson42: OK, those lines have been removed, but I'm not sure it's improved. Those "options" are still hardcoded in the mpas_atmphys_vars module, but that was the original design. The other MYNN scheme should still work.

mgduda: Agreed that it's not optimal to have these options hard-wired in `mpas_atmphys_vars`. Once we've got a bit more time, I suppose we can take a closer look at how to handle those in a better way.

mgduda:
I think the code is generally looking good from a software perspective. With PR [#1460](https://github.com/MPAS-Dev/MPAS-Model/pull/1460), which went in to the v8.4.1 release and modified packages for various surface fields, I think there will now be merge conflicts with the `develop` branch. @joeolson42 would you be able to rebase this PR branch onto the HEAD of the `develop` branch to resolve these conflicts? Or, if you'd like, I can help with this, as I'm familiar with the changes from #1460 that caused the merge conflicts.

Also, to accommodate changes in #1460, I think we'll need to add `'sf_mynnsfclay'` to the list of valid surface layer schemes in the [logic for the sfclayer package](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.1/src/core_atmosphere/Registry.xml#L488-L490).

joeolson42:
I'll take care of it. I really like the idea of the sfclayer package--thanks for that! I'll fix the logic after the merge.

joeolson42:
@mgduda About our conversation about QKE advection... It would be nice to get that in as part of this PR, but I'm not sure how best to do it. It would be nice to keep TKE advection as a namelist option in case it has numerical stability problems in regional domains, like it did in WRF (especially for level 3). But if we follow the way it was done in WRF:

1.  we used an additional array (QKE_ADV) which is what was actually advected,
2. then QKE was set equal to QKE_ADV at the beginning of the MYNN-EDMF,
3. then QKE_ADV was set equal to QKE at the end of the MYNN-EDMF 
4. (2) and (3) were only performed if the namelist variable "tke_advect" was set to true

This worked and we could continue with this method, but I was wondering if you had a cleaner idea that wouldn't require an additional array. Something to think about.

PR #1456, opened 2026-05-19
Feature/regex match speedup

yjwu890355:
Speed up regex matching in diagnostic output

Compile the diagnostic output regex pattern once and reuse it across all stream
name checks instead of calling regcomp() for every comparison. This avoids
repeated regex compilation overhead in regex_matching.c, which could account for
more than 30% of execution time.

Timer information shows 10x speedup in 'diagnostic_fields' on CWA Fujitsu FX1000 machine

Credit: Yu-Tze Hong <aspen.hong@gmail.com>

### DIFF
--- a/src/framework/mpas_stream_list.F
+++ b/src/framework/mpas_stream_list.F
@@ -223,9 +223,18 @@ module mpas_stream_list
     !>  will return .false.
     !
     !-----------------------------------------------------------------------
+
+    ! Optimized; changes from original:
+    !   1. Pattern is compiled ONCE before the node loop (not once per node).
+    !   2. Compilation uses the cross-call cache in regex_matching.c, so each
+    !      unique pattern string is compiled at most once for the entire run.
+    !   3. regex_free_cached is a no-op (cache owns lifetime) but is called at
+    !      every exit point for forward-compatibility and readability.
+    !   4. Three interface blocks replace the single check_regex_match interface.
+
     logical function MPAS_stream_list_query(list, streamPattern, stream, ierr) result(found) !{{{
 
-        use iso_c_binding, only: c_char, c_int
+        use iso_c_binding, only: c_char, c_int, c_intptr_t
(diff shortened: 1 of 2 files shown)


mgduda:
@yjwu890355 Thank you for identifying this opportunity to speed up regular expression matching. Under the assumption that most calls to the `check_regex_match` function will not be using a regular expression, but will instead be checking a simple non-regex string, I think the changes in this commit would offer similar speed-ups but with far less code: ae7a3895 .

Would you be willing to consider the code changes in commit ae7a3895 instead?

yjwu890355:
Yes, thank you for the suggestion. Based on my testing, this approach also gives about a 10x speedup in diagnostic_fields (200s ->14s) on the CWA Fujitsu FX1000 machine. I agree that commit ae7a389 is a simpler and more effective solution.

mgduda:
> Yes, thank you for the suggestion. Based on my testing, this approach also gives about a 10x speedup in diagnostic_fields (200s ->14s) on the CWA Fujitsu FX1000 machine. I agree that commit [ae7a389](https://github.com/MPAS-Dev/MPAS-Model/commit/ae7a389594be1800f5735d3078e5d42c606b8349) is a simpler and more effective solution.

I've just opened PR #1466 with the simplified optimization. Please let me know how you and Yu-Tze Hong would like to be credited (name, GitHub ID, institution, etc.). Thanks so much again for finding this optimization!

yjwu890355:
Thanks for opening PR #1466 and for the simplified implementation.

For the credit, please use:

Ying-Jhang Wu (@yjwu890355), Central Weather Administration (CWA), Taipei, Taiwan
Yu-Tze Hong (@aspenhong), Central Weather Administration (CWA), Taipei, Taiwan

Thanks again!

mgduda:
Since PR #1466 has been merged to `develop`, I'll go ahead and close this PR.

PR #1457, opened 2026-05-22
Add a time-dependent function for CO2 (Update PR #1173)

jihyeonjang:
This PR builds on #1173, which proposed a smoothly increasing, time-dependent CO2 concentration function. The function form is a close fit for past CO2 growth and extrapolates it for the future. 

This update introduces a new radiation utility module for subroutines or functions used by both LW and SW radiation schemes. 

Summary of changes: 
- rebase PR #1173 changes onto the current develop branch 
- add a new utility module (`mpas_atmphys_radiation_utils.F`)
- move `radconst` and `co2_estimate` into `mpas_atmphys_radiation_utils`
- rename `co2` to `co2_val` to avoid conflict with the non-1D `co2` array used by the CAM radiation scheme
- compute `co2_val` in `radconst` 
- call `radconst` before `rrtmg_lwrad` and `rrtmg_swrad` 
(Moving `radconst` from `mpas_atmphys_driver_radiation_sw.F90` to `mpas_atmphys_radiation_utils.F90`  removes the build dependency of LW on the SW module.)

A future update will implement reading GHG values from prescribed tables, following the WRF approach. The table year-dependent GHG is based on observed values for past years and official climate scenarios for future years.

### DIFF
--- a/src/core_atmosphere/physics/Makefile
+++ b/src/core_atmosphere/physics/Makefile
@@ -46,6 +46,7 @@ OBJS = \
 	mpas_atmphys_manager.o             \
 	mpas_atmphys_o3climatology.o       \
 	mpas_atmphys_packages.o            \
+	mpas_atmphys_radiation_utils.o     \
 	mpas_atmphys_rrtmg_lwinit.o        \
 	mpas_atmphys_rrtmg_swinit.o        \
 	mpas_atmphys_sfc_diagnostics.o     \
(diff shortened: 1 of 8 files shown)


dudhia (approved):

weiwangncar:
@jihyeonjang Creating a separate module for radiation utility is good. I would prefer to have CO2 computed from a different routine than radconst. This will keep radconst the same as in WRF, and later when a table version of CO2 is added, no change would be required to modify radconst. But I'm ok if you and Michael are ok with the current solution.

dudhia:
I had similar thoughts to Wei as CO2 and radconst are not physically related.

jihyeonjang:
Thank you, Wei and Jimy, for your comments. I thought `radconst` is a subroutine that produces constants for the radiation scheme, which is why I added the CO2 function in `radconst`. But I agree with Wei's suggestion. As she mentioned, to keep `radconst` the same as in WRF, I will modify the code so that the CO2 function is handled seperately from `radconst`.

weiwangncar (approved):

weiwangncar:
@jihyeonjang Thanks for improving my old PR.

jihyeonjang:
@weiwangncar Thank you so much for the quick review!

dudhia (approved):

PR #1458, opened 2026-05-22
Fix bug in registry-generated code to set up derived dimensions

mgduda:
This PR fixes a bug in registry-generated code for setting up derived dimensions.

Previous work (in PR #1451, commit ae45a181) introduced changes to eliminate memory leaks that occurred while setting up derived dimensions. The approach taken in that work involved deallocating dimensions after they were allocated and added to the dimension pool, since the pool assigns the value of the dimension to its own internal memory, and it is up to calling code to free any memory allocated for the dimension argument to `mpas_pool_add_dimension`. However, as an unintended side effect, any later derived dimensions that depended on a dimension whose value came from a namelist or other registry definiton would generate a segmentation fault after referencing an unassociated pointer to the now deallocated dimension.

For example, in the `init_atmosphere` core, the `nVertLevels` dimension is set based on the namelist option `config_nvertlevels`, and the `nVertLevelsP1` dimension is subsequently set to `nVertLevels+1`.

This PR fixes this issue by re-setting pointers to point to pool-owned memory after a dimension has been allocated, added to the dimension pool, and freed.

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -1087,6 +1087,8 @@ int parse_dimensions_from_registry(ezxml_t registry)/*{{{*/
 				fortprintf(fd, "         call mpas_pool_add_dimension(dimensionPool, '%s', %s)\n", dimname, dimname);
 
 				fortprintf(fd, "         deallocate(%s)\n", dimname);
+				fortprintf(fd, "         ! Now let %s point to pool memory so the dimension can be referenced later in this routine as needed\n", dimname);
+				fortprintf(fd, "         call mpas_pool_get_dimension(dimensionPool, '%s', %s)\n", dimname, dimname);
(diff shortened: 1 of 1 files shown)


mgduda:
For the record, it's entirely my fault that this issue made its way into the `develop` branch in the first place -- I should have tested more thoroughly before merging PR #1451!

src/tools/registry/gen_inc.c
abishekg7: The changes look good overall and able to build and run without issues. Just had a small comment re. indentation of this line in the generated code.

mgduda: Although this line isn't part of any of my changes in this PR, I can push an update to correct the indentation of this line as well as line 1082.

mgduda: I've added a separate commit to address indentation issues in the generated Fortran code.

abishekg7 (approved):
Looks good!

PR #1459, opened 2026-05-26
Cryptic soil layer indexing bug

mpaiao:
This fixes a rather cryptic bug, that occurs when the number of soil layers is not the default (4 layers) for Noah-MP, which results in the model not conserving water. 

**1. Soil layer indexing bug.** Noah-MP uses positive indices for soil layers, and negative indices for snowpack layers. In sub-routine `lsm_noahmp_toMPAS`, the code copies the depths from Noah-MP to a local structure (`zsnsoxy`) that assumes indices going from 1 to the number of combined snowpack and soil layers. However, the original code had one error in the conversion logic, which coincidentally worked for the specific case in which the number of soil layers is exactly the number of snowpack ayers plus 1. The proposed change follows the logic used elsewhere in the code and ensures the conversion always works.

**2. Hardcoded number of soil plus snow levels.** File `Registry_noahmp.xml` originally defines dimension `nzSnowLevels` (total number of soil plus snowpack layers) as 7. This works when using the default number of soil layers, but causes segmentation violations otherwise.

### DIFF
--- a/src/core_atmosphere/physics/Registry_noahmp.xml
+++ b/src/core_atmosphere/physics/Registry_noahmp.xml
@@ -16,7 +16,7 @@
       <dim name="nSnowLevels" definition="3"
                 description="The number of snow layers used by the NOAHMP land-surface scheme"/>
 
-      <dim name="nzSnowLevels" definition="7"
+      <dim name="nzSnowLevels" definition="nSoilLevels+nSnowLevels"
                 description="The number of snow layer depths used by the NOAHMP land-surface scheme"/>
 
  </dims>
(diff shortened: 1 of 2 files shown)


mpaiao:
This PR mirrors the one I submitted to [MONAN](https://github.com/monanadmin/MONAN-Model/pull/32). 

I found this bug as I am working to allow the total number of soil layers (and their thickness) more flexible in MONAN. Happy to point to the subsequent PR if interested in implementing it in MPAS too.

cenlinhe (approved):
This bug fix looks good to me.

mgduda:
@mpaiao I think we're now planning for an MPAS v8.4.1 bugfix release within the next week, and we could include this fix in that release. Would you be able to rebase this PR branch off of the `v8.4.0` tag? There's no need to create a new PR, and you could just push the rebased branch to `mpaiao-pr-nsoil-hotfix` to update this PR. If you have any questions, or if I can help at all, just let me know!

mpaiao:
@mgduda I think this is now rebased off of release-v8.4.0, but let me know if this does not look correct.

mgduda (approved):
@mpaiao The rebased branch looks good to me. Thanks so much for tracking down this issue and for providing a fix!

PR #1460, opened 2026-05-26
Registry allocates surface arrays based on new sfclayer package not PBL choice

dudhia:
When the LES options are run in real-data cases, the sfclayer physics is used but not the PBL physics.
Many (over 30) 2d arrays were allocated based on PBL packages when their true dependence is on the sfclayer package.
This caused a failure due to using unallocated arrays.
The fix is to create and use instead a sfclayer package in the Registry.
(Note: in development code PBL was on in namelist to enable these arrays but by-passed for LES options, 
now we need PBL to be off when LES is chosen which could require a safety check).
This enables real-data LES applications to run.

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -485,6 +485,9 @@
                 <package name="les"
                          description="Variables and options associated with LES models"
                          active_when="trim(config_les_model) /= 'none'"/>
+                <package name="sfclayer"
+                         description="Variables and options associated with surface layer physics"
+                         active_when="(trim(config_physics_suite) /= 'none' .and. trim(config_sfclayer_scheme) == 'suite') .or.  any(trim(config_sfclayer_scheme) == [ character(len=StrKind) :: 'sf_monin_obukhov', 'sf_monin_obukhov_rev', 'sf_mynn' ])"/>
         </packages>
 
 
@@ -2750,7 +2753,7 @@
 
                 <var name="hpbl" type="real" dimensions="nCells Time" units="m"
                      description="Planetary Boundary Layer (PBL) height"
-                     packages="bl_mynn_in;bl_ysu_in"/>
+                     packages="bl_mynn_in;bl_ysu_in;sfclayer"/>
 
 		<var name="kzh" type="real" dimensions="nVertLevels nCells Time" units="m^{2} s^{-1}"
                      description="vertical diffusion coefficient of potential temperature"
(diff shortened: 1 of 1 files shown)


mgduda:
@dudhia It seems that we may also need [`ch` and `qcg`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Registry.xml#L3008-L3014) to be associated with the new `sfclayer` package if we would like to run with the MYNN surface layer scheme but not the MYNN PBL scheme. What do you think?

dudhia:
> @dudhia It seems that we may also need [`ch` and `qcg`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Registry.xml#L3008-L3014) to be associated with the new `sfclayer` package if we would like to run with the MYNN surface layer scheme but not the MYNN PBL scheme. What do you think?

I agree. Did you see if there more needed with that sfclayer option?

mgduda:
> > @dudhia It seems that we may also need [`ch` and `qcg`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Registry.xml#L3008-L3014) to be associated with the new `sfclayer` package if we would like to run with the MYNN surface layer scheme but not the MYNN PBL scheme. What do you think?
> 
> I agree. Did you see if there more needed with that sfclayer option?

Those seemed to be the only two additional variables that needed to be available when using the MYNN surface layer scheme as far as I could tell.

dudhia:
> > > @dudhia It seems that we may also need [`ch` and `qcg`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Registry.xml#L3008-L3014) to be associated with the new `sfclayer` package if we would like to run with the MYNN surface layer scheme but not the MYNN PBL scheme. What do you think?
> > 
> > 
> > I agree. Did you see if there more needed with that sfclayer option?
> 
> Those seemed to be the only two additional variables that needed to be available when using the MYNN surface layer scheme as far as I could tell.

I am not seeing your changes, so should I update my PR?

mgduda:
> > > > @dudhia It seems that we may also need [`ch` and `qcg`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Registry.xml#L3008-L3014) to be associated with the new `sfclayer` package if we would like to run with the MYNN surface layer scheme but not the MYNN PBL scheme. What do you think?
> > > 
> > > 
> > > I agree. Did you see if there more needed with that sfclayer option?
> > 
> > 
> > Those seemed to be the only two additional variables that needed to be available when using the MYNN surface layer scheme as far as I could tell.
> 
> I am not seeing your changes, so should I update my PR?

Sorry... I wasn't sure if you'd prefer to update the branch or if I should do so. If you don't mind, I can amend this PR branch.

dudhia:
> > > > > @dudhia It seems that we may also need [`ch` and `qcg`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/Registry.xml#L3008-L3014) to be associated with the new `sfclayer` package if we would like to run with the MYNN surface layer scheme but not the MYNN PBL scheme. What do you think?
> > > > 
> > > > 
> > > > I agree. Did you see if there more needed with that sfclayer option?
> > > 
> > > 
> > > Those seemed to be the only two additional variables that needed to be available when using the MYNN surface layer scheme as far as I could tell.
> > 
> > 
> > I am not seeing your changes, so should I update my PR?
> 
> Sorry... I wasn't sure if you'd prefer to update the branch or if I should do so. If you don't mind, I can amend this PR branch.

Go ahead. I didn't know you could do that.

mgduda:
After a closer look, I think there are two issues with the current code that prevent it from working as expected:
1. The new `sfclayer` package should probably be active when `trim(config_sfclayer_scheme) /= 'off'` rather than when `trim(config_sfclayer_scheme) /= 'none'`, since `off` is how a user is expected to indicate that no scheme is to be used.
2. Right now, we evaluate the Registry-generated package logic before physics suites have been set up, and so, for example, `config_sfclayer_scheme` will be `suite` rather than `off` when evaluating package logic if a user sets `config_physics_suite = 'none'` in their namelist; this would lead to an active `sfclayer` package even when no physics have been selected.

Addressing the first issue in this PR should be easy. For the second issue, we could either add a separate commit to this PR, or we could open an new PR just to change the order in which physics suites and Registry-generated package logic are evaluated.

dudhia:
If config_sfclayer_scheme is not in the namelist, I guess it would be 'suite'? So with this PR we would be allocating them when not needed.
I think for config_les_model we have 'none' as a choice and default, and for some other options I see.

mgduda:
So that we don't need to make changes to the order of package setup (physics suites, Registry-generated logic), I think the following would work for the `active_when` specification for the new `sfclayer` package:
```fortran
(config_physics_suite /= 'none' .and. config_sfclayer_scheme == 'suite') .or. any(config_sfclayer_scheme == [ character(len=StrKind) :: 'sf_monin_obukhov', 'sf_monin_obukhov_rev', 'sf_mynn' ])
```
Essentially, the `sfclayer` package is active if any non-`none` physics suite has been selected (without an override of the surface layer scheme), or if the surface layer scheme has been explicitly set to any of the possible non-`off` options.

dudhia (commented):
Looks OK to me.

mgduda (approved):

PR #1461, opened 2026-05-27
Fix for multiply occurring nml options in `active_when` attribute

abishekg7:
### Problem

When declaring a package with the optional `active_when` attribute in Registry.xml, if a namelist option appeared more than once in an `active_when` attribute (e.g., config_foo > 0 .and. config_foo < 10), the code generation logic in `gen_inc.c` would produce duplicate Fortran variable declarations and `mpas_pool_get_config` calls in the generated `setup_packages.inc`. This caused compiler errors associated the generated code.

### Solution

The solution involves modifying the logic in `package_logic_routine` and `gen_pkg_debug_info` in `gen_inc.c` to identify unique namelist options in the `active_when` attributes. 

This PR introduces a simple helper function, `add_unique_key_to_list`, to add unique namelist options to a string list. This list is then iterated over by subsequent logic in order to ensure that there are no multiply defined variables in the generated Fortran code.

Presently, the `MAX_LIST_SIZE` macro caps the maximum number of unique namelist options allowed in the `active_when` attribute at 20.

### DIFF
--- a/src/tools/registry/gen_inc.c
+++ b/src/tools/registry/gen_inc.c
@@ -28,11 +28,13 @@ int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
                           const char *packagename, const char *packagewhen,
                           ezxml_t registry);
 void gen_pkg_debug_info(FILE *fd, regex_t *preg, ezxml_t registry,
-                        const char *packagename, const char *packagewhen);
+                        const char *packagename, const char *packagewhen,
+						char **nmloptions_list, int num_nmloptions);
 
 #define NUM_MODIFIED_ATTRS 2
 #define NUM_IGNORED_ATTRS 9
 #define NUM_NUMERIC_ATTRS 1
+#define MAX_LIST_SIZE 20
 
 static const char *NUMERIC_ATTRS[NUM_NUMERIC_ATTRS] = {
 	"missing_value"
@@ -2727,6 +2729,46 @@ int generate_package_logic(ezxml_t registry)/*{{{*/
 
 /******************************************************************************
  *
+ * add_unique_key_to_list
+ *
+ * Adds a string to a list of strings, if it is not already present.
+ * Allocates a copy of the key string and stores it at the next available
+ * index in string_list.
+ *
+ * Inputs:
+ *   string_list - an array of char pointers to store strings
+ *   key - a null-terminated string containing the key to add
+ *   num_keys - the current number of keys stored in string_list
+ *
+ * Outputs:
+ *   string_list - updated with the new key appended, if not a duplicate
+ *   num_keys - incremented by 1 if the key was successfully added
+ *
+ * Return value: 0 if the key was added successfully, 1 if the key was
+ *   already present in the list, and -1 if the list is full.
+ *
+ ******************************************************************************/
+int add_unique_key_to_list(char **string_list, const char *key, int *num_keys)
+{
+
+	for (int i = 0; i < *num_keys; i++) {
+		if (strcmp(string_list[i], key) == 0) {
+			return 1;
+		}
+	}
+
+	if (*num_keys < MAX_LIST_SIZE) {
+		string_list[*num_keys] = strdup(key);
+		(*num_keys)++;
+		return 0;
+	} else {
+		return -1;
+	}
+}
+
+
+/******************************************************************************
+ *
  * package_logic_routine
  *
  * Generates code for the Fortran routine 'setup_X_package' that defines the active
@@ -2756,7 +2798,9 @@ int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
 	ezxml_t packages_xml, package_xml;
 	char *match;
 	regoff_t next;
-
+	char *nmloptions_list[MAX_LIST_SIZE];
+	int num_nmloptions = 0;
+	int ierr;
 
 	fortprintf(fd, "\n");
 	fortprintf(fd, "   !\n");
@@ -2777,9 +2821,23 @@ int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
 	fortprintf(fd, "      logical, pointer :: %sActive\n", packagename);
 	fortprintf(fd, "\n");
 
+	fprintf(stdout, "Parsing active_when string for package %s: %s\n", packagename, packagewhen);
 	next = 0;
 	while ((match = nmlopt_from_str(preg, packagewhen, &next)) != NULL) {
 		const char *nmltype;
+		ierr = add_unique_key_to_list(nmloptions_list, match, &num_nmloptions);
+		if (ierr < 0) {
+			// Error case: too many unique namelist options in active_when attribute.
+			fprintf(stderr, "Error: Too many unique namelist options specified in active_when"
+				    " attribute for %s package. Maximum allowed is %d.\n", packagename, MAX_LIST_SIZE);
+			free(match);
+			for (int i = 0; i < num_nmloptions; i++) free(nmloptions_list[i]);
+			return 1;
+		} else if (ierr == 1) {
+			// Namelist option already in list, so we can skip it.
+			free(match);
+			continue;
+		}
 
 		nmltype =  nmlopt_type(registry, match);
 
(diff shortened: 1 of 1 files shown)


src/tools/registry/gen_inc.c
```diff
@@ -2731,6 +2728,51 @@ int generate_package_logic(ezxml_t registry)/*{{{*/
...
+{
+	int i;
+
+	for (i = 0; i < *num_options; i++) {
+		if (strcmp(nmloptions_list[i], option) == 0) {
+			return 1;
+		}
+	}
+
+	if (*num_options < 20) {
```
mgduda: Rather than using a hardwired value of 20, perhaps the maximum size of the list could be defined as a constant or a macro at the top of this file?

src/tools/registry/gen_inc.c
```diff
@@ -2762,7 +2804,8 @@ int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
 	ezxml_t packages_xml, package_xml;
 	char *match;
 	regoff_t next;
-
+	char **nmloptions_list = malloc(20 * sizeof(char*));
```
mgduda: If the maximum size of the list is fixed at compile time, and it it's relatively small, you could save an allocation and free by declaring `nmloptions_list` here as:
```c
char *nmloptions_list[MAX_LIST_SIZE];
```

src/tools/registry/gen_inc.c
```diff
@@ -2935,22 +2986,26 @@ const char * nmlopt_type(ezxml_t registry, const char *nmlopt)/*{{{*/
...
 	regoff_t next = 0;
+	int i;
 
 	fortprintf(fd, "      PACKAGE_LOGIC_PRINT('')\n");
 	fortprintf(fd, "      PACKAGE_LOGIC_PRINT(\"  %s is active when (%s)\")\n", packagename, packagewhen);
 	fortprintf(fd, "      PACKAGE_LOGIC_PRINT('    namelist settings:')\n");
 	fortprintf(fd, "      PACKAGE_LOGIC_PRINT('    ------------------')\n");
 
-	while ((match = nmlopt_from_str(preg, packagewhen, &next)) != NULL) {
+	for(i = 0; i < num_options; i++) {
```
mgduda: For consistency with earlier code, `i` could be declared as an `int` in the `for` loop here.

src/tools/registry/gen_inc.c
```diff
@@ -2783,9 +2826,15 @@ int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
 	fortprintf(fd, "      logical, pointer :: %sActive\n", packagename);
 	fortprintf(fd, "\n");
 
+	fprintf(stdout, "Parsing active_when string for package %s: %s\n", packagename, packagewhen);
 	next = 0;
 	while ((match = nmlopt_from_str(preg, packagewhen, &next)) != NULL) {
 		const char *nmltype;
+		if(add_namelist_option_to_list(nmloptions_list, match, &num_options) != 0) {
```
mgduda: If we hit an error condition (most likely, a full list), it might be best to free up memory and return from the `package_logic_routine` function with a non-zero return code.

mgduda (changes requested):

src/tools/registry/gen_inc.c
abishekg7: Addressed these suggestions in the new commit. And also tested the failure case when number of unique nml options > 20 ( or macro definition)

src/tools/registry/gen_inc.c
```diff
@@ -2948,9 +3007,10 @@ void gen_pkg_debug_info(FILE *fd, regex_t *preg, ezxml_t registry,
 	fortprintf(fd, "      PACKAGE_LOGIC_PRINT('    namelist settings:')\n");
 	fortprintf(fd, "      PACKAGE_LOGIC_PRINT('    ------------------')\n");
 
-	while ((match = nmlopt_from_str(preg, packagewhen, &next)) != NULL) {
+	for(int i = 0; i < num_nmloptions; i++) {
```
mgduda: Can we use whitespace for this `for` loop that's consistent with the other `for` loops (e.g., on line 2886 of this PR)?

src/tools/registry/gen_inc.c
```diff
@@ -2783,9 +2824,24 @@ int package_logic_routine(FILE *fd, regex_t *preg, const char *corename,
...
+		ierr = add_namelist_option_to_list(nmloptions_list, match, &num_nmloptions);
+		if( ierr < 0) {
+			// Error case: too many unique namelist options in active_when attribute.
+			fprintf(stderr, "Error: Too many unique namelist options specified in active_when"
+				    " attribute for %s package. Maximum allowed is %d.\n", packagename, MAX_NML_LIST_SIZE);
+			free(match);
+			for (int i = 0; i < num_nmloptions; i++) free(nmloptions_list[i]);
+			return 1;
+		}
+		else if( ierr == 1) {
```
mgduda: Other code in this file, and other C code in MPAS, tends to place `else` and `else if` on the same line as the closing `}` for the preceding block. Should we do the same here for consistency?

Also, could we follow the style of a single space after `if`, `else`, `for`, `while`, etc. and before the `(`, and to not place whitespace immediately to the right of `(` and to the left of `)`?

src/tools/registry/gen_inc.c
```diff
@@ -2731,6 +2729,47 @@ int generate_package_logic(ezxml_t registry)/*{{{*/
...
+ *
+ * Outputs:
+ *   nmloptions_list - updated with the new option appended, if not a duplicate
+ *   num_nmloptions - incremented by 1 if the option was successfully added
+ *
+ * Return value: 0 if the option was added successfully, 1 if the option was
+ *   already present in the list, and -1 if the list is full.
+ *
+ ******************************************************************************/
+int add_namelist_option_to_list(char **nmloptions_list, const char *option,
```
mgduda: Just for consideration, this function (and the other code being added in this PR to handle lists of strings) is not specific to namelist options, and so we could generalize the names to reflect the applicability to general strings.

mgduda: Of course, local variable names can reflect their actual use.

src/tools/registry/gen_inc.c
```diff
@@ -2731,6 +2729,47 @@ int generate_package_logic(ezxml_t registry)/*{{{*/
...
+ * Inputs:
+ *   string_list - an array of char pointers to store strings
+ *   key - a null-terminated string containing the key to add
+ *   num_keys - the current number of keys stored in string_list
+ *
+ * Outputs:
+ *   string_list - updated with the new key appended, if not a duplicate
+ *   num_keys - incremented by 1 if the key was successfully added
+ *
+ * Return value: 0 if the option was added successfully, 1 if the option was
```
mgduda: We may want to change the word "option" here to "key" or "string" to be consistent with the terminology used elsewhere in the documentation block.

src/tools/registry/gen_inc.c
```diff
@@ -2731,6 +2729,47 @@ int generate_package_logic(ezxml_t registry)/*{{{*/
...
+ *   already present in the list, and -1 if the list is full.
+ *
+ ******************************************************************************/
+int add_unique_key_to_list(char **string_list, const char *key,
+                                int *num_keys)
```
mgduda: `num_keys` might fit on the line above while still keeping the line less than 80 chars.

mgduda (changes requested):
I added just two minor comments, and otherwise, I think this is ready to go as an open PR after tidying up the commit history and adding a PR description.

abishekg7:
Addressed the last two comments and squashed to a single commit. I think the change set is small enough for one commit, but can also move the new function definition to its separate commit.

mgduda:
@abishekg7 Could you rebase this PR branch off of the `v8.4.0` tag? That will allow us to merge this to the `hotfix-v8.4.1` branch without bringing in any other changes that weren't in the v8.4.0 release.

abishekg7:
> @abishekg7 Could you rebase this PR branch off of the `v8.4.0` tag? That will allow us to merge this to the `hotfix-v8.4.1` branch without bringing in any other changes that weren't in the v8.4.0 release.

Rebased and force pushed.

mgduda (approved):
Thanks for the fix for this issue and also for the thorough PR description!

thomasuhoh:
This commit seems to break the compilation with the NVHPC compiler:

mpifort -D_MPI -DCPRPGI -DCORE_ATMOSPHERE -DMPAS_NAMELIST_SUFFIX=atmosphere -DMPAS_EXE_NAME=atmosphere_model -DMPAS_OPENACC -DSINGLE_PRECISION -DMPAS_NATIVE_TIMERS -DMPAS_GIT_VERSION>
NVFORTRAN-S-0290-Unexpected continuation line (./inc/setup_packages.inc: 107)
0 inform, 0 warnings, 1 severes, 0 fatal for atm_setup_sfclayer_package
make[3]: *** [Makefile:97: mpas_atm_core_interface.o] Error 2

With MPAS 8.4.0, everything is okay.

abishekg7:
> This commit seems to break the compilation with the NVHPC compiler:
> 
> mpifort -D_MPI -DCPRPGI -DCORE_ATMOSPHERE -DMPAS_NAMELIST_SUFFIX=atmosphere -DMPAS_EXE_NAME=atmosphere_model -DMPAS_OPENACC -DSINGLE_PRECISION -DMPAS_NATIVE_TIMERS -DMPAS_GIT_VERSION> NVFORTRAN-S-0290-Unexpected continuation line (./inc/setup_packages.inc: 107) 0 inform, 0 warnings, 1 severes, 0 fatal for atm_setup_sfclayer_package make[3]: *** [Makefile:97: mpas_atm_core_interface.o] Error 2
> 
> With MPAS 8.4.0, everything is okay.

Thanks for the report, @thomasuhoh! We also noticed it yesterday, and will have a fix soon. Were you able to isolate the nvhpc build error to this commit in particular, or did you notice this with either v8.4.1 release or the current HEAD of develop branch?

thomasuhoh:
I noticed it with the v8.4.1 release downloaded from [https://github.com/MPAS-Dev/MPAS-Model/releases](url)

mgduda:
@thomasuhoh There are a couple of workarounds that you can try (either on its own should be sufficient, and you don't need to use them both together):
1. Add `GEN_F90=true` to your build command. This will pre-process source files with `cpp` rather than letting the `nvfortran` compiler handle the pre-processing, which is what seems to be the source of the issue.
2. Change `fortprintf` to `fprintf` on [line 3002 of `src/tools/registry/gen_inc.c`](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.1/src/tools/registry/gen_inc.c#L3002). This will avoid line wrapping in the generated source code, and that should avoid issues in the `nvfortran` compiler's pre-processing of the line-wrapped `PACKAGE_LOGIC_PRINT` macro.

PR #1463, opened 2026-06-01
Add a dm_info object to the argument list of core mpas_setup_packages_function

mgduda:
This PR adds a `dm_info` object to the argument list of the `mpas_setup_packages_function` core interface function. Adding a `dm_info` object allows a core's package setup function to make use of MPI when setting up packages. For example, a core may read data from a file only on MPI rank 0 and broadcast the resulting values to other MPI ranks rather than having each MPI rank independently open and read the file.

In addition to updating the `mpas_setup_packages_function` abstract interface in `mpas_core_types.inc` with an additional `dminfo` argument, this PR updates the call to `domain_ptr % core % setup_packages` in `mpas_init` to include the `dminfo` argument, and it also updates the `*_setup_packages` functions in each core's interface module.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core_interface.F
+++ b/src/core_atmosphere/mpas_atm_core_interface.F
@@ -99,7 +99,7 @@ module atm_core_interface
    !>  not allocated until after this routine has been called.
    !
    !-----------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


abishekg7 (approved):
Looks good. Checked with gnu, intel and nvhpc builds. And also with atmosphere, init_atmosphere, sw cores.

PR #1464, opened 2026-06-02
Changes to filename_template checks 

abishekg7:
## Issue:
Previously, if two active, output stream definitions in the streams.<CORE> file specified the same string as the `filename_template` attribute, the corresponding MPAS CORE would crash with an error message. This was also the case for when one of the streams, sharing the `filename_template`, included an inactive package, hence effectively making it inactive. The existing logic to perform stream uniqueness checks in `src/framework/xml_stream_parser.c` did not have the information about which packages associated with a stream were active, and hence did not account for inactive packages. 

## Solution
This PR introduces a new subroutine `MPAS_stream_mgr_check_filename_template` in `src/framework/mpas_stream_manager.F` to check whether two active streams might potentially be accessing the same file (via `filename_template`) resulting in file conflicts. More specifically, this subroutine checks that there are no active output streams that share an identical `filename_template` attribute with any other active input or output streams in the stream manager. An active stream in this context is a stream with attribute `active_stream = .true.` and is not deactivated by an inactive package and is either an input or output stream with at least one active alarm. This routine can be called from within `MPAS_stream_mgr_validate_streams` or separately as needed.

This PR also introduces a new function 'MPAS_stream_mgr_get_num_alarms', which returns the number of active alarms for a given stream in the specified direction. 

The previous logic to check for unique `filename_template` attributes in `src/framework/xml_stream_parser.c` has been removed, while retaining the checks for unique stream names in the same location. Note that the previous logic only check for conflicts between two output streams, whereas the new logic also checks for potential conflicts between an input stream and an output stream accessing the same file.

### DIFF
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -593,7 +593,7 @@ module mpas_stream_manager
         type (MPAS_streamManager_type), intent(inout) :: manager
         integer, intent(out), optional :: ierr
 
-        integer :: threadNum, err_local,err_local1, err_local2
+        integer :: threadNum, err_local
         character (len=StrKIND) :: message, streamID
         type (MPAS_stream_list_type), pointer :: stream1_cursor, stream2_cursor
         type (MPAS_TimeInterval_type)  :: clock_interval
@@ -619,9 +619,9 @@ module mpas_stream_manager
                 stream1_pkg_active = stream_active_pkg_check(stream1_cursor)
 
                 stream1_input_nalarms = MPAS_stream_mgr_get_num_alarms(manager, trim(stream1_cursor % name), &
-                                                                       MPAS_STREAM_INPUT, err_local1)
+                                                                       MPAS_STREAM_INPUT)
                 stream1_output_nalarms = MPAS_stream_mgr_get_num_alarms(manager, trim(stream1_cursor % name), &
-                                                                        MPAS_STREAM_OUTPUT, err_local2)
+                                                                        MPAS_STREAM_OUTPUT)
 
                 stream1_input = ((stream1_cursor % direction == MPAS_STREAM_INPUT .or. &
                                 stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
@@ -644,9 +644,9 @@ module mpas_stream_manager
                 do while (MPAS_stream_list_query(manager % streams, streamID, stream2_cursor))
 
                     stream2_input_nalarms = MPAS_stream_mgr_get_num_alarms(manager, trim(stream2_cursor % name), &
-                                                                           MPAS_STREAM_INPUT, err_local1)
+                                                                           MPAS_STREAM_INPUT)
                     stream2_output_nalarms = MPAS_stream_mgr_get_num_alarms(manager, trim(stream2_cursor % name), &
-                                                                            MPAS_STREAM_OUTPUT, err_local2)
+                                                                            MPAS_STREAM_OUTPUT)
 
                     stream2_pkg_active = stream_active_pkg_check(stream2_cursor)
                     stream2_input = ((stream2_cursor % direction == MPAS_STREAM_INPUT .or. &
@@ -1906,28 +1906,31 @@ module mpas_stream_manager
        type (MPAS_streamManager_type), intent(in) :: manager
        character(len=*), intent(in) :: streamID
        integer, intent(in) :: direction
-       integer, intent(out) :: ierr
+       integer, intent(out), optional :: ierr
 
        type (mpas_stream_list_type), pointer :: streamCursor
+       integer :: err_local
 
        numAlarms = 0
-       ierr = MPAS_STREAM_MGR_NOERR
+       err_local = MPAS_STREAM_MGR_NOERR
        nullify(streamCursor)
 
        if ( mpas_stream_list_query(manager % streams, streamID, streamCursor) ) then
           if (direction == MPAS_STREAM_INPUT) then
-            numAlarms = MPAS_stream_list_length(streamCursor % alarmList_in, ierr)
+            numAlarms = MPAS_stream_list_length(streamCursor % alarmList_in, err_local)
           else if (direction == MPAS_STREAM_OUTPUT) then
-            numAlarms = MPAS_stream_list_length(streamCursor % alarmList_out, ierr)
+            numAlarms = MPAS_stream_list_length(streamCursor % alarmList_out, err_local)
           else
-            ierr = MPAS_STREAM_MGR_ERROR
+            err_local = MPAS_STREAM_MGR_ERROR
             call MPAS_stream_mesg(manager % errorLevel, 'Invalid direction encountered in MPAS_stream_mgr_get_num_alarms')
           end if
        else
-          ierr = MPAS_STREAM_MGR_ERROR
+          err_local = MPAS_STREAM_MGR_ERROR
           call MPAS_stream_mesg(manager % errorLevel, 'Stream ''' // trim(streamID) // ''' &
                                 does not exist. Cannot retrieve number of alarms in stream.')
        end if
+
+       if (present(ierr)) ierr = err_local
 
     end function MPAS_stream_mgr_get_num_alarms
 
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -594,7 +594,6 @@ module mpas_stream_manager
         integer, intent(out), optional :: ierr
 
         character(len=StrKIND) :: stream1_name, stream2_name, stream1_filename, stream2_filename
-        character(len=StrKIND) :: filename_interval
         integer :: threadNum, err_local
         character (len=StrKIND) :: message, streamID
         type (MPAS_stream_list_type), pointer :: stream1_cursor, stream2_cursor
@@ -621,7 +620,7 @@ module mpas_stream_manager
                              (stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
 
             call mpas_log_write('Stream 1 '//trim(stream1_name)//' filename '//trim(stream1_filename)//' &
-             active = $l output = $l  packages_active $l',logicArgs=(/stream1_active, stream1_output, stream1_pkg_active/))
+             & active = $l output = $l  packages_active $l',logicArgs=(/stream1_active, stream1_output, stream1_pkg_active/))
 
             if ( stream1_active .and. stream1_output .and. stream1_pkg_active) then
             
@@ -639,12 +638,12 @@ module mpas_stream_manager
 
                         stream2_filename = stream2_cursor % filename_template
                         call mpas_log_write('Stream 2 '//trim(stream2_name)//' filename '//trim(stream2_filename)//' &
-                            active = $l output = $l pkg_active: $l',logicArgs=(/stream2_active, stream2_output, stream2_pkg_active/))
+                            & active = $l output = $l pkg_active: $l',logicArgs=(/stream2_active, stream2_output, stream2_pkg_active/))
                         
                         if ( trim(stream1_filename) == trim(stream2_filename) ) then
                             message = 'Found identical values of the filename_template attribute for multiple active output streams &
-                                       (' // trim(stream1_name) // ' and ' // trim(stream2_name) // ') in streams.<CORE>. This may &
-                                        result in file conflicts.'
+                                       & (' // trim(stream1_name) // ' and ' // trim(stream2_name) // ') in streams.<CORE>. This may &
+                                       & result in file conflicts.'
                             call mpas_log_write(message, MPAS_LOG_CRIT)
                         end if
 
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -44,6 +44,7 @@ module mpas_stream_manager
               MPAS_stream_mgr_get_next_field, &
               MPAS_stream_mgr_stream_exists, &
               MPAS_stream_mgr_get_stream_interval, &
+              MPAS_stream_mgr_check_valid_record_interval, &
               MPAS_get_stream_filename, &
               MPAS_build_stream_filename
 
@@ -593,11 +594,17 @@ module mpas_stream_manager
         type (MPAS_streamManager_type), intent(inout) :: manager
         integer, intent(out), optional :: ierr
 
-        integer :: threadNum
+        integer :: threadNum, err_local,err_local1, err_local2
+        real (RKIND) :: dt_i, dt_o
         character (len=StrKIND) :: message, streamID
         type (MPAS_stream_list_type), pointer :: stream1_cursor, stream2_cursor
+        type (MPAS_TimeInterval_type)  :: clock_interval
         logical :: stream1_pkg_active, stream2_pkg_active
+        !character (len=StrKIND) :: stream1_input_interval, stream2_input_interval, stream1_output_interval, stream2_output_interval
+        logical :: stream1_input, stream2_input
         logical :: stream1_output, stream2_output
+        logical :: stream1_input_interval, stream1_output_interval
+        logical :: stream2_input_interval, stream2_output_interval
 
         STREAM_DEBUG_WRITE('-- Called MPAS_stream_mgr_check_filename_template() for all streams')
 
@@ -613,25 +620,53 @@ module mpas_stream_manager
             do while (MPAS_stream_list_query(manager % streams, streamID, stream1_cursor))
 
                 stream1_pkg_active = stream_active_pkg_check(stream1_cursor)
-                stream1_output = (stream1_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
-                                 (stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
+                ! call mpas_stream_mgr_get_property(manager, stream1_cursor % name, MPAS_STREAM_PROPERTY_RECORD_INTV, &
+                !                                   stream1_input_interval, MPAS_STREAM_INPUT, err_local)
+                stream1_input_interval = MPAS_stream_mgr_check_valid_record_interval(manager, trim(stream1_cursor % name), &
+                                                                                    MPAS_STREAM_INPUT, err_local1)
+                stream1_output_interval = MPAS_stream_mgr_check_valid_record_interval(manager, trim(stream1_cursor % name), &
+                                                                                    MPAS_STREAM_OUTPUT, err_local2)
 
-                if (.not. stream1_cursor % active_stream .or. .not. stream1_output .or. .not. stream1_pkg_active) then
+                stream1_input = ((stream1_cursor % direction == MPAS_STREAM_INPUT .or. &
+                                stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                stream1_input_interval)
+                stream1_output = ((stream1_cursor % direction == MPAS_STREAM_OUTPUT .or. &
+                                 stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                 stream1_output_interval)
+
+                call mpas_log_write('Stream 1 '//trim(stream1_cursor % name)//' filename '//trim(stream1_cursor % filename_template)//' &
+                & active = $l input = $l output = $l  packages_active $l',logicArgs=(/stream1_cursor % active_stream, stream1_input, stream1_output, stream1_pkg_active/))
+
+                if (.not. stream1_cursor % active_stream &
+                    .or. .not. (stream1_input .or. stream1_output) &
+                    .or. .not. stream1_pkg_active) then
+                    !call mpas_log_write('cycle1')
                     cycle
                 end if
 
                 nullify(stream2_cursor)
                 do while (MPAS_stream_list_query(manager % streams, streamID, stream2_cursor))
 
-                    stream2_pkg_active = stream_active_pkg_check(stream2_cursor)
-                    stream2_output = (stream2_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
-                                     (stream2_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
+                    stream2_input_interval = MPAS_stream_mgr_check_valid_record_interval(manager, trim(stream2_cursor % name), MPAS_STREAM_INPUT, err_local1)
+                    stream2_output_interval = MPAS_stream_mgr_check_valid_record_interval(manager, trim(stream2_cursor % name), MPAS_STREAM_OUTPUT, err_local2)
 
+                    stream2_pkg_active = stream_active_pkg_check(stream2_cursor)
+                    stream2_input = ((stream2_cursor % direction == MPAS_STREAM_INPUT .or. &
+                                    stream2_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                    stream2_input_interval)
+                    stream2_output = ((stream2_cursor % direction == MPAS_STREAM_OUTPUT .or. &
+                                     stream2_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                     stream2_output_interval)
+
+                     call mpas_log_write('Stream 2 '//trim(stream2_cursor % name)//' filename '//trim(stream2_cursor % filename_template)//' &
+                        & active = $l input = $l output = $l pkg_active: $l',logicArgs=(/stream2_cursor % active_stream, stream2_input, stream2_output, stream2_pkg_active/))
                     ! uniqueness_check has already checked that two different streams do not have the same name
-                    if (trim(stream1_cursor % name) == trim(stream2_cursor % name) &
-                        .or. .not. stream2_cursor % active_stream &
-                        .or. .not. stream2_output &
-                        .or. .not. stream2_pkg_active) then
+                    if (trim(stream1_cursor % name) == trim(stream2_cursor % name) & ! Skip comparing the stream to itself
+                        .or. .not. stream2_cursor % active_stream & ! Skip if Stream 2 is not active
+                        .or. .not. (stream2_input .or. stream2_output) &
+                        .or. (stream1_input .and. stream2_input) & ! Skip if Streams 1 and 2 are both input streams
+                        .or. .not. stream2_pkg_active) then ! Skip if Stream 2 is inactive due to inactive package
+                        !call mpas_log_write('cycle2')
                         cycle
                     end if
 
@@ -641,7 +676,6 @@ module mpas_stream_manager
                                   ', in streams.<CORE>. This may result in file conflicts.'
                         call mpas_log_write(message, messageType=MPAS_LOG_ERR)
                         if (present(ierr)) ierr = MPAS_STREAM_MGR_ERROR
-                        return
                     end if
 
                 end do
@@ -1856,6 +1890,50 @@ module mpas_stream_manager
 
     end function MPAS_stream_mgr_get_stream_interval
 
+
+
+    function MPAS_stream_mgr_check_valid_record_interval(manager, streamID, direction, ierr) result(validInterval)
+
+       implicit none
+
+       logical :: validInterval
+
+       type (MPAS_streamManager_type), intent(in) :: manager
+       character(len=*), intent(in) :: streamID
+       integer, intent(in) :: direction
+       integer, intent(out) :: ierr
+
+       type (mpas_stream_list_type), pointer :: streamCursor, alarmCursor
+
+       validInterval = .false.
+       ierr = MPAS_STREAM_MGR_NOERR
+       nullify(streamCursor)
+       !call mpas_set_timeInterval(interval, dt = 0.0_RKIND)
+
+       if ( mpas_stream_list_query(manager % streams, streamID, streamCursor) ) then
+
+          if (direction == MPAS_STREAM_INPUT) then
+
+            call mpas_check_alarms(manager % streamClock, streamCursor % alarmList_in % head, validInterval)
+
+          else if (direction == MPAS_STREAM_OUTPUT) then
+
+            call mpas_check_alarms(manager % streamClock, streamCursor % alarmList_out % head, validInterval)
+          else
+
+             ierr = MPAS_STREAM_MGR_ERROR
+             call MPAS_stream_mesg(manager % errorLevel, 'Invalid direction encountered in MPAS_stream_mgr_get_stream_interval')
+
+          end if
+
+       else
+
+          ierr = MPAS_STREAM_MGR_ERROR
+          call MPAS_stream_mesg(manager % errorLevel, 'Stream ''' // trim(streamID) // ''' does not exist. Cannot retrieve stream input/output intervals')
+
+       end if
+
+    end function MPAS_stream_mgr_check_valid_record_interval
 
     !-----------------------------------------------------------------------
     !  routine MPAS_stream_mgr_set_property_int
--- a/src/framework/mpas_stream_manager.F
+++ b/src/framework/mpas_stream_manager.F
@@ -567,6 +567,7 @@ module mpas_stream_manager
         ! Only check filename_template uniqueness when validating all streams 
         if (.not. present(streamID)) then
             call MPAS_stream_mgr_check_filename_template(manager, ierr=err_local)
+            if (present(ierr)) ierr = err_local
         end if
 
     end subroutine MPAS_stream_mgr_validate_streams!}}}
@@ -592,13 +593,15 @@ module mpas_stream_manager
         type (MPAS_streamManager_type), intent(inout) :: manager
         integer, intent(out), optional :: ierr
 
-        integer :: threadNum, err_local
+        integer :: threadNum
         character (len=StrKIND) :: message, streamID
         type (MPAS_stream_list_type), pointer :: stream1_cursor, stream2_cursor
         logical :: stream1_pkg_active, stream2_pkg_active
         logical :: stream1_output, stream2_output
 
         STREAM_DEBUG_WRITE('-- Called MPAS_stream_mgr_check_filename_template() for all streams')
+
+        if (present(ierr)) ierr = MPAS_STREAM_LIST_NOERR
 
         threadNum = mpas_threading_get_thread_num()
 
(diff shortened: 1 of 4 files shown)


src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,103 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+    !>  routine can be called from within MPAS_stream_mgr_validate_streams or 
+    !>  separately as needed. 
+    !
+    !-----------------------------------------------------------------------
+    subroutine MPAS_stream_mgr_check_filename_template(manager, ierr)!{{{
+
+        implicit none
+
+        type (MPAS_streamManager_type), intent(inout) :: manager
+        integer, intent(out), optional :: ierr
```
mgduda, line 602: It doesn't look like `ierr`, which has `intent(out)`, is ever initialized or set in this routine.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,103 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
 #endif
 
         end do
+        
+        ! Only check filename_template uniqueness when validating all streams 
+        if (.not. present(streamID)) then
+            call MPAS_stream_mgr_check_filename_template(manager, ierr=err_local)
```
mgduda, line 569: If `MPAS_stream_mgr_check_filename_template` returns a non-success error code, it doesn't look like that will translate into a non-success return code from `MPAS_stream_mgr_validate_streams`.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,103 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+            
+            stream1_name = stream1_cursor % name
+            stream1_active = stream1_cursor % active_stream
+            stream1_filename = stream1_cursor % filename_template
+            stream1_pkg_active = stream_active_pkg_check(stream1_cursor)
+            stream1_output = (stream1_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
+                             (stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
+
+            call mpas_log_write('Stream 1 '//trim(stream1_name)//' filename '//trim(stream1_filename)//' &
+             active = $l output = $l  packages_active $l',logicArgs=(/stream1_active, stream1_output, stream1_pkg_active/))
```
mgduda:
```
mpas_stream_manager.F:624:14:

  624 |              active = $l output = $l  packages_active $l',logicArgs=(/stream1_active, stream1_output, stream1_pkg_active/))
      |              1
Warning: Missing ‘&’ in continued character constant at (1) [-Wampersand]
```

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,103 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+                    stream2_pkg_active = stream_active_pkg_check(stream2_cursor)
+                    stream2_output = (stream2_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
+                             (stream2_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
+
+                    ! uniqueness_check has already checked that two different streams do not have the same name
+                    if ( trim(stream1_name) /= trim(stream2_name) .and. stream2_active .and. stream2_output .and. stream2_pkg_active) then
+
+                        stream2_filename = stream2_cursor % filename_template
+                        call mpas_log_write('Stream 2 '//trim(stream2_name)//' filename '//trim(stream2_filename)//' &
+                            active = $l output = $l pkg_active: $l',logicArgs=(/stream2_active, stream2_output, stream2_pkg_active/))
```
mgduda:
```
mpas_stream_manager.F:642:29:

  642 |                             active = $l output = $l pkg_active: $l',logicArgs=(/stream2_active, stream2_output, stream2_pkg_active/))
      |                             1
Warning: Missing ‘&’ in continued character constant at (1) [-Wampersand]
```

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,103 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+                        
+                        if ( trim(stream1_filename) == trim(stream2_filename) ) then
+                            message = 'Found identical values of the filename_template attribute for multiple active output streams &
+                                       (' // trim(stream1_name) // ' and ' // trim(stream2_name) // ') in streams.<CORE>. This may &
+                                        result in file conflicts.'
```
mgduda:
```
Warning: Missing ‘&’ in continued character constant at (1) [-Wampersand]
mpas_stream_manager.F:647:41:

  647 |                                         result in file conflicts.'
      |                                         1
Warning: Missing ‘&’ in continued character constant at (1) [-Wampersand]
```

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,103 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+    !-----------------------------------------------------------------------
+    subroutine MPAS_stream_mgr_check_filename_template(manager, ierr)!{{{
+
+        implicit none
+
+        type (MPAS_streamManager_type), intent(inout) :: manager
+        integer, intent(out), optional :: ierr
+
+        character(len=StrKIND) :: stream1_name, stream2_name, stream1_filename, stream2_filename
+        character(len=StrKIND) :: filename_interval
```
mgduda: It doesn't look like `filename_interval` is used in this routine.

src/framework/mpas_stream_manager.F
mgduda: I suppose we'll be removing these calls to `mpas_log_write` in the final iteration of the code, and that might allow some of the local variables (e.g., `stream1_active`) to be eliminated.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,103 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+            stream1_active = stream1_cursor % active_stream
+            stream1_filename = stream1_cursor % filename_template
+            stream1_pkg_active = stream_active_pkg_check(stream1_cursor)
+            stream1_output = (stream1_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
+                             (stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
+
+            call mpas_log_write('Stream 1 '//trim(stream1_name)//' filename '//trim(stream1_filename)//' &
+             active = $l output = $l  packages_active $l',logicArgs=(/stream1_active, stream1_output, stream1_pkg_active/))
+
+            if ( stream1_active .and. stream1_output .and. stream1_pkg_active) then
```
mgduda: Although I haven't tried it in a local copy, it might lead to more readable code (at least, require one less level of indentation for code beginning on line 628) if we were to use a `cycle` statement here; e.g.,
```fortran
            if (.not. stream1_active &
                .or. .not. stream1_output &
                .or. .not. stream1_pkg_active) then
                cycle
            end if
```

mgduda (changes requested):

src/framework/mpas_stream_manager.F
abishekg7: Would it be okay to terminate execution within `MPAS_stream_mgr_check_filename_template` in the case of invalid stream definitions, without returning an error code? 

I see that otherwise the abort happens in `mpas_init` with the call to `mpas_dmpar_global_abort`. However, this seems like a generic error message, and I'm not sure if I can pass an error message up two subroutines to print.

mgduda: I think it would be best to write an error message (but not a _critical error message_) to the log file in this routine, and to return an error code so that the calling routine has a chance to add context to the error log, clean up the state of the model and files, etc.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,104 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+        if (present(ierr)) ierr = MPAS_STREAM_LIST_NOERR
+
+        threadNum = mpas_threading_get_thread_num()
+
+        if ( threadNum == 0 ) then
+
+            streamID = '.*' !  query all streams
+
+            nullify(stream1_cursor)
+            do while (MPAS_stream_list_query(manager % streams, streamID, stream1_cursor))
```
abishekg7, line 624: Removed the `ierr` argument in calls to `MPAS_stream_list_query` as it might not actually return any error codes.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,104 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+
+        integer :: threadNum
+        character (len=StrKIND) :: message, streamID
+        type (MPAS_stream_list_type), pointer :: stream1_cursor, stream2_cursor
+        logical :: stream1_pkg_active, stream2_pkg_active
+        logical :: stream1_output, stream2_output
+
+        STREAM_DEBUG_WRITE('-- Called MPAS_stream_mgr_check_filename_template() for all streams')
+
+        if (present(ierr)) ierr = MPAS_STREAM_LIST_NOERR
```
mgduda: Would it be better to set `ierr` to `MPAS_STREAM_MGR_NOERR` since this is a stream manager routine (and since the caller of this routine, `MPAS_stream_mgr_validate_streams`, is expected to return `MPAS_STREAM_MGR_NOERR` if no error occurred)?

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,104 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+
+                    call mpas_log_write('Stream 2 '//trim(stream2_cursor % name)//' filename '//trim(stream2_cursor % filename_template)//' &
+                        & active = $l output = $l pkg_active: $l',logicArgs=(/stream2_cursor % active_stream, stream2_output, stream2_pkg_active/))
+
+                    if (trim(stream1_cursor % filename_template) == trim(stream2_cursor % filename_template)) then
+                        message = 'Found identical values of the filename_template attribute for multiple &
+                                  active output streams (' // trim(stream1_cursor % name) // ' and ' // &
+                                  trim(stream2_cursor % name) // ') in streams.<CORE>. This may result in &
+                                  file conflicts.'
+                        call mpas_log_write(message)
```
mgduda, line 675: I think it would be better to write this as an error message (i.e., with `messageType=MPAS_LOG_ERR`).

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,104 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+
+                stream1_pkg_active = stream_active_pkg_check(stream1_cursor)
+                stream1_output = (stream1_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
+                                 (stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
+
+                call mpas_log_write('Stream 1 '//trim(stream1_cursor % name)//' filename '//trim(stream1_cursor % filename_template)//' &
+                & active = $l output = $l  packages_active $l',logicArgs=(/stream1_cursor % active_stream, stream1_output, stream1_pkg_active/))
+
+                if (.not. stream1_cursor % active_stream .or. .not. stream1_output &
+                    .or. .not. stream1_pkg_active) then
```
mgduda: Just a minor comment on style: we could consider placing all conditions on the same line (in which case, the line would still not be longer than several of the lines above in this function), or we could place each condition on its own line as is done further below on lines 635-638.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,104 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+                    end if
+
+                    call mpas_log_write('Stream 2 '//trim(stream2_cursor % name)//' filename '//trim(stream2_cursor % filename_template)//' &
+                        & active = $l output = $l pkg_active: $l',logicArgs=(/stream2_cursor % active_stream, stream2_output, stream2_pkg_active/))
+
+                    if (trim(stream1_cursor % filename_template) == trim(stream2_cursor % filename_template)) then
+                        message = 'Found identical values of the filename_template attribute for multiple &
+                                  active output streams (' // trim(stream1_cursor % name) // ' and ' // &
+                                  trim(stream2_cursor % name) // ') in streams.<CORE>. This may result in &
+                                  file conflicts.'
```
mgduda:
```
mpas_stream_manager.F:649:35:
  
  649 |                                   file conflicts.'                     
Warning: Missing '&' in continued character constant at (1) [-Wampersand]
```
Generally, it's probably preferable to have fewer (or even no) breaks in error messages so that one can `grep` for error messages in the code. But, we do need to keep to the 132-column limit for source lines, and so breaking at places not in a string literal is probably best where possible.

mgduda (changes requested):

src/framework/mpas_stream_manager.F
abishekg7: Done. But also the longer mpas_log_write statements were temporary and have been removed with the next commit.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,96 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+                if (.not. stream1_cursor % active_stream .or. .not. stream1_output .or. .not. stream1_pkg_active) then
+                    cycle
+                end if
+
+                nullify(stream2_cursor)
+                do while (MPAS_stream_list_query(manager % streams, streamID, stream2_cursor))
+
+                    stream2_pkg_active = stream_active_pkg_check(stream2_cursor)
+                    stream2_output = (stream2_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
+                                     (stream2_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
```
mgduda: I may be misreading the logic, but it looks like we will only catch cases where two output (or input-output) streams share a `filename_template`. However, I think we would also want to catch cases where an active input stream shares a `filename_template` with an active output (or input-output) stream.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,96 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+                        cycle
+                    end if
+
+                    if (trim(stream1_cursor % filename_template) == trim(stream2_cursor % filename_template)) then
+                        message = 'Found identical values of the filename_template attribute for multiple active output streams, ' &
+                                  // trim(stream1_cursor % name) // ' and ' // trim(stream2_cursor % name) // &
+                                  ', in streams.<CORE>. This may result in file conflicts.'
+                        call mpas_log_write(message, messageType=MPAS_LOG_ERR)
+                        if (present(ierr)) ierr = MPAS_STREAM_MGR_ERROR
+                        return
```
mgduda: Rather than returning on the first pair of matching `filename_template` values, we could instead report all errors before returning. Admittedly, it may not be common for someone to define more than two streams that share a `filename_template` (or more than one pair of streams that conflict), but I think it's a small enough change to report all errors before returning that it's worth doing.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,96 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+            streamID = '.*' !  query all streams
+
+            nullify(stream1_cursor)
+            do while (MPAS_stream_list_query(manager % streams, streamID, stream1_cursor))
+
+                stream1_pkg_active = stream_active_pkg_check(stream1_cursor)
+                stream1_output = (stream1_cursor % direction == MPAS_STREAM_OUTPUT) .or. &
+                                 (stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT)
+
+                if (.not. stream1_cursor % active_stream .or. .not. stream1_output .or. .not. stream1_pkg_active) then
```
mgduda: Thinking about it a bit more, perhaps we could also skip (for the purpose of finding duplicate `filename_template` values) streams whose `input_interval` or `output_interval` is `none`, since that stream -- although potentially active -- would never be read or written?

mgduda (changes requested):

mgduda:
I've run a few quick tests, and it looks like the core logic is now functioning as expected. I had just a couple of ideas that might clean up the implementation:
1. The outer loop (with `stream1_cursor`) could `cycle` on anything other than an active output stream, and that might obviate the need for checks like `(stream1_input .and. stream2_input)` in the inner loop.
2. If the inner loop (with `stream2_cursor`) iterated only over streams that occur in the stream list after `stream1_cursor`, that would eliminate duplicate error messages (e.g., "Found identical values ... for ... foobar_1 and foobar_2" and later "Found identical values ... for ... foobar_2 and foobar_1").

abishekg7:
> I've run a few quick tests, and it looks like the core logic is now functioning as expected. I had just a couple of ideas that might clean up the implementation:
> 
> 1. The outer loop (with `stream1_cursor`) could `cycle` on anything other than an active output stream, and that might obviate the need for checks like `(stream1_input .and. stream2_input)` in the inner loop.
> 2. If the inner loop (with `stream2_cursor`) iterated only over streams that occur in the stream list after `stream1_cursor`, that would eliminate duplicate error messages (e.g., "Found identical values ... for ... foobar_1 and foobar_2" and later "Found identical values ... for ... foobar_2 and foobar_1").

I've incorporated the 2nd idea in the most recent commit. But having either the outer loop or the inner loop consider only output streams, appears to miss some filename conflicts, due to the directionality introduced by idea #2.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,126 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+    !  routine MPAS_stream_mgr_check_filename_template
+    !
+    !> \brief Check for identical filename templates in active I/O streams.
+    !> \author Abishek Gopal
+    !> \date   June 3 2026
+    !> \details
+    !>  Checks that there are no identical filename templates among active I/O
+    !>  streams in the stream manager, which may lead to file conflicts. This
+    !>  routine can be called from within MPAS_stream_mgr_validate_streams or
+    !>  separately as needed.
```
mgduda: Could we add a sentence or two about the return values from this routine? That may be especially relevant, since the return value of this routine may be used directly as the return value to `MPAS_stream_mgr_validate_streams`.

It might also be helpful to add a bit more detail as to what exactly this routine is checking. For example, the description "among active I/O streams" suggests that two active input streams might be caught by this routine.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,126 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+    !
+    !-----------------------------------------------------------------------
+    subroutine MPAS_stream_mgr_check_filename_template(manager, ierr)!{{{
+
+        implicit none
+
+        type (MPAS_streamManager_type), intent(inout) :: manager
+        integer, intent(out), optional :: ierr
+
+        integer :: threadNum, err_local,err_local1, err_local2
```
mgduda: Could you take another pass through the code changes in this PR and regularize the whitespace? For example, there's no space between `err_local` and `err_local1` here. Other places, the whitespace is inconsistent within parentheses (e.g., lines 608 and 612).

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,126 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+
+            nullify(stream1_cursor)
+            do while (MPAS_stream_list_query(manager % streams, streamID, stream1_cursor))
+
+                stream1_pkg_active = stream_active_pkg_check(stream1_cursor)
+
+                stream1_input_nalarms = MPAS_stream_mgr_get_num_alarms(manager, trim(stream1_cursor % name), &
+                                                                       MPAS_STREAM_INPUT, err_local1)
+                stream1_output_nalarms = MPAS_stream_mgr_get_num_alarms(manager, trim(stream1_cursor % name), &
+                                                                        MPAS_STREAM_OUTPUT, err_local2)
```
mgduda: It doesn't look like the `err_local1` or `err_local2` return error codes are ever used. It might simplify the code to make `ierr` an `optional` argument to `MPAS_stream_mgr_get_num_alarms`, and to not pass any corresponding actual argument in these calls. That would also allow the local variables `err_local1` and `err_local2` to be removed.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,126 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+
+                stream1_input = ((stream1_cursor % direction == MPAS_STREAM_INPUT .or. &
+                                stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                stream1_input_nalarms > 0)
+                stream1_output = ((stream1_cursor % direction == MPAS_STREAM_OUTPUT .or. &
+                                 stream1_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                 stream1_output_nalarms > 0)
```
mgduda: I may be mistaken, but with the exception of `stream1_input`, the above `stream1_*` variables are only used until the point where we enter the second loop over streams, and so it might be possible to simply declare the variables `pkg_active` (rather than `stream1_pkg_active` and `stream2_pkg_active`), etc. This would cut down on the number of local variables that the reader would need to consider when looking through this code.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,126 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+                        .or. (stream1_input .and. stream2_input) & ! Skip if Streams 1 and 2 are both input streams
+                        .or. .not. stream2_pkg_active) then ! Skip if Stream 2 is inactive due to inactive package
+                        cycle
+                    end if
+
+                    if (trim(stream1_cursor % filename_template) == trim(stream2_cursor % filename_template)) then
+                        message = 'Found identical values of the filename_template attribute for multiple active output streams, ' &
+                                  // trim(stream1_cursor % name) // ' and ' // trim(stream2_cursor % name) // &
+                                  ', in streams.<CORE>. This may result in file conflicts.'
+                        call mpas_log_write(message, messageType=MPAS_LOG_ERR)
```
mgduda, line 678: The error message can be quite long when printed to a single line, e.g.,
```
ERROR: Found identical values of the filename_template attribute for multiple active output streams, restart and output, in streams.<CORE>. This may result in file conflicts.
```
It might be better to explicitly write a multi-line error message with multiple calls to `mpas_log_write`.

mgduda (changes requested):

abishekg7:
I have addressed the remaining suggestions in the last few commits.

mgduda:
@abishekg7 As of now, there are quite a few messages that are written to the log file even when there are no issues with the streams. For example, here's what I get from the log file for the `atmosphere` core:
```
 Stream 1 invariant filename invariant.nc  active = T input = F output = F  packages_active T
 Stream 1 input filename x1.40962.init.nc  active = T input = T output = F  packages_active T
 Stream 2 restart filename restart.**-**-**_**.**.**.nc  active = T input = T output = T pkg_active: T
 Stream 2 iau filename x1.40962.AmB.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
 Stream 2 lbc_in filename lbc.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
 Stream 2 ugwp_oro_data_in filename x1.40962.ugwp_oro_data.nc  active = T input = T output = F pkg_active: F
 Stream 2 ugwp_ngw_in filename ugwp_limb_tau.nc  active = T input = T output = F pkg_active: F
 Stream 2 da_state filename mpasout.**-**-**_**.**.**.nc  active = T input = T output = T pkg_active: F
 Stream 2 output filename history.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
 Stream 2 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
 Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
 Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
 Stream 1 restart filename restart.**-**-**_**.**.**.nc  active = T input = T output = T  packages_active T
 Stream 2 iau filename x1.40962.AmB.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
 Stream 2 lbc_in filename lbc.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
 Stream 2 ugwp_oro_data_in filename x1.40962.ugwp_oro_data.nc  active = T input = T output = F pkg_active: F
 Stream 2 ugwp_ngw_in filename ugwp_limb_tau.nc  active = T input = T output = F pkg_active: F
 Stream 2 da_state filename mpasout.**-**-**_**.**.**.nc  active = T input = T output = T pkg_active: F
 Stream 2 output filename history.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
 Stream 2 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
 Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
 Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
 Stream 1 iau filename x1.40962.AmB.**-**-**_**.**.**.nc  active = T input = T output = F  packages_active F
 Stream 1 lbc_in filename lbc.**-**-**_**.**.**.nc  active = T input = T output = F  packages_active F
 Stream 1 ugwp_oro_data_in filename x1.40962.ugwp_oro_data.nc  active = T input = T output = F  packages_active F
 Stream 1 ugwp_ngw_in filename ugwp_limb_tau.nc  active = T input = T output = F  packages_active F
 Stream 1 da_state filename mpasout.**-**-**_**.**.**.nc  active = T input = T output = T  packages_active F
 Stream 1 output filename history.**-**-**_**.**.**.nc  active = T input = F output = T  packages_active T
 Stream 2 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
 Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
 Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
 Stream 1 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T  packages_active T
 Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
 Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
 Stream 1 surface filename x1.40962.sfc_update.nc  active = T input = F output = F  packages_active T
 Stream 1 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T  packages_active F
```
Could you update the PR so that these messages are not written?

src/framework/mpas_stream_manager.F
mgduda: @abishekg7 It looks like this suggestion is still pending.

abishekg7: The previous message has actually been modified slightly, and it is split up into two calls to mpas_log_write. Let me know if it needs further changes.

abishekg7:
> @abishekg7 As of now, there are quite a few messages that are written to the log file even when there are no issues with the streams. For example, here's what I get from the log file for the `atmosphere` core:
> 
> ```
>  Stream 1 invariant filename invariant.nc  active = T input = F output = F  packages_active T
>  Stream 1 input filename x1.40962.init.nc  active = T input = T output = F  packages_active T
>  Stream 2 restart filename restart.**-**-**_**.**.**.nc  active = T input = T output = T pkg_active: T
>  Stream 2 iau filename x1.40962.AmB.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
>  Stream 2 lbc_in filename lbc.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
>  Stream 2 ugwp_oro_data_in filename x1.40962.ugwp_oro_data.nc  active = T input = T output = F pkg_active: F
>  Stream 2 ugwp_ngw_in filename ugwp_limb_tau.nc  active = T input = T output = F pkg_active: F
>  Stream 2 da_state filename mpasout.**-**-**_**.**.**.nc  active = T input = T output = T pkg_active: F
>  Stream 2 output filename history.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
>  Stream 2 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
>  Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
>  Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
>  Stream 1 restart filename restart.**-**-**_**.**.**.nc  active = T input = T output = T  packages_active T
>  Stream 2 iau filename x1.40962.AmB.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
>  Stream 2 lbc_in filename lbc.**-**-**_**.**.**.nc  active = T input = T output = F pkg_active: F
>  Stream 2 ugwp_oro_data_in filename x1.40962.ugwp_oro_data.nc  active = T input = T output = F pkg_active: F
>  Stream 2 ugwp_ngw_in filename ugwp_limb_tau.nc  active = T input = T output = F pkg_active: F
>  Stream 2 da_state filename mpasout.**-**-**_**.**.**.nc  active = T input = T output = T pkg_active: F
>  Stream 2 output filename history.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
>  Stream 2 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
>  Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
>  Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
>  Stream 1 iau filename x1.40962.AmB.**-**-**_**.**.**.nc  active = T input = T output = F  packages_active F
>  Stream 1 lbc_in filename lbc.**-**-**_**.**.**.nc  active = T input = T output = F  packages_active F
>  Stream 1 ugwp_oro_data_in filename x1.40962.ugwp_oro_data.nc  active = T input = T output = F  packages_active F
>  Stream 1 ugwp_ngw_in filename ugwp_limb_tau.nc  active = T input = T output = F  packages_active F
>  Stream 1 da_state filename mpasout.**-**-**_**.**.**.nc  active = T input = T output = T  packages_active F
>  Stream 1 output filename history.**-**-**_**.**.**.nc  active = T input = F output = T  packages_active T
>  Stream 2 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: T
>  Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
>  Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
>  Stream 1 diagnostics filename diag.**-**-**_**.**.**.nc  active = T input = F output = T  packages_active T
>  Stream 2 surface filename x1.40962.sfc_update.nc  active = T input = F output = F pkg_active: T
>  Stream 2 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
>  Stream 1 surface filename x1.40962.sfc_update.nc  active = T input = F output = F  packages_active T
>  Stream 1 diag_ugwp filename diag_ugwp.**-**-**_**.**.**.nc  active = T input = F output = T  packages_active F
> ```
> 
> Could you update the PR so that these messages are not written?

Done.

src/framework/mpas_stream_manager.F
```diff
@@ -563,10 +563,131 @@ subroutine MPAS_stream_mgr_validate_streams(manager, streamID, ierr)!{{{
...
+                    pkg_active = stream_active_pkg_check(stream2_cursor)
+                    stream2_input = ((stream2_cursor % direction == MPAS_STREAM_INPUT .or. &
+                                    stream2_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                    input_nalarms > 0)
+                    stream2_output = ((stream2_cursor % direction == MPAS_STREAM_OUTPUT .or. &
+                                     stream2_cursor % direction == MPAS_STREAM_INPUT_OUTPUT) .and. &
+                                     output_nalarms > 0)
+
+                     call mpas_log_write('Stream 2 '//trim(stream2_cursor % name)//' filename '//trim(stream2_cursor % filename_template)//' &
+                        & active = $l input = $l output = $l pkg_active: $l',logicArgs=(/stream2_cursor % active_stream, stream2_input, stream2_output, pkg_active/))
```
mgduda, line 664: I wonder if we might be better off omitting these messages from the log file. As they are now, they might not be particularly valuable to most users. Here's what I'm seeing in a run of the `init_atmosphere` core:
```
 Stream 2 output filename init.nc  active = T input = F output = T pkg_active: T
 Stream 2 ugwp_oro_data filename ugwp_oro_data.nc  active = T input = F output = T pkg_active: F
 Stream 2 surface filename sfc_update.nc  active = T input = F output = T pkg_active: F
 Stream 2 lbc filename lbc.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
 Stream 2 ugwp_oro_data filename ugwp_oro_data.nc  active = T input = F output = T pkg_active: F
 Stream 2 surface filename sfc_update.nc  active = T input = F output = T pkg_active: F
 Stream 2 lbc filename lbc.**-**-**_**.**.**.nc  active = T input = F output = T pkg_active: F
```

mgduda (changes requested):

src/framework/mpas_stream_manager.F
abishekg7: I left this in for debugging purposes until this PR reaches the approval stage. Can remove it if there are no other changes to the main logic required.

mgduda: It's possible that the logic may still need some adjustment. When I have the following two streams defined
```
<immutable_stream name="restart"
                  type="input;output"
                  filename_template="restart.$Y-$M-$D_$h.$m.$s.nc"
                  input_interval="initial_only"
                  output_interval="6:00:00" />

<stream name="foo"
        type="input"
        filename_template="restart.$Y-$M-$D_$h.$m.$s.nc"
        input_interval="6:00:00" >

        <file name="stream_list.atmosphere.output"/>
</stream>
```
I don't get any runtime errors, though I would have expected there to be since the `restart` stream is an `input;output` stream and it shares a `filename_template` with the `foo` stream.

abishekg7: Fixed in [77c20a3](https://github.com/MPAS-Dev/MPAS-Model/pull/1464/commits/77c20a3de7840b1d9d396f7cdb56f55dc332b8ff)

PR #1465, opened 2026-06-04
Fix broken builds with GEN_F90=true by adding missing include paths to Makefiles

mgduda:
This PR fixes broken builds with `GEN_F90=true` by adding missing include paths to the Makefile recipes that used when MPAS is compiled with `GEN_F90=true`.


abishekg7 (approved):
Checked with different compilers and COREs. Seems fine.

PR #1466, opened 2026-06-05
Provide optimized code path in check_regex_match function for non-regex strings

mgduda:
This PR provides an optimized code path in the `check_regex_match` function for non-regex strings. When the `pattern` argument to the `check_regex_match` function in `regex_matching.c` is not a regular expression, a simple string comparison is sufficient and the cost of compiling a regular expression can be avoided.

Thanks to Yu-Tze Hong (@aspenhong) and Ying-Jhang Wu (@yjwu890355) from the Central Weather Administration (CWA), Taipei, Taiwan, for identifying this opportunity for optimization and for providing the initial code changes.

### DIFF
--- a/src/framework/regex_matching.c
+++ b/src/framework/regex_matching.c
@@ -34,6 +34,23 @@ static int possible_bre(const char *s)
 }
 
 
+/******************************************************************************
+ *
+ * check_regex_match
+ *
+ * Determine whether an input string 'str' matches the Basic Regular Expression
+ * (BRE) 'pattern'.
+ *
+ * Inputs:
+ *   pattern - a null-terminated string that may contain any valid BRE,
+ *             including a simple string
+ *   str     - a null-terminated string to be checked against pattern
+ *
+ * Return value: 1 if the input string str matches the BRE pattern,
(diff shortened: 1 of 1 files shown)


mgduda:
The changes in this PR offer an alternative to those in PR #1456.

jim-p-w (approved):
Looks good to me.

src/framework/regex_matching.c
```diff
@@ -1,14 +1,54 @@
...
+	for (; *s != '\0'; s++) {
+		if (strchr(BRE_SPECIAL_CHARS, *s) != NULL) {
+			return 1;
+		}
+	}
+	return 0;
+}
+
+
+void check_regex_match(const char * pattern, const char * str, int *imatch)
```
islas, line 54: I think it may be worth adding some function description to `check_regex_match()` especially as the internal helper function has details on the return value that differ slightly from the return values of `imatch` from the primary function in this file.

mgduda: I've just added documentation for the `check_regex_match` function with commit 384dff94.

islas (approved):

PR #1467, opened 2026-06-05
Add packages for hydrometeors in the init_atmosphere core

mgduda:
This PR defines new packages for hydrometeors, one package per hydrometeor, in the `init_atmosphere` core, and it provides two means of activating these packages.
    
1) If the new namelist option `config_active_hydrometeors` is set to `'detect_automatically'`, packages will be set up by scanning through the fields in the initial intermediate file, with packages being activated only for those hydrometeors that are present in the file.    
    
2) Alternatively, and to avoid the need to scan through all fields in the initial intermediate file, `config_active_hydrometeors` may be set to to a semicolon-separated list of zero or more hydrometeor names (e.g., `'qc;qr'`). Packages for those specified hydrometeors will be active, and the packages for all unspecified hydrometeors will be inactive.
    
A new subroutine, `setup_hydrometeor_packages`, handles the details of the hydrometeor package setup, and this routine is called from `init_atm_setup_packages` only if `config_init_case` is 7 or 9, and if `met_stage_out` is true.

### DIFF
--- a/src/core_init_atmosphere/Registry.xml
+++ b/src/core_init_atmosphere/Registry.xml
@@ -217,6 +217,11 @@
                      description="Whether to use specific-humidity as the first-guess moisture variable. If this option is False, relative humidity will be used."
                      possible_values="true or false"/>
 
+                <nml_option name="config_active_hydrometeors"   type="character"   default_value="detect_automatically"
+                     units="-"
+                     description="The set of hydrometeor species that are to be processed from input intermediate files when generating ICs and LBCs for real-data simulations."
+                     possible_values="`detect_automatically', or a semicolon-separated list of hydrometeor names (e.g., `qc;qr'"/>
+
         </nml_record>
 
         <nml_record name="vertical_grid" in_defaults="true">
@@ -387,6 +392,8 @@
                 <package name="first_guess_field" description="3-d atmospheric or land-surface fields on first-guess levels"/>
                 <package name="microphysics_aerosols" description="Initialization of GOCART-based water- and ice-friendly aerosols for Thompson and TEMPO microphysics"/>
                 <package name="noahmp" description="Whether to process, read, and write static fields only required by Noah-MP"/>
+                <package name="qc" description="Whether QC is available for inclusion in ICs and LBCs"/>
+                <package name="qr" description="Whether QR is available for inclusion in ICs and LBCs"/>
         </packages>
 
 
@@ -1159,10 +1166,12 @@
                              description="Water vapor mixing ratio"/>
 
                         <var name="qc" array_group="moist" units="kg kg^{-1}"
-                             description="Cloud water mixing ratio"/>
+                             description="Cloud water mixing ratio"
+                             packages="qc"/>
 
                         <var name="qr" array_group="moist" units="kg kg^{-1}"
-                             description="Rain water mixing ratio"/>
+                             description="Rain water mixing ratio"
+                             packages="qr"/>
 
                         <var name="nifa" array_group="number" units="nb kg^{-1}"
                              description="Gocart ice-friendly aerosol number concentration"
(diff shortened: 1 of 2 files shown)


src/core_init_atmosphere/Registry.xml
```diff
@@ -217,6 +217,11 @@
                      description="Whether to use specific-humidity as the first-guess moisture variable. If this option is False, relative humidity will be used."
                      possible_values="true or false"/>
 
+                <nml_option name="config_active_hydrometeors"   type="character"   default_value="detect_automatically"
+                     units="-"
+                     description="The set of hydrometeor species that are to be processed from input intermediate files when generating ICs and LBCs for real-data simulations."
+                     possible_values="`detect_automatically', or a semicolon-separated list of hydrometeor names (e.g., `qc;qr'"/>
```
abishekg7, line 223: Would it be simpler to use 'auto' instead of 'detect_automatically'?

mgduda: Given that this is the default (and therefore, users wouldn't need to type it themselves), I'd lean towards the more verbose of the two options, i.e., `detect_automatically`.

src/core_init_atmosphere/mpas_init_atm_core_interface.F
```diff
@@ -523,6 +719,82 @@ function init_atm_setup_block(block) result(ierr)
...
+   !
+   !-----------------------------------------------------------------------
+   subroutine mpas_split_string_new(string, delimiter, subStrings)
+
+      implicit none
+
+      ! Arguments
+      character(len=*), intent(in) :: string
+      character, intent(in) :: delimiter
+      character(len=:), dimension(:), allocatable, intent(inout) :: subStrings
```
abishekg7, line 747: Could declaring this with intent(out) work instead? I see that the compiler will automatically deallocate an allocatable with intent(out) before entering the routine. https://fortran-lang.org/learn/best_practices/allocatable_arrays/

src/core_init_atmosphere/mpas_init_atm_core_interface.F
```diff
@@ -523,6 +719,82 @@ function init_atm_setup_block(block) result(ierr)
...
+   !>  delimiter characters.
+   !>
+   !>  If more than one delimiter character appears consecutively, no empty string
+   !>  in the output subStrings array is generated.
+   !>
+   !>  The length of the strings in the output subStrings argument is equal to the
+   !>  length of the longest substring in the input string.
+   !
+   !-----------------------------------------------------------------------
+   subroutine mpas_split_string_new(string, delimiter, subStrings)
```
abishekg7, line 740: Also wondering if this subroutine might be better placed in some common utils module

src/core_init_atmosphere/mpas_init_atm_core_interface.F
```diff
@@ -316,9 +319,202 @@ function init_atm_setup_packages(configs, streamInfo, packages, iocontext, dminf
...
+      else
+
+         call mpas_log_write('Setting up hydrometeor packages from config_active_hydrometeors list: ' &
+                             //trim(config_active_hydrometeors))
+
+         call mpas_split_string_new(trim(config_active_hydrometeors), ';', hydrometeors)
+
+         do i = 1, size(hydrometeors)
+            select case (trim(hydrometeors(i)))
+               case ('qc')
```
abishekg7, line 487: This part of the code is case sensitive to the values specified in `config_active_hydrometeors`. Is that okay?

src/core_init_atmosphere/mpas_init_atm_core_interface.F
mgduda: The commit message for 3bc08ca9 does recognize that this routine might ultimately be migrated to the `mpas_string_utils` module. I'll replace 3bc08ca9 with a commit that adds this routine directly to `mpas_string_utils.F`.

src/core_init_atmosphere/mpas_init_atm_core_interface.F
mgduda: Could you do some testing to verify that this works in practice with the GNU, Intel, and NVHPC compilers?

src/core_init_atmosphere/mpas_init_atm_core_interface.F
mgduda: I'm fine with keeping this case-sensitive for now.

src/core_init_atmosphere/mpas_init_atm_core_interface.F
abishekg7: will check

abishekg7: I'm fine with the PR as it is, and will approve for now. But can check this with various compilers for future reference.

abishekg7 (approved):

src/core_init_atmosphere/mpas_init_atm_core_interface.F
mgduda: Ok, I'll merge this PR as it is now, and in a future PR, we can move `mpas_split_string_new` to the `mpas_string_utils` module. In that future PR, we can switch to the use of `intent(out)` for the `subStrings` argument, and we can also add unit tests for `mpas_split_string_new` to the `test` core.

PR #1472, opened 2026-06-15
Build option to download and build PT-SCOTCH automatically

abishekg7:
Previously, using PT-SCOTCH for online graph partitioning with MPAS required the user to build and install it themselves and set SCOTCH=<path> environment variable before building MPAS. While this is a completely functional option, many users may prefer the convenience of an automatic fetch and build process. 

This PR adds an option to automatically fetch and build PT-SCOTCH under `src/external` during the MPAS build. The previously active `SCOTCH` variable is repurposed as  a `true or false` build option to toggle linking PT-SCOTCH library with the MPAS build. Previously, `SCOTCH` pointed to the PT-SCOTCH installation path. This is now available through the `SCOTCH_ROOT` variable. This allows the user to still link MPAS with custom PT-SCOTCH installations.

There are a few different build + link scenarios with these two variables

1. When `SCOTCH=false` or unspecified, the regular MPAS build process continues without attempting to build or link against PT-SCOTCH.

```
make gnu CORE=init_atmosphere
```

2. When `SCOTCH=true` and `SCOTCH_ROOT` is not set or unset, the build process automatically fetches and builds the PT-SCOTCH graph-partitioning library, followed by the build of the MPAS model, and then the PT-SCOTCH library is linked as a static library.

```
make gnu CORE=init_atmosphere SCOTCH=true
```

3. When `SCOTCH=true` and `SCOTCH_ROOT` points to the location of an existing PT-SCOTCH installation, the regular MPAS build process completes and then the PT-SCOTCH library is linked as a static library.

```
export SCOTCH_ROOT=/path/to/scotch/installation
make gnu CORE=init_atmosphere SCOTCH=true
```

4. When `SCOTCH=true` and `SCOTCH_ROOT` is set but does not point to the location of an existing PT-SCOTCH installation, the build process exits with an error. 

##
Notes:

- A new Externals.cfg is added under the src directory so that both the atmosphere and init_atmosphere CORES can use this external

- If the automatic PT-SCOTCH fetch and build is successful, it is installed under `src/external/scotch/install`. And the libraries under the `lib` subdirectory. However, in the case of a user-specified PT-SCOTCH installation, we search both `$(SCOTCH_ROOT)/lib` and `$(SCOTCH_ROOT)/lib64` subdirectories to find the relevant libraries.

- `SCOTCH=$(SCOTCH)` is added to the options tracked by `rebuild_check`, so switching SCOTCH on/off between builds is detected as an incompatible-option change (triggering the clean/rebuild prompt or AUTOCLEAN).

- Add a scotch cleanup step (rm -r build install) to the clean target so make clean removes the downloaded/built SCOTCH tree along with the other external libs.

###
Current limitations

- CMake (>= 3.10) is presently required in order to build PT-SCOTCH using the automatic fetch + build method. In addition, all the usual dependencies of PT-SCOTCH need to be present in the environment prior to the build.

- We presently support only the linking of static PT-SCOTCH libraries with MPAS. Shared libraries are currently not supported.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -777,6 +777,34 @@ endif
 	LIBS += $(NCLIB)
 endif
 
+export SCOTCH ?= false
+ifeq "$(SCOTCH)" "true"
+	ifeq ($(wildcard $(SCOTCH_ROOT)), )
+		export SCOTCH_RELPATH=src/external/scotch
+		export SCOTCH_ROOT=${CURDIR}/${SCOTCH_RELPATH}/install
+		export SCOTCH_LIB_DIR=lib
+	else
+		ifneq ($(wildcard $(SCOTCH_ROOT)/lib/libptscotch.a), )
+			SCOTCH_LIB_DIR=lib64
+		else ifneq ($(wildcard $(SCOTCH_ROOT)/lib64/libptscotch.a), )
+			SCOTCH_LIB_DIR=lib64
+		else
+			$(error Could not find SCOTCH library in $(SCOTCH_ROOT)/lib or $(SCOTCH_ROOT)/lib64)
+		endif
+	endif
+	SCOTCH_INCLUDES += -I$(SCOTCH_ROOT)/include
+	SCOTCH_LIBS += -L$(SCOTCH_ROOT)/$(SCOTCH_LIB_DIR) -lptscotch -lscotch  -lptscotcherr -lm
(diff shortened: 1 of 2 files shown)


abishekg7:
Tested cases 1, 2, 3, 4 with `CORE=init_atmosphere` with gnu, intel and nvhpc cpu builds + runs on Derecho.

PR #1473, opened 2026-06-15
Atmosphere/package eval order

mgduda:
This PR incorporates several small modifications to the registry-defined package logic processing in the atmosphere core, especially concerning the `sfclayer` package, ultimately allowing for improved evaluation of physics namelist options in package logic.

### DIFF
--- a/src/core_atmosphere/mpas_atm_core_interface.F
+++ b/src/core_atmosphere/mpas_atm_core_interface.F
@@ -134,11 +134,6 @@ module atm_core_interface
       integer :: local_ierr
 
 
-      ierr = atm_setup_packages_when(configs, packages)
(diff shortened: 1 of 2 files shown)


src/core_atmosphere/mpas_atm_core_interface.F
```diff
@@ -260,6 +254,11 @@ function atm_setup_packages(configs, streamInfo, packages, iocontext, dminfo) re
 
 #endif
 
+      ierr = atm_setup_packages_when(configs, packages)
```
abishekg7, line 257: Not sure if users/community are expected to add new logic in this function, but adding a brief comment about this call needing to follow any new logic might be useful.

PR #995, opened 2022-10-07
Add 3-D grid analysis nudging FDDA to MPAS-A

jherwehe:
This EPA_FDDA PR will add 3-D grid analysis nudging FDDA to MPAS-A as 
an option to assist in keeping retrospective simulations dynamically 
on track to improve evaluation against observations.  This FDDA 
development for MPAS-A (Bullock et al., 2018) nudges temperature, 
humidity, and wind toward target values and is based on the comparable 
FDDA feature available in WRF, including MPAS-A versions of many of the 
WRF FDDA options.  Enabling grid analysis nudging FDDA in MPAS-A 
maintains high fidelity to the reference fields while still conserving 
mass.

An extra step using init_atmosphere_model is necessary to generate the 
FDDA file required for input to a subsequent MPAS-A simulation with FDDA 
enabled.  A script (provided to the MPAS-A maintainers) creates the FDDA 
file by running init_atmosphere_model in a loop to create an initial 
conditions file for each specified time (currently at 6-h intervals) of 
the available reference driving data and then utilizes open source netCDF 
Operator tools (NCO; https://nco.sourceforge.net/) to concatenate those
init files and add variable descriptions and mesh information.

**New stream file added to streams.atmosphere under \<streams\>:**
<immutable_stream name="fdda"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;type="input"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;filename_template="x4.163842.US_fdda.jul2016.nc"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;input_interval="6:00:00"/>

**New namelist.atmosphere options added under &physics:**
&nbsp;&nbsp;&nbsp;config_fdda_scheme = 'analysis'
&nbsp;&nbsp;&nbsp;config_fdda_int = 21600.
[...]

PR #1481, opened 2026-07-02
Revised PR#995 for adding analysis grid nudging capability

maywswong:
This PR is created to merge code changes related to adding an analysis grid nudging capability in MPAS-A. The core of the code changes was presented in Bullock et al. (2018) and PR #995.  In this PR, several updates and revisions are made to PR #995 and are as follow.

(1) Added capabilities in init_atmosphere_model to generate the target "FDDA" analysis files that are required for input to atmosphere_model when the analysis nudging option is turned on, instead of using an offline script as presented in PR #995. See below for instructions on generating target files using init_atmosphere.

(2) Calls to the computation of the nudging tendencies are implemented similar to those for other additional tendency terms, such as those from Incremental Analysis Update (IAU), instead of adding the nudging tendencies inside the physics as implemented in PR #995.

(3) Renamed new variables in PR #995 (XX_fdda_new, XX_fdda_old) to a more concise two-time-level variables (XX_fdda). Also renamed the namelist record in namelist.atmosphere from &fdda to &nudging.

========= **Instructions for generating and using files with nudging target fields**  ========= 

To generate the nudging target files using init_atmosphere (e.g., from six-hourly analyses):

a. Create your `init.nc` file as usual. This step needs to be done prior to generating the nudging files because `init.nc` contains the vertical and horizontal mesh information needed for interpolating the analyses onto the MPAS mesh.

b. In `namelist.init_atmosphere`, set `config_init_case = 11`, and set `config_start_time` and `config_stop_time` to reflect the beginning and end of your simulation period. The corresponding preprocessed six-hourly intermediate files should reside in the run directory, or as defined by `config_met_prefix` (similar to when generating lateral boundary condition files for a regional simulation).

c. In `streams.init_atmosphere`, add the immutable output stream for the target files:

```
<immutable_stream name="fdda"
                  type="output"
                  filename_template="fdda.$Y-$M-$D_$h.$m.$s.nc"
                  filename_interval="output_interval"
                  packages="fdda"
                  clobber_mode="replace_files"
                  output_interval="06:00:00" />
```

d. When running MPAS-atmosphere, use the same `filename_template` when defining the immutable input stream in `streams.atmosphere`:

```
<immutable_stream name="fdda"
                  type="input"
                  filename_template="fdda.$Y-$M-$D_$h.$m.$s.nc"
                  input_interval="6:00:00" />
```

e. When running MPAS-atmosphere, add the following example namelist record to use nudging option:

```
&nudging
    config_fdda_scheme = 'analysis'
    config_fdda_t = true
    config_fdda_q = false
    config_fdda_uv = true
    config_fdda_uv_in_pbl = false
    config_fdda_t_in_pbl = false
    config_fdda_q_in_pbl = false
/
```
Namelist option `config_fdda_scheme` set to `'analysis'` (`'off'`) turns on (off) the nudging option. 

===================================================================== 

**Reference:**
Bullock Jr., O. R., H. Foroutan, R. C. Gilliam, and J. A. Herwehe, 2018:
   Adding four-dimensional data assimilation by analysis nudging to the
   Model for Prediction Across Scales – Atmosphere (version 4.0). Geosci.
   Model Dev., 11, 2897–2922. https://doi.org/10.5194/gmd-11-2897-2018

**Related PR:** https://github.com/MPAS-Dev/MPAS-Model/pull/995


src/core_init_atmosphere/mpas_init_atm_cases.F
```diff
@@ -6903,6 +6967,841 @@ real (kind=RKIND) function atm_get_sounding_2( variable, height )
    end function atm_get_sounding_2
 
 !-----------
+<<<<<<< HEAD
```
mgduda: I think you'll need to remove the git conflict markers here and lines 7803-7804.

src/core_atmosphere/physics/mpas_atmphys_todynamics.F
```diff
@@ -26,7 +26,11 @@ module mpas_atmphys_todynamics
 
  implicit none
  private
+<<<<<<< HEAD
+ public:: physics_get_tend, pre_physics_get_tend, post_physics_get_tend, tend_toEdges
+=======
  public:: physics_get_tend, pre_physics_get_tend, post_physics_get_tend
+>>>>>>> 91c5eac175eebeaf4206bacd5cb50c39dff3c152
```
mgduda: This looks like an unresolved git merge conflict.

src/core_atmosphere/Registry.xml
```diff
@@ -1790,6 +1893,22 @@
...
+                <var name="theta_fdda" type="real" dimensions="nVertLevels nCells Time" units="K"
+                     description="potential temperature for FDDA grid nudging"/>
+
+                <var name="u_fdda" type="real" dimensions="nVertLevels nCells Time" units="m s^{-1}"
+                     description="zonal wind component for FDDA grid nudging"/>
+
+                <var name="v_fdda" type="real" dimensions="nVertLevels nCells Time" units="m s^{-1}"
+                     description="meridional wind component for FDDA grid nudging"/>
+
+        </var_struct>
```
mgduda: So that these fields aren't allocated for simulations that don't make use of FDDA, we could define a new package and attach it to these fields.

src/core_atmosphere/physics/mpas_atmphys_driver.F
```diff
@@ -177,7 +177,7 @@ subroutine physics_driver(domain,itimestep,xtime_s)
     config_pbl_scheme        .ne. 'off' .or. & 
     config_radt_lw_scheme    .ne. 'off' .or. &
     config_radt_sw_scheme    .ne. 'off' .or. &
-    config_sfclayer_scheme   .ne. 'off') then
+    config_sfclayer_scheme   .ne. 'off') then
```
mgduda: It doesn't look like any of the changes in `mpas_atmphys_driver.F` are related to FDDA, so perhaps we could revert them.

src/core_atmosphere/Registry.xml
```diff
@@ -1170,6 +1261,18 @@
 #endif
 		</stream>
 
+		<stream name="fdda" 
+                        type="input" 
+                        filename_template="x1.40962.fdda.nc"
```
mgduda: Just a small usability suggestion: we could set the default `filename_template` here to match the default `filename_template` used for the "fdda" stream in the init_atmosphere core (i.e., `fdda.$Y-$M-$D_$h.$m.$s.nc`).

PR #1483, opened 2026-07-03
Reduce the constant used to create rain number concentration from cloud-water autoconversion

AndersJensen-NOAA:
Reduce the constant used to create rain number concentration from cloud-water autoconversion to be consistent with WRF for Thompson microphysics.

Here's the value used in WRF:
https://github.com/wrf-model/WRF/blob/06d4240ae989cc3e50af412bb472df3d9048783c/phys/module_mp_thompson.F#L2191

A value of 200 results in reduced warm-rain precipitation, particularly along the Pacific coast. Testing done by Trude Eidhammer.


weiwangncar (approved):
It increases the domain averaged rainfall by a small amount in a 3 km domain test near Singapore.

PR #1485, opened 2026-07-08
Replaced log gamma function with intrinsic

robertmullis:
This is a simple fix to replace the existing problematic log gamma function with the Fortran intrinsic. We performed a test run, and this did give slightly different results, but given the simplicity and equivalence of the function, we believe they are ultimately doing the same thing.

- This is my first PR to MPAS. Please advise on any processes. I have another PR coming with similar numerical changes.


robertmullis:
Updating this to include the incomplete gamma function. Ignore this for now.

PR #1486, opened 2026-07-08
Replaced problematic incomplete gamma function with modernized version from SLATEC (copyright free).

robertmullis:
This is a fix to replace the legacy incomplete gamma function with a modernized version from the SLATEC library, which is in the public domain. This also replaces the problematic log gamma function with the LOG_GAMMA Fortran intrinsic. Unit testing was performed on the function by compiling the source file against stand-in framework modules and comparing a series of values between the new and original implementations. 

The implementations had a maximum relative difference of 4.1e-7, and the new implementation had a greater accuracy when compared to the Wolfram Alpha result for the same inputs (~1e-14 relative).

The functions were removed outright from `[src/core_atmosphere/physics/physics_wrf/module_mp_thompson.F]` because they were legacy code that was not called anywhere else.

- This is my first PR to MPAS. Please advise on any processes for the future.

### DIFF
--- a/src/core_atmosphere/physics/mpas_atmphys_functions.F
+++ b/src/core_atmosphere/physics/mpas_atmphys_functions.F
@@ -14,11 +14,10 @@
 
  implicit none
  private
- public:: gammln,gammp,wgamma,rslf,rsif
+ public:: gammp,wgamma,rslf,rsif
 
 
  contains
-
 
 !=================================================================================================================
 !NOTE: functions rslf and rsif are taken from module_mp_thompson temporarily for computing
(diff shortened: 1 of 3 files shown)


jeguerra-2025:
Hi @robertmullis,

Similarly to your other PR#1487, we really appreciate this effort to bring these processing elements completely into the public domain. However, we will need to discuss a more complete testing protocol here. Because these functions alter the behavior of the microphysics, changes here could be significant. Can you please elaborate more on the testing that you have performed? What would be your recommendation for a more comprehensive test set here?

PR #1496, opened 2026-07-24
Change default value of atmosphere core config_number_of_sub_steps from 2 to 4

mgduda:
This PR changes the default value of the atmosphere core namelist option `config_number_of_sub_steps` from 2 to 4 in the `Registry.xml` file. This change addresses an issue in convection-permitting simulations where noise can sometimes appear in the jet streams using the previous default value (`config_number_of_sub_steps = 2`). The noise, when present, also produces unphysical features in horizontal kinetic energy spectra at high wavenumbers.

_Note: Increasing the number of sub-steps from 2 to 4 can be expected to increase the computational cost of the dynamics by ~10%, and the computational cost of a full-physics simulation by ~5%._

### DIFF
--- a/src/core_atmosphere/Registry.xml
+++ b/src/core_atmosphere/Registry.xml
@@ -101,10 +101,10 @@
                      description="Whether to super-cycle scalar transport"
                      possible_values="Logical values"/>
 
-                <nml_option name="config_number_of_sub_steps" type="integer" default_value="2"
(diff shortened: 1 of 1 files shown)


jeguerra-2025 (approved):
This is a fine default setting if it promotes an overall more stable time integration of the model. I'm confident it does. But, if it results in a ~10% runtime penalty, can we look for a way of compensating for that added cost? Does the added stability allow us to use a slightly longer time step for example?

PR #1499, opened 2026-08-05
Fix for LES tke in real-data cases

dudhia:
For the prognostic tke option, initial tke was zero. This leads to no tke generation because source terms depend on non-zero tke through the K diffusion coefficients. The fix is to set a small minimum value (currently chosen as 0.01 m2/s) on the K coefficients. Two lines are added in the dissipation module.


dudhia:
Not sure what branch to merge to.

skamaroc (approved):
This change is necessary as stated in the pull request and the commit message.  The lower bound on the tke is consistent with practices in other models (e.g. WRF) and is in the correct location in the MPAS code base.

Issue #1501, opened 2026-08-13
CMake configure runs `checkout_externals` unconditionally


fmahebert:
**Problem:**

The CMake in `src/core_atmosphere/CMakeLists.txt` runs `checkout_externals` unconditionally on every configure. Logging the run shows the default (update) path makes three network round-trips per external:

```
git fetch --quiet --tags origin                  ← network
git ls-remote --exit-code --heads origin <tag>   ← network
git ls-remote --exit-code --heads origin <tag>   ← network (identical, repeated)
git checkout --quiet <tag>                       ← onto the commit it is already on, for most re-configures
```

With two externals (`MMM-physics` @ `20250616-MPASv8.3`, `UGWP` @ `MPAS_20241223`) that is six GitHub round-trips, measuring in at 3–4 s on my dev system, for _cold and warm configures_ alike, pulling data for tags that are likely already up-to-date locally.

Besides the time cost, the other consequence is that configure fails with no network. This blocks air-gapped builds, sandboxed CI, or any build behind a flaky network.

**Proposed fix:**

Use `checkout_externals --status` to check whether the externals match their `Externals.cfg` versions. This operates locally, with no network, in essentially ~0 s. Therefore, we can run it first and only fall through to the update when a component is missing, on the wrong version, or locally modified.

guoqing-noaa:
I would advocate we adopt the standard git submodule method for this situation.

fmahebert:
@guoqing-noaa while I'm not an MPAS model expert, I think that could work, and would have the benefit of a more standardized approach. However, note that we can improve the situation with less disruption to the user community by patching the current logic around git fetching — that's what PR 1502 is doing.

fmahebert:
Resolved by #1502

PR #1502, opened 2026-08-13
Skip externals update when already up to date

fmahebert:
This PR is a possible fix to https://github.com/MPAS-Dev/MPAS-Model/issues/1501

The PR branch first runs `checkout_externals --status` to check whether the external versions match what is in `Externals.cfg`. Only if there is a mismatch do we fall back to remote git operations to update the local copy. We determine the mismatch by parsing the output of the status command.

Tested by running locally and observing (a) out-of-date externals are still updated from upstream, and (b) up-to-date externals are now verified locally and network access is skipped.

This saves 3-4 seconds of configure time on my dev system.


jim-p-w (approved):
When both of the external repositories are checked out and are clean, cloning the external repositories is skipped.
If either of the repositories is not checked out, the missing repo does get cloned.

If one repo has modifications and the other repo isn't checked out, nothing happens and cmake fails. I think that is a problem with the checkout_externals mechanism and is outside the scope of this change.

mgduda:
@fmahebert The changes look good to me. Could you add some detail to the commit message? Generally, we'd like a one-line summary (80 characters or less), followed by a blank line, and then one or more paragraphs explaining the changes in the commit (wrapping lines at 80 characters).

fmahebert:
@mgduda no problem; I've just force-pushed a change to rephrase the commit message. Please let me know if this is an appropriate level of detail, I'm happy to add or remove if you recommend it.

mgduda:
@fmahebert The new commit message looks great -- thanks for the update! (For a bit more background, since we use merge commits in the MPAS-Model, individual commits are preserved, and in case a commit gets cherry-picked onto another branch, it can become detached from the PR that introduced that commit; so, including details of the code changes in individual commit messages is helpful.)

mgduda (approved):

fmahebert:
Thanks for the context @mgduda, that policy makes sense.

PR #1504, opened 2026-08-19
Use the file() CREATE_LINK command to create symlinks when running cmake

jim-p-w:
This PR replaces the cmake execute_process() command with the file() command to create symlinks.

Using execute_process to make symlinks is expensive; it spawns a new cmake process.
The file() command is more efficient; it makes the symlink in the currently executing cmake process.

This was tested by running cmake directly against a clone of the MPAS-Model repository, as well
as running cmake against a clone of the mpas-bundle repository (which include MPAS-Model).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 ## MPAS-Model
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.14)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Functions/MPAS_Functions.cmake)
 get_mpas_version(MPAS_VERSION)
--- a/src/core_atmosphere/CMakeLists.txt
+++ b/src/core_atmosphere/CMakeLists.txt
@@ -388,8 +388,8 @@ set(PHYSICS_WRF_DATA_DIR ${mpas_data_SOURCE_DIR}/atmosphere/physics_wrf/files)
 file(GLOB PHYSICS_WRF_DATA RELATIVE ${PHYSICS_WRF_DATA_DIR} "${PHYSICS_WRF_DATA_DIR}/*")
 file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/${PROJECT_NAME}/core_atmosphere)
 foreach (data_file IN LISTS PHYSICS_WRF_DATA)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink ${PHYSICS_WRF_DATA_DIR}/${data_file}
-                    ${CMAKE_BINARY_DIR}/${PROJECT_NAME}/core_atmosphere/${data_file})
+    file(CREATE_LINK ${PHYSICS_WRF_DATA_DIR}/${data_file}
+         ${CMAKE_BINARY_DIR}/${PROJECT_NAME}/core_atmosphere/${data_file} SYMBOLIC)
 endforeach ()
 install(DIRECTORY ${PHYSICS_WRF_DATA_DIR}/ DESTINATION ${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}/core_atmosphere)
 


mgduda:
@svahl991 -- @jim-p-w suggested adding you as a reviewer as well. If you'd like, I can first add you as a collaborator, then add you as a formal reviewer. Otherwise, if you'd just like to comment on the PR, that should be fine, too.

src/core_atmosphere/CMakeLists.txt
```diff
@@ -388,8 +388,7 @@ set(PHYSICS_WRF_DATA_DIR ${mpas_data_SOURCE_DIR}/atmosphere/physics_wrf/files)
 file(GLOB PHYSICS_WRF_DATA RELATIVE ${PHYSICS_WRF_DATA_DIR} "${PHYSICS_WRF_DATA_DIR}/*")
 file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/${PROJECT_NAME}/core_atmosphere)
 foreach (data_file IN LISTS PHYSICS_WRF_DATA)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink ${PHYSICS_WRF_DATA_DIR}/${data_file}
-                    ${CMAKE_BINARY_DIR}/${PROJECT_NAME}/core_atmosphere/${data_file})
+    file(CREATE_LINK ${PHYSICS_WRF_DATA_DIR}/${data_file} ${CMAKE_BINARY_DIR}/${PROJECT_NAME}/core_atmosphere/${data_file} SYMBOLIC)
```
jim-p-w: Should I break this into two lines, e.g.
```
    file(CREATE_LINK ${PHYSICS_WRF_DATA_DIR}/${data_file}
         ${CMAKE_BINARY_DIR}/${PROJECT_NAME}/core_atmosphere/${data_file} SYMBOLIC)
```

svahl991:
> @svahl991 -- @jim-p-w suggested adding you as a reviewer as well. If you'd like, I can first add you as a collaborator, then add you as a formal reviewer. Otherwise, if you'd just like to comment on the PR, that should be fine, too.

This looks fine to me, but I will also pass this buck down the line to @fmahebert, who was the one who put together all the PR's similar to this one in the JEDI repositories.

fmahebert (commented):
This matches what I'd expect based on what I did in JEDI.

My only suggestion would be to make sure `cmake_minimum_required(VERSION 3.14)` is specified, for `CREATE_LINK` support.

Thanks, @jim-p-w !

(Context for those not following JEDI repo developments: doing this change across most of jedi-bundle repos eliminated ~4000 subprocess launches. That saves ~7 seconds of CMake configuration time on a Apple M2 based dev system, and >60 seconds on one HPC system with a network filesystem (and perhaps policies that limit process launching?). The benefit in any particular repo would be proportional to the number of symlinks/processes involved in the configuration.)

jim-p-w:
@fmahebert Thanks for the suggestion. I changed `cmake_minimum_required` from 3.12 to 3.26 (which was released in 2023 and is the oldest version currently available on derecho).

fmahebert:
@jim-p-w sounds good, thanks — for those of at JCSDA, we've had access to CMake 3.27+ since spack-stack 1.8 from two years ago, so shouldn't be a problem.

fmahebert (approved):
non-expert review

CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 ## MPAS-Model
-cmake_minimum_required(VERSION 3.12)
+# cmake version in ncarenv/23.06
+cmake_minimum_required(VERSION 3.26)
```
mgduda, line 3: Is CMake 3.26 required by the other changes in this PR? If not, perhaps there's no need to otherwise require a newer version of CMake.

jim-p-w: file(CREATE_LINK) was introduced in cmake 3.14, so cmake_minimum_required must be increased.
As noted above, cmake 3.26 was released in 2023.

mgduda: Would it make more sense to set the minimum required version to 3.14 then?

jim-p-w: Yes, I changed the minimum version to 3.14. I have tested with cmake version 3.14. ~~I have not tried to build with cmake 3.14, I currently don't have access to any systems with that version.~~

mgduda:
@jim-p-w This change looks good to me, and I'd only suggest combining the two commits from this PR into a single commit before we merge.

jim-p-w:
> @jim-p-w This change looks good to me, and I'd only suggest combining the two commits from this PR into a single commit before we merge.

New commit has been pushed.

CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 ## MPAS-Model
-cmake_minimum_required(VERSION 3.12)
+# cmake version in ncarenv/23.06
```
mgduda: I'd suggest removing this comment.

jim-p-w: Good catch.

mgduda (approved):

PR #1507, opened 2026-09-03
Fix qv index lookup for lbc_scalars

jihyeonjang:
This PR fixes the qv constituent index lookup in the `init_atm_case_lbc` subroutine. 

The `init_atm_case_lbc` subroutine retrieves the `lbc_scalars` array from the `lbc_state` pool, but obtains the `qv` constituent index from the `state` pool using `index_qv`. The `scalars` and `lbc_scalars` arrays are separate variable arrays, and their constituent indices are generated independently. Therefore, the `index_lbc_qv` should be retrieved from `lbc_state` and used when indexing `lbc_scalars`.

Since `index_qv` and `index_lbc_qv` are expected to always be the same (both equal to 1), this change would not affect the lateral boundary conditions generated for regional MPAS simulations.


mgduda (approved):
Nice work spotting this issue, and thanks for the clear PR description!

PR #1512, opened 2026-09-08
Optimize ACC port of atm_advance_acoustic_step_work:3976

abishekg7:
This PR introduces optimizations for the OpenACC port of atm_advance_acoustic_step_work: 3976. The table below lists the timings for a real, global 30km experiment on A100 GPU with the nvhpc.

For nvhpc gpu runs, -gpu=math_uniform is introduced as a build flag to ensure optimizations are bit-identical, and the we report the numbers using NV_ACC_TIME=1. The GPU runs are on 1 Derecho GPU node, using 1 A100 via 1 MPI task.

The numbers reported for the CPU runs with gnu and intel compilers use the newly-added timers local to this region, and are averaged across three runs. The CPU runs use a single derecho CPU node each, fully subscribed to 128 MPI tasks.

Version | GPU kernel time (ms) |   | CPU timer - gnu |   | CPU timer - intel25
-- | -- | -- | -- | -- | --
base | 12.825 |   | 3.71329 |   | 3.41218
Opt 1 | 10.954 |   | 3.846676667 |   | 3.7927
Opt 1 + macros | 10.952 |   |   |   | 3.515413333
Opt 2 + macro | 9.354 |   | 3.80428 |   | 3.436093333
Opt 3 | 9.276 |   | 3.691653333 |   | 3.41279

TODO: Add @Pranay-Reddy-Kommera as primary commit author

This work was completed in part at the NCAR/NLR/NOAA Open Hackathon, part of the Open Hackathons program. The authors would like to acknowledge OpenACC-Standard.org for their support.

### DIFF
--- a/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
+++ b/src/core_atmosphere/dynamics/mpas_atm_time_integration.F
@@ -3972,7 +3972,7 @@ module atm_time_integration
       end if
 
 !$OMP BARRIER
-
+      call mpas_timer_start('atm_advance_acoustic_step_3976')
       !$acc parallel default(present)
       !$acc loop gang worker private(ts,rs)
       do iCell=cellSolveStart,cellSolveEnd  ! loop over all owned cells to solve
@@ -4104,7 +4104,7 @@ module atm_time_integration
 
       end do !  end of loop over cells
       !$acc end parallel
-
+      call mpas_timer_stop('atm_advance_acoustic_step_3976')
 
    end subroutine atm_advance_acoustic_step_work
 


src/core_atmosphere/dynamics/mpas_atm_time_integration.F
```diff
@@ -4046,8 +4057,25 @@ subroutine atm_advance_acoustic_step_work(nCells, nEdges, nCellsSolve, cellStart
                           + cofwt(k-1,iCell)*(etp(k-1)*ts(k-1)+etm(k-1)*rtheta_pp(k-1,iCell)))
             end do
 
-            ! tridiagonal solve sweeping up and then down the column
+#ifdef MPAS_OPENACC
```
mgduda, line 4060: It may be that we wouldn't lose much (any?) CPU performance if we were to move this directive below the `k`-loop (so, move it to line ~4066), and to simply use `rho_pp(k, iCell)` and `rtheta_pp(k, iCell)` on lines 4111 and 4113. This would allow us to eliminate the need for the `RHO_PP` and `RTHETA_PP` macros.

mgduda: Using a regional domain with 65094 cells on one CPU core, there does appear to be a slight CPU performance boost (~5%) with this proposed change; this was using the Intel 2025.3.2 compilers. However, the result could be different with other compilers or with more or fewer cells per MPI task.

abishekg7: Thanks! Will also check this with the global runs.

abishekg7: For the global real case, with 655362 cells, I'm noticing a 1.8% slowdown with `intel/2025.2.1` and 4.4% slowdown with `gnu/12.2.0`

mgduda: For my regional test case using the GNU 14.3.0 compilers, I'm seeing a ~1% speedup with the proposed changes when running with 1 MPI task, and about a ~3% speedup when running with 128 MPI tasks (though based on just one run from each combination of code version and MPI task count).

Since the code region in question only accounts for a few percent of the total time integration time, I'd say any speedup or slowdown on the order of a few percent is probably negligible, and the benefit in the way of simpler code is probably good enough reason to go with the proposed code changes.

abishekg7: Thanks for looking into it! I have modified commit `167ed` to include these changes. And I think the PR is in otherwise a decent shape for review.

